PR #2, opened 2017-05-19
Fv couplings

IvarStefansson:
First version of the coupling and grid elimination stuff.


src/porepy/numerics/fv/test_coupler_mpfa.py
```diff
@@ -0,0 +1,133 @@
...
+        kxx = np.ones(g.num_cells)
+
+        if all(g.cell_centers[0, :] < 0.0001):
+            perm = second_order_tensor.SecondOrderTensor(3,
+                                                         kxx / 100, kxx, kxx)
+        else:
+            perm = second_order_tensor.SecondOrderTensor(3,
+                                                         kxx * np.power(100, g.dim < 3))
+
+        d['k'] = perm
```
keileg: I am not happy with such non-descriptive names, but we should set standards for this on a more general level.

src/porepy/numerics/fv/test_coupler_mpfa.py
```diff
@@ -0,0 +1,133 @@
...
+
+from porepy.grids.coarsening import *
+from porepy.grids import grid_bucket, remove_grids
+from porepy.fracs import meshing  # split_grid
+
+from porepy.numerics.mixed_dim import coupler
+import copy
+from porepy.numerics.fv import mpfa, tpfa_coupling, tpfa
+from porepy.numerics.mixed_dim import condensation
+
```
keileg: Reorganize imports (generally), in line with https://www.python.org/dev/peps/pep-0008/#imports

src/porepy/grids/grid_bucket.py
```diff
@@ -582,3 +597,13 @@ def compute_geometry(self, is_embedded=True, is_starshaped=False):
                             is_starshaped=is_starshaped) for g, _ in self]
 
 #------------------------------------------------------------------------------#
+
+    def copy(self):
```
keileg: Comments.

src/porepy/grids/grid_bucket.py
```diff
@@ -556,6 +578,36 @@ def assign_node_ordering(self):
                 # Assign new value
                 n['node_number'] = counter
                 counter += 1
+#------------------------------------------------------------------------------#
+
+    def update_node_ordering(self, removed_number):
```
keileg: What happens if more than one node is removed?

src/porepy/grids/remove_grids.py
```diff
@@ -0,0 +1,72 @@
+from porepy.grids import grid_bucket
```
IvarStefansson: No, not really. I agree that that would make sense. Should I just move it?

src/porepy/numerics/fv/mpfa.py
```diff
@@ -64,84 +150,15 @@ def mpfa(g, k, bnd, faces=None, eta=0, inverter='numba'):
...
-
-        # Assemble the right hand side and solve
-        rhs = q + div * bound_flux * bound_vals
-        x = sps.linalg.spsolve(A, rhs)
-        f = flux * x - bound_flux * bound_vals
-
-    """
-
-    """
-    Method properties and implementation details.
```
IvarStefansson: No, that is just a mistake. Sorry, I'll reinsert it.

src/porepy/grids/grid_bucket.py
IvarStefansson: I have only removed one node at a time so far. But for large fracture sets I agree that it would be useful to be able to do multiple nodes at a time. I'll look into it.

keileg: Or, just leave it as an assumption, just for now. I lean towards this option.

PR #9, opened 2017-05-23
Changed bound_stress/flux sign

rbe051:
Changed the sign of boundary stress. We now assume the discretization does not care which part goes to the lhs and which part goes to the rhs. 

The sign is now defined from the following equation:
T = stress * cell_disp + bound_stress * bound_disp

Note that we now have to add a negative sign to the rhs when we solve elasticity/flow problem:
rhs = - div * bound_stress * bound_disp
d = np.linalg.solve((div*stress).A , rhs.A)


rbe051:
Merged wrong way

PR #18, opened 2017-06-01
Add fracture id to 2d fracture grids.

keileg:
We need something similar for 2d domains, and for Cartesian fractures.


keileg:
I will cover this now, and then merge.

PR #19, opened 2017-06-01
Config file

keileg:
Config file to replace passing around gmsh_path. Also added an option num_processors (for simple parallel processing, hopefully to be added in the future).

TODO: 1) Does this make sense?
2) If yes, purge gmsh_path from all examples.


keileg:
The approach works fine; so I will merge soon, but will wait a day or two.

Examples still need need to be reworked.

92runarlb:
I tested this a bit, and it works nicely. The documentation seems sufficient, if we targeting more advanced users.

PR #24, opened 2017-06-19
Added possibility in coupler to store several data dictionaries.

rbe051:
We need some way to store data for different solvers in the grid_bucket. E.g., if we want to solve elasticity and flow on the same grid bucket, the data keys "bc" and "bc_val" are the same for both problems.

As a solution to this problem, I suggest we add dictionaries to the data for each grid. The data for one node can then look like this:
data ={'node_number': 0, 'flow':{'bc', ...}, 'elasticity': {'bc',...}}

I'm not sure if this is the most elegant way to do this, so I'm open to other suggestions.

### DIFF
--- a/src/porepy/numerics/mixed_dim/coupler.py
+++ b/src/porepy/numerics/mixed_dim/coupler.py
@@ -11,7 +11,7 @@ class Coupler(object):
 
 #------------------------------------------------------------------------------#
 
-    def ndof(self, gb):
+    def ndof(self, gb, data_key=None):
         """
         Store in the grid bucket the number of degrees of freedom associated to
         the method.
@@ -22,14 +22,16 @@ class Coupler(object):
         gb: grid bucket.
 
         """
+        if data_key is None:
+            gb.add_node_prop('dof')
 
-        gb.add_node_prop('dof')
         for g, d in gb:
-            d['dof'] = self.solver.ndof(g)
+            k_d = self.__get_data(d, data_key)
+            k_d['dof'] = self.solver.ndof(g)
 
 #------------------------------------------------------------------------------#
 
-    def matrix_rhs(self, gb, matrix_format="csr"):
+    def matrix_rhs(self, gb, matrix_format="csr", data_key=None):
         """
         Return the matrix and righ-hand side for a suitable discretization, where
         a hierarchy of grids are considered. The matrices are stored in the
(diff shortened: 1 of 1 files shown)


keileg:
I understood little of your data_key (should it be 'flow' /'elasticity'), but some comments may help.

Wrt multi-physics / competing keywords, I agree with the need for something like this, and storing it in the bucket seems reasonable. Actually, I handled the same problem ad hoc earlier today, and that was less than beautiful. I am not sure about the nested dictionaries, but neither can I see better approaches right now. 

Is it an option to let the discretization classes decide which subdictionary to access? A case in point in the Biot discretization, which at certain points need both flow, mechanics, and then some extra information for the coupling fields. 

Not sure if this was useful, but we can discuss it tomorrow.

src/porepy/numerics/mixed_dim/coupler.py
```diff
@@ -101,13 +112,24 @@ def split(self, gb, key, values):
...
+
         for g, d in gb:
-            i = d['node_number']
-            d[key] = values[slice(dofs[i], dofs[i+1])]
+            k_data = self.__get_data(data, data_key)
+            i = data['node_number']
+            k_data[key] = values[slice(dofs[i], dofs[i+1])]
 
 #------------------------------------------------------------------------------#
+    def __get_data(self, data, data_key):
```
keileg: The idea is to either return the sub-dictionary, or the dictionary itself if data_key is not found? This should mean that for methods that do not need the subdict, we can just store the parameters in `data`, forget about keywords like `flow`, and everything will work as before?

If correct, this is elegant, but it can lead to errors that are hard to debug. For instance, if I misspell `flow`, or any of a number of other scenarios. I can't see better ways of doing it, but I suggest we take some precautions where possible, such as surrounding L131 with a try-except.

src/porepy/numerics/mixed_dim/coupler.py
```diff
@@ -11,7 +11,7 @@ def __init__(self, solver, coupling=None):
 
 #------------------------------------------------------------------------------#
 
-    def ndof(self, gb):
+    def ndof(self, gb, data_key=None):
```
keileg: It is not clear to me what is `data_key` and who access this? Is it the discretization who decides?

src/porepy/numerics/mixed_dim/coupler.py
keileg: @alessiofumagalli , input on this?

keileg: @IvarStefansson @alessiofumagalli : Conflicting keywords will also arise for flow and transport. Have you experience with this, and if so, how did you handled it.

rbe051: a short example of how this would be used: 
` gb = create_grid()
  # Assign parameters
    gb.add_node_props(['flow', 'elasticity'])
    for g, d in gb:
        if g.dim==3:
            elasticity_dict = _get_data_elasticity(g)
            d['elasticity'] = elasticity_dict
        flow_dict = _get_data_flow(g)
        d['flow'] = flow_dict
   # Choose and define the solvers and coupler
    solver_flow = mpfa.Mpfa()
    solver_elasticity = fvf.FracturedMpsa()

    coupling_conditions = tpfa_coupling.TpfaCoupling(solver_flow)
    solver_flow = coupler.Coupler(solver_flow, coupling_conditions)
    A, b = solver_flow.matrix_rhs(gb, data_key='flow')
`

So, the user is responsible to define and give the correct key-word. We could fix the subdict names for different solvers (e.g., flow, transport, mech), but this would loose a bit of the generality. What if I want to run two flow simulations on the same bucket, but with different fracture apertures? 

As you say, if we don't need subdict we can just drop it and store the data directly in the `data`

keileg: Perhaps it makes sense. But what do you do when you need more than one subdict within a solver, e.g. Biot? 

I'm also not convinced to tie this too tightly with the GridBucket, since the transition to a single grid should be seamless.

rbe051: I've not though about the case where you need more than one subdict within a solver. I guess you could have one key-word per subdict (`Biot().discretize(g, data, flow_key_word='flow', mech_key_word='mech', compr_key_work='compr')`), but it starts to become a bit cumbersome. 

The transition to a single grid is quite simple:
`
    g = gb.grids_of_dimension(3)[0]
    
    data = gb.graph.node[g]
    A_e, b_e = solver_elasticity.matrix_rhs(g, data['elasticity'])
    sol = sps.linalg.spsolve(A_e.copy(), b_e)
`

keileg:
I formally close this one. The functionality will instead be provided by a new Data class.

PR #25, opened 2017-06-20
Names

IvarStefansson:
New naming conventions:
f -> source
k -> perm
beta_n -> discharge
c -> stiffness
a -> aperture
Simplification of tensors.

I think I have changed (almost) all occurrences in the code, there are definitely some left in the documentation. I have not moved the couplings, did we agree on how to do this?


rbe051:
Why discharge and not flux?

keileg:
This should be seen in connection with #24 - we don't want to rewrite all data structures twice in a week. Runar and I have some ideas on how to approach #24, so let's discuss tomorrow.

IvarStefansson:
"Discharges" because "flux" is taken in the FV discretizations

keileg:
This is superseded by #28 The branch will be kept and extended.

PR #30, opened 2017-06-24
Move examples

keileg:
Changes:
1) All examples are moved from the library to examples.
2) The test script searches in examples as well, although discovery may not be optimal yet (__init__ is missing). 
3) We have a policy for testing of examples: If the script starts with test_, pytest should see it. See Readme in examples for details.
4) I believe gmsh is installed on travis during testing (sorry to anyone getting emails for all pushes, I did not make it on the first attempt). Thus we can run more general integration test, keeping in mind it should not take too long.

Tasks for everyone: Find examples you have made, and 
1) make them run
2) consider whether it should be made into a test. This may require that we include vtk in the dependencies, but there are some technical difficulties there (I cannot find vtk on pypi, only conda, and I am hesitant to jump ship to conda right now).


keileg:
Formally closed. Will squeeze commits into more appropriate chunks, and then merge into develop.

PR #32, opened 2017-06-28
Params

keileg:
Aparently working version of new Parameter class. Hopefully, all numerical methods have been dealt with.

Need for input / changes:
1) Run scripts to modify. In particular in examples.
2) Room for improvement in the Parameter class? I'm not completely sure about the way default parameters are treated, but I believe it is okay.


src/porepy/params/data.py
```diff
@@ -0,0 +1,510 @@
...
+            return default * np.ones(self._num_cells)
+
+        if isinstance(self._aperture, np.ndarray):
+            # Hope that the user did not initialize as array with wrong size
+            return self._apertures
+        else:
+            return self._apertures * np.ones(self._num_cells)
+
+    def _set_aperture(self, val):
+        if (isinstance(val, np.ndarray) and np.any(val < 0)) or val < 0:
```
rbe051: This does not work if you send inn an array with only positive values.

What is the reason for not doing: `if np.any(val < 0):` ? This seems to work just fine for me

alessiofumagalli:
Can you remind me why some parameters have a get functions and some others are accessed without?

src/porepy/params/data.py
keileg: I agree. Will fix.

keileg:
Get functions are used to access multi-physics fields. Should we have get for all? I am not opposed, it may reduce confusion, but also increase complexity.

alessiofumagalli:
when you say increase complexity, you mean we need to write more set and get functions?
also, can we make a running example with grid bucket and a running example with multi-physics?

rbe051:
I'm working on an example

alessiofumagalli:
also, is there a specific reason to keeping a dictionary as input with parameter class as the only object?

rbe051:
But it is not the only input. We also have node_number. And I guess we also want it to store, e.g., solution data?

alessiofumagalli:
I see, and now it makes even more sense to group physical parameters in a single class.

keileg:
Regarding get/set: Should we rename Parameters._get_aperture() to get_aperture() to get a consistent naming schemes? In this case, we can access the aperture either as an attribute (param.apertures), or by using the get method (right now, we can get it by the _get-method, but that is of course not the intention).

alessiofumagalli:
I agree to remove the underscore

keileg:
I am merging this one now.

PR #35, opened 2017-07-03
mcolon should return empty array instead of None

rbe051:
Previously:
```python
a = np.array([1, 2])
b = np.array([1, 2])
c = mcolon.mcolon(a, b)
assert c is None
```

Now: 
```python
a = np.array([1, 2])
b = np.array([1, 2])
c = mcolon.mcolon(a, b)
assert c.size==0
```

Any objections?


keileg:
Looks reasonable to me. I can't think of anywhere that will give immediate problems, so I vote for merge if the unit tests do not break.

PR #38, opened 2017-07-19
Re structuring solvers

rbe051:


src/porepy/grids/coarsening.py
```diff
@@ -128,15 +131,21 @@ def tpfa_matrix(g, perm=None, faces=None):
 
     """
     if perm is None:
-        perm = tensor.SecondOrder(g.dim,np.ones(g.num_cells))
+        perm = tensor.SecondOrder(g.dim, np.ones(g.num_cells))
```
IvarStefansson: Permeability tensor of dimension g.dim may be dangerous for tpfa if the grid is a lower-dimensional part of a grid bucket. In this case, something like gb.dim_max is preferable.

rbe051: Does a normal grid, say CartGrid, have g.dim_max?

IvarStefansson: No, don't think so. Also, in this particular case it doesn't matter, since unit permeability is assigned.

Issue #43, opened 2017-08-21
Missing faces (cells?) on fractures


rbe051:
I came over a case where a face in the 3D grid is missing. The face should be a fracture face. The face exist in the 2D grid (as a cell), however, the 2D cell does not have a mapping to any face in the 3D domain. I have checked that it is not any problem with the mapping, and indeed the 3D face does not exists. I assume this also means that the two cells that should be connected to the face (note that this is before we split the grid) are also missing. 

Either, we are not reading the output from gmsh improperly, or gmsh removes this cell for some reason...

```python
import numpy as np

from porepy.fracs import fractures, meshing
from porepy.viz.exporter import export_vtk


def create_fractures(data):
    data = np.atleast_2d(data)
    centers = data[:, 0:3]
    maj_ax = data[:, 3]
    min_ax = data[:, 4]
    maj_ax_ang = data[:, 5]
    strike_ang = data[:, 6]
    dip_ang = data[:, 7]
    if data.shape[1] == 9:
        num_points = data[:, 8]
    else:
        num_points = 16 * np.ones(data.shape[0])

    frac_list = []

    for i in range(maj_ax.shape[0]):
        frac_list.append(fractures.EllipticFracture(centers[i, :],
                                                    maj_ax[i],
                                                    min_ax[i],
                                                    maj_ax_ang[i],
                                                    strike_ang[i],
                                                    dip_ang[i],
                                                    num_points[i]))
    fracture_network = fractures.FractureNetwork(frac_list)
    return fracture_network


frac_list = np.array([[2000, 2000, 2000, 1500, 1500, 0, 100, 80],
                      [1800, 1800, 2150, 700, 700, 0, 170, 70],
                      [2650, 2200, 1700, 850, 850, 0, 130, 70],
                      [2100, 2100, 2000, 1000, 1000, 0, 60, 80]])

box = {'xmin': 0, 'ymin': 0, 'zmin': 0,
       'xmax': 4000, 'ymax':  4000, 'zmax': 5000}
f = create_fractures(frac_list)
g = meshing.simplex_grid(f, box)
g.assign_node_ordering()

export_vtk(g, 'temp')
`

rbe051:
This seems to have gotten worse with some update... A test for missing face_cells mapping has been added in the mesh-generation for the branch compressible_flow. All unit-tests were running fine after this. But then develop was merged into compressible_flow. After that, several tests fails due to missing face_cells mapping.

keileg:
[...]

PR #45, opened 2017-10-09
Compressible flow

rbe051:
Pull-request for Darcy-flow equation solver (note that the name of the branch is something else, as I continued working in an old branch as this branch contained the foundation of the Darcy solver). 

The branch is not completely ready to merge yet (missing e.g., documentation), but I wanted too get some feedback before continuing. 

See numerics.darcyEq for the suggestion for a darcy solver, and the tutorial folder for a jupyter example for how to use it.


rbe051:
Travis seems to fail due to issue #43

rbe051:
The tests that fails on travis seems to run fine on my computer. I tried both gmsh 2.8.3 and 2.16.0

Do the tests run fine on your computers?

src/porepy/numerics/darcyEq.py
```diff
@@ -0,0 +1,175 @@
...
+        self.parameters = {'file_name': physics}
+        self.parameters['folder_name'] = 'results'
+
+    def solve(self):
+        self.p = sps.linalg.spsolve(self.lhs, self.rhs)
+        return self.p
+
+    def reassemble(self):
+        """
+        reassemble matrices. This must be called between every time step to
```
keileg: Right hand side, yes, but there is no need to reassemble the discretization between time steps?

src/porepy/numerics/darcyEq.py
```diff
@@ -0,0 +1,175 @@
...
+        return self.lhs, self.rhs
+
+    def source_disc(self):
+        if isinstance(self.grid(), GridBucket):
+            source_discr = source.IntegralMultiDim(physics=self.physics)
+        else:
+            source_discr = source.Integral(physics=self.physics)
+        return source_discr
+
+    def flux_disc(self):
```
keileg: Here, more flexibility would of course be welcome. But I see this is a place where multiple physics (pressure and diffusion, say) may crash with the desire for many numerical methods.

src/porepy/numerics/darcyEq.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        lhs_flux, rhs_flux = self._discretize(self.flux_disc())
+        lhs_source, rhs_source = self._discretize(self.source_disc())
+        assert lhs_source.nnz == 0, 'Source lhs different from zero!'
+        self.lhs = lhs_flux
+        self.rhs = rhs_flux + rhs_source
+        return self.lhs, self.rhs
+
+    def source_disc(self):
+        if isinstance(self.grid(), GridBucket):
```
keileg: This will fail if the user gives a porepy.GridBucket (short hand namespace). I believe equivalent problems can be found in several other places. We should find a more general way of doing this, perhaps by always referring to the short hand version.

src/porepy/numerics/darcyEq.py
```diff
@@ -0,0 +1,175 @@
...
+            self._data['param'] = Parameters(self.grid())
+        self._data['param'].set_tensor(self.physics, self.permeability())
+        self._data['param'].set_porosity(self.porosity())
+        self._data['param'].set_bc(self.physics, self.bc())
+        self._data['param'].set_bc_val(self.physics, self.bc_val())
+        self._data['param'].set_source(self.physics, self.source())
+        self._data['param'].set_aperture(self.aperture())
+
+
+if __name__ == '__main__':
```
keileg: Do you plan to keep this, or is it testing?

src/porepy/numerics/darcyEq.py
IvarStefansson: I can think of cases where lhs reassembly is necessary. One example is if you want to use the same setup with different parameters, say change permeability in one fracture to study impact of connectivity

rbe051: Could add something like: 
def reasemblerhs(self):
or 
def reasemble(self, only_rhs=False)

src/porepy/numerics/darcyEq.py
rbe051: I discussed something similar with Ivar earlier this week. If the discretization of a single grid, e.g., tpfa.Tpfa took a tuple as input ( (grid, data) ) we would avoid this problem as the structure for calling mixed dim solvers and mono dim solvers would agree ( Tpfa( (g, data)), or TpfaMultiDim( gb)).

src/porepy/numerics/darcyEq.py
rbe051: What kind of flexibillity would you like to have?

src/porepy/numerics/darcyEq.py
rbe051: testing

rbe051:
I pulled the travis setup file from develop, and python 3.* works fine, however python 2.7 fails. I think it is because in one of the test I call unittest.__init__()

keileg:
Regarding unittest: Can you use setUp() instead of __init__? 

Possibly see also this https://stackoverflow.com/questions/17353213/init-for-unittest-testcase .

src/porepy/numerics/pde_solver.py
```diff
@@ -0,0 +1,237 @@
...
+
+from porepy.grids import structured
+from porepy.grids.grid_bucket import GridBucket
+from porepy.params.data import Parameters
+from porepy.params import tensor
+from porepy.params import bc
+from porepy.numerics.mixed_dim.solver import Solver
+
+
+class AbstractSolver(object):
```
keileg: This, and its children, are time stepping schemes, whereas Parabolic and Darcy are mainly concerned with spatial discretization. How do we combine the two? Is it possible?

src/porepy/numerics/darcy.py
```diff
@@ -0,0 +1,122 @@
...
+import scipy.sparse as sps
+
+from porepy.numerics.fv import tpfa, source, fvutils
+from porepy.grids.grid_bucket import GridBucket
+from porepy.params import bc, tensor
+from porepy.params.data import Parameters
+from porepy.viz.exporter import export_vtk
+
+
+class Darcy():
```
keileg: Why does this not inherit from parabolicSolver? Because of no time stepping?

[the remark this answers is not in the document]
rbe051: Yes, maybe this would be a good idea, but I would like to wait a bit to see how this works out before we implement this

rbe051: yes. 

You could imagine that it inherited from ParabolicProblem, and gave out a zero time-stepping lhs and rhs. But maybe it would be better to add a new class called Elliptic problem which it inherits from (or just call it Elliptic instead of Darcy)?

src/porepy/numerics/fv/source.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def matrix_rhs(self, g, data):
+        param = data['param']
+        sources = param.get_source(self)
+        lhs = sps.csc_matrix((g.num_cells, g.num_cells))
+        assert sources.size == g.num_cells, 'There should be one soure value for each cell'
+        return lhs, sources
+
+
+class Density(Solver):
```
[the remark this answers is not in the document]
rbe051: Density multiply the source by the cell volume, and Integral does not.

Lets say the you set your Parameter.set_source('flow', 1), and the cell volume is 0.5. 
Integral will then set a source of 1 while Density will 'integrate' 1 over the cell volume, so it sets a source of 0.5

src/porepy/numerics/fv/source.py
```diff
@@ -0,0 +1,78 @@
...
+        param = data['param']
+        sources = param.get_source(self)
+        lhs = sps.csc_matrix((g.num_cells, g.num_cells))
+        assert sources.size == g.num_cells, 'There should be one soure value for each cell'
+        rhs = sources * g.cell_volumes
+
+        return lhs, rhs
+
+
+class IntegralMultiDim(Solver):
```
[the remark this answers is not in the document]
rbe051: yes, but for now it follows the convention of TpfaMultiDim

src/porepy/numerics/pde_solver.py
rbe051: AbstractSolver takes a Parabolic as an input in the initialization

src/porepy/numerics/darcy.py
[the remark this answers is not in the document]
keileg: Then perhaps first add comments in the local function, and later migrate to superclass? Or do you think tutorial suffices?

keileg: Yes, call it elliptic instead of Darcy. Will inheritance make sense, e.g. avoid cut and paste?

src/porepy/numerics/fv/source.py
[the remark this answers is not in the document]
keileg: Do we really want both these features? I see nothing but confusion here - last time I fixed a bug related to densities vs integrated was literally ten minutes ago.

src/porepy/numerics/fv/source.py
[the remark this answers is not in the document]
keileg: @IvarStefansson We should fix this.

src/porepy/numerics/pde_solver.py
keileg: Perfect.

src/porepy/numerics/darcy.py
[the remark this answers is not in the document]
rbe051: I think the tutorial suffices, but the comments do not. But will add them anyway

PR #56, opened 2017-11-02
Method in coupler to obtain global degrees of freedom for a given grid

keileg:
Needed e.g. for linear solvers.

Also restructured the Coupler somewhat, fixed what seemed like a bug in TpfaMultiDim (I'm not sure why this did not show up before, there seems to be an infinite loop between the coupler and its solver), and added module level comments in


IvarStefansson:
Good! The grid-dof function is really useful. And I like the module level comments. The TPFAMultiDim dof function update applies to MPFAMultiDim as well, right? (and possibly elsewhere)
First thought on the "Java style", "in the Java sense" comments is that it might be good to avoid them:
1) Feels out of place (it's called porePY, after all).
2) Requires familiarity with Java to be informative.

keileg:
Java style is simply my way of referring to the concept of contracts, none
of the other languages I know have them. Porepy will forever stay free from
Java.

PR #63, opened 2017-11-08
Update class names

alessiofumagalli:
small updates to be conforming with __MixDim, add also IntegralDFN


alessiofumagalli:
fixing a problem with Python2.7, the commit is on the way

PR #64, opened 2017-11-09
Introduce the dual elliptic class and tests

alessiofumagalli:
- introduce the source for dual formulation problems
- remove the source term from the dual vem
- introduce the dual elliptic class
- introduce some test for the latter

### DIFF
--- a/src/porepy/numerics/elliptic.py
+++ b/src/porepy/numerics/elliptic.py
@@ -8,7 +8,7 @@ import numpy as np
 import scipy.sparse as sps
 
 from porepy.numerics.fv import tpfa, source, fvutils
-import porepy.numerics.vem as vem
+from porepy.numerics.vem import vem_dual, vem_source
 from porepy.grids.grid_bucket import GridBucket
 from porepy.params import bc, tensor
 from porepy.params.data import Parameters
@@ -64,7 +64,6 @@ class Elliptic():
         self.lhs = []
         self.rhs = []
         self.x = []
-        self.is_x_split = False
 
         self.parameters = {'file_name': physics}
         self.parameters['folder_name'] = 'results'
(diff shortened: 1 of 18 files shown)


src/porepy/numerics/elliptic.py
```diff
@@ -43,7 +44,7 @@ class Elliptic():
             Returns Flux discretization object (E.g., Tpfa)
     grid(): Returns: the Grid or GridBucket
     data(): Returns: Data dictionary
-    split(name): Assignes the pressure self.p to the data dictionary at each
+    split(name): Assignes the solution self.solution to the data dictionary at each
```
rbe051: should it be self.x and not self.solution ?

src/porepy/numerics/elliptic.py
```diff
@@ -77,53 +83,115 @@ def reassemble(self):
...
+            self.split(self.pressure_name)
+        else:
+            self._data[self.pressure_name] = self.x
+
+    def discharge(self, discharge_name='discharge'):
+        if self.is_GridBucket:
+            fvutils.compute_discharges(self.grid(), self.physics,\
+                                       self.pressure_name)
+        else:
+            pass
```
rbe051: I will fix this

src/porepy/numerics/elliptic.py
```diff
@@ -77,53 +83,115 @@ def reassemble(self):
...
+                                      self.projected_discharge_name)
+        else:
+            discharge = self._data[self.discharge_name]
+            projected_discharge = self._flux_disc.project_u(self._gb,
+                                                          discharge, self._data)
+            self._data[self.projected_discharge_name] = projected_discharge
 
+#------------------------------------------------------------------------------#
 
 class EllipticData():
```
rbe051: You don't need to change anything here?

src/porepy/numerics/elliptic.py
alessiofumagalli: yes, you're right.

src/porepy/numerics/elliptic.py
alessiofumagalli: well, nothing so far :)

PR #67, opened 2017-11-12
Speedup of export of 3d grids.

keileg:
Created a dedicated function to sort point on a face plane in 3d.
Speedup due to skipping of tests etc. Looks to be at least 70% decrease
in time.


keileg:
I believe the implementation is correct, and it surely is faster. Merging.

PR #68, opened 2017-11-12
Linsolve

keileg:
Tentative structure for iterative linear solvers for elliptic problems.

Much work is needed wrt tuning of parameters to get a reasonable solver in all cases, but for this, we need experience. In particular, I'm not sure about the maximum system size for direct solvers - it may be better to set this to infinite and let the user suffer the consequences for now.

Also note that for dual problems, we keep direct solver only for now.


keileg:
I will look into trouble with Travis tomorrow.

src/porepy/numerics/elliptic.py
```diff
@@ -71,8 +74,54 @@ def __init__(self, gb, data=None, physics='flow'):
...
+        # Discretize
+        tic = time.time()
+        print('Discretize')
+        self.lhs, self.rhs = self.reassemble()
+        print('Done. Elapsed time ' + str(time.time() - tic))
+
+        # Solve
+        tic = time.time()
+        print('Linear solver')
+        ls = LSFactory()
```
rbe051: Is there a reason to initialize LSFactory here or could it be done in the else statement?

src/porepy/numerics/elliptic.py
```diff
@@ -71,8 +74,54 @@ def __init__(self, gb, data=None, physics='flow'):
...
+
+        Returns:
+            np.array: Pressure state.
+
+        """
+
+        # Discretize
+        tic = time.time()
+        print('Discretize')
+        self.lhs, self.rhs = self.reassemble()
```
rbe051: We want to add way to solve without reasembling (And at some point I will change reasemble to assemble), but this branch is probably not the place

src/porepy/numerics/elliptic.py
```diff
@@ -141,6 +190,66 @@ def save(self, variables, save_every=None):
             variables = {k: self._data[k] for k in variables if k in self._data}
         export_vtk(self.grid(), file_name, variables, folder=folder_name)
 
+
+    ### Helper functions for linear solve below
+    def _setup_preconditioner(self):
```
rbe051: I assume you have some plan for the functions below, but you do realize this will never be called as you have commented it out?

src/porepy/numerics/linalg/linsolve.py
```diff
@@ -0,0 +1,237 @@
...
+
+@author: Eirik Keilegavlen
+"""
+import numpy as np
+import scipy.sparse.linalg as spl
+
+try:
+    import pyamg
+except ImportError:
+    " Could not import the pyamg package. pyamg sover will not be available."
```
rbe051: Typo, second sentence

keileg: Yes, they are part of the search for a decent solver. To be included, I hope, if not then purged later.

PR #70, opened 2017-11-13
Update exporter

alessiofumagalli:
Please try to merge this update and use it. 
There are two example in 
- examples/example1/benchmark1/test_vem.py
- examples/papers/arXiv:1711.01818/example_5_2_1.py
The tests fail because of the new interface is not updated everywhere. Once we finalise the interface I'll fix the other files and merge.
Ciao


rbe051:
I like this very much! I expect most of the tests that imports exporter and fail do in fact not need to import it at all

alessiofumagalli:
You're right Runar, probably we simply need to remove them.

alessiofumagalli:
I'm updating also the elliptic and parabolic and compressible. soon I'll send the new version, please try before accept the new version.

keileg:
Makes a lot of sense to me. 

I'm working on an update to the sort_points function, but that I can do independently.

alessiofumagalli:
the branch does not fail any more. I need another approval to merge it :)

PR #71, opened 2017-11-13
Update the vem source for the dfn case

alessiofumagalli:
very simple update to fix an error for the vem source term in dfn setting

### DIFF
--- a/src/porepy/numerics/vem/vem_source.py
+++ b/src/porepy/numerics/vem/vem_source.py
@@ -59,6 +59,13 @@ class IntegralMixDim(SolverMixDim):
 class IntegralDFN(SolverMixDim):
     def __init__(self, dim_max, physics='flow'):
         # NOTE: There is no flow along the intersections of the fractures.
+        # In this case a mixed solver is considered. We assume only two
+        # (contiguous) dimensions active. In the higher dimensional grid the
+        # physical problem is discretise with a vem solver and in the lower
+        # dimensional grid Lagrange multiplier are used to "glue" the velocity
+        # dof at the interface between the fractures.
+        # For this reason the matrix_rhs and ndof need to be carefully taking
+        # care.
 
         self.physics = physics
         self.dim_max = dim_max
--- a/src/porepy/numerics/vem/vem_source.py
+++ b/src/porepy/numerics/vem/vem_source.py
@@ -66,10 +66,19 @@ class IntegralDFN(SolverMixDim):
         self.discr = Integral(self.physics)
         self.coupling_conditions = None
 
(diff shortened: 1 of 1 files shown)


alessiofumagalli:
Since it's a branch-to-fix-something, I'll merge it in a couple of hours if I don't get any feedback

src/porepy/numerics/vem/vem_source.py
```diff
@@ -66,19 +66,28 @@ def __init__(self, dim_max, physics='flow'):
...
         self.coupling_conditions = None
 
-        kwargs = {"discr_ndof": self.discr.ndof,
+        kwargs = {"discr_ndof": self.__ndof__,
                   "discr_fct": self.__matrix_rhs__}
         self.solver = Coupler(coupling = None, **kwargs)
         SolverMixDim.__init__(self)
 
+    def __ndof__(self, g):
+        # The highest dimensional problem has the standard number of dof
```
keileg: Why would this be different in different dimensions? I don't see it, so at least a comment is due.

alessiofumagalli: it's again the fact of dfn source term. in the intersection we have only num_cells dof while in the fractures num_cells+num_faces. it's because the equations are not uniforms on each dimension

alessiofumagalli: it's the same concept of __matrix__rhs

keileg:
Then I vote for specification + making this an option. If you just read
this part of the code, it looks like you're working with a heterogeneous
discretizaiton. cf our discussion last week.

man. 13. nov. 2017 kl. 14:03 skrev Alessio Fumagalli <
notifications@github.com>:

alessiofumagalli:
OK, I can improve the comments. Do you mean make the full dfn setting as optional with a single flag? Is it possible but not the scope of this branch, I guess :)

keileg:
I think it's the same comment as last week regarding dfn that was not dfn.
If the function is related to a specific, and non-standard, modeling choice
this ought to be reflected in function name and comments. Does that make
sense?

alessiofumagalli:
I agree, we need to find a name convention able to handle these cases.. But I'm very bad doing these mix(ing) :P

alessiofumagalli:
I'll merge the branch keeping in mind your comments

keileg:
@alessiofumagalli What you are doing is in essence to merge two grids with a full continuity condition, and without a lower-dimensional grid as a buffer inbetween. We can imagine doing the same with two 3d grids (don't know how to do this in practice right now, but that's a minor issue) - and then gluing them together with mortars. So, when the mortars are more or less clear, we should revisit this and see if we can make more sense of the code in that way.

PR #86, opened 2017-11-23
Various updates mainly to Model classes.

keileg:
* Changed names of solvers: ParabolicProblem-> ParabolicModel etc.
* Added data so that we can solve a heat equation with reasonable parameter names.
* Various minor changes when preparing PorePy paper.


keileg:
Well, this was fun. Thanks for unit tests, everyone. 

Can I do pde_solver -> time_stepper as well?

rbe051:
pde_solver -> time_stepper er bra

keileg:
Then I have no more planned changes, so please have a look.

src/porepy/params/data.py
```diff
@@ -131,6 +131,56 @@ def set_fluid_compr(self, val):
         self._fluid_compr = val
     fluid_compr = property(get_fluid_compr, set_fluid_compr)
 
+#----------- Fluid density - treated as constant for now
+    def get_fluid_density(self):
+        if hasattr(self, '_fluid_density'):
+            return self._fluid_density
+        else:
+            return 1
```
rbe051: Why not make it a cellwise quantity? At some point in the near future it will be changed anyway.

keileg:
I'm not sure if transport will be conservative in that case. Otherwise I
agree.

rbe051:
If you have a compressible flow it will not be conservative _unless_ you evaluate the densities at cell values (since div(v) != 0). 

Also, you might wish to use the fluid density for something else than transport. Anyway, I think this should be the users problem

PR #93, opened 2017-12-02
Dfn y isect

keileg:
Several changes related to fracture extrusion, merging of 1D grids and dfn meshing.


keileg:
I think we can go with this now, but let me know your opinion, @alessiofumagalli

alessiofumagalli:
I see that the branch fails in many places.. we should go through and check why, I try to have a look

alessiofumagalli:
the problem has been fixed and travis agrees. I merge it

PR #94, opened 2017-12-05
Dbaroli/docker integration

DavideBaroliUniLu:
I regenerate the Dockerfile and create a readme as you request.
Could you check the new deployment status (it takes time since I have base planning on dockerhub).
The image is pushing on dockerhub named "pmgbergen/porepy:latest".
The dimension is 3.5 GB. It is comprehensive of all dev dependencies, vtk and Gmsh. 
If you want to change the GMSH version you can easily do it " the line 59 of Dockerfile.
ENV GMSH_VERSION 2.11.0 " 
The system packages are comprehensive of the opencascade library so that if you upgrade gmsh then you should not require any extra dep. 
If you need further information or you need a rebase of this branch  (before pushing) let me know.
If you need to use jupyter, you should add in pip list in Dockerfile  (add in line 92 of Dockerfile) [ Maybe we want a different custom image for demonstration]. 
If you have in the container jupyter  and you want to use jupyter through it. In Linux, you need to do:
docker run --name notebook -w /home/porepy -v $(pwd):/home/porepy/shared -d -p 127.0.0.1:8888:8888 docker.io/pmgbergen/porepy:latest 'jupyter-notebook --ip=0.0.0.0'
For other cases, similarly, you could look at fenicsproject [http://fenics.readthedocs.io/projects/containers/en/latest/jupyter.html]


keileg:
This looks good to me. Are you confident it will work, or should we get someone to try it first?

The readme was good, but I think I will move the detailed explanation to a separate file, and then link from the main file. We try to keep the central Readme simple, so that it easy to get an overview.

DavideBaroliUniLu:
@keileg  the deployment at docker hub is successful. We can modify travis: pulling the image and the run the test suite. 
Under section script of travis you can add  (Travis use docker only with linux env).
` if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
        docker run -u root docker.io/docker pull pmgbergen/porepy:latest /bin/sh -c "cd /home/porepy/porepy-src; python setup.py test";
      fi `

DavideBaroliUniLu:
Running the unit test suite in the container i get just this assertion false:
Output: 1 failed, 483 passed, 16 warnings in 144.75 seconds
 AssertionError: assert False
`        +  where False = <function allclose at 0x7fd56fffe0d0>(matrix([[ -2.00000000e+00,   0.00000000e+00,   2.00000000e+00,\n          -5.55111512e-16,   0.00000000e+00],\n        [...000e+00],\n        [  0.00000000e+00,  -4.99600361e-16,   2.00000000e+00,\n           0.00000000e+00,  -2.00000000e+00]]), array([[ -2.00000000e+00,   0.00000000e+00,   2.00000000e+00,\n          0.00000000e+00,   0.00000000e+00],\n       [  0...00000e+00],\n       [  0.00000000e+00,   2.22044605e-16,   2.00000000e+00,\n          0.00000000e+00,  -2.00000000e+00]]), 1e-15, 1e-15)
E        +    where <function allclose at 0x7fd56fffe0d0> = np.allclose `
The test with the expection is:
test/unit/test_hybrid_vem.py line 190  AssertionError

keileg:
@alessiofumagalli could you have a look?

DavideBaroliUniLu:
I am performing a small change to allow user to add custom library to conda env. When one has such docker container and wants to costumize it. It is possible to commit it locally and also to push on dockerhub. Please could you provide me you name of account on dockerhub ( email: davide.baroli@uni.lu).

alessiofumagalli:
is there a way I can run it on my laptop, so I can check why the test fails.
thanks

DavideBaroliUniLu:
This is M computed : [[-3.4976884   3.4976884   0.          0.        ]
 [ 3.4976884  -7.93596501  4.43827662  0.        ]
 [ 0.          4.43827662 -9.65880718  5.22053056]
 [ 0.          0.          5.22053056 -5.22053056]]
I use `import pdbpp; pdbpp.set_trace()`

keileg:
@alessiofumagalli What is the status on this one. Can we merge when I have modified the readme?

PR #95, opened 2017-12-06
Enumath porepy

alessiofumagalli:
- a couple of examples for the ENUMATH porepy paper
- import filter for fractures defined through a csv file


IvarStefansson:
I wonder whether it would make sense to make use of the EllipticDataAssigner class?

alessiofumagalli:
can you point an example?

alessiofumagalli:
@IvarStefansson , I've used the EllipticDataAssigner class. Let me know. Thanks

src/porepy/fracs/non_conforming.py
```diff
@@ -473,12 +473,25 @@ def update_cell_faces(g, delete_faces, new_faces, in_combined, fn_orig,
     # relations
     if in_combined[0] < in_combined[-1]:
         for i in range(deleted_2_new_faces.size):
-            deleted_2_new_faces[i] = new_faces[np.arange(in_combined[i],
-                                                         in_combined[i+1])]
+            if in_combined[i] == in_combined[i+1]:
```
keileg: Please add comment on what happens here. It is difficult enough, without having to guess what we were thinking months before.

src/porepy/fracs/importer.py
```diff
@@ -11,6 +12,56 @@
 
 #------------------------------------------------------------------------------#
 
+def dfm_from_csv(file_name, tol=1e-4, **mesh_kwargs):
```
keileg: Naming is critical here. First, I suppose 3d should be somewhere in the name, and that the existing function mesh_from_csv should have 2d in it. Or we want a wrapper function which splits into 2D and 3d, but maybe not.

Then dfm: I know you want to differ from dfn, and I support it. My initial thought is, if specifications are needed, it is dfn that is the special case.

Also, can you split the function into two:  1) Read csv and return FractureNetwork. 2) Given the network, create mesh by calling upon meshing.simplex_grid? Then it will be easy to extend this to dfn if desirable. 

BTW, you don't need to impose boundaries + find and split intersections, this is being taken care of by the meshing functions already.

src/porepy/fracs/non_conforming.py
```diff
@@ -554,6 +567,10 @@ def update_cell_faces(g, delete_faces, new_faces, in_combined, fn_orig,
         # Note use of original coordinates here.
         ci_coord = node_coord_orig[:, fn_orig[:, ci]]
         # Coordinates of the nodes of the first new face
+        if new_faces_loc.size == 0:
+            print(deleted_2_new_faces, cf_2_f[cf[i]])
```
keileg: Is this intentional?

alessiofumagalli: it is not, removed

src/porepy/fracs/importer.py
alessiofumagalli: I changed some of the names to be more consistent.
I split the function indicated in two

I don't see where the comment in BTW is in the code. Can you be more specific?

src/porepy/fracs/non_conforming.py
alessiofumagalli: I already forgot :(

src/porepy/fracs/simplex.py
```diff
@@ -143,7 +143,7 @@ def tetrahedral_grid(fracs=None, box=None, network=None, subdomains=[], **kwargs
 
     # Invert phys_names dictionary to map from physical tags to corresponding
     # physical names
-    phys_names = {v[0]: k for k, v in phys_names.items()}
+    phys_names = {v: k for k, v in phys_names.items()}
```
keileg: You need to update to meshio 1.10, then the code will work as it was.

keileg:
Your last commit is not necessary (actually it is wrong), but you need to update to the latest version of meshio (1.10). They did a change to the way physical names from gmsh is treated; this causes your problem.

keileg:
It seems we're in good shape, except from the last two commits. I will force-push a retraction of those, and then merge.

PR #106, opened 2017-12-19
Fractured mpsa

rbe051:
Low pressure stimulation


src/porepy/numerics/parabolic.py
```diff
@@ -64,30 +65,41 @@ def space_disc(self):
     problem.solve()
     '''
 
-    def __init__(self, gb, physics='transport', **kwargs):
+    def __init__(self, gb, physics='transport',time_step=1.0, end_time=1.0, **kwargs):
```
rbe051: I added time_step and end_time to the init.
This way the number of cases where you have to overload Parabolic is reduced drastically

src/porepy/numerics/parabolic.py
```diff
@@ -64,30 +65,41 @@ def space_disc(self):
...
         self.physics = physics
-        self._data = dict()
+        self._data = kwargs.get('data', dict())
+        self._time_step = time_step
+        self._end_time = end_time
         self._set_data()
+
         self._solver = self.solver()
-        self._solver.parameters['store_results'] = True
+        self._solver.parameters['store_results'] = False
```
rbe051: Can we remove this parameter variable?

src/porepy/numerics/parabolic.py
```diff
@@ -103,7 +115,11 @@ def step(self):
     def update(self, t):
         'Update parameters to time t'
         for g, d in self.grid():
-            d['problem'].update(t)
+            d[self.physics + '_data'].update(t)
+
```
rbe051: The data assigner should now be stored with keyword, e.g., 'flow_data'. I think this is quite consistent with the physics consept used in the parameter class

src/porepy/params/bc.py
```diff
@@ -62,8 +62,15 @@ def __init__(self, g, faces=None, cond=None):
...
+            if faces.dtype==bool:
+                if faces.size != self.num_faces:
+                    raise ValueError('''When giving logical faces, the size of
+                                        array must match number of faces''')
+                faces = np.argwhere(faces)
             if not np.all(np.in1d(faces, bf)):
                 raise ValueError('Give boundary condition only on the boundary')
+            if isinstance(cond, str):
+                cond = [cond] * faces.size
             if faces.size != len(cond):
```
rbe051: first case: bc accepts boolean array
second case: if a string is given, it assumes this is the boundary condition assigned to all given faces.

Ok?

src/porepy/numerics/parabolic.py
keileg: Fair enough.

src/porepy/numerics/parabolic.py
keileg: We don't use it, do we? If not, then clearly.

src/porepy/numerics/parabolic.py
keileg: Good.

src/porepy/params/bc.py
keileg: Yes, I've been wanting to do this.

keileg:
There are some issues with meshio that I hope to fix soon, so no merging until then. I will take care of it when we are done.

keileg:
Travis is failing; my guess is it has to do with compute_discharge(). @rbe051 could you have a look?

rbe051:
I think commit 3ba07ae should fix this. I will merge if tests pass

PR #110, opened 2018-01-03
if variable not found in grid, skip!

rbe051:
I sometimes have variables that only live on some grids (e.g. mechanical displacement only exist in 3D).
  
I'm not sure if we want to do this change. What do you think?

### DIFF
--- a/src/porepy/viz/exporter.py
+++ b/src/porepy/viz/exporter.py
@@ -302,6 +302,8 @@ class Exporter():
 
         if data is not None:
             for name_field, values_field in data.items():
+                if values_field is None:
+                    continue
                 dataVTK = ns.numpy_to_vtk(values_field.ravel(order='F'),
                                           deep=True, array_type=vtk.VTK_DOUBLE)
                 dataVTK.SetName(str(name_field))


src/porepy/viz/exporter.py
```diff
@@ -302,6 +302,8 @@ def _write_vtk(self, data, name, g_VTK):
 
         if data is not None:
             for name_field, values_field in data.items():
+                if values_field is None:
+                    continue
                 dataVTK = ns.numpy_to_vtk(values_field.ravel(order='F'),
```
rbe051: Do you see any problem by not write a variable to all dimensions?

One problem I see is that you now will not get an error message if you try to export a non-existing variable

keileg:
Is the variable added to the GridBucket, but is None on some grids? If so, we can add a check if the node_props exists (or something, @alessiofumagalli is morally responsible for function names in that class). I would like to get an error message if I miss on a variable name.

rbe051:
It should work for something like this: 
```python
gb.add_node_props(['pressure'])
exporter.write_vtk(['pressure'])
```
I think it still should give an error if pressure is not added to the data, but I can check tomorrow

rbe051:
As I though: 
```python
gb.meshing.simplex_grid(...)
exporter.write_vtk(['pressure'])
```
gives an error, but 
As I though: 
```python
gb.meshing.simplex_grid(...)
gb.add_node_props(['pressure'])
exporter.write_vtk(['pressure'])
```
works. If there are no further objections to this, I will merge this to develop

keileg:
Go ahead.

tor. 4. jan. 2018 kl. 14:23 skrev rbe051 <notifications@github.com>:

PR #120, opened 2018-02-16
Mechanics bc

keileg:
I need the option for rolling boundary conditions in MPSA. 

@mstarnoni Can you please check bc.py that I did not mess up?

@rbe051 , @IvarStefansson : I messed up the merge a little but, partly my own fault, partly something that I don't understand (still my fault, though). The troublesome modules were bc (has been expanded, I lost part of the functionality in develop for a while), mpsa (functions discretize and matrix_rhs), and fvutils (compute_discharges). I'm pretty sure everything is back again, but please verify.


mstarnoni:
bc.py seems ok

rbe051:
Looks ok to me. Am I correct to believe that this only allows for rolling boundary conditions along the coordinate axis?

src/porepy/numerics/fv/mpsa.py
```diff
@@ -41,7 +41,7 @@ def ndof(self, g):
 
 #------------------------------------------------------------------------------#
 
-    def matrix_rhs(self, g, data, discretize=True):
+    def matrix_rhs(self, g, data):
```
rbe051: Why is discretize removed ? will this file even run now?

rbe051:
There is something wrong with vectorial bc. It calls g.get_boundary_faces() which does not exist anymore.

VectorialBc also miss unittests which should be added before a merge

mstarnoni:
@rbe051 you're right, it only covers that case by now.

Issue #92, opened 2017-12-02
Bug in fracs.split_grid()


keileg:
For certain fracture networks, an error occurs in the function duplicate_nodes(), around L480. The symptom 
is indexing with an index iv that is too long compare to the array length. 

It seems the bug was introduced in d34fff4, or one of the preceding updates.

I have not managed to recreate the error on a small network (<40 fractures 3D), let me know when you need a case.

rbe051:
Reverted to an older version of split_grid. Let me know if this works

keileg:
I ran into trouble in split_nodes. Apparently, I end up with a node that has no cells (the actual error message comes from __extract_submatrix which cannot handle empty matrices). Runscript for debugging to come.

keileg:
Also, my impression is that the memory usage was quite heavy. The simulation that crashed had a peak memory (in split_fractures()) of at least 8GB for a mesh with ~250.000 fractures. I'm not sure if this is unavoidable, but it is a lot.

rbe051:
This is related to the speed issue. The old version of split_grid we reverted to uses a lot of memory (thus is also slow), due to the slicing of the matrices. When I find the bug in the optimized split_grid, I will cut the memory need considerably.

keileg:
I tried again, without detecting the issue. This thing feels like a ghost.

I will close this now, and reopen if trouble reappears.

rbe051:
Do we want to merge the optimized split grid back into develop?

keileg:
No, the bug is back again. I will keep you posted.

keileg:
If you consider the notebook just pushed to branch flow_transport_pp, path examples/papers/arxiv171000460 , there is a create_grid function there. When I ran that function in optimize split grid, I got into trouble.

Also note you may need to use updated version of gmsh (3.06 rather than 2.11) to create grid. I will change install instructions.

rbe051:
Everything works fine for me......

keileg:
Interesting. I will look into it. Sigh.

rbe051:
Are you feeling brave enough to try to merge the optimization back in? It would be nice to cut the gridding time with an order of magnitude

keileg:
This was the strange behavior in splitting of the fractures, that only
showed up occasionally on my computer? If so, let's give it a try.

Eirik

rbe051:
Ok, I will test the branch again on my computer, and if it works, I will merge it

rbe051:
[...]

PR #121, opened 2018-02-19
optimizing split_grid module

rbe051:
new attempt at merging split_grid optimization. Hopefully the ghost has vanished


rbe051:
Fixes Issue #92

PR #122, opened 2018-03-12
Update dfn comparison examples

alessiofumagalli:
Finalise the updates for using DFN module. Update the tests for the relative paper.


keileg:
Looks okay to me. We may want to revisit file formats when we eventually revise the importer, but that is not for today.

Feel free to merge when you are finished.

PR #126, opened 2018-03-27
Exporter update

alessiofumagalli:
update the exported such that each dimension is saved in a single file


alessiofumagalli:
I need to fix the test_vtk in case vtk is not present in the installation

alessiofumagalli:
please check that on your installation the tests works

PR #129, opened 2018-05-04
Update develop

alessiofumagalli:
small updates for the develop branch


keileg:
I think you are the only one who have used this function, so if you are happy with it, go ahead and merge.

PR #134, opened 2018-07-02
Travis test for python 3.7

keileg:


keileg:
This is lagged far beyond, so I close it, and return to the issue in a new patch.

PR #152, opened 2018-09-11
Robin bc

rbe051:
Implemented Robin conditions for the MPFA and MPSA discretization. In the process I also found the bug in MPSA which gave sub-optimal convergence rates for the MPSA with Neumann. The issue was that at the interior faces the asymmetric part of the stress tensor cancel by construction, so it was just thrown away in the code. At the Neumann boundaries we then have two options: 
1: Include the asymmetric part for the discretization of these faces
2: Also throw away the asymmetric part of the given traction, or force the user to only give asymmetric part (I don't like this option too much)

Therefore, I ended up implementing option 3, and included the asymmetric part for all faces (I'm lazy). 
Apparently this breaks down in some cases (some of the unittests fails) because the local system becomes singular. I am not sure if this is because we included the asymmetric part to the interior faces (which would be strange since these should cancel), or because they are included to the exterior faces.

Also, I am afraid the robin implementation increases the memory use by 1/3, as we have to create all the local matrices also for the Robin condition. I guess if we are a bit clever we can mitigate this issue.

Another task that is left before a merge is to implement robin for vector bc (should be trivial?) and possible for TPFA also.


src/porepy/numerics/fv/mpfa.py
```diff
@@ -549,35 +554,52 @@ def _mpfa_local(g, k, bnd, eta=None, inverter="numba", apertures=None):
...
         (sgn[0], (subcell_topology.subfno, subcell_topology.cno))
     ).tocsr()
     # The cell centers give zero contribution to flux continuity
     nk_cell = sps.coo_matrix(
         (np.zeros(1), (np.zeros(1), np.zeros(1))),
         shape=(subcell_topology.num_subfno, subcell_topology.num_cno),
     ).tocsr()
-    del sgn
+
+    # For the Robin codition the distance from the cell centers to face centers
```
keileg: condition

src/porepy/numerics/fv/mpfa.py
```diff
@@ -598,18 +620,25 @@ def _mpfa_local(g, k, bnd, eta=None, inverter="numba", apertures=None):
     bound_exclusion = fvutils.ExcludeBoundaries(subcell_topology, bnd, g.dim)
 
     # No flux conditions for Dirichlet boundary faces
```
keileg: Does this also mean we assign Robin conditions on the rows that represent Neumann conditions?

Update comment, please.

src/porepy/numerics/fv/mpfa.py
```diff
@@ -598,18 +620,25 @@ def _mpfa_local(g, k, bnd, eta=None, inverter="numba", apertures=None):
...
-    nk_grad = bound_exclusion.exclude_dirichlet(nk_grad)
-    nk_cell = bound_exclusion.exclude_dirichlet(nk_cell)
-    # No pressure condition for Neumann boundary faces
-    pr_cont_grad_all = pr_cont_grad
-    pr_cont_grad = bound_exclusion.exclude_neumann(pr_cont_grad)
-    pr_cont_cell = bound_exclusion.exclude_neumann(pr_cont_cell)
+    nk_grad_n = bound_exclusion.exclude_rob_dir(nk_grad_all)
```
keileg: I think somewhere we need 10 lines of documentation of the logic in the implementation

src/porepy/numerics/fv/mpsa.py
```diff
@@ -138,6 +138,181 @@ def rhs(self, g, bound_stress, bc_val, f):
 # ------------------------------------------------------------------------------#
 
 
+class RobinMpsa(Mpsa):
+    """
+    Subclass of MPSA for discretizing a domain with robin boundary conditinos.
```
keileg: conditions. 
Robin (capital R)

src/porepy/numerics/fv/mpsa.py
```diff
@@ -138,6 +138,181 @@ def rhs(self, g, bound_stress, bc_val, f):
...
+                (np.ones(rob_ind.size), (np.arange(rob_ind.size), rob_ind)),
+                (rob_ind.size, g.num_faces * g.dim),
+            )
+            -sgn_rob * bound_stress_external[rob_ind]
+        )
+
+        b_matrix = sps.vstack((d_b, d_t), format="csr")
+
+        data["b_e"] = b_matrix
+        data["A_e"] = A
```
keileg: Does not this conflict with the operator way of thinking about the discretization? Can we derive a 'stress' etc? If not, then a better explanation of the method is necessary.

It may be that what we really need is a more nuanced view of variables internal (or on the boundary) to a grid. We usually think that there is a single variable on all cells / faces / .., but what you have here is a cell centered displacement on all cells, combined with displacement on certain faces. I'm not sure how to include such options in a neat way. Suggestions?

src/porepy/numerics/fv/mpsa.py
```diff
@@ -528,7 +703,7 @@ def given_slip_distance(self, g, stress, bound_stress, faces=None):
 # ------------------------------------------------------------------------------#
 
 
-def mpsa(g, constit, bound, eta=None, inverter=None, max_memory=None, **kwargs):
+def mpsa(g, constit, bound, eta=None, alpha=None, inverter=None, max_memory=None, **kwargs):
```
keileg: Same comment on alpha as in MPFA, a more descriptive name would be much better.

src/porepy/numerics/fv/fvutils.py
```diff
@@ -689,8 +689,42 @@ def __init__(self, subcell_topology, bound, nd):
                 shape=(row_dir.size, num_subfno),
             ).tocsr()
 
-        elif self.bc_type == "vectorial":
+            col_dir = np.argwhere([not it for it in bound.is_rob[fno]])
```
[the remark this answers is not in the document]
rbe051: I think this must be a mistake. I did not do anything with vectorial.

src/porepy/numerics/fv/fvutils.py
```diff
@@ -689,8 +689,42 @@ def __init__(self, subcell_topology, bound, nd):
...
+            ).tocsr()
+
+            col_rob = np.argwhere([it for it in  bound.is_rob[fno]])
+            row_rob = np.arange(col_rob.size)
+            self._keep_robin = sps.coo_matrix(
+                (np.ones(row_rob.size), (row_rob, col_rob.ravel("C"))),
+                shape=(row_rob.size, num_subfno),
+            ).tocsr()
+            
+        elif self.bc_type == "vectorial":
```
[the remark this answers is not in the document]
rbe051: nope

src/porepy/numerics/fv/fvutils.py
```diff
@@ -940,6 +995,138 @@ def exclude_neumann_nd(self, other):
...
+            exclude_neu_dir_nd = sps.kron(sps.eye(self.nd), self._exclude_rob_dir)
+
+        elif self.bc_type == "vectorial":
+            raise NotImplementedError()
+
+        return exclude_neu_dir_nd * other
+
+    def keep_robin_nd(self, other):
+        """ Keep Roben conditions for vector equations (elasticity).
+        See above method without _nd suffix for description.
```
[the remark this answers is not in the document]
rbe051: Its probably below

src/porepy/numerics/fv/mpsa.py
rbe051: This is a implementation of Robin condition on a face level. This implementation does not know about the subfaces. I guess this class can be left out of this pull-request

src/porepy/numerics/fv/fvutils.py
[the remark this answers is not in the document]
rbe051: oh, the git diff made me confused. Is there anything to comment? It is the exactly same thing as above. Just replaced is_neu or is_dir with is_rob

src/porepy/numerics/fv/mpfa.py
rbe051: I guess the rough idea is given for the mpfa discretization, but the discretization is messy, so it would not be bad to add some more

keileg:
Also see input on code quality from Codacy, https://app.codacy.com/app/keileg/porepy/pullRequest?prid=2171303

keileg:
I'd say leave it for now. The point should be to improve the quality of the code that we're actually using; unit tests are just a single purpose tool. And, as you point out, it will take ages to fix these. The same to some degree applies to runscripts.

src/porepy/numerics/fv/mpsa.py
```diff
@@ -945,38 +969,54 @@ def mpsa_elasticity(g, constit, subcell_topology, bound_exclusion, eta, inverter
     hook_normal = sps.coo_matrix(
         (np.ones(ind_f.size), (np.arange(ind_f.size), ind_f)),
         shape=(ind_f.size, ind_f.size),
-    ) * (ncsym + ncasym)
+    ) * (ncsym_all + ncasym)
```
rbe051: when adding the asymmetric part of the stress we sometimes get singular matrices in the local system. It seems to happen when all subfaces of a node has a Neumann conditions (but if this is the only time it happens I don't know). 

@keileg I am adding the asymmetric part right here. Is this wrong? Should there be a transpose? a factor 1/2? I tried to compare it with the MPSA paper, but it is hard to understand which terms in the paper corresponds to which terms in the code.

src/porepy/numerics/fv/mpsa.py
```diff
@@ -945,38 +969,54 @@ def mpsa_elasticity(g, constit, subcell_topology, bound_exclusion, eta, inverter
...
     del ind_f
     # The final expression of Hook's law will involve deformation gradients
     # on one side of the faces only; eliminate the other one.
     # Note that this must be done before we can pair forces from the two
     # sides of the faces.
-    hook = __unique_hooks_law(ncsym, ncasym, subcell_topology, nd)
+    hook = __unique_hooks_law(ncsym_all, ncasym, subcell_topology, nd)
 
+    # Pair the forces from each side
+    ncsym_all = subcell_topology.pair_over_subfaces_nd(ncsym_all + ncasym)
```
rbe051: when adding the asymmetric part of the stress we sometimes get singular matrices in the local system. It seems to happen when all subfaces of a node has a Neumann conditions (but if this is the only time it happens I don't know). 

@keileg I am adding the asymmetric part right here. Is this wrong? Should there be a transpose? a factor 1/2? I tried to compare it with the MPSA paper, but it is hard to understand which terms in the paper corresponds to which terms in the code. Comparing with line 972 the term ncsym_all + ncasym is correct, but it drives my crazy trying to figure out why we get singular local systems

src/porepy/numerics/fv/fvutils.py
```diff
@@ -671,94 +671,98 @@ def __init__(self, subcell_topology, bound, nd):
         fno = subcell_topology.fno_unique
         num_subfno = subcell_topology.num_subfno_unique
 
+        self.fno = fno
```
rbe051: This whole class (ExcludeBoundaries) is a monster! I did manage to clean up the __init__, however it still have 14 (!) methods that basically does the same. We are apparently doing something wrong with the boundary conditions

keileg: I agree with both. It started out as a simple concept, but then it snowballed. Feel free to create an issue on it.

PR #156, opened 2018-09-13
unittest for vectorial boundary conditions

keileg:
Old branch by @mstarnoni


keileg:
@mstarnoni It actually was useful, so I merged it into develop.

PR #157, opened 2018-09-17
Robin vectorial

rbe051:
Added Robin conditions to create_rhs_nd in MPSA

@mstarnoni, I rewrote most of the create_rhs_nd function in MPSA. 
For Neumann and Dirichlet conditions (including rolling) I believe it should give the same result as before my changes, but I will wait with the merge untill you approve


keileg:
Looks good to me. Even Codacy is positive, I believe that is a first!

I presume it is the black formatting that gives all the format changes? Hopefully, this will leave a lighter trace as we have accessed most of the files.

rbe051:
Yes, it is black. I think I have run black on all the files now, so hopefully this will help

PR #158, opened 2018-09-17
Codacy fix

rbe051:
Fixed all assertions for Unittests so codacy do not complain

### DIFF
--- a/test/integration/test_TpfaMonoCoupling.py
+++ b/test/integration/test_TpfaMonoCoupling.py
@@ -43,7 +43,7 @@ class TestTpfaCouplingDiffGrids(unittest.TestCase):
 
         # test pressure
         for g, d in gb:
-            assert np.allclose(d["pressure"], xmax - g.cell_centers[0])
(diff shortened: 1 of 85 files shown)


keileg:
Great! Although Codacy is ungrateful.

keileg:
Any idea why coverage declined?

rbe051:
Good question.. I will invesstigate

rbe051:
It is the test that I removed that did no testing. I removed it because it had no asserts

keileg:
How can that decrease effective coverage with 0.5%?

rbe051:
I don't know how coveralls calculate this percentage. But if you look at the previous commit ffc2bfc it passed the coveralls.

It is strange that it drops this much, since the removed test does the same as the one I kept. 
However, I can't figure out in coveralls where the percentage drops. Should it not say so somewhere?

rbe051:
I think there must have been a bug in Codacy. Now everything is fine. Closing pull-request

PR #169, opened 2018-10-17
Fix neumann boundary

rbe051:
An attempt to fix the Neumann boundary conditions. Many of the changes here are done by bull request #168, but had to merge it into this branch to make the unittests run.

The mpsa_partial_update is not running properly. I think this is because when you are doing a local update, we are not able to tell if a Neumnan boundary should have the asymmetric part or not. I don't have any good solutions for this right now.


src/porepy/numerics/fv/mpsa.py
```diff
@@ -1040,6 +1016,15 @@ def mpsa_elasticity(
...
 
+    # To avoid singular matrices we are not abe to add the asymetric part of the stress
+    # tensor to the Neumann and Robin boundaries for nodes that only has more
+    # Neumann-boundary faces than gradients. This will typically happen in the
+    # corners where you only can have one gradient for the node. Normally if you
+    # have at least one internal face connected to the node you are should be safe.
+    _eliminate_ncasym_neumann(
+        ncasym, subcell_topology, bound_exclusion, cell_node_blocks, nd
+    )
+
```
rbe051: @keileg : Do you agree when the local system becomes singular? I am not 100 percent sure this is correct, but it is valid at least for nodes with only 1 subcell, that is, in the corners. In 3D this condition will also be violated at edges of the unit cube if there is a Neuman condition on both sides of the edge

keileg: I think it would be most useful if you walk me through this. The same applies for the comment regarding partial update - we clearly need either to fix those, or disable the unit tests, but I don't immediately see what you think is the problem.  I haven't looked at that code for well over a year, so that may be part of the problem.

I suggest we discuss this at some point.

alessiofumagalli:
@rbe051 , is this only mpsa?

rbe051:
@alessiofumagalli yes, mpfa and tpfa does not have this problem, and should work fine

rbe051:
@keileg I am ready to merge this branch. Any objections?

alessiofumagalli:
@rbe051 , yes mine. I'm currently merging the other one :P

alessiofumagalli:
now you're free to merge

PR #173, opened 2018-11-16
update the coarsening strategy for the new interface, activate in exa…

alessiofumagalli:
…mple1/benchmark1/test_vem two coarsening test


keileg:
Looks good to me. @IvarStefansson Could you please have a look at this and check if it will inflict on your effort on the parameters? I don't think so, but your task is hard enough without us intervening.

IvarStefansson:
This should not affect me to any significant degree, so please go ahead!

alessiofumagalli:
done @IvarStefansson , thanks

PR #176, opened 2018-11-30
Mpsa subface formulation

rbe051:
The main part of this pull request is an expansion of Mpsa so that we can retrieve the subcell and subface discretizations. This enables us to specify the boundary conditions for each subface instead of on each face. Further it let us obtain the subface tractions instead of the face traction. This increased accuracy is crucial if you are interested in the boundary (contact mechanics). If you don't care about the subfaces everything should work as before. Mpsa will only return the subfaces if the boundary-condition is given for the subfaces instead of faces.

This pull request also includes some improvements in the subface-displacement reconstruction, notably a chang from 'C' ordering to 'F' and from giving the displacement per gradient on the subfaces, to 1 displacement per subface (the displacement is averaged over the displacement from the gradient on the two sides. At the continuity point this is the same)

We have also included two interface-laws for the elasticity problem, a Robin coupling and a full continuity coupling.

Before accepting this pull request have a look in fv_elliptic and comment on the classes of FVElliptic and FVEllipticVector. Should we have both? Should they be renamed?


src/porepy/numerics/fv/fv_elliptic.py
```diff
@@ -344,3 +347,429 @@ def enforce_neumann_int_bound(
         """
         # Operation is void for finite volume methods
         pass
+
+
+class FVVectorElliptic(pp.numerics.mixed_dim.solver.Solver):
```
keileg: Do you actually use this class for anything? If not, it feels like a premature abstraction that may cause more harm than good in the future. As an example, it indicates that mappings between grids in different dimensions is straightforward for elasticity - fact is that nobody knows this.

Unless I am overlooking something here, my preference would be to include what is needed to get contact mechanics working into mpsa, and then remove the rest - but save the commit id.

keileg: Nothing should inherit from the old solver anymore - the whole concept should be purged. This PR is not the place for that, though.

src/porepy/numerics/fv/fvutils.py
```diff
@@ -188,9 +186,11 @@ def compute_dist_face_cell(
 
     _, cols = np.meshgrid(subcell_topology.subhfno, np.arange(dims))
     cols += matrix_compression.rldecode(np.cumsum(blocksz) - blocksz[0], blocksz)
-    eta_vec = eta * np.ones(subcell_topology.fno.size)
-    # Set eta values to zero at the boundary
-    if not eta_at_bnd:
+    if np.asarray(eta).size == subcell_topology.num_subfno_unique:
```
keileg: Is eta interpreted according to its length? Documentation is needed at the function level.

src/porepy/numerics/fv/fvutils.py
```diff
@@ -537,12 +537,16 @@ def map_hf_2_f(fno, subfno, nd):
     fno face numbering in sub-cell topology based on unique subfno
     subfno sub-face numbering
     nd dimension
-
+    g (opitonal) If a grid is supplied the method defines all mappings
```
keileg: What does 'all mappings' refer to?

src/porepy/numerics/fv/fvutils.py
```diff
@@ -713,17 +705,18 @@ def __init__(self, subcell_topology, bound, nd):
 
     def _basis_transformation(self, basis):
```
keileg: What is this function doing?

src/porepy/numerics/fv/fvutils.py
```diff
@@ -1223,3 +1216,22 @@ def compute_darcy_flux(
         # Remove contribution directly on the boundary faces.
         induced_flux[g_h.tags["fracture_faces"]] = 0
         gb.node_props(g_h)[d_name] += induced_flux
+
+
+def boundary_to_sub_boundary(bound, subcell_topology):
+    """
+    Convert a boundary condition defined for faces to a boundary condition defined by
+    subfaces.
```
keileg: A modified copy of the boundary condition object is returned?

src/porepy/numerics/fv/mpsa.py
```diff
@@ -13,20 +13,14 @@
 import logging
 import porepy as pp
 
-from porepy.numerics.fv import fvutils
-from porepy.utils import matrix_compression, mcolon, sparse_mat
-from porepy.grids import structured, partition
-from porepy.params import tensor, bc
```
keileg: It is good to see those old imports disappearing.

src/porepy/numerics/fv/mpsa.py
```diff
@@ -1329,13 +1253,10 @@ def reconstruct_displacement(g, subcell_topology, eta=None, eta_at_bnd=False):
...
-        eta (float, range=[0,1]): Optional. Parameter determining the point at which the
-            displacement is evaluated. If eta is not given the method will call
-            fvutils.determine_eta(g) to set it.
-        eta_at_bnd (bool): Optional, defaults to False.
-            In MPSA eta is only enforced at internal faces, while is always set to
-            eta=0 at the boundary. eta_at_bnd=True will override this behaviour,
-            and give the displacement at the points given by eta, also for the boundary.
+        eta (float or ndarray, range=[0,1]): Optional. Parameter determining the point
+            at which the displacement is evaluated. If eta is a nd-array it should be on
+            the size of subcell_topology.num_subfno. If eta is not given the method will
```
keileg: This also means that etas that vary on the sub-face level must be set according to the numbering in that array? Sound like one really has to know what one is doing, but it has to be that way, I suppose. The alternative of leaving it to fvutils is fully acceptable.

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -334,7 +334,325 @@ def assemble_matrix_rhs(
...
+        # If master and slave is the same grid, they should contribute to the same
+        # row and coloumn. When the assembler assigns matrix[idx] it will only add
+        # the slave information because of duplicate indices (master and slave is the same).
+        # We therefore write the both master and slave info to the slave index.
+        if g_master == g_slave:
+            master_ind = 1
+        else:
+            master_ind = 0
+
+        self.discr_master.assemble_int_bound_displacement_trace(
```
keileg: Now I see where you get the FVEllipticVectorial stuff from, some of those functions are indeed necessary. Nevertheless, I think those should be moved to mpsa - in particular since it seems unlikely we will invent and implement a new fv discretization for elasticity anytime soon.

src/porepy/params/bc.py
```diff
@@ -8,7 +8,36 @@
...
+    def copy(self):
+        """
+        Create a deep copy of the boundary condition.
+
+        Returns:
+            bc: A deep copy of self. All attributes will also be copied.
+
+        """
+        # We don't call the init since we don't have access to the grid.
+        # Maybe this is bad style.
```
keileg: Quite likely. Let us see if anyone will ever notice.

test/integration/test_mpsa_partial.py
```diff
@@ -4,144 +4,6 @@
 import porepy as pp
 
 
-class TestUpdateDisc(unittest.TestCase):
```
keileg: What happened to the test?

src/porepy/grids/fv_sub_grid.py
```diff
@@ -0,0 +1,54 @@
+import numpy as np
+import scipy.sparse as sps
+import porepy as pp
+
+
+class FvSubGrid(pp.Grid):
```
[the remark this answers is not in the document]
rbe051: I wrote this function exactly with the proper sub-grid in mind. In 2D it was straight-forward, but in 3D it was a bit more tricky, and I abounded this function and just forgot about it. 

I can not decide if it has the right of life or not. It might be a convenient way to set the subface-boundary conditions if you don't know anything about SubfaceTopology, however, then you are not able to interpret the output of Mpsa which is on the subface level. 

I think the cleanest would be to actually construct the sub-cell grid, let mpsa/mpfa work on this, and return the discretization on the sub-grid. This would simplify mpsa/mpfa considerably and possible make the code readable. But as you say, this is a project in its own

src/porepy/grids/mortar_grid.py
```diff
@@ -297,7 +297,7 @@ def _check_mappings(self, tol=1e-4):
 # ------------------------------------------------------------------------------#
 
 
-class BoundaryMortar(object):
+class BoundaryMortar(MortarGrid):
```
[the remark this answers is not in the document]
rbe051: For now, yes, Ideally, no. 

It should be quite straight forward to remove the BoundaryMortar, however, MortarGrid was written with only mixed-dim couplings in mind, which means that there still are a few differences that needs to be sorted out first.

src/porepy/numerics/fv/fv_elliptic.py
rbe051: I kind off agree, but I don't see the harm in the abstraction. If you want to couple different dimensions you first need to write a mixed dim interface law, and if you do that you know enough to know what you are doing.  But I agree most (or all) can be moved to the Mpsa-class. However, I think we should add a VectorEllipticDiscretization to the mixed-dim folder (this folder should not be called mixed-dim, but that is an other issue) as @IvarStefansson  as plans of finite elements for elasticity.

src/porepy/numerics/fv/mpsa.py
rbe051: Yes, eta has to be set according to SubcellTopology. I can't really think of any other way of setting this. I guess if you want to set the continuity points on the subcells it can be required that you know what you are doing

src/porepy/numerics/fv/fv_elliptic.py
rbe051: Also, this specify the contract between the node-discretization and interface law. Keeping this separate from mpsa should be beneficial.

test/integration/test_mpsa_partial.py
rbe051: Good catch. It must have been deleted by an accident

rbe051: Never mind, I now remember. It was deleted on purpose. This test does exactly the same as the other class in the same file. There was a purpose of two tests for the early versions of displacement reconstruction, but at the current version there is no difference between the two tests.

rbe051:
I'm ready to merge this if you agree @keileg @IvarStefansson

IvarStefansson:
> IvarStefansson

There is a new Python 3.5 issue in test_elliptic_interface_laws.py. This just shows the importance of fixing the other one (#177), as it increases the risk of overlooking further 3.5 failures.

rbe051:
> > IvarStefansson
> 
> There is a new Python 3.5 issue in test_elliptic_interface_laws.py. This just shows the importance of fixing the other one (#177), as it increases the risk of overlooking further 3.5 failures.

It should work now. It was the recurring error of node_numbers switching randomly in 3.5 making the tests fail at random

PR #178, opened 2018-12-03
make the assembler more fast by avoiding useless empty sparse matrix creation

alessiofumagalli:
very easy branch, but I notices a speed up of 10 in the assembly procedure. Now most of the time is spent in the actual discretization and not in other stuff. 
Ciao ciao

### DIFF
--- a/src/porepy/numerics/mixed_dim/assembler.py
+++ b/src/porepy/numerics/mixed_dim/assembler.py
@@ -120,12 +120,20 @@ class Assembler(pp.numerics.mixed_dim.AbstractAssembler):
             (g1, data_1, data_edge, local_matrix)
 
         """
+        # Define the matrix format, common for all the sub-matrices
+        if matrix_format == "csr":
+            sps_matrix = sps.csr_matrix
+        elif matrix_format == "csc":
+            sps_matrix = sps.csc_matrix
+        else: # coo
+            sps_matrix = sps.coo_matrix
+
 
         # Initialize the global matrix.
-        matrix, rhs, block_dof, full_dof = self._initialize_matrix_rhs(gb, variables)
+        matrix, rhs, block_dof, full_dof = self._initialize_matrix_rhs(gb, variables, sps_matrix)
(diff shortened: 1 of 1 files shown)


src/porepy/numerics/mixed_dim/assembler.py
```diff
@@ -120,12 +120,20 @@ def assemble_matrix_rhs(
             (g1, data_1, data_edge, local_matrix)
 
         """
+        # Define the matrix format, common for all the sub-matrices
+        if matrix_format == "csr":
+            sps_matrix = sps.csr_matrix
+        elif matrix_format == "csc":
+            sps_matrix = sps.csc_matrix
+        else: # coo
+            sps_matrix = sps.coo_matrix
```
keileg: Are you sure coo is a good default value?

src/porepy/numerics/mixed_dim/assembler.py
```diff
@@ -156,7 +164,13 @@ def assemble_matrix_rhs(
 
                             # Assign values in global matrix
                             var_key_name = self._variable_term_key(term, row, col)
-                            matrix[var_key_name][ri, ci] += loc_A
+                            # Check if the current block is None or not, it could
```
keileg: Please explain why it can be none (initialization)

alessiofumagalli:
note: the speed up was problem specific with 1 2d grid, 84 2d grid and 85 0d grid. but i guess it was mostly related to problems with many elements in the grid bucket

src/porepy/numerics/mixed_dim/assembler.py
alessiofumagalli: not really, should we put csr?

src/porepy/numerics/mixed_dim/assembler.py
alessiofumagalli: i just realized by running the tests

alessiofumagalli:
python 3.6 works, 3.5 has the same problem as develop

Issue #177, opened 2018-12-03
Python 3.5 Travis failure


IvarStefansson:
In Python 3.5, the tutorial on dfn meshing fails:

`/porepy/fracs/split_grid.py", line 242, in duplicate_faces
    update_values[i] = gh.tags[key][frac_id]
IndexError: index 256 is out of bounds for axis 0 with size 256`

Moreover, the unit test test_tag_gmsh.py fails randomly in 3.5, and the tests take longer than usual on Python 3.6.

keileg:
@alessiofumagalli : From what I can see, you made the code and the test. Would you care to have a look?

IvarStefansson:
Note that the tutorial only seems to fail on the cron runs.

alessiofumagalli:
@IvarStefansson , should we close this issue?

PR #180, opened 2018-12-05
Issue177

IvarStefansson:
Fixed the two problems described in #177 :
   In test_tag_gmsh.py, we now check the ordering of the two 1d neighbors of the 2d grid, since this is what determines the order of the splitting of the 2d faces.
   Since the DFN meshing tutorial is obsolete, it was moved from tutorials, and thereby not checked in the cron builds.


keileg:
This is a very good solution.

keileg:
Resolves #177

PR #188, opened 2018-12-09
Remove old model classes, elliptic_assembler, cleanup tests

keileg:
Main changes:
1) All the old model classes are gone, as are their disabled tests.
2) EllipticAssembler, Solver, Coupler and some related classes are deleted.
3) EllipticDiscretization is moved to interface_laws.
4) There is a new module, test/common.py for functionality that is common to several tests. For the moment, it contains methods to solve mixed-dim elliptic equations.

Q: I have not yet touched the time_stepper. Should we get rid of that one as well, when we're at it?


keileg:
Regarding test/common.py, I think the placement is okay, it is common for unit and integration, but I was not happy with the name - perhaps test_common.py would have been better? Or commons? Suggestions and comments are most welcome.

rbe051:
> Regarding test/common.py, I think the placement is okay, it is common for unit and integration, but I was not happy with the name - perhaps test_common.py would have been better? Or commons? Suggestions and comments are most welcome.

Isn't utils the name we give when we don't have a better name for it? I think test/commons is good. It is clear that this is test commons because its placement in the folder structure

src/porepy/numerics/fracture_deformation.py
```diff
@@ -1,365 +1,11 @@
...
-
-        self.d_n[fi] += self.fracture_dilation(slip_d, fi)
-        self.d_n[fi_left] += self.fracture_dilation(slip_d, fi_left)
-
-        assert np.all(self.d_n[fi] > -1e-6)
-
-        self.x[:, fi] += slip_vec
-        self.x[:, fi_left] -= slip_vec
+""" This is a resurrection of a small part of the old fracture_deformation
+model. It contains a single function that should really be placed somewhere els.
```
rbe051: And i'm sure we have three different versions of this hidden away in dark places of the code

test/common.py
```diff
@@ -0,0 +1,73 @@
+""" Functionailty that is common for several tests, and therfore stored in a
+common module.
+"""
+import numpy as np
+
+import porepy as pp
+
+
+def setup_flow_assembler(gb, method, data_key=None, coupler=None):
+    """ Setup a standard assembler for the flow problem for a given grid bucket.
```
rbe051: Good!

rbe051: This should maybe go into the new folder we discussed on Thursday. It would make @alessiofumagalli  happy

test/integration/test_TpfaMonoCoupling.py
```diff
@@ -33,27 +33,21 @@ def test_two_cart_grids(self):
...
-        tpfa = pp.Tpfa(key)
-        for g, d in gb:
-            d[discretization_key] = tpfa
 
-        for _, d in gb.edges():
-            d[discretization_key] = pp.FluxPressureContinuity(key, tpfa)
+        for e, d in gb.edges():
+            mg = d["mortar_grid"]
+            d[pp.PARAMETERS] = pp.Parameters(mg, ["flow"], [{}])
             pp.params.data.add_discretization_matrix_keyword(d, "flow")
```
rbe051: This looks much cleaner now

IvarStefansson:
> > Regarding test/common.py, I think the placement is okay, it is common for unit and integration, but I was not happy with the name - perhaps test_common.py would have been better? Or commons? Suggestions and comments are most welcome.
> 
> Isn't utils the name we give when we don't have a better name for it? I think test/commons is good. It is clear that this is test commons because its placement in the folder structure

I agree. In #189, I added test/test_utils.py. Discussion with @keileg concluded with our preferring this over just utils.py, since it will make the tests more readable.

test/common.py
IvarStefansson: I agree. But I'm not entirely convinced about "mortar_solution". Since the node variable is "pressure", it would be more consistent using something more physics specific. "flux" and "mortar_flux" come to mind.

test/common.py
```diff
@@ -0,0 +1,73 @@
...
+                g1: ('pressure', "diffusive"),
+                g2: ('pressure', "diffusive"),
+                e: ('mortar_solution', coupler),
+                }}
+        d[pp.DISCRETIZATION_MATRICES] = {"flow": {}}
+
+    assembler = pp.Assembler()
+    return assembler
+
+def solve_pressure_distribute(gb, assembler):
```
IvarStefansson: I'm particularly fastidious about names today, so I'd prefer solve_and_distribute_pressure.

src/porepy/numerics/fracture_deformation.py
keileg: I will let it live for now, then we can move it later on.

keileg:
I will kill the time stepper, clean up some names, and then merge.

PR #189, opened 2018-12-10
Biot

IvarStefansson:
Splitting of Biot class into DivD, GradP and BiotStabilization. Thus, the assembler can assemble the individual terms using the assemble_matrix_rhs methods of three new classes and Mpsa, Mpfa and MassMatrix. The idea is to discretize all mechanics terms (the three new + Mpsa) by a call to the _discretize_mech method of the old Biot class.
Testing could be improved, documentation should be fairly good.

Added test_utils.py to the test folder.
Note the inconsistency in source term treatment between mpsa and other discretizations, see slack channel #biot.

### DIFF
--- a/src/porepy/numerics/fv/biot.py
+++ b/src/porepy/numerics/fv/biot.py
@@ -8,12 +8,14 @@ from porepy.numerics.mixed_dim.solver import Solver
 
 
 class Biot(Solver):
-    def __init__(self, eta=None):
-        """ Set default values for some parameters used in discretization.
+    def __init__(self, mechanics_keyword="mechanics", flow_keyword="flow"):
+        """ Set the two keywords.
 
+        The keywords are used to access and store parameters and discretization
+        matrices.
         """
-        defaults = {"fluid_compr": 0, "fluid_viscosity": 1, "biot_alpha": 1}
-        self.defaults = defaults
+        self.mechanics_keyword = mechanics_keyword
+        self.flow_keyword = flow_keyword
 
     def ndof(self, g):
         """ Return the number of degrees of freedom associated wiht the method.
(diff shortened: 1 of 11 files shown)


src/porepy/numerics/fv/biot.py
```diff
@@ -213,18 +208,22 @@ def assemble_matrix(self, g, data):
...
     def _discretize_compr(self, g, data):
-        parameter_dictionary = data[pp.PARAMETERS]["flow"]
-        matrix_dictionary = data[pp.DISCRETIZATION_MATRICES]["flow"]
-        compr = parameter_dictionary["fluid_compressibility"]
-        poro = parameter_dictionary["porosity"]
-        matrix_dictionary["compr_discr"] = sps.dia_matrix(
-            (g.cell_volumes * compr * poro, 0), shape=(g.num_cells, g.num_cells)
+        """
+        TODO: Sort out time step (inconsistent with MassMatrix).
+        """
```
rbe051: In my opinion the time-step disappears from the MassMatrix

src/porepy/numerics/fv/biot.py
```diff
@@ -608,10 +611,344 @@ def compute_stress(self, g, u, data):
...
+                             stored in the matrix dictionary."""
+            )
+        biot_alpha = data[pp.PARAMETERS][self.keyword]["biot_alpha"]
+        return matrix_dictionary["grad_p"] * biot_alpha
+
+    def assemble_rhs(self, g, data):
+        """ Return the zero right-hand side for a discretization of the pressure
+        gradient term.
+
+        @Runar: Is it correct that this is zero.
```
rbe051: Probably not

src/porepy/numerics/fv/biot.py
```diff
@@ -608,10 +611,344 @@ def compute_stress(self, g, u, data):
...
+        g: grid, or a subclass.
+
+        Return
+        ------
+        dof: the number of degrees of freedom.
+
+        """
+        return g.num_cells
+
+    def extract_displacement(self, g, solution_array, d):
```
rbe051: It seems a bit strange that all of these classes shall have an extract_displacement/pressure method.
It tells me that we still can make improvements in the code structure.

src/porepy/numerics/fv/biot.py
keileg: I prefer to put it on the flow term.

src/porepy/numerics/fv/biot.py
```diff
@@ -311,14 +311,16 @@ def _discretize_mech(self, g, data):
...
+        matrices_m = data[pp.DISCRETIZATION_MATRICES][self.mechanics_keyword]
+        matrices_f = data[pp.DISCRETIZATION_MATRICES][self.flow_keyword]
+        bound_mech = parameters_m["bc"]
+        bound_flow = parameters_f["bc"]
+        constit = parameters_m["fourth_order_tensor"]
 
-        eta = data.get("eta", 0)
-        inverter = data.get("inverter", None)
+        eta = parameters_m.get("mpsa_eta", fvutils.determine_eta(g))
+        inverter = parameters_m.get("inverter", None)
```
keileg: Is it better to use 'mpsa_inverter'  to make the usage more specific? I added 'mpfa_inverter' to mpfa.py over in #188 , somehow that functionality had died during some updates.

src/porepy/numerics/fv/biot.py
keileg: It will have a contribution from boundary terms, similar to standard mpfa/mpsa. This may require some thinking.

src/porepy/numerics/fv/biot.py
keileg: I agree, this will hopefully evolve as we get more experience.

src/porepy/numerics/fv/biot.py
```diff
@@ -608,10 +611,344 @@ def compute_stress(self, g, u, data):
...
+            data (dictionary): With data stored.
+
+        Returns:
+            np.ndarray: Zero right hand side vector with representation of boundary
+                conditions.
+        """
+        return np.zeros(self.ndof(g))
+
+
+class BiotStabilization(pp.numerics.mixed_dim.VectorEllipticDiscretization):
```
keileg: I guess it will be up to me to write the explanation of what on earth this term really is. I will return to it one day I feel particularly brave.

src/porepy/numerics/fv/biot.py
rbe051: I think it then would be better with block_inverter or local_inverter

PR #194, opened 2019-01-11
Biot rhs

IvarStefansson:
Important note:
The time step was removed from MassMatrix.

Otherwise, there are few updates in fv/biot.py:
DivD class returns the rhs
Fixed two bugs in the Biot rhs

utils/test_biot.py was extended with a test of the rhs. See that file for how to assemble Biot with a non-zero rhs. In particular, the assemble_rhs functions of the classes Mpfa and MassMatrix must be adjusted to account for the time step and the rhs, respectively.

### DIFF
--- a/src/porepy/numerics/fv/biot.py
+++ b/src/porepy/numerics/fv/biot.py
@@ -53,6 +53,8 @@ class Biot:
 
         TODO: Boundary effects of coupling terms.
 
+        There is an assumption on constant mechanics BCs, see DivD.assemble_matrix().
+
         Parameters:
             g: grid, or subclass, with geometry fields computed.
             data: dictionary to store the data terms. Must have been through a
@@ -72,10 +74,10 @@ class Biot:
 
         matrices_m = data[pp.DISCRETIZATION_MATRICES][self.mechanics_keyword]
         matrices_f = data[pp.DISCRETIZATION_MATRICES][self.flow_keyword]
-        biot_alpha = data[pp.PARAMETERS][self.mechanics_keyword]["biot_alpha"]
+
+        dt = data[pp.PARAMETERS][self.flow_keyword]["time_step"]
         p_bound = (
-            -div_flow * matrices_f["bound_flux"] * p
-            - matrices_m["bound_div_d"] * d * biot_alpha
+            -div_flow * matrices_f["bound_flux"] * p * dt
         )
         s_bound = -div_mech * matrices_m["bound_stress"] * d
         return np.hstack((s_bound, p_bound))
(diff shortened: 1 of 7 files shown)


src/porepy/numerics/fv/biot.py
```diff
@@ -198,7 +202,7 @@ def assemble_matrix(self, g, data):
                 [A_mech, matrices_m["grad_p"] * biot_alpha],
                 [
                     matrices_m["div_d"] * biot_alpha * d_scaling,
-                    matrices_f["mass"] + dt * A_flow + matrices_f["biot_stabilization"],
+                    matrices_f["mass"] + dt * A_flow + biot_alpha * matrices_f["biot_stabilization"],
```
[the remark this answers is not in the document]
keileg: What do you mean?

src/porepy/numerics/fv/biot.py
```diff
@@ -386,7 +392,7 @@ def build_rhs_normals_single_dimension(dim):
         num_dir_subface = (
             bound_exclusion_mech.exclude_neu.shape[1]
             - bound_exclusion_mech.exclude_neu.shape[0]
-        ) * nd
+        )
```
keileg: Is this a bugfix?

src/porepy/numerics/fv/biot.py
```diff
@@ -843,7 +850,26 @@ def assemble_rhs(self, g, data):
...
                 conditions.
         """
-        return np.zeros(self.ndof(g))
+        parameter_dictionary = data[pp.PARAMETERS][self.keyword]
+        matrix_dictionary = data[pp.DISCRETIZATION_MATRICES][self.keyword]
+
+        # For IE and constant BCs, the boundary part cancels, as the contribution from
+        # successive timesteps (n and n+1) appear on the rhs with opposite signs. For
+        # transient BCs, use the below with the appropriate version of d_bound_i.
+        d_bound_1 = parameter_dictionary["bc_values"]
```
keileg: Is this correct? Will not d_bound_1 == d_bound_0

src/porepy/numerics/fv/biot.py
rbe051: Yes, bound_exlusioin.exlude_neu (and _dir, etc) changed from 1 per face to nd per face at some point (related to mpsa always getting BoundaryConditionVectorial).

src/porepy/numerics/fv/biot.py
rbe051: I belive it is correct. If you have constant boundary conditions  d_bound_1 == d_bound_0, but this is not true for time dependent boundary conditions. Time dependent boundary conditions are not supported yet, but Ivar and I agreed it was best to still write the equation down to make it explicit

src/porepy/numerics/fv/biot.py
rbe051: Yes, because bound_exclusion has changed from operating

src/porepy/numerics/fv/biot.py
keileg: I should have read the comments more carefully then. Good.

src/porepy/numerics/fv/biot.py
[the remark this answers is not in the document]
rbe051: In the previous version the stabilization term did not include biot alpha.
I multiplied the stabilization term with the biot alpha.
Do you agree that the stabilization term should include a biot alpha?

keileg:
Yes.

PR #197, opened 2019-01-18
Gradp

mstarnoni:
New discretization of the gradp term in mpsa-Biot.

### DIFF
--- a/src/porepy/numerics/fv/biot.py
+++ b/src/porepy/numerics/fv/biot.py
@@ -191,7 +191,8 @@ class Biot:
         # Put together linear system
         A_flow = div_flow * matrices_f["flux"]
         A_mech = div_mech * matrices_m["stress"]
-
+        grad_p = div_mech * (data['grad_p_jumps'] + data['grad_p_force'])
+        
         # Time step size
         dt = param[self.flow_keyword]["time_step"]
 
@@ -199,12 +200,12 @@ class Biot:
         # Matrix for left hand side
         A_biot = sps.bmat(
             [
-                [A_mech, matrices_m["grad_p"] * biot_alpha],
+                [A_mech, grad_p],
                 [
                     matrices_m["div_d"] * biot_alpha * d_scaling,
                     matrices_f["mass"]
                     + dt * A_flow
-                    + biot_alpha * matrices_f["biot_stabilization"],
+                    + data['stabilization',
                 ],
             ]
         ).tocsr()
@@ -365,48 +366,13 @@ class Biot:
         num_subhfno = subcell_topology.subhfno.size
 
         num_nodes = np.diff(g.face_nodes.indptr)
-        sgn = g.cell_faces[subcell_topology.fno, subcell_topology.cno].A
 
-        # The pressure gradient term in the mechanics equation is discretized
-        # as a force on the faces. The right hand side is thus formed of the
-        # normal vectors.
-        def build_rhs_normals_single_dimension(dim):
-            val = (
-                g.face_normals[dim, subcell_topology.fno]
-                * sgn
-                / num_nodes[subcell_topology.fno]
-            )
-            mat = sps.coo_matrix(
-                (val.squeeze(), (subcell_topology.subfno, subcell_topology.cno)),
-                shape=(subcell_topology.num_subfno, subcell_topology.num_cno),
-            )
-            return mat
-
-        rhs_normals = build_rhs_normals_single_dimension(0)
-        for iter1 in range(1, nd):
-            this_dim = build_rhs_normals_single_dimension(iter1)
-            rhs_normals = sps.vstack([rhs_normals, this_dim])
-
-        rhs_normals = bound_exclusion_mech.exclude_dirichlet(rhs_normals)
-
-        num_dir_subface = (
-            bound_exclusion_mech.exclude_neu.shape[1]
-            - bound_exclusion_mech.exclude_neu.shape[0]
-        )
-        # No right hand side for cell displacement equations.
-        rhs_normals_displ_var = sps.coo_matrix(
-            (
-                nd * subcell_topology.num_subfno - num_dir_subface,
-                subcell_topology.num_cno,
-            )
-        )
-
-        # Why minus?
-        rhs_normals = -sps.vstack([rhs_normals, rhs_normals_displ_var])
-        del rhs_normals_displ_var
+        num_subfno_unique = subcell_topology.num_subfno_unique
+        num_subfno = subcell_topology.num_subfno
+        num_cno = subcell_topology.num_cno
 
         # Call core part of MPSA
-        hook, igrad, rhs_cells, cell_node_blocks, hook_normal = mpsa.mpsa_elasticity(
+        hook, igrad, rhs_cells, cell_node_blocks = mpsa.mpsa_elasticity(
             g, constit, subcell_topology, bound_exclusion_mech, eta, inverter
         )
 
@@ -426,24 +392,6 @@ class Biot:
         # Discretization of boundary values
         bound_stress = hf2f * hook * igrad * rhs_bound
 
-        # Face-wise gradient operator. Used for the term grad_p in Biot's
-        # equations.
-        rows = fvutils.expand_indices_nd(subcell_topology.cno, nd)
-        cols = np.arange(num_subhfno * nd)
-        vals = np.tile(sgn, (nd, 1)).ravel("F")
-        div_gradp = sps.coo_matrix(
-            (vals, (rows, cols)),
-            shape=(subcell_topology.num_cno * nd, num_subhfno * nd),
-        ).tocsr()
-
-        #        del hook, rhs_bound
-        del rows, cols, vals
-
-        grad_p = div_gradp * hook_normal * igrad * rhs_normals
-        # assert np.allclose(grad_p.sum(axis=0), np.zeros(g.num_cells))
-
-        del hook_normal, div_gradp
-
         div = self._subcell_gradient_to_cell_scalar(g, cell_node_blocks)
 
         div_d = div * igrad * rhs_cells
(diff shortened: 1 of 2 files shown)


src/porepy/numerics/fv/biot.py
```diff
@@ -199,12 +201,10 @@ def assemble_matrix(self, g, data):
         # Matrix for left hand side
         A_biot = sps.bmat(
             [
-                [A_mech, matrices_m["grad_p"] * biot_alpha],
+                [A_mech, grad_p],
```
IvarStefansson: Don't we want to keep the biot alpha factor on grad_p?

src/porepy/numerics/fv/biot.py
```diff
@@ -191,6 +191,8 @@ def assemble_matrix(self, g, data):
         # Put together linear system
         A_flow = div_flow * matrices_f["flux"]
         A_mech = div_mech * matrices_m["stress"]
+        grad_p = div_mech * (matrices_m["grad_p_jumps"] + matrices_m["grad_p_force"])
+        stabilization = matrices_f["biot_stabilization"]
```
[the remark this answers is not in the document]
mstarnoni: They should always be combined. It's the feature of the method. grad_p_force is the force on the faces from one side only, grad_p_jumps represents the force due to a pressure imbalance between the cell-centers, that is [[n alpha p]].

src/porepy/numerics/fv/biot.py
```diff
@@ -453,15 +400,131 @@ def build_rhs_normals_single_dimension(dim):
...
 
+        # ----------------------------------------------------------------------------
+
+        # Biot discretization
+
+        # ----------------------------------------------------------------------------
+
+        # Take Biot's alpha as a tensor
+        alpha = parameters_m["biot_alpha"]
+        alpha_tensor = pp.SecondOrderTensor(2, alpha * np.ones(g.num_cells))
```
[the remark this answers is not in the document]
mstarnoni: my mistake, replace with nd.

src/porepy/numerics/fv/biot.py
```diff
@@ -453,15 +400,131 @@ def build_rhs_normals_single_dimension(dim):
...
+        # Compute nAlpha product same as nK in Darcy
+        nAlpha_grad, cell_node_blocks, sub_cell_index = fvutils.scalar_tensor_vector_prod(
+            g, alpha_tensor, subcell_topology
+        )
+
+        # transfer nAlpha to a face-based
+        unique_nAlpha_grad = subcell_topology.pair_over_subfaces(nAlpha_grad)
+
+        # convenience method for reshaping nAlpha from face-based
+        # to component-based
```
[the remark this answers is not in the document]
mstarnoni: This is to build a block diagonal sparse matrix compatible with igrad * rhs_units, that is first all x-component, then y, and z

src/porepy/numerics/fv/biot.py
```diff
@@ -453,15 +400,131 @@ def build_rhs_normals_single_dimension(dim):
...
+            bound_exclusion_mech.exclude_neu.shape[1]
+            - bound_exclusion_mech.exclude_neu.shape[0]
+        )
+
+        # No right hand side for cell displacement equations.
+        rhs_units_displ_var = sps.coo_matrix(
+            (nd * num_subfno - num_dir_subface, num_subfno_unique * nd)
+        )
+
+        # Why minus?
```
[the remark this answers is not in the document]
mstarnoni: this was a comment left from previous implementations, but thanks :)

src/porepy/numerics/fv/biot.py
```diff
@@ -453,15 +400,131 @@ def build_rhs_normals_single_dimension(dim):
...
+            return mat
+
+        sc2c = build_sc2c_single_dimension(0)
+        for i in range(1, nd):
+            this_dim = build_sc2c_single_dimension(i)
+            sc2c = sps.vstack([sc2c, this_dim])
+
+        # computation of imbalance coefficients,
+        # that is jumps in cell-centers pressures
+        grad_p_jumps = hf2f * hook * igrad * rhs_units * unique_nAlpha_grad * sc2c
```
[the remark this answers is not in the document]
mstarnoni: this is jumps in [[n alpha p]].

src/porepy/numerics/fv/biot.py
```diff
@@ -453,15 +400,131 @@ def build_rhs_normals_single_dimension(dim):
...
+        # that is the actual force on the face
+        grad_p_force = hf2f * map_unique_subfno * nAlpha_grad * sc2c
+
+        matrices_m["grad_p_jumps"] = grad_p_jumps
+        matrices_m["grad_p_force"] = grad_p_force
+
+        stabilization = div * igrad * rhs_units * unique_nAlpha_grad * sc2c
+        # include in the stabilization term also the "hydrostatic term?
+        # I believe so but this needs some more testing
+        stabilization += div * nAlpha_grad * sc2c
```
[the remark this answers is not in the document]
mstarnoni: the "hydrostatic" term is grad_p_force.
The notation is by analogy with the new mpfa discretization with gravity.

src/porepy/numerics/fv/biot.py
mstarnoni: in the new discretization alpha is already inside grad_p

IvarStefansson: Ok, good!

src/porepy/numerics/fv/biot.py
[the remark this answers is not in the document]
rbe051: So from what I understand the divergence of grad_p_force should be zero? That is:
 div_mech * grad_p_force * pressure == 0

rbe051:
I fixed my comment about time discretization of the stabilization

src/porepy/numerics/fv/biot.py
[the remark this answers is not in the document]
keileg: Then please leave a comment.

src/porepy/numerics/fv/biot.py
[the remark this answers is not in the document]
keileg: Please find a better name - the analogue is not clear, in particular since nobody outside our group has thought about mpfa and gravity in that particular way.

src/porepy/numerics/fv/biot.py
```diff
@@ -453,15 +405,131 @@ def build_rhs_normals_single_dimension(dim):
...
+        matrices_f["biot_stabilization"] = stabilization
+
+        # ----------------------------------------------------------------------------
+
+        # End of Biot discretization
```
keileg: This is not necessary, at most leave a single line comment.

src/porepy/numerics/fv/biot.py
```diff
@@ -453,15 +405,131 @@ def build_rhs_normals_single_dimension(dim):
...
         matrices_m["bound_div_d"] = bound_div_d
 
+        # ----------------------------------------------------------------------------
+
+        # Biot discretization
+
+        # ----------------------------------------------------------------------------
+
+        # Take Biot's alpha as a tensor
+        alpha = parameters_m["biot_alpha"]
```
keileg: biot_alpha or biot_coefficient? I think I lean towards the latter, but I am open for other opinions.

src/porepy/numerics/fv/biot.py
```diff
@@ -983,4 +1051,17 @@ def assemble_rhs(self, g, data):
...
                 conditions.
         """
-        return np.zeros(self.ndof(g))
+        parameter_dictionary = data[pp.PARAMETERS][self.keyword]
+        matrix_dictionary = data[pp.DISCRETIZATION_MATRICES][self.keyword]
+
+        # The stabilization is the pressure contribution to the div u part of the
+        # fluid mass conservation, thus need a right hand side in the implicit Euler
+        # discretization.
+        pressure_0 = parameter_dictionary["state"]
```
keileg: This is the pressure at the previous time step? If so, I think it should not be in the parameter dictionary. @IvarStefansson @rbe051 What do you say?

src/porepy/numerics/fv/fvutils.py
```diff
@@ -538,6 +538,19 @@ def expand_indices_nd(ind, nd, direction="F"):
     return new_ind
 
 
+def expand_indices_incr(ind, dim, increment):
+
+    # Convenience method for duplicating a list, with a certain increment
```
keileg: It is not a list, but an np array.

src/porepy/numerics/fv/fvutils.py
```diff
@@ -538,6 +538,19 @@ def expand_indices_nd(ind, nd, direction="F"):
     return new_ind
 
 
+def expand_indices_incr(ind, dim, increment):
```
keileg: Documentation needed.

src/porepy/numerics/fv/fvutils.py
```diff
@@ -619,6 +632,74 @@ def vector_divergence(g):
...
+    _, j = np.meshgrid(subcell_topology.subhfno, np.arange(nd))
+    sum_blocksz = np.cumsum(blocksz)
+    j += pp.utils.matrix_compression.rldecode(sum_blocksz - blocksz[0], blocksz)
+
+    # Distribute faces equally on the sub-faces
+    num_nodes = np.diff(g.face_nodes.indptr)
+    normals = (
+        g.face_normals[:, subcell_topology.fno] / num_nodes[subcell_topology.fno]
+    )
+    if apertures is not None:
```
keileg: Why do we do this? I have the feeling someone else than @mstarnoni need to comment.

src/porepy/numerics/fv/biot.py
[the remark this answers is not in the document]
keileg: .. And if they are always combined, is there a reason to keep them as two separate fields?

src/porepy/numerics/fv/biot.py
```diff
@@ -365,48 +370,13 @@ def _discretize_mech(self, g, data):
...
-        )
 
-        # Why minus?
-        rhs_normals = -sps.vstack([rhs_normals, rhs_normals_displ_var])
-        del rhs_normals_displ_var
+        num_subfno_unique = subcell_topology.num_subfno_unique
+        num_subfno = subcell_topology.num_subfno
+        num_cno = subcell_topology.num_cno
 
         # Call core part of MPSA
```
keileg: Will this tacitly also construct the mpsa discretization matrices, and if so, do we risk overwriting values? 

@IvarStefansson, @rbe051 : We need to be careful here, at the very least, we should do a proper documentation on the relation between the components that together form mpsa for biot.

src/porepy/numerics/fv/biot.py
```diff
@@ -453,15 +405,131 @@ def build_rhs_normals_single_dimension(dim):
...
-        stabilization = div * igrad * rhs_normals
-
         matrices_m["stress"] = stress
         matrices_m["bound_stress"] = bound_stress
-        matrices_m["grad_p"] = grad_p
         matrices_m["div_d"] = div_d
-        matrices_f["biot_stabilization"] = stabilization
         matrices_m["bound_div_d"] = bound_div_d
 
+        # ----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
keileg: Also, it is wrong, div_d is part of the Biot discretization as well.

src/porepy/numerics/fv/biot.py
keileg: We should anyhow not have a parameter or variable called state.

keileg:
I used the 'request changes' option in the review, just to see what happens.

src/porepy/numerics/fv/biot.py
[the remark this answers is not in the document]
rbe051: If grad_p_force is what I think it is, it should be two separate fields

src/porepy/numerics/fv/biot.py
rbe051: Yes, but we don't (possible) overwrite before line 458. And you only overwrite if you use the same keyword. With the new parameter/keyword-structure I think it is fair to leave this to the user

rbe051: We might want to revisit the idea of separate keywords for input (parameters) and output (discretization_matrices)

src/porepy/numerics/fv/biot.py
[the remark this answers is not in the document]
mstarnoni: The divergence of grad_p_force is NOT zero.
They are always combined so my sense it that we can group together into a single variable grad_p

src/porepy/numerics/fv/biot.py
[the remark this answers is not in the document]
mstarnoni: It is defined as a scalar right now. For example, it is used as a scalar in the alpha div_u term.

src/porepy/numerics/fv/biot.py
```diff
@@ -393,143 +388,184 @@ def _discretize_mech(self, g, data):
...
-        # Compute nAlpha product same as nK in Darcy
-        nAlpha_grad, cell_node_blocks, sub_cell_index = fvutils.scalar_tensor_vector_prod(
-            g, alpha_tensor, subcell_topology
-        )
+        num_nodes = np.diff(g.face_nodes.indptr)
+
+        # Step 1
+
+        # Take Biot's alpha as a tensor
+        alpha_tensor = tensor.SecondOrder(nd, alpha * np.ones(g.num_cells))
```
rbe051: This is old syntax and should not run in this branch. If it is running on your computer it is because python still has access to the old version of porepy you have on your computer. Same goes for line 483 (should be alpha_tensor.values)

src/porepy/numerics/fv/biot.py
```diff
@@ -393,143 +388,184 @@ def _discretize_mech(self, g, data):
...
 
         rhs_units = build_rhs_units_single_dimension(0)
-
+        
         for i in range(1, nd):
             this_dim = build_rhs_units_single_dimension(i)
             rhs_units = sps.block_diag([rhs_units, this_dim])
 
-        rhs_units = bound_exclusion_mech.exclude_dirichlet(rhs_units)
+        rhs_units = bound_exclusion.exclude_dirichlet_nd(rhs_units)
```
rbe051: bound_exclusion.exclude_dirichlet_nd is outdated

PR #200, opened 2019-01-23
Multiphysics

alessiofumagalli:
Merge some left over commits from the branch. They deal with dfn stuff.


src/porepy/numerics/mixed_dim/pressure_trace.py
```diff
@@ -0,0 +1,186 @@
...
+The module contains the class that implement pressure trace discretization,
+no PDE are associated with this discretization.
+"""
+
+import numpy as np
+import scipy.sparse as sps
+
+import porepy as pp
+
+class PressureTrace(
```
rbe051: I guess this is a function used for the DFN methods to hold the pressure at the interface between several grids? I was just wondering if PressureTrace is the best name since it does not have a function assemble_int_bound_pressure_trace ?

alessiofumagalli: and also the name "pressure" is not really the best..

alessiofumagalli:
it's again the ordering given from networkx which is not stable in python 3.5 that make the test to fail. this means not very robust test, but I see this as an additional motivation to kick out python 3.5. opinions?

rbe051:
Can you fix the test? I guess it might take a few months before we drop 3.5, and you don't want to wait that long before you merge this in

alessiofumagalli:
I-ll try to fix the problem, it should not be too difficult.

src/porepy/numerics/mixed_dim/pressure_trace.py
```diff
@@ -0,0 +1,186 @@
...
+
+import numpy as np
+import scipy.sparse as sps
+
+import porepy as pp
+
+
+class PressureTrace(pp.Assembler):
+    """ Implement a pressure trace discretization, no PDE are associated with this discretization.
+    Only the interface operator are given. A possible usage is for the co-dimensional objects in a
```
keileg: I don't understand the purpose here. You want to map cell center pressures to the boundary? Why do you need a separate discretization for this? And which numerical method are you assuming? It will not be consistent with a FV approach.

keileg: It does not capture the pressure difference from cell center to the face caused by the flux over the face. The point is, the pressure trace operator is already defined for all EllipticDiscretization methods, this is what is used in mixed-dimensional couplings.

alessiofumagalli: the PressureTrace class is like the classes in the files fv_elliptic.py or dual_elliptic.py. The thing is, I need a class that realise no-equation in the domain (the matrix and rhs are empty) but give some coupling conditions for the higher dimensional domain. This is an approach similar to what we have used in the DFN paper we did. The variable in the lower dimensional domain can be used as Lagrange multiplier to couple the flow. The name might be misleading, thou

alessiofumagalli:
it's not clear to me why the continuous-integration/travis-ci/push keeps failing, @keileg any suggestion?

src/porepy/numerics/mixed_dim/pressure_trace.py
keileg: I think I understand. The name is misleading, and the documentation must be improved.

alessiofumagalli: Sure, I'll do it

keileg:
Are we ready to merge?

alessiofumagalli:
it depends if you like or not the name.. it's the last issue

keileg:
Merge when you feel ready

alessiofumagalli:
OK, I'm currently cleaning up the multilayer. This after, but for sure by today.

PR #204, opened 2019-01-30
Multilayer

alessiofumagalli:
In this branch the multi layer model (rock matrix + damaged layer + fault, the last two co-dimensional). Everything is contained in the corresponding "paper" folder. Soon the submission to arxiv and the new name for the folder. Some small fix in the library.


src/porepy/grids/mortar_grid.py
```diff
@@ -78,6 +78,9 @@ def __init__(self, dim, side_grids, face_cells, name=""):
         self.cell_volumes = np.hstack(
             [g.cell_volumes for g in self.side_grids.values()]
         )
+        self.cell_centers = np.hstack(
```
keileg: Are we introducing more of a geometry for the mortar grid? I see no problems right now, and this is probably a reasonable approach, but I want to understand what is going on.

src/porepy/numerics/vem/dual_elliptic.py
```diff
@@ -289,7 +289,7 @@ def _velocity_dof(self, g, g_m, grid_swap):
         return hat_E_int
 
     def assemble_int_bound_flux(
-        self, g, data, data_edge, grid_swap, cc, matrix, self_ind
+        self, g, data, data_edge, grid_swap, cc, matrix, self_ind, coefficient = 1.
     ):
```
keileg: 1) This breaks the uniform interface of all EllipticDiscretizations. I don't like it at all, with a unit default or not. What is the case where you need to do this? If it is a special case dedicated for a paper, and I will quite likely characterize it as such, the preferred solution is to write a dedicated subclass with the desired behavior.
2) If we are to do this, you will need to allow it for fv methods as well. And documentation of what actually happens, and which equation is solved must be in place. But don't bother until you have convinced me that this should actually be included.
3) You need to run black.

src/porepy/numerics/vem/dual_elliptic.py
```diff
@@ -318,17 +318,19 @@ def assemble_int_bound_flux(
...
         # The matrix must be the VEM discretization matrix.
         mg = data_edge["mortar_grid"]
 
         hat_E_int = self._velocity_dof(g, mg, grid_swap)
-        cc[self_ind, 2] += matrix[self_ind, self_ind] * hat_E_int
+        cc[self_ind, 2] += coefficient * matrix[self_ind, self_ind] * hat_E_int
 
     def assemble_int_bound_source(
-        self, g, data, data_edge, grid_swap, cc, matrix, self_ind
+        self, g, data, data_edge, grid_swap, cc, matrix, self_ind, coefficient = 1.
```
keileg: The above comment of course applies to all methods.

src/porepy/grids/mortar_grid.py
alessiofumagalli: OK, I'll put a comment. Also, I'm not tracking this attribute for changing in the mortar grid. I should modify the corresponding tests to do it..

src/porepy/numerics/vem/dual_elliptic.py
alessiofumagalli: OK, I see it will be difficult to convince you to but this update in the core :P
I would probably write a subclass in the paper folder, if this will be needed for other applications in the future we might move the modifications in the core.

src/porepy/grids/mortar_grid.py
alessiofumagalli: no need for the test after a careful check

src/porepy/numerics/vem/dual_elliptic.py
alessiofumagalli: Actually it's quite annoying to "just" inherit and make the change.. Cause I need to inherit again to obtain the implementation of the MVEM.. I can propose the change also for the FV solver? I checked it and it looks less cumbersome

keileg:
I agree with Ivar.

keileg:
Also see the codacy comments, in particular on the importer

alessiofumagalli:
> Also see the codacy comments, in particular on the importer

I opened a Pandora jar

alessiofumagalli:
Hey, can I merge it?

IvarStefansson:
> 
> 
> > Also see the codacy comments, in particular on the importer
> 
> I opened a Pandora jar

I know, and it seems to reveal only a limited number of issues at a time.

IvarStefansson:
> 
> 
> Hey, can I merge it?

As far as I am concerned, yes.

PR #206, opened 2019-01-31
Handle empty data for vtk export of single grid

keileg:

PR #209, opened 2019-02-04
when exporting a single grid with empty data, gives an empty dictiona…

alessiofumagalli:
…ry and not an empty list. Travis is not catching this problems since vtk is not installed.

Sorry Eirik, my bad..


keileg:
Didn't we discuss this related to #206?

alessiofumagalli:
yes, but I did a mistake. In the case of a single grid we should convert to a dictionary and not to a list. In fact the data in this case are passed as: {"field_name": field_values}

alessiofumagalli:
@keileg , can I proceed?

keileg:
I'm okay with merging if you know what you are doing

PR #211, opened 2019-02-04
Biot fixes

rbe051:
Neumann Boundary-conditions for Biot.

As a bonus, contact interface laws (contact mechanics for Biot) is also included.

### DIFF
--- a/src/porepy/numerics/fv/biot.py
+++ b/src/porepy/numerics/fv/biot.py
@@ -437,11 +437,10 @@ class Biot:
             g, subcell_topology, alpha, bound_exclusion_mech
         )
 
-        # Note that sgn_diag_F only flips the sign of boundary terms, see comment above.
         if subface_rhs:
-            grad_p = hook * igrad * rhs_jumps + sgn_diag_F * grad_p_face
+            grad_p = hook * igrad * rhs_jumps + grad_p_face
         else:
-            grad_p = hf2f * (hook * igrad * rhs_jumps + sgn_diag_F * grad_p_face)
+            grad_p = hf2f * (hook * igrad * rhs_jumps + grad_p_face)
         stabilization = div * igrad * rhs_jumps
 
         # We obtain the reconstruction of displacments. This is equivalent as for
@@ -603,13 +602,13 @@ class Biot:
         # Then the stress equilibrium for the Neumann subfaces.
         # Then the Robin subfaces.
         # And last, the displacement continuity on both internal and external subfaces.
-        rhs_int = bound_exclusion_mech.exclude_boundary(rhs_units)
-        rhs_neu = bound_exclusion_mech.keep_neumann(sgn_diag_C * rhs_units)
-        rhs_rob = bound_exclusion_mech.keep_robin(sgn_diag_C * rhs_units)
+        rhs_int = bound_exclusion.exclude_boundary(rhs_units)
+        rhs_neu = bound_exclusion.keep_neumann(sgn_diag_C * rhs_units)
+        rhs_rob = bound_exclusion.keep_robin(sgn_diag_C * rhs_units)
 
         num_dir_subface = (
-            bound_exclusion_mech.exclude_neu_rob.shape[1]
-            - bound_exclusion_mech.exclude_neu_rob.shape[0]
+            bound_exclusion.exclude_neu_rob.shape[1]
+            - bound_exclusion.exclude_neu_rob.shape[0]
         )
 
         # No right hand side for cell displacement equations.
@@ -646,7 +645,8 @@ class Biot:
         del vals, rows, cols
         
         # Prepare for computation of grad_p_face term  
-        grad_p_face = map_unique_subfno * nAlpha_grad * sc2c 
+        # Note that sgn_diag_F might only flip the boundary signs. See comment above.
+        grad_p_face = sgn_diag_F * map_unique_subfno * nAlpha_grad * sc2c 
 
         return rhs_jumps, grad_p_face
            
--- a/src/porepy/numerics/fv/biot.py
+++ b/src/porepy/numerics/fv/biot.py
@@ -75,10 +75,23 @@ class Biot:
         matrices_m = data[pp.DISCRETIZATION_MATRICES][self.mechanics_keyword]
         matrices_f = data[pp.DISCRETIZATION_MATRICES][self.flow_keyword]
 
+        bound_stress = matrices_m["bound_stress"]
+        bound_flux = matrices_f["bound_flux"]
+        if bound_stress.shape[0] != g.dim * g.num_faces:
+            # If the boundary conditions are given on the
+            # subfaces we have to map them to the faces
+            hf2f_nd = pp.fvutils.map_hf_2_f(g=g)
+            hf2f = pp.fvutils.map_hf_2_f(nd=1, g=g)
+            bound_stress = hf2f_nd * bound_stress
+            bound_flux = hf2f * bound_flux
+
         dt = data[pp.PARAMETERS][self.flow_keyword]["time_step"]
-        p_bound = -div_flow * matrices_f["bound_flux"] * p * dt
-        s_bound = -div_mech * matrices_m["bound_stress"] * d
-        return np.hstack((s_bound, p_bound))
+        p_bound = -div_flow * bound_flux * p * dt
+        s_bound = -div_mech * bound_stress * d
+        # Note that the following is zero only if the previous time step is zero.
+        # See comment in the DivD class
+        div_d_rhs = -0*matrices_m["bound_div_d"] * d
+        return np.hstack((s_bound, p_bound + div_d_rhs))
 
     def rhs_time(self, g, data):
         """ Time component of the right hand side (dependency on previous time
@@ -194,12 +207,24 @@ class Biot:
         matrices_m = data[pp.DISCRETIZATION_MATRICES][self.mechanics_keyword]
         matrices_f = data[pp.DISCRETIZATION_MATRICES][self.flow_keyword]
         # Put together linear system
-        A_flow = div_flow * matrices_f["flux"]
-        A_mech = div_mech * matrices_m["stress"]
+        if matrices_m["stress"].shape[0] != g.dim * g.num_faces:
+            # If we give the boundary conditions for subfaces, the discretization
+            # will also be returned for the subfaces. We therefore have to map
+            # everything to faces before we proceeds.
+            hf2f_nd = pp.fvutils.map_hf_2_f(g=g)
+            hf2f = pp.fvutils.map_hf_2_f(nd=1, g=g)
+            stress = hf2f_nd * matrices_m["stress"]
+            grad_p = hf2f_nd * matrices_m["grad_p"]
+            flux = hf2f * matrices_f["flux"]
+        else:
+            stress = matrices_m["stress"]
+            flux = matrices_f["flux"]
+            grad_p = matrices_m["grad_p"]
+        A_flow = div_flow * flux
+        A_mech = div_mech * stress
+        grad_p = div_mech * grad_p
         stabilization = matrices_f["biot_stabilization"]
 
-        grad_p = div_mech * matrices_m["grad_p"]
-        
         # Time step size
         dt = param[self.flow_keyword]["time_step"]
 
@@ -359,15 +384,17 @@ class Biot:
 
         # Define subcell topology
         subcell_topology = fvutils.SubcellTopology(g)
+        # If g is not already a sub-grid we create one
+        if bound_mech.num_faces == subcell_topology.num_subfno_unique:
+            subface_rhs = True
+        else:
+            # And we expand the boundary conditions to fit the sub-grid
+            bound_mech = pp.fvutils.boundary_to_sub_boundary(bound_mech, subcell_topology)
+            subface_rhs = False
+
         # Obtain mappings to exclude boundary faces for mechanics
-        bound_mech_sub = fvutils.boundary_to_sub_boundary(bound_mech, subcell_topology)
         bound_exclusion_mech = fvutils.ExcludeBoundaries(
-            subcell_topology, bound_mech_sub, nd
-        )
-        # ... and flow
-        bound_flow_sub = fvutils.boundary_to_sub_boundary(bound_flow, subcell_topology)
-        bound_exclusion_flow = fvutils.ExcludeBoundaries(
-            subcell_topology, bound_flow_sub, nd
+            subcell_topology, bound_mech, nd
         )
 
         # Call core part of MPSA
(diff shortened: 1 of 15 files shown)


src/porepy/numerics/fv/fv_elliptic.py
```diff
@@ -142,6 +142,10 @@ def assemble_matrix(self, g, data):
 
         div = pp.fvutils.scalar_divergence(g)
         flux = matrix_dictionary["flux"]
+        if flux.shape[0] != g.num_faces:
```
keileg: Is this a good way of doing this? I would prefer having a keyword that specifies that the operation should be done, if possible? 
Anyhow, method documentation  is necessary, so is comments in the code for anyone not very familiar with the mpfa implementation.

src/porepy/numerics/fv/fv_elliptic.py
```diff
@@ -169,6 +173,9 @@ def assemble_rhs(self, g, data):
             self.discretize(g, data)
 
         bound_flux = matrix_dictionary["bound_flux"]
+        if g.dim > 0 and bound_flux.shape[0] != g.num_faces:
```
keileg: Same comment as above.

src/porepy/numerics/fv/fvutils.py
```diff
@@ -584,6 +584,7 @@ def map_hf_2_f(fno=None, subfno=None, nd=None, g=None):
     ).tocsr()
     return hf2f
 
+
```
keileg: Does black do this? 

Codacy had quite a few complaints on trailing whitespaces etc.

src/porepy/numerics/fv/mpfa.py
```diff
@@ -463,6 +463,15 @@ def _local_discr(self, g, k, bnd, eta=None, inverter="numba", apertures=None):
         # discretization.
         subcell_topology = fvutils.SubcellTopology(g)
 
+        # Below, the boundary conditions should be defined on the subfaces.
```
keileg: This is a change in the functionality of mpfa, is it not? I am okay with it, but documentation is necessary.

src/porepy/numerics/fv/mpfa.py
```diff
@@ -463,6 +463,15 @@ def _local_discr(self, g, k, bnd, eta=None, inverter="numba", apertures=None):
...
+        if bnd.num_faces == subcell_topology.num_subfno_unique:
+            # The boundary conditions is already given on the subfaces
+            subface_rhs = True
+        else:
+            # If bnd is not already a sub-face_bound we extend it
+            bnd = pp.fvutils.boundary_to_sub_boundary(bnd, subcell_topology)
```
keileg: When I see how much this cleaned up below, I really like it.

src/porepy/numerics/fv/mpfa.py
```diff
@@ -691,24 +674,42 @@ def _local_discr(self, g, k, bnd, eta=None, inverter="numba", apertures=None):
...
+        # )
         # For Dirichlet faces, simply recover the boundary condition
         bound_pressure_face_dir = sps.diags(bnd.is_dir.astype(np.int))
 
         bound_pressure_face = (
             bound_pressure_face_dir + remove_not_neumann * bound_pressure_face_neu
         )
 
+        if not subface_rhs:
+            # In this case we set the value at a face, thus, we need to distribute the
```
keileg: If returning boundary discretization in terms of subfaces is an option, how do you distinguish between different subfaces on a face? Say I want to assign a bc u = x + y, and use a certain eta for the continuity point, can I then assign the right bc value to the right subface?

src/porepy/numerics/fv/mpfa.py
```diff
@@ -828,9 +837,9 @@ def _create_bound_rhs(
         is_rob = np.logical_and(bnd.is_rob, np.logical_not(bnd.is_internal))
 
         fno = subcell_topology.fno_unique
-        num_neu = np.sum(is_neu[fno])
-        num_dir = np.sum(is_dir[fno])
-        if not num_rob == np.sum(is_rob[fno]):
+        num_neu = np.sum(is_neu)
+        num_dir = np.sum(is_dir)
+        if not num_rob == np.sum(is_rob):
             raise AssertionError()
```
keileg: Error description please.

src/porepy/numerics/fv/mpsa.py
```diff
@@ -384,6 +384,14 @@ def assemble_int_bound_displacement_trace(
                 * matrix_dictionary["bound_displacement_face"]
                 * proj_int.T
             )
+            cc[2, 2] -= (
```
keileg: This imposes a symmetric stress? Short comment please.

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -723,4 +725,260 @@ def assemble_matrix_rhs(
...
+
+        """
+        matrix_dictionary_edge = data_edge[pp.DISCRETIZATION_MATRICES][self.keyword]
+
+        self.discretize(g_master, g_slave, data_master, data_slave, data_edge)
+
+        if not g_master.dim == g_slave.dim:
+            raise AssertionError("Slave and master must have same dimension")
+
+        master_ind = 0
```
keileg: This I like, I should have used it in similar functions.

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -723,4 +725,260 @@ def assemble_matrix_rhs(
...
+            data_edge: Data dictionary for the edge between the subdomains
+            matrix_master: original discretization for the master subdomain
+            matrix_slave: original discretization for the slave subdomain
+
+        """
+        matrix_dictionary_edge = data_edge[pp.DISCRETIZATION_MATRICES][self.keyword]
+
+        self.discretize(g_master, g_slave, data_master, data_slave, data_edge)
+
+        if not g_master.dim == g_slave.dim:
```
keileg: This means that for a fracture problem, g_master == g_slave? Not necessarily, but in practice.

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -723,4 +725,260 @@ def assemble_matrix_rhs(
...
+        # EK: For some reason, the following lines were necessary to apease python
+        rhs = np.empty(3, dtype=np.object)
+        rhs[master_ind] = np.zeros(matrix[master_ind, master_ind].shape[1])
+        rhs[slave_ind] = np.zeros(matrix[slave_ind, slave_ind].shape[1])
+        rhs[2] = np.zeros(self.ndof(mg))
+
+        # The convention, for now, is to put the master grid information
+        # in the first column and row in matrix, slave grid in the second
+        # and mortar variables in the third
+        # If master and slave is the same grid, they should contribute to the same
```
keileg: Sounds like the assembler should be reworked slightly. We can discuss this later. I presume this may also cause problems for periodic bcs for a flow problem.

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -723,4 +725,260 @@ def assemble_matrix_rhs(
...
+        # row and coloumn. When the assembler assigns matrix[idx] it will only add
+        # the slave information because of duplicate indices (master and slave is the same).
+        # We therefore write the both master and slave info to the slave index.
+        if g_master == g_slave:
+            master_ind = 1
+        else:
+            master_ind = 0
+
+        # Obtain the contribution of the cell centered pressure on the displacement
```
keileg: I am very happy with the level of comments here.

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -723,4 +725,260 @@ def assemble_matrix_rhs(
...
+            g_master, data_edge, matrix, False, master_ind
+        )
+        self.discr_slave.enforce_neumann_int_bound(
+            g_slave, data_edge, matrix, True, slave_ind
+        )
+
+        return matrix, rhs
+
+
+class DivU_StressMortar(RobinContactBiotPressure):
```
keileg: Is mpsa.py the right module for this function? Not that I have an immediate better suggestion.

src/porepy/utils/sort_points.py
```diff
@@ -184,16 +184,13 @@ def update_queue(pair_0, pair_1):
             )
         )
         hit_old = np.logical_and.reduce(
-            (
-                is_ordered,
-                np.any(t == q[0], axis=0),
-                np.any(t == q[1], axis=0),
-            )
+            (is_ordered, np.any(t == q[0], axis=0), np.any(t == q[1], axis=0))
```
keileg: It seems we have slightly different versions of black - both the old and the new versions are obviously automatically formatted, but in different ways.

Anyhow, I can see there is some cleaning to do in this file; I will take care of it.

test/integration/test_elliptic_interface_laws.py
```diff
@@ -132,21 +154,281 @@ def assign_parameters(self, gb, mortar_weight, robin_weight, rhs):
...
+            RW = sps.diags(robin_weight)
+            data = {
+                "mortar_weight": [MW] * mg.num_cells,
+                "robin_weight": [RW] * mg.num_cells,
+                "robin_rhs": np.tile(rhs, (mg.num_cells)),
+            }
+            pp.initialize_data(mg, d, self.kw, data)
+
+
+class TestBiotTwoGridCoupling(unittest.TestCase):
```
keileg: While we are waiting for someone to write a user manual - haha: It seems this is the best example on how to set up the contact problem. Could you make a comment to that effect, or to a better place, somewhere in the code, maybe somewhere in biot.py, so that people like me can find it?

src/porepy/numerics/fv/biot.py
```diff
@@ -75,10 +75,23 @@ def rhs_bound(self, g, data):
         matrices_m = data[pp.DISCRETIZATION_MATRICES][self.mechanics_keyword]
         matrices_f = data[pp.DISCRETIZATION_MATRICES][self.flow_keyword]
 
+        bound_stress = matrices_m["bound_stress"]
+        bound_flux = matrices_f["bound_flux"]
+        if bound_stress.shape[0] != g.dim * g.num_faces:
```
keileg: Same comment as for mpfa/mpsa: If bcs by subface are permissible, this must be documented. See also below.

src/porepy/numerics/fv/biot.py
```diff
@@ -243,8 +269,7 @@ def _discretize_mech(self, g, data):
         Discretization of poro-elasticity by the MPSA-W method.
 
         Implementation needs (in addition to those mentioned in mpsa function):
-            1) Fields for non-zero boundary conditions. Should be simple.
-            2) Split return value grad_p into forces and a divergence operator,
+            1) Split return value grad_p into forces and a divergence operator,
```
keileg: Tada!

src/porepy/numerics/fv/biot.py
```diff
@@ -841,6 +892,68 @@ def assemble_rhs(self, g, data):
...
         return np.zeros(self.ndof(g))
 
+    def assemble_int_bound_displacement_trace(
+        self, g, data, data_edge, grid_swap, cc, matrix, self_ind
+    ):
+        """ Assemble the contribution from the pressure to the the trace of the
+        displacement on internal boundaries.
+
+        The intended use is when the internal boundary is coupled to another
+        node. Specific usage depends on the
```
keileg: Node in the grid_bucket sense.

src/porepy/numerics/fv/fv_elliptic.py
rbe051: I think this is the most consistent way of defining it as of now. Specifying values on subfaces is for expert users anyhow. I think you are right about including it as a keyword, but I think it is necessary to have a new discussion about the parameter class an whether this fits into that one.

src/porepy/numerics/fv/mpfa.py
rbe051: It is not a change in the functionality, it is just rewriting the code. We now have bnd.is_neu everywhere instead of bnd.is_neu[subcell_topology.fno_unique].

src/porepy/numerics/fv/mpfa.py
rbe051: exactly! It is basically just committing to the idea of subfaces, also on the boundary

src/porepy/numerics/fv/mpfa.py
rbe051: Short answer; yes.
Long answer: I think this is where all your complaints about documentation above leads to. This part should be properly documented, partly here, but it is also common with MPSA. The difficulties arises since this is more a hack to be able to set subface boundary conditions, and will only be used by people that are _very_ familiar with mps/fa. To make it something other than a hack I think we would need to introduce a proper concept of a subgrid (not just subcell_topology). The boundary condition will be defined at the continuity points according to subcell_topology.subfno_unique. So for your example:
```python
# Find the continuity points
eta = 1/3
cont_pnt = g.face_centers[:g.dim, s_t.fno_unique] + eta * (
    g.nodes[:g.dim, s_t.nno_unique] - g.face_centers[:g.dim, s_t.fno_unique]
    )
bc_val = cont_pnt[0] + count_pnt[1]
```

src/porepy/numerics/fv/mpsa.py
rbe051: The comment is in the else part of the if/else. Will add a comment here also

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
rbe051: No, g_master can be a different domain than g_slave, both in theory and practice.

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
rbe051: this is the elliptic_interface_law_module. Not sure if this is the right one either

test/integration/test_elliptic_interface_laws.py
rbe051: Do you want it immediately? Otherwise, I think I will make a tutorial. Also the code I'm writing for the paper will probably be a good place to look when that is public

src/porepy/numerics/fv/mpfa.py
keileg: But can you also specify bcs per halfface?

rbe051: yes

src/porepy/numerics/fv/fv_elliptic.py
keileg: Then, can you leave it as a implementation note at the top of some function (not in the method documentation, but right below)?

test/integration/test_elliptic_interface_laws.py
keileg: If you make a tutorial, I will be very happy. Runscript is also very good.

keileg:
Except my wish for an implementation note on subface bcs, I'm very happy with this PR.

test/integration/test_elliptic_interface_laws.py
rbe051: Biot with Neumann boundaries will at least become a tutorial. Contact mechanics might be a tad to complicated to fit nicely into a tutorial

rbe051:
Happy @keileg ?

keileg:
Merge when you're done.

PR #215, opened 2019-02-12
Update benchmark

alessiofumagalli:
in this branch we bring back to the current version of pp the work done for the benchmark 3d. It should be mostly local to the related folder.


src/porepy/numerics/fv/fvutils.py
```diff
@@ -1310,6 +1310,7 @@ def map_subgrid_to_grid(g, loc_faces, loc_cells, is_vector):
 def compute_darcy_flux(
     gb,
     keyword="flow",
+    keyword_store=None,
```
rbe051: I think we need a consistent way of doing this. Not this PR though, but something to think about

IvarStefansson:
Any idea about the 3.5 Travis failure?

keileg:
Travis failed on setup before porepy. I tried to restart the job, fingers crossed

src/porepy/numerics/fv/fvutils.py
keileg: I agree, this is rudimentary at the moment. We should also implement restart. I have some ideas about combining this with vtk output, but ideas are most welcome. Not in this PR, though.

PR #221, opened 2019-02-20
Improved vtk time step exporter

keileg:
The exporter used the time step both to count time (as displayed by paraview) and to extend file names to differ between time steps. The latter required the time step to be an int. We remove this, let the file extension by default run from 0 (but can be overridden), and set the time as displayed by paraview when exporting the vtk.

The setup works on my local setup (no unit tests for vtk are run on travis).

### DIFF
--- a/src/porepy/viz/exporter.py
+++ b/src/porepy/viz/exporter.py
@@ -231,6 +231,12 @@ class Exporter:
         if self.fixed_grid:
             self._update_gb_VTK()
 
+        # Counter for time step. Will be used to identify files of individual time step,
+        # unless this is overridden by optional parameters in write_vtk
+        self._time_step_counter = 0
+        # Storage for file name extensions for time steps
+        self._exported_time_step_file_names = []
+
     # ------------------------------------------------------------------------------#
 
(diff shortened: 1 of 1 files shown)


src/porepy/viz/exporter.py
```diff
@@ -310,19 +333,13 @@ def write_pvd(self, time):
...
-
         if self.is_GridBucket:
-            [
-                o_file.write(fm % (time[t], self._make_file_name(self.name, t, dim)))
-                for t in time_step
-                for dim in self.dims
-            ]
+            for time, fn in zip(timestep, file_extension):
+                for dim in self.dims:
+                    o_file.write(fm % (time, self._make_file_name(self.name, fn, dim)))
```
alessiofumagalli: more readable :)

alessiofumagalli:
It looks ok, is this part mandatory or we can still use what we already have?

src/porepy/viz/exporter.py
keileg: Indeed.

keileg:
Yes, the old setup should still work.

PR #223, opened 2019-02-21
Flux projections from mortars in fv_elliptic should integrate, not average

keileg:


rbe051:
If you think this is correct it is good. Ok, with merge when the unittests run

keileg:
This turned out to be rather involved. I will close the PR now, and then possibly return to it later.

PR #231, opened 2019-02-27
fix exporter due to a small bug introduced for non-time dependent pro…

alessiofumagalli:
…blems

It should be immediate, just a small fix


keileg:
What does it do in practice?

alessiofumagalli:
for steady problem the pvd had wrong names of the vtu files

PR #233, opened 2019-02-27
name fix in dfn_transport discretization.py

alessiofumagalli:


alessiofumagalli:
not really needed a review

PR #234, opened 2019-03-05
Bug fix in mpfa

IvarStefansson:
The sign of cell contribution to face pressure reconstruction is now compatible with the face contribution. This part might lack proper testing, since the contribution is zero for Cartesian grids.

### DIFF
--- a/src/porepy/numerics/fv/mpfa.py
+++ b/src/porepy/numerics/fv/mpfa.py
@@ -694,6 +694,7 @@ class Mpfa(FVElliptic):
             subface_rhs,
         )
         # Discretization of boundary values
+
         bound_flux = darcy * igrad * rhs_bound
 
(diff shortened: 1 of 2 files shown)


src/porepy/numerics/fv/mpfa.py
```diff
@@ -730,6 +730,7 @@ def _local_discr(
         sgn_arr[bound_faces] = g.cell_faces[bound_faces].sum(axis=1).A.ravel()
```
keileg: Could you leave one line of comment why the sign corrections are necessary, and what they do?

IvarStefansson: The honest answer is that I'm not entirely sure, but now it's consistent with the face contribution. Do you want me to write this as a comment?

PR #241, opened 2019-03-22
Normal diffusivity

IvarStefansson:
1) Moves the factor 2 from the Robin_coupling to the run scripts (when assigning normal_diffusivity on the edges, it should be multiplied by two).
2) Making sure we use the aperture, not the cross-sectional area ("epsilon", in the terminology of the benchmark paper, or "specific volume" in our new terminology(!)), for scaling of the assigned normal_diffusivity

@alessiofumagalli,  could I ask you to review this thoroughly, please?


examples/papers/dfn_transport/example2/pot.py
```diff
@@ -190,39 +139,17 @@ def main():
...
         for method in methods_alessio:
-            data = (
-                folder_in
-                + method
-                + "/"
-                + "Cmax_"
-                + grid[0]
-                + ".csv"
-            )
+            data = folder_in + method + "/" + "Cmax_" + grid[0] + ".csv"
```
keileg: This looks to me like we run different versions of black. Can you please check your versions? I would like to understand better how this thing evolves.

test/integration/test_mpfaMultiDim.py
```diff
@@ -34,9 +34,7 @@ def setup_cart_2d(nx):
     for e, d in gb.edges():
         # Compute normal permeability
         gl, _ = gb.nodes_of_edge(e)
-
-        d_l = gb.node_props(gl)
-        kn = 1.0 / np.mean(d_l[pp.PARAMETERS][kw]["aperture"])
+        kn = 1.0 / aperture
```
keileg: I prefer defining common variables outside loops, instead of defining inside, and then pick it up later. But that is a matter of taste.

test/integration/test_tpfaMultiDim.py
```diff
@@ -24,11 +24,11 @@ def setup_2d_1d(nx, simplex_grid=False):
 
     gb.compute_geometry()
     gb.assign_node_ordering()
+    aperture = 0.01 / np.max(nx)
```
keileg: .. I mean, like this.

test/integration/test_mpfaMultiDim.py
IvarStefansson: I agree. I just didn't bother to make more changes than I had to...

IvarStefansson: And it's not purely a matter of taste. I would argue that defining variables multiple times increases the risk of bugs.

examples/papers/dfn_transport/example2/pot.py
alessiofumagalli: black, version 18.9b0

examples/papers/flow_upscaling/example1/data.py
```diff
@@ -50,13 +50,11 @@ def flow(gb, model, data):
...
     for e, d in gb.edges():
-        g_l = gb.nodes_of_edge(e)[0]
         mg = d["mortar_grid"]
         check_P = mg.slave_to_mortar_avg()
-
-        aperture = gb.node_props(g_l, pp.PARAMETERS)[model_data]["aperture"]
-        gamma = check_P * np.power(aperture, 1 / (2.0 - g.dim))
-        kn = data["kf"] / mu * np.ones(mg.num_cells) / gamma
+        aperture = data["aperture"]
+        gamma = check_P * aperture
```
alessiofumagalli: gamma does not involve the power term, is this correct?

IvarStefansson: I don't know where the name "gamma" comes from. But the aperture which the normal permeability should be scaled with has dimensions of length, and should not have be raised to the power (dim_max - dim), no.

PR #243, opened 2019-03-25
Biot

rbe051:
Fixed sign mistake in Biot as well as some minor improvements


src/porepy/numerics/fv/biot.py
```diff
@@ -1078,14 +1078,22 @@ def assemble_rhs(self, g, data):
...
         rhs_bound = (
             -matrix_dictionary["bound_div_d"] * (d_bound_1 - d_bound_0) * biot_alpha
         )
 
         # Time part
-        d_cell = parameter_dictionary["state"]
+        if "bc_values" in parameter_dictionary["state"]:
+            d_cell = parameter_dictionary["state"]["displacement"]
+        else:
+            d_cell = parameter_dictionary["state"]
```
rbe051: @IvarStefansson @keileg : 
1: What do you think about this way of setting the information from the previous time step (dictionary["state"]["'displacement"] instead of dictionary["state"]. The reason for this is that I also need access to dictionary["state"]["bc_values"] in time dependent problems)

I don't think this is the perfect way to do this, but it fits into the framework.
2:
Should we remove the if-else statement and require the parameter_dictionary["state"] to be a dictionary?

src/porepy/numerics/mixed_dim/assembler.py
```diff
@@ -327,12 +327,22 @@ def assemble_matrix_rhs(
...
+                    # nodes
+                    if matrix[mat_key_master][mi, si] is None:
+                        matrix[mat_key_master][mi, si] = tmp_mat[0, 1]
+                    else:
+                        matrix[mat_key_master][mi, si] += tmp_mat[0, 1]
+
+                    if matrix[mat_key_slave][si, mi] is None:
+                        matrix[mat_key_slave][si, mi] = tmp_mat[1, 0]
+                    else:
+                        matrix[mat_key_slave][si, mi] += tmp_mat[1, 0]
```
rbe051: @keileg What do think about this? This allows for communication between master and slave without going through a mortar variable. I had to do this for my coupled biot-contact mechanics-flow in fractures model to work (I could probably get around it, but this implementation was very convenient). Should we keep this or revert it back to how it was?

keileg: In principle I am not at all happy with allowing for direct communication, it allows for all kinds of hacks that are not present in the equations. The functional framework requires that a mortar variable is involved in the communication, and I would like to stick with that.

Also, having spent the day working on the equations for contact mechanics, with and without flow, it is not clear to me why you need it - from what I can see, this should all work out with communication via the mortars. Which term is it you want to do this for?

keileg:
@alessiofumagalli : Please run black before merging.

src/porepy/numerics/fv/biot.py
IvarStefansson: 1) I could think of alternative placements. The current is
data["PARAMETERS"]["state"]
I would say that 
data["state"]
is more natural, in a way. Or even specify to "variable_state" or something. In this case, we might also consider using upper case for consistency.
2) I agree with Eirik.

rbe051: > 1.: Do we have a structure for storing primary (and secondary) variables during time stepping? If not, I think this is a fair approach, in line with other things we are doing.
no we do not

> 2. If we go with the state, I definitely vote for requiring that the structure be followed.
I agree

> 1. I could think of alternative placements. The current is
>    data["PARAMETERS"]["state"]
>    I would say that
>    data["state"]
>    is more natural, in a way. Or even specify to "variable_state" or something. In this case, we might also consider using upper case for consistency.

Not variable_state. 1. bc_values is not a variable. 2. Not only variables are time dependent. E.g., apertures for slip, permeabilities for two phase flow...

src/porepy/numerics/mixed_dim/assembler.py
rbe051: It is not strictly needed, but used for convenience. Also, the assembler gives the impression that this is allowed: the coupling discretization gets a 3x3 block matrix and nothing tells you that you can not fill out all the blocks. But I agree, lets remove this from this PR

PR #248, opened 2019-04-05
update the fracture_sets with the new interface for the importing + o…

alessiofumagalli:
…ther updates for flow_upscaling


src/porepy/fracs/fractures_2d.py
```diff
@@ -479,6 +480,27 @@ def plot(self, **kwargs):
         """
         pp.plot_fractures(self.domain, self.pts, self.edges, **kwargs)
 
+    def to_csv(self, file_name):
```
keileg: I agree, this is something we should have.

alessiofumagalli: :D great! thanks

PR #251, opened 2019-04-11
Pressure reconstruction

rbe051:
Pressure reconstruction for mpfa on all subfaces. Previously it was only done on the boundary faces.


test/unit/test_pressure_reconstruction.py
```diff
@@ -0,0 +1,760 @@
+import numpy as np
```
keileg: There is a test file, uni/test_fv_boundary_pressure_reconstrution or similar, which already contains test. Merging that file with this one would be good.

PR #255, opened 2019-05-15
State keyword

keileg:


src/porepy/utils/common_constants.py
```diff
@@ -65,7 +65,7 @@
 MILLIDARCY = MILLI * DARCY
 
 PASCAL = 1.0
-BAR = 101325 * PASCAL
+BAR = 100000 * PASCAL
 ATMOSPHERIC_PRESSURE = BAR
```
jhabriel: Eirik,

The atmospheric pressure on line 69 should be:

ATMOSPHERIC_PRESSURE = 101325 * PASCAL

PR #258, opened 2019-05-23
State keyword

IvarStefansson:
Introduced the pp.STATE keyword in the Biot class. The idea is to store data from previous time steps in d[pp.STATE]. 
We introduce the variable to the discretisation (on initialisation), so that it can access it during discretisation and assembly. For Biot, this means that a Biot object will access the pressure as
p = d[pp.STATE][self.scalar_variable]
Furthermore, parameters from previous may also be stored. If the old flow BC values are needed, they should be stored in
d[pp.STATE][self.flow_keyword]["bc_values"].

A version of the above documentation/explanation is also found in data.py, where I also included the function set_state, intended for initialising or updating the state (similar to initialize_data).


src/porepy/numerics/fv/biot.py
```diff
@@ -104,21 +111,23 @@ def rhs_time(self, g, data):
         Parameters:
             g: grid, or subclass, with geometry fields computed.
             data: dictionary to store the data terms. Must have been through a
-                call to discretize() to discretization of right hand side.
-            state: np.ndarray optional, solution vector from previous time
-                step. Defaults to zero.
+                call to discretize() to discretization of right hand side. May
+                contain the field "STATE", storing the solution vectors from previous
```
rbe051: "STATE" -> pp.STATE

src/porepy/numerics/fv/biot.py
```diff
@@ -965,6 +974,21 @@ class DivD(
...
 
+    def __init__(
+        self, keyword="mechanics", variable="displacement", mortar_variable="traction"
+    ):
+        """ Set the two keywords.
+
+        The keywords are used to access and store parameters and discretization
+        matrices.
+        """
+        self.keyword = keyword
```
rbe051: Should call parent init

src/porepy/numerics/fv/biot.py
```diff
@@ -1164,6 +1188,16 @@ class BiotStabilization(
     """ Class for the stabilization term of the Biot equation.
     """
 
+    def __init__(self, keyword="mechanics", variable="pressure"):
```
rbe051: parent init

src/porepy/numerics/fv/biot.py
```diff
@@ -1079,14 +1103,14 @@ def assemble_rhs(self, g, data):
...
-        d_bound_0 = parameter_dictionary["state"]["bc_values"]
+        d_bound_0 = data[pp.STATE][self.keyword]["bc_values"]
         biot_alpha = parameter_dictionary["biot_alpha"]
         rhs_bound = (
             -matrix_dictionary["bound_div_d"] * (d_bound_1 - d_bound_0) * biot_alpha
         )
 
         # Time part
-        d_cell = parameter_dictionary["state"]["displacement"]
+        d_cell = data[pp.STATE][self.variable]
```
rbe051: You have data[pp.STATE][self.keyword]["bc_values"], but here it is only data[pp.STATE][self.variable]. I understand why, but it is a bit inconsistent

src/porepy/params/data.py
```diff
@@ -50,6 +50,17 @@
...
+
+Also contains a function for setting the state. The state is all data associated with
+the previous time step or iteration, and is stored in data[pp.STATE]. The solution of a
+variable is stored in
+
+data[pp.STATE][variable_name],
+
+whereas data such as BC values are stored similarly to in the Parameters class, in
+
+data[pp.STATE][keyword]["bc_values"].
```
rbe051: Maybe
data[pp.STATE][variable_name]["values"] and
data[pp.STATE][variable_name]["bc_values"] ??
or:
data[pp.STATE][keyword][variabel_name] and
data[pp.STATE][keyword]["bc_values"] ??

or: 
data[pp.STATE][variable_name]
data[pp.STATE][bc_values_name] ??

PR #271, opened 2019-05-31
Ad assembler

rbe051:
I don't know if we actually wan to merge this. I think it either has to be better integrated into the assembler (especially considering the latest updates to the assembler), be added as a separate assembler, or just be hidden in some run-script.


src/porepy/grids/grid_bucket.py
```diff
@@ -1059,6 +1059,72 @@ def all_dims(self):
...
+        """
+        Get the cell volumes of all cells of the grid bucket, considering a loop
+        on all the grids.  It is possible to specify a condition based on the
+        grid to select some of them.
+
+        Parameter:
+            cond: optional, predicate with a grid as input.
+
+        Return:
+            num_cells: the total number of cells of the grid bucket.
```
keileg: The documentation is not right, is it?

src/porepy/grids/grid_bucket.py
```diff
@@ -1059,6 +1059,72 @@ def all_dims(self):
...
+        """
+        Get the face centers of all faces of the grid bucket, considering a loop
+        on all the graph nodes.  It is possible to specify a condition based on the
+        grid to select some of them.
+
+        Parameter:
+            cond: optional, predicate with a grid as input.
+
+        Return:
+            num_cells: the total number of cells of the grid bucket.
```
keileg: Same comment regarding documentation

src/porepy/numerics/fv/tpfa.py
```diff
@@ -80,6 +80,8 @@ def discretize(self, g, data, faces=None):
         if g.dim == 0:
             matrix_dictionary["flux"] = sps.csr_matrix([0])
             matrix_dictionary["bound_flux"] = 0
+            matrix_dictionary["bound_pressure_cell"] = sps.csr_matrix([1])
```
keileg: Good.

keileg:
Is it correct that it will make assembly of AD systems much easier, but that you still need to know what you are doing? I think it is worthwhile to bring it into the assembler - at least until we have a better place to put nonlinear stuff, then we may adjust, or kill, as we learn more about this.

rbe051:
> Is it correct that it will make assembly of AD systems much easier, but that you still need to know what you are doing? I think it is worthwhile to bring it into the assembler - at least until we have a better place to put nonlinear stuff, then we may adjust, or kill, as we learn more about this.

Yes, I guess you can say this is for more advanced users. I don't think it should be to difficult to use, but lets say it is still under heavy development and the syntax and behavior will change. The only part I am a bit worried about is the ordering of the system.  The ad variables need to have the same ordering as the block matrices, and this is now given by how networkx iterate over the grid bucket. We could let the assembler give the ordering, but then the assembler also have to initiate the ad variables. I think it might be worthwhile to have a separate  ad assembler for the non-linear stuff. 

This was mostly random thoughts I have made, but let us keep it like this for now.

PR #273, opened 2019-06-07
Flow upscaling update

alessiofumagalli:
few updates for the dfn_transport and pp paper


alessiofumagalli:
Very little has changed in the core (do not consider the files in dfn_transport), basically only the coarsening. We have saved also the original grid, before doing the actual coarsening, to be able to use Ivar matlab scripts from the benchmark which require simplicial grids.

PR #276, opened 2019-06-13
Enhance 1d grid intersections

rbe051:
I enhanced the merge_1d_grids so that it can merge general 1d grids.

I belive it previously had an assumption on the node, face, cell ordering that were not true for intersecting 1d grids. But please review this carefully, as I might have been unlucky.


src/porepy/fracs/mortars.py
```diff
@@ -233,35 +240,36 @@ def match_grids_1d(new_1d, old_1d, tol):
...
-    if switch_new:
-        new_ind = new_ind[::-1]
-    switch_old = old_ind[0] > old_ind[-1]
-    if switch_old:
-        old_ind = old_ind[::-1]
-
-    diff_new = np.diff(new_ind)
-    diff_old = np.diff(old_ind)
-    new_in_full = rldecode(np.arange(diff_new.size), diff_new)
-    old_in_full = rldecode(np.arange(diff_old.size), diff_old)
```
rbe051: I replace 100% of this function. I belive the issue might partly be in these two lines. This seems to assume that the the nodes and cells are ordered as given from pp.TensorGrid. Also, I belive there is some issues with non_conforming.merge_1d_grids(..) when the grid has two nodes at the same point (i.e. a 0d intersection).

I believe the new code can handle this

keileg:
Looks good to me. I added some comments while reading through the code, and also moved a function in geometry.intersections, so that it is close to its natural cousins.

PR #279, opened 2019-06-20
Mpfa gravity

mstarnoni:
Added consistent discretization of MPFA in the presence of gravity.
The div_g-discretizer works similar to grad_p in Biot.
A boolean variable gravity (default False) is set to indicate whether the div_g-term is to be returned.


mstarnoni:
All mpfa integration tests now fail because there are 5 values to unpack instead of 4. Should I modify all these tests adding a blank space in place of div_g, i.e. flux, bound_flux, _, _, _ 
 = pp.Mpfa("flow")._local_discr(...
or should I change the implementation?

src/porepy/numerics/fv/fvutils.py
```diff
@@ -584,6 +584,34 @@ def map_hf_2_f(fno=None, subfno=None, nd=None, g=None):
     ).tocsr()
     return hf2f
 
+def sc_2_c(nd, sub_cell_index, cell_node_blocks, num_subhfno):
```
rbe051: This function does exactly the same as the function: map_sc_2_c ?
If so, can we merge them, e.g., include the length of the vector as dim:
 def map_sc_2_c(nd, sub_cell_index, cell_index, dim):?

Or call the functions sc_2_c_scalar and sc_2_c_vector?

src/porepy/numerics/fv/fvutils.py
```diff
@@ -584,6 +584,34 @@ def map_hf_2_f(fno=None, subfno=None, nd=None, g=None):
...
+    """
+
+    num_cells = cell_node_blocks.max() + 1
+
+    rows = sub_cell_index.ravel('F')
+    cols = expand_indices_nd(cell_node_blocks, nd)
+    vals = np.ones(num_subhfno)
+    sc2c = sps.coo_matrix((vals, (rows, cols)),
+                               shape=(num_subhfno,
+                                      num_cells * nd)).tocsr()
```
rbe051: From the dimensions of this matrix, it looks like sc2c is a mapping from cells to subcells. See also comment below

src/porepy/numerics/fv/mpfa.py
```diff
@@ -870,7 +904,116 @@ def _local_discr(
             pressure_trace_bound = hf2f * area_mat * pressure_trace_bound * hf2f.T
             pressure_trace_cell = hf2f * area_mat * pressure_trace_cell
 
-        return flux, bound_flux, pressure_trace_cell, pressure_trace_bound
+        if gravity:
+            # in this case, we consider the full darcy's law in the presence of gravity
+            # we apply the discrete flux operator to the gravity term
+            discr_div_g = self.discretize_div_g(g, subcell_topology, bound_exclusion, darcy, igrad, nk_grad_all, nk_grad_paired)
+
+            # Output should be on cell-level (not sub-cell)
```
rbe051: Input?

src/porepy/numerics/fv/mpfa.py
```diff
@@ -870,7 +904,116 @@ def _local_discr(
...
+            sc2c = fvutils.sc_2_c(g.dim, sub_cell_index, cell_node_blocks[0], subcell_topology.subhfno.size)
+
+            div_g = hf2f * discr_div_g * sc2c
```
rbe051: Gravity should be cell-centered? So sc2c is a mapping from cells -> sub cells?

src/porepy/numerics/fv/mpfa.py
```diff
@@ -870,7 +904,116 @@ def _local_discr(
...
+        else:
+            div_g = None
+        
+        return flux, bound_flux, div_g, pressure_trace_cell, pressure_trace_bound
+
+    def discretize_div_g(self, g, subcell_topology, bound_exclusion, darcy, igrad, nk_grad_all, nk_grad_paired):
+        """
+        Consistent discretization of div_g-term in MPFA-O method.
+        For more details, see Starnoni et al (2019), Consistent discretization
```
rbe051: Do we want to write out the full reference?

src/porepy/numerics/fv/mpfa.py
```diff
@@ -870,7 +904,116 @@ def _local_discr(
...
+        Thus (i)-(iii) can be inverted to express the additional pressure gradients
+        due to imbalance in gravitational forces as in terms of the cell center variables.
+        Thus we can compute the basis functions 'div_g_jumps' on the sub-cells.
+        To ensure flux continuity, as soon as a convention is chosen for what side
+        the flux evaluation should be considered on, an additional term, called
+        'div_g_face', is added to the full flux. This latter term represents the flux
+        due to cell-center gravity acting on the face from the chosen side.
+        The pair subfno_unique-unique_subfno gives the side convention.
+        The full flux on the face is therefore given by
+        q = flux * p + bound_flux * p_b + (div_g_jumps + div_g_face) * g
```
rbe051: Maybe mention that the output is div_g = div_g_jumps + div_g_faces

test/unit/test_mpfa_gravity.py
```diff
@@ -0,0 +1,113 @@
...
+
+p = p0 + sympy.Piecewise(((1-y)*g1, y>=0.5), (0.5*g1+(0.5-y)*g2, y<0.5))
+gx = sympy.diff(p, x)
+gy = sympy.diff(p, y)
+
+p_f = sympy.lambdify((x, y), p, 'numpy')
+gx_f = sympy.lambdify((x, y), gx, 'numpy')
+gy_f = sympy.lambdify((x, y), gy, 'numpy')
+
+deviation_from_plane_tol=1e-5-5
```
rbe051: I think it would be better to put all of these "global variables" in the __init__ of the test class

src/porepy/numerics/fv/fvutils.py
mstarnoni: they do both mapping from subcells to cells, I'm pretty sure there is a way to merge them

src/porepy/numerics/fv/mpfa.py
```diff
@@ -166,6 +178,9 @@ def mpfa(
                 will be fluxes induced by the prescribed pressure. Incorporation as
                 a right hand side in linear system by multiplication with
                 divergence operator.
+            scipy.sparse.csr_matrix (shape num_faces, num_cells*dim): discretization of
```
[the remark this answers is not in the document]
rbe051: I was thinking about it, but could not agree with myself. This is whatever modifies the pressure in Darcy's law right? :
u = K grad (p - g)
I'm not sure about source as this is so much linked to the mass conservation equation. But I do agree that a more general name than gravity would be better.

src/porepy/numerics/fv/mpfa.py
```diff
@@ -870,7 +904,116 @@ def _local_discr(
...
+        # as a force on the faces. The right hand side is thus formed of the
+        # unit vector.           
+        vals = np.ones(num_subfno_unique)
+        rows = subcell_topology.subfno_unique 
+        cols = subcell_topology.subfno_unique
+        rhs_units = sps.coo_matrix((vals, (rows, cols)), 
+                                 shape=(num_subfno_unique,
+                                        num_subfno_unique))
+
+        rhs_units = bound_exclusion.exclude_dirichlet(rhs_units)
```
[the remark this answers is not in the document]
rbe051: Good catch. I belive this will not work; the ordering of the equations change when applying a Robin condition (see line 797 in mpfa and comment above). I made a commit that I think fixes this

src/porepy/numerics/fv/fvutils.py
mstarnoni: We can merge the two functions but we would not merge effectively nothing, that is we would still maintain the current code lines, and just add an if statement for vector/scalar problems.
These two functions do the same thing but on different problems, so also the ordering is different.
One (sc_2_c) maps a vector quantity (gravity) for a scalar problem (flow).
The other(map_sc_2_c) maps a scalar quantity (pressure) for a vector problem (mechanics).
I would rather change names (map_sc_2_c_vec_2_scal and map_sc_2_c_scal_2_vec) than implement another if statement.

src/porepy/numerics/fv/mpfa.py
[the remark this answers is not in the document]
rbe051: Maybe body force? 

I have an other question. How do I define the direction the gravity term works in?

src/porepy/numerics/fv/fvutils.py
mstarnoni: or map_sc_2_c_scalar and map_sc_2_c_vector as Runar suggested

src/porepy/numerics/fv/mpfa.py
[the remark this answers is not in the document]
mstarnoni: technically it is a cell-center vector source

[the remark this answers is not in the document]
mstarnoni: > Maybe body force?
> 
> I have an other question. How do I define the direction the gravity term works in?

right now gravity vector is defined in the runscript

src/porepy/numerics/fv/mpfa.py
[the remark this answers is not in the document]
mstarnoni: perfect I integrated it and works! Thanks

src/porepy/numerics/fv/mpfa.py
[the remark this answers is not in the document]
rbe051: Ahh, I understand. This is nice

src/porepy/numerics/fv/fvutils.py
rbe051: I understand. Then I lean towards two different functions. But their names should be updated.

If we follow the convention from mortar grids we should drop "map" and 2->to (even if I would argue that to-> 2 in mortar grid, but that is an other discussion :P ) 
Further, a scalar quantity is the default (e.g., boundary condition), so how about the names
cell_vector_to_subcell and cell_to_subcell_vector?

mstarnoni: they have dimensions of (num_subcells * nd, num_cells * nd) and (num_subcells * nd, num_cells) respectively, so I would call them cell_vector_to_subcell_vector and cell_scalar_to_subcell_vector.
We might also drop the last vector as they both map to a subcell vector (n_K in flow and alpha_grad in Biot).

Issue #281, opened 2019-06-30
Missing "compressible_flow_with_automatic_differentiation" tutorial


femorode:
I noticed that the "automatic_differentiation.ipynb" tutorial references the "compressible_flow_with_automatic_differentiation" tutorial, which is appears to be missing in the tutorials or examples folder.

keileg:
@rbe051 : Do you know the status of this one?

rbe051:
It seems that the tutorial was outdated at some point and moved to examples/examples10. I have revived the tutorial and made a pull request for it. It should be merged back into develop within a day or two.

rbe051:
Fixed by pull request #282

PR #282, opened 2019-07-01
Revieved compressible flow tutorial

rbe051:
revived the compressible flow tutorial with AD. This is in response to Issue #281 Also fixed all the unittests and failing tutorials after the STATE keyword was introduced.

@alessiofumagalli, can you have a look at the changes I made in plot_grid() ?

@keileg, you are mostly added since you were involved in the issue


keileg:
Very good. Thanks.

PR #288, opened 2019-07-12
fix STATE variable in flux reconstruction

alessiofumagalli:


alessiofumagalli:
@IvarStefansson , not really much to say.. but I'll wait few days before the merging

PR #289, opened 2019-07-12
update benchmark 3d files according to the new implementation of pp

alessiofumagalli:


alessiofumagalli:
@IvarStefansson , not much to say.. in few days I'll do the merge

examples/papers/arXiv_1809_06926/case1/data.py
```diff
@@ -41,15 +41,16 @@ def add_data(gb, data, solver_name):
...
             "porosity",
             "Aavatsmark_transmissibilities",
         ]
     )
     for g, d in gb:
-        d["low_zones"] = low_zones(g)
         d["Aavatsmark_transmissibilities"] = True
 
+        low_zone = low_zones(g)
+        d[pp.STATE]["low_zones"] = low_zone
```
IvarStefansson: I don't like that we have to put everything that we want to export in STATE. Could the exporter check both in the highest level and in STATE? What do you think, @keileg?

alessiofumagalli: I agree with out.. Let's wait @keileg opinion

keileg: State is to me primary variables, possibly also secondary variables and other derived quantities. This looks like a parameter. Modifying the exporter sounds better - we could even ask it to look for specified state and parameter keywords separately, but I'm not sure that is what we want.

alessiofumagalli: then the idea is to look both in `d` and in `d[pp.STATE]` for the variables we want to export? I would support this idea and I can implement it. Let me know

PR #303, opened 2019-08-16
Transient contact condition

IvarStefansson:
Two main changes:
   Initial state subtracted for the tangential displacement jumps in the contact discretization
   Fixed sign mistake in DivU

### DIFF
--- a/src/porepy/models/contact_mechanics_model.py
+++ b/src/porepy/models/contact_mechanics_model.py
@@ -12,8 +12,12 @@ undergo major changes (or be deleted).
 import numpy as np
 import scipy.sparse as sps
 from scipy.spatial.distance import cdist
+import logging
 
 import porepy as pp
+
+# Module-wide logger
+logger = logging.getLogger(__name__)
 
 
 class ContactMechanics:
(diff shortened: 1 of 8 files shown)


src/porepy/models/contact_mechanics_model.py
```diff
@@ -430,15 +430,13 @@ def newton_iteration(assembler, setup, u0, tol=1e-14, solver=None):
     A, b = assembler.assemble_matrix_rhs()
     print("max A: {0:.2e}".format(np.max(np.abs(A))))
     print(
-        "max: {0:.2e} and min: {1:.2e} A sum: ".format(
+        "max: {0:.2e} and min: {1:.2e} A sum.".format(
```
rbe051: I think you can remove these print statements, or maybe better change then to  logger.debug(".......")

src/porepy/numerics/contact_mechanics/contact_conditions.py
```diff
@@ -116,21 +120,39 @@ def discretize(self, g_h, g_l, data_h, data_l, data_edge):
...
         # two sides (which is really a difference since one of the sides have
         # its sign switched).
-        displacement_jump_global_coord = (
+        # The tangential displacements are relative to the initial state, which in the
+        # transient case equals the previous time step.
+        previous_displacement_iterate = data_edge[pp.STATE]["previous_iterate"][
+            self.mortar_displacement_variable
+        ]
+        previous_displacement_time = data_edge[pp.STATE][
+            self.mortar_displacement_variable
```
rbe051: is self.mortar_displacement_keyword a better name?

IvarStefansson: I don't think so, seeing as this is the stored variable.

src/porepy/models/contact_mechanics_model.py
IvarStefansson: Good idea!

PR #315, opened 2019-09-09
Mixed dim flow tutorial

IvarStefansson:
Expanded the single-phase flow tutorial to a simple 2d case with one fracture. Also simplified the geometry of the single-dimensional example.


IvarStefansson:
> One question: In the conventions, do we want to be even clearr that there is no hidden scaling of permeabilities?

We want to be as clear as possible. But I'm not sure exactly where you mean. Do you have a suggestion for a rephrasing/sentence to be added?

keileg:
Something like: 'To be clear, the permeabilities specified within and between subdomains are absolute, that is, no aperture scaling of permeabilities takes place inside the code'?

PR #326, opened 2019-10-19
New parameters

IvarStefansson:
Added UnitFluid and very minor updates to the rock parameters. 

Only one question, I think: Should we rename water.py to fluid.py, since it now contains two different fluids (one of which isn't water)?


src/porepy/params/water.py
```diff
@@ -3,12 +3,47 @@
...
+        if theta_ref is None:
+            self.theta_ref = 20 * (pp.CELSIUS)
+        else:
+            self.theta_ref = theta_ref
+
+        self.COMPRESSIBILITY = 1 / pp.PASCAL
+        self.BULK = 1 / self.COMPRESSIBILITY
+
+    def thermal_expansion(self, delta_theta):
+        """ Units: m^3 / m^3 K, i.e. volumetric """
```
keileg: Documentation on units I like a lot!

PR #330, opened 2019-11-11
Simpler domain boundaries

rbe051:
I got tired of making the domain boundary dictionaries


keileg:
Looks good if we want something like this, but I am a bit hesitant: The trend over the last year has been to remove implicit definitions of parameters, such as 'minimal coordinates of the bounding box are zero unless otherwise stated'. 

@IvarStefansson what do you think?

IvarStefansson:
> 
> 
> Looks good if we want something like this, but I am a bit hesitant: The trend over the last year has been to remove implicit definitions of parameters, such as 'minimal coordinates of the bounding box are zero unless otherwise stated'.
> 
> @IvarStefansson what do you think?

I'm tempted to vote to allow this, as long as it's obvious from the documentation what happens. I think the documentation needs to be updated at least for the two fracture network classes, possibly in other places as well.

rbe051:
I would argue that this is ok as it is the same syntax as CartGrid, StructureSimplexGrid, TensorGrid...

We could do something like [xmin, xmax, ymin, ymax], but then it should also be changed in CartGrid, etc.

src/porepy/fracs/utils.py
```diff
@@ -127,3 +127,48 @@ def snap_fracture_set_2d(pts, edges, snap_tol, termination_tol=1e-2, max_iter=10
...
+    ValueError('Unknown size of domain. domain_to_dict is only implemented in 2d or 3d')
+    """
+
+    """
+    """
+    if domain is None:
+        return domain
+    elif isinstance(domain, dict):
+        return domain
+    else:
```
keileg: This covers the case where domain is a list or an ndarray? It is very confusing.

src/porepy/fracs/utils.py
```diff
@@ -127,3 +127,48 @@ def snap_fracture_set_2d(pts, edges, snap_tol, termination_tol=1e-2, max_iter=10
         logger.warning("Fracture snapping failed to converge")
         logger.warning("Residual: " + str(diff))
         return pts, False
+
+
+def domain_to_dict(domain):
```
keileg: The function name is misleading, it may not return a dictionary.

src/porepy/fracs/utils.py
```diff
@@ -127,3 +127,48 @@ def snap_fracture_set_2d(pts, edges, snap_tol, termination_tol=1e-2, max_iter=10
...
+    """ Transform input to a valid domain dictionary. If domain is already a dictionary,
+    return the dictionary.
+
+    Parameters:
+        domain (None, dict, ndarray, list): Specify the domain either as None, a dictionary,
+        a list domain = [xmax, ymax] (or domain = [xmax, ymax ,zmax])), or a ndarray.
+    Returns:
+        dict: A dictionary. If input domain is None or a dictionary, return intput. If input
```
keileg: I tried to write down the types here: 

domain_to_dict(domain: typing.Union[typing.List, typing.Dict[str, float], np.ndarray, None]) -> typing.Union[typing.Dict[str, float], None, np.ndarray]:

This is a mess, but it is not clear that it is caused by this PR; most of the problems can be traced way back into the network specification. On the other hand, adding a list did not help.

src/porepy/fracs/fractures.py
```diff
@@ -641,7 +641,7 @@ def __init__(self, fractures=None, domain=None, verbose=0, tol=1e-8):
 
         # Initialize with an empty domain. Can be modified later by a call to
         # 'impose_external_boundary()'
-        self.domain = domain
+        self.domain = pp.fracs.utils.domain_to_dict(domain)
```
keileg: @IvarStefansson is right, documentation of options for domain is needed (also independent of this PR).

test/integration/test_fracture_mesh_generation.py
```diff
@@ -688,6 +688,18 @@ def test_fracture_cut_not_split_by_domain(self):
         gb = network.mesh(mesh_args)
         self.assertTrue(len(gb.grids_of_dimension(2)) == 1)
 
+    def test_no_dictionary_boundary(self):
```
keileg: Very good.

src/porepy/fracs/utils.py
rbe051: See next comment

src/porepy/fracs/utils.py
rbe051: I agree with all of the points here. I wrote this function for FractureNetwork2d and everything was neat and simple. Then I though, this should be easy to include for the 3d case as well. Wrong! There are so many different possibilities to set the boundaries, and they do not agree between 2d and 3d, which caused this mess of I/O arguments. 

Because of this, I think we do not need another way to set the boundary, and I subtract this pull-request. 

However, I am still fed up with making the domain boundary dictionary. Is there a reason we actually want this as a dictionary and not just as a list [xmin, xmax, ymin, ymax....] ?

keileg: I agree, this is a mess. There is no particular reason for the dict, except that the dictionary makes variable definitions more explicit. This is a real worry to me, I would never be able to remember whether the order is [xmin, xmax, ymin, ymax] or [xmin, ymin, xmax, ymax]. 

How about namedtuples:
from collections import namedtuple
Domain = namedtuple('domain', ['xmin', 'xmax', 'ymin', 'ymax', 'zmin', 'zmax'], defaults=(None, None))
d = Domain(0, 1, -3, 2)
d[0] = 0
d.ymin = -3

One possible issue is that the defaults parameter to namedtuple is new to python3.7, the construction in 3.6 (which I think we will still support for a while, or?) is a bit more cumbersome, see e.g. https://stackoverflow.com/questions/11351032/namedtuple-and-default-values-for-optional-keyword-arguments

This should take care of list and dictionaries. None is still an option, or perhaps we can integrate that into the namedtuple? 

Numpy arrays must still be an option, and I am afraid they will also be introduced to 2d networks at some point, to account for non-box domains there as well. But, if we can go from 4 to 2 formats, where the second option can be considered advanced usage, that is great progress.

keileg: BoxDomain may be a better name than Domain.

PR #331, opened 2019-11-11
Fixed bug where geometry values in grid bucket used syntax from networkx

rbe051:
A trivial fix, but I'll let you approve @keileg


rbe051:
I added a bugfix to g.compute_geometry() to this pull request.

The old version gives wrong face normals for left boundary for the case:
```python
g = pp.CartGrid([1, 1], [5000, 5000])
g.compute_geometry()
g.face_normals[:, 0]
```

PR #340, opened 2019-12-13
Add Metis instructios to Readme.md

ScottHMcKean:
Add some basic instructions for the installation of Metis for new users
based on my experience as a linux/python newbie.


keileg:
Thanks, this can indeed be useful. However, do you think you can move it to Install.md instead of Readme? We are trying to keep a reasonable length of the Readme.

Again, thanks!

ScottHMcKean:
Agreed! I'll modify my pull request tonight

PR #341, opened 2019-12-18
Model adjustments

IvarStefansson:
Keep track of error and number of iterations for each time step when solving models.
Add temperature source discretization to thm_model.

### DIFF
--- a/src/porepy/numerics/contact_mechanics/contact_conditions.py
+++ b/src/porepy/numerics/contact_mechanics/contact_conditions.py
@@ -87,7 +87,9 @@ class ColoumbContact:
 
         # Numerical parameter, value and sensitivity is currently unknown.
         # The thesis of Huber is probably a good place to look for information.
-        c_num = 100
+        c_num = parameters_l[self.keyword].get(
+            "contact_mechanics_numerical_parameter", 100
+        )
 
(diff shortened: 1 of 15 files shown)


src/porepy/numerics/contact_mechanics/contact_conditions.py
```diff
@@ -87,7 +87,9 @@ def discretize(self, g_h, g_l, data_h, data_l, data_edge):
 
         # Numerical parameter, value and sensitivity is currently unknown.
         # The thesis of Huber is probably a good place to look for information.
-        c_num = 100
+        c_num = parameters_l[self.keyword].get(
+            "contact_mechanics_numerical_parameter", 100
+        )
```
keileg: Very good.

src/porepy/numerics/contact_mechanics/contact_conditions.py
```diff
@@ -336,7 +338,7 @@ def _penetration(self, Tn, un, cn):
 
         """
         # Not sure about the sensitivity to the tolerance parameter here.
-        tol = 1e-8 * cn
+        tol = 1e-12 * cn
```
keileg: We should look deeper into this, but let's keep it as you suggest for now.

PR #343, opened 2020-01-02
Fix temperature bc assignment in thm model

keileg:
The thm model called the bc_values and bc_type methods for the scalar variable instead of temperature.


IvarStefansson:
Thanks for catching this! And sorry for introducing it in the first place...

src/porepy/models/thm_model.py
```diff
@@ -149,9 +149,9 @@ def set_temperature_parameters(self) -> None:
         for g, d in self.gb:
             # By default, we set the same type of boundary conditions as for the
             # pressure problem, that is, zero Dirichlet everywhere
```
IvarStefansson: This should obviously read "temperature/energy problem"

PR #358, opened 2020-03-12
Dfn 1d

alessiofumagalli:
add the possibility to create a dfn from 1d fractures

### DIFF
--- a/src/porepy/fracs/fractures_2d.py
+++ b/src/porepy/fracs/fractures_2d.py
@@ -295,6 +295,32 @@ class FractureNetwork2d(object):
         else:
             return G
 
+    def connected_networks(self):
+        """
+        Return all the connected networks as separate networks.
+
+        """
+
+        # extract the graph by splitting the intersections
+        graph, network = self.as_graph()
+        sub_networks = []
+
+        # loop on all the sub-graphs obtained
+        sub_graphs = nx.connected_component_subgraphs(graph)
+        for sub_graph in sub_graphs:
+            # get all the data for the current sub-network
+            # NOTE: standard numpy conversion does not work for a graph with
+            # one edge
+            edges = np.empty((2, sub_graph.number_of_edges()), dtype=np.int)
+            for idx, (u, v) in enumerate(sub_graph.edges):
+                edges[:, idx] = [u, v]
+            # for compability we keep the same points as the original graph
+            pts = network.pts
+            # create the sub network
+            sub_networks += [FractureNetwork2d(pts, edges, network.domain, network.tol)]
+
+        return np.asarray(sub_networks, dtype=np.object)
+
     def split_intersections(self, tol=None):
         """ Create a new FractureSet, with all fracture intersections removed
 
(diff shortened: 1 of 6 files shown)


src/porepy/fracs/fractures_2d.py
```diff
@@ -177,9 +179,12 @@ def mesh(self, mesh_args, tol=None, do_snap=True, constraints=None, **kwargs):
 
         if do_snap and p is not None and p.size > 0:
             p, _ = pp.frac_utils.snap_fracture_set_2d(p, e, snap_tol=tol)
-        grid_list = porepy.fracs.simplex.triangle_grid(
-            p, e[:2], self.domain, tol=tol, constraints=constraints, **mesh_args
-        )
+        if dfn:
+            grid_list = pp.fracs.simplex.line_grid_embedded(p, e[:2], self.domain, tol=tol, **mesh_args)
```
keileg: Have you used black?

src/porepy/fracs/fractures_2d.py
```diff
@@ -295,6 +300,48 @@ def as_graph(self, split_intersections=True):
         else:
             return G
 
+    def connected_networks(self):
+        """
+        Return all the connected networks as separate networks.
```
keileg: Proper documentation needed if the function is to be included.

src/porepy/fracs/fractures_2d.py
```diff
@@ -295,6 +300,48 @@ def as_graph(self, split_intersections=True):
         else:
             return G
 
+    def connected_networks(self):
```
keileg: I am not sure this is something we really need - you can easily do this from the user side. I am afraid of overloading on utility functions. So unless you can convince me otherwise, I do not want to include this.

keileg: And yes, I am aware the same can probably be said about a lot of the other functions in this class.

src/porepy/fracs/fractures_2d.py
```diff
@@ -295,6 +300,48 @@ def as_graph(self, split_intersections=True):
...
+            for idx, (u, v) in enumerate(sub_graph.edges):
+                edges[:, idx] = [u, v]
+            # for compability we keep the same points as the original graph
+            pts = network.pts
+            # create the sub network
+            sub_networks += [FractureNetwork2d(pts, edges, network.domain, network.tol)]
+
+        return np.asarray(sub_networks, dtype=np.object)
+
+    def purge_pts(self):
```
keileg: Same comment as above - do we really need this?

src/porepy/fracs/simplex.py
```diff
@@ -157,61 +157,68 @@ def triangle_grid(points, edges, domain, constraints=None, **kwargs):
     """
     logger.info("Create 2d mesh")
```
keileg: This we should discuss. First, all the splitting functionality will be moved into FractureNetwork2d, so I am very hesitant to allow for changes now. 

Second, I have no clue what your changes here are doing; granted this was not a well documented function, but you sure didn't help out.

PR #360, opened 2020-03-27
Unstructured grid convergence (WIP)

haakon-e:
# Overview
I have implemented a method to construct a mapping between from each grid cell on a fine grid bucket to the cell it is geometrically contained in on a coarse grid bucket. It is intended to work with unstructured meshes created by gmsh. A key assumption is that the each cell in the refined grid is fully contained within one cell on the coarse grid.

A wrapper method allows for any subclass of `pp.models.ContactMechanics` to be run iteratively over `n` grids of increasing degree of refinement, with errors computed relative to the solution on the finest grid.

# Details
The PR includes three main contributions:
1. A wrapper for the python module "gmsh" (which again is a wrapper for the real gmsh) that runs the gmsh command "RefineBySplitting". This is entirely equivalent to opening your mesh in the gmsh program, then pressing the 'refine by splitting' button.
2. A method to construct a mapping from each grid cell in a fine grid to the cell in the coarse grid that geometrically contains the fine cell. Because we assume the entire fine cell is contained in the coarse cell, we do this check by running `point_in_polyhedron`, with the coarse cell as the polyhedron and the cell center of the fine cell as the point.
In addition, I have a wrapper that enables this procedure for each grid in a grid bucket.
The resulting mapping is stored as a node property on the appropriate node on the coarse grid bucket.
3. A method to compute grid errors, where a refined grid bucket is used as reference solution. You can select which variables to compute errors over.
In addition, for your convenience, I have added a pretty flexible wrapper to run a complete convergence study with `n` grid refinements using any subclass of `ContactMechanics` coupled with an appropriate run method like `pp.run_time_dependent_model`

# Methods and location in porepy
First, a note: I have added new python files where I deemed appropriate, but this is of course just a suggestion from me.
#### 1. The first contribution is located in `pp.utils.grid_refinement.py`
https://github.com/pmgbergen/porepy/blob/c465ca0482b4b006d083669cee6e662acb01546e/src/porepy/utils/grid_refinement.py#L28-L33
This method needs a pair `(in_file, network)`. The first is the path to a `.geo` file that describes the domain. The second is a `pp.FractureNetwork3d` (or 2d).
#### 2. The second contribution is also located in `pp.utils.grid_refinement.py`
https://github.com/pmgbergen/porepy/blob/c465ca0482b4b006d083669cee6e662acb01546e/src/porepy/utils/grid_refinement.py#L135-L140
https://github.com/pmgbergen/porepy/blob/c465ca0482b4b006d083669cee6e662acb01546e/src/porepy/utils/grid_refinement.py#L89-L92
#### 3. The third contribution is located in `pp.utils.grid_convergence.py` and `pp.models.run_convergence_study.py`
https://github.com/pmgbergen/porepy/blob/c465ca0482b4b006d083669cee6e662acb01546e/src/porepy/utils/grid_convergence.py#L26-L32
https://github.com/pmgbergen/porepy/blob/c465ca0482b4b006d083669cee6e662acb01546e/src/porepy/models/run_convergence_study.py#L72-L82

## End note:
I was brave enough to "sneak in" two completely unrelated changes in this PR.
1. (ddf0dc53df44dfb9609978ff6866937ef8937a2b): scaling coefficients are set from `params`, if present.
2. (c465ca0482b4b006d083669cee6e662acb01546e): Updated documentation for `create_grid` and initialized the grid bucket to `self.gb = None`. The idea is that all class variables should be initialized in the `__init__` method, and this suggestion formalises that (albeit for only one variable).


src/porepy/utils/grid_convergence.py
```diff
@@ -0,0 +1,130 @@
...
+            if var not in check_keys:
+                logger.info(f"{var} not present on grid number "
+                            f"{node_number} of dim {g.dim}.")
+                continue
+
+            # Compute errors relative to the reference grid
+            # TODO: Should the solution be divided by g.cell_volumes or similar?
+            # TODO: If scaling is used, consider that - or use the export-ready variables,
+            #   'u_exp', 'p_exp', etc.
+            sol = states[var].reshape((var_dof, -1), order='F').T  # (num_cells x var_dof)
```
keileg: Does this assume cell variables (with one dof per cell)?

src/porepy/utils/grid_convergence.py
```diff
@@ -0,0 +1,130 @@
...
+            mapped_sol: np.ndarray = mapping.dot(sol)  # (num_cells x variable_dof)
```
keileg: mapping is csc? so dot is the same as *? I've learned to be afraid of np.dot, so I want to be sure.

src/porepy/models/run_convergence_study.py
```diff
@@ -0,0 +1,217 @@
...
+        list of (n+1) grid buckets in increasing order
+        of refinement (reference grid last)
+    errors : List[dict]
+        List of (n) dictionaries, each containing
+        the error for each variable on each grid.
+        Each list entry corresponds index-wise to an
+        entry in gb_list.
+    """
+
+    # 1. Step: Create n grids by uniform refinement.
```
keileg: This kind of documentation I like.

src/porepy/models/run_convergence_study.py
```diff
@@ -0,0 +1,217 @@
...
+        num_refinements=n_refinements,
+    )
+    for _gb in gb_list:
+        pp.contact_conditions.set_projections(_gb)
+
+    return gb_list
+
+
+def run_model_for_convergence_study(
+        model: Type[ContactMechanics],
```
keileg: Can it instead inherit from AbstractModel?

keileg:
One more thought: Should the new functions have tests?

src/porepy/models/run_convergence_study.py
```diff
@@ -0,0 +1,217 @@
...
+    TypeVar,
+    Union,
+)
+from pathlib import Path
+from pprint import pformat
+import functools
+import inspect
+
+import pendulum
+import porepy as pp
```
alessiofumagalli: all these import are standard or we need to install new modules? it looks like you use a lot of modules never used in pp

src/porepy/models/run_convergence_study.py
```diff
@@ -0,0 +1,217 @@
...
+
+from porepy.models.contact_mechanics_model import ContactMechanics
+from porepy.utils.grid_refinement import gb_coarse_fine_cell_mapping, refine_mesh
+from porepy.utils.grid_convergence import grid_error
+
+
+logger = logging.getLogger(__name__)
+
+
+def gb_refinements(
```
alessiofumagalli: is this function works also for dfn? i.e., without the higher dimension grid?

src/porepy/models/run_convergence_study.py
```diff
@@ -0,0 +1,217 @@
...
+        network=network,
+        num_refinements=n_refinements,
+    )
+    for _gb in gb_list:
+        pp.contact_conditions.set_projections(_gb)
+
+    return gb_list
+
+
+def run_model_for_convergence_study(
```
alessiofumagalli: are you computing the L2 errors? if so it should be clear, otherwise is a bit misleading

src/porepy/models/run_convergence_study.py
```diff
@@ -0,0 +1,217 @@
...
+    # --- SETUP AND SOLVE ---
+    # -----------------------
+    newton_params = newton_params if newton_params else {}
+
+    for gb in gb_list:
+        setup = model(params=params)
+        # Critical to this step is that setup.prepare_simulation() (specifically setup.create_grid())
+        # doesn't overwrite that we manually set a grid bucket to the model.
+        setup.gb = gb
+        pp.contact_conditions.set_projections(setup.gb)
```
alessiofumagalli: this only work for contact mechanics?

src/porepy/utils/grid_convergence.py
```diff
@@ -0,0 +1,130 @@
...
+            sol_ref = states_ref[var].reshape((var_dof, -1), order='F').T  # (num_cells x var_dof)
+
+            # axis=0 gives component-wise norm.
+            absolute_error = np.linalg.norm(mapped_sol - sol_ref, axis=0)
+            norm_ref = np.linalg.norm(sol_ref, axis=0)
```
alessiofumagalli: again make clear that you are doing the L2 error of cell-based variables

src/porepy/utils/grid_convergence.py
```diff
@@ -0,0 +1,130 @@
...
+
+            if np.any(norm_ref < 1e-10):
```
alessiofumagalli: do you want to punt the 1e-10 in a kwargs so a user can change it? if so add this to the doc of the function

src/porepy/utils/grid_refinement.py
```diff
@@ -0,0 +1,239 @@
...
+
+        # Update pointer to which cell centers to use as test points
+        in_poly_ids = test_cells_ptr[in_poly]  # id of cells inside this polyhedron
+        test_cells_ptr = test_cells_ptr[~in_poly]  # Keep only cells not inside this polyhedron
+
+        # Update mapping
+        indices = np.append(indices, in_poly_ids)
+        indptr = np.append(indptr, indptr[-1] + in_poly_ids.size)
+
+    data = np.ones(indices.size)
```
alessiofumagalli: do you want to have a measure ration between the reference and the mapped one? just proposing, maybe it's out of scope

src/porepy/models/run_convergence_study.py
keileg: Good point. This module should not be imported in pp.__init__ - that is probably not relevant anyhow, and usage and installation of the new modules should be documented. is it pip

src/porepy/models/run_convergence_study.py
keileg: I would think not - if there are assumptions on the distribution of grids between the dimensions, these should be easy to remove.

src/porepy/models/run_convergence_study.py
keileg: I think we can generalize it to AbstractModel. Then you will need to implement a new model, say FlowModel, to use it. In practice, this will be a lightweight wrapper around your simulation script, which standardizes some names etc.

src/porepy/models/run_convergence_study.py
keileg: I agree.

I also think we can generalize the error computations if need be, but that is not for this PR, nor the task of @haakon-e

src/porepy/utils/grid_refinement.py
keileg: @alessiofumagalli I have no idea what you mean here.

src/porepy/models/run_convergence_study.py
keileg: To be clear, for this PR, I think we should generalize to AbstractModel (should be straightforward), but leave it at that.

src/porepy/models/contact_mechanics_biot_model.py
```diff
@@ -40,8 +40,8 @@ def __init__(self, params=None):
         self.scalar_parameter_key = "flow"
 
         # Scaling coefficients
-        self.scalar_scale = 1
-        self.length_scale = 1
+        self.scalar_scale = params.get('scalar_scale', 1)
+        self.length_scale = params.get('length_scale', 1)
```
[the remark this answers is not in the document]
IvarStefansson: Yes, this has been my guiding principle as well. I must admit there are quite a few additional parameters in the subclasses I'm currently using. But I think this actually reinforces my view; there are so many different possible modifications we might want to do that we can't include all.

src/porepy/utils/grid_refinement.py
```diff
@@ -0,0 +1,239 @@
...
+            # Make polyhedron from node coordinates
+            # Polyhedron defined as a list of nodes defining its (convex) faces.
+            # Assumes simplexes: Every node except one defines every face.
+            assert (num_nodes == 4), "We assume simplexes in 3D (i.e. 4 nodes)"
+            node_coords = nodes[:, nodes_idx]
+
+            ids = np.arange(num_nodes)
+            polyhedron = [node_coords[:, ids != i] for i in np.arange(num_nodes)]
+            test_points = cells_ref[:, test_cells_ptr]  # Test only points not inside another polyhedron.
+            in_poly = pp.geometry_property_checks.point_in_polyhedron(
```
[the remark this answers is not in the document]
IvarStefansson: We could think of investigating whether the cells of the refined grids can be tagged with their "parent" coarse cells in gmsh. Then we wouldn't have to identify them geometrically

src/porepy/utils/grid_convergence.py
```diff
@@ -0,0 +1,130 @@
...
+    n_variables = len(variable)
+
+    errors = {}
+
+    grids = gb.get_grids()
+    grids_ref = gb_ref.get_grids()
+    n_grids = len(grids)
+
+    for i in np.arange(n_grids):
+        g, g_ref = grids[i], grids_ref[i]
```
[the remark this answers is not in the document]
IvarStefansson: There has to be a way of passing the order of the fractures (as in the list from which the network is constructed) through gmsh and tag the fractures. If this is not supported yet (in porepy, that is), I think we need to add it

src/porepy/utils/grid_convergence.py
IvarStefansson: No, the purpose of line 102 is precisely to allow for vectors, right?

src/porepy/models/run_convergence_study.py
IvarStefansson: I think some of the imports could be avoided.

src/porepy/utils/grid_convergence.py
IvarStefansson: The lesson from the contact mechanics is that we don't want "random" tolerances scattered around the code. I suggest a variable "tol" (or better, a more specific name. "norm_tol"), which is set at the top of the function (or as a kwarg) and documented.

src/porepy/utils/grid_convergence.py
keileg: Sorry - I meant face variables

src/porepy/utils/grid_refinement.py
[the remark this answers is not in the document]
keileg: It would be surprising if gmsh does not keep track of this. Getting hold of the information is another matter.

src/porepy/models/contact_mechanics_biot_model.py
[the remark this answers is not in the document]
keileg: agreed.

keileg:
@haakon-e @alessiofumagalli @IvarStefansson : To summarize status, this is work with high potential, that we really want to add to the code with some modifications. However, it should not be the task of Håkon to fix this - he has a thesis to write up. Thus the responsibility to improve this PR is on the rest of us. I will try to take the lead, feel free to commit @alessiofumagalli and @IvarStefansson

src/porepy/models/contact_mechanics_biot_model.py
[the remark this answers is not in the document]
haakon-e: The reason I made this particular suggestion is that these parameters are set in the `__init__` method. This is contrary to any other method in `ContactMechanics` and `ContactMechanicsBiot`, which by nature of not being `__init__` are more natural to overwrite when subclassing.

#### Why?
Many constants (parameter names, etc) are set in the `__init__` method, which makes it natural to subclass, but then call by cooperative subclassing (by `super().__init__`.
Because this is set to 1, when I subclass, I will have to either
* set these to parameters __again__ after I have called `super().__init__`
OR
* bypass this `__init__` method by implementing its other functionality, then manually call `ContactMechanics.__init__()`.

The latter option is of course infeasible and super confusing. But, I would argue that the first option is rather confusing as well. I can only speak for myself however saying that this caused quite a few unexpected results at some point due to these being set back to 1 after I thought I had set them to some other value.

src/porepy/utils/grid_refinement.py
```diff
@@ -0,0 +1,239 @@
...
+import porepy as pp
+from porepy.fracs.simplex import tetrahedral_grid_from_gmsh
+from porepy.fracs.meshing import grid_list_to_grid_bucket
+import numpy as np
+import scipy.sparse as sps
+
+logger = logging.getLogger(__name__)
+
+
+def refine_mesh(
```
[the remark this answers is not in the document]
haakon-e: Resolved by 3c720e4

src/porepy/models/run_convergence_study.py
```diff
@@ -0,0 +1,217 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def gb_refinements(
+        network: Union[pp.FractureNetwork3d, pp.FractureNetwork2d],
+        gmsh_folder_path: Union[str, Path],
+        mesh_args: dict,
+        n_refinements: int = 0
+) -> List[pp.GridBucket]:
```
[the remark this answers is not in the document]
haakon-e: Could you review my new suggestion (see 3c826a6), using generators?

At this point, `refine_mesh_by_splitting` should be optimized. I have not done that for `run_model_for_convergence_study` yet (since it needs the finest grid anyways).

src/porepy/utils/grid_refinement.py
alessiofumagalli: it took me a while to understand it for me as well.. I mean, a projection matrix that map variables from the original grid to the refined one (maybe the vice versa, as well)

keileg:
@alessiofumagalli @IvarStefansson @haakon-e : I will do changes to the file structure in this PR, and clean up some old mess in pp along the way. So, no commits to this branch until further notice, please.

keileg:
Update: I will also try to grow this into a general-purpose mesh refinment study, using Håkon's work as a starting point. The design idea is similar to what Ivar did for the models: Make skeleton classes for grid generation and refinement, as well as post processing, and let the problem itself be based on a model.

As a test case, I'm thinking to use what @jhabriel is doing for error estimates. The process ahead will thus be to make a somewhat coarse interface, and refine it as we get more experience. 

@alessiofumagalli : Are you doing any convergence tests these days? 

@haakon-e @IvarStefansson : Input from you based on your project may become interesting, but I don't think it will be relevant for you to get involved in the design itself; your studies are too far advanced for it to make sense to use this new framework.

Time scale: I'll try to prioritize this next week.

keileg:
I have finally found the time to pick this up again, and will soon push changes. The changes that will enter the code are:
1) There will be a GridSequenceFactory that encapsulates the method for grid refinement by splitting, together with other refinement types. 
2) Some of the computations of errors will be added to utils.errors
3) The methodology to run a convergence test will for the moment not be added. The proposed method only covers a very special type of convergence test, which in some sense is incompatible with other, equally relevant types of convergence tests. So, until the abstraction becomes clearer, my suggestion is that the main function for convergence tests enters into runscripts.

Comments @IvarStefansson @haakon-e @alessiofumagalli

PR #362, opened 2020-03-30
Fix 0d grid tpfa

alessiofumagalli:
avoid the request of having set the permeability and bc for a 0d grid


keileg:
You sure you want to include all commits in this PR? The description does not match at all.

alessiofumagalli:
Sorry, my mistake it was only the last commit. I'll fix it now.

PR #365, opened 2020-03-30
consider subprocess python package in calling gmsh instead of os, it …

alessiofumagalli:
…should handle also path with spaces

### DIFF
--- a/src/porepy/grids/gmsh/gmsh_interface.py
+++ b/src/porepy/grids/gmsh/gmsh_interface.py
@@ -2,6 +2,7 @@
 
 import numpy as np
 import os
+import subprocess
 
 # meshio has changed the name of the module taking care of gmsh import.
 # Ensure compatibility with both versions with a try-except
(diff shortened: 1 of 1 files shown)


haakon-e:
If I may suggest using gmsh's own python wrapper for this.
This is what I did to mesh (and refine mesh) in my other PR:
https://github.com/pmgbergen/porepy/blob/437db014d542f45d0a24214a434024c60a72e75c/src/porepy/utils/grid_refinement.py#L63-L71
This is quite clean and easy to understand, in my opinion.

Note that for this, we need the SDK version of gmsh instead of the current PorePy-gmsh requirements. Either way, it will be relevant for my PR for us to decide which way to go on this.
https://github.com/pmgbergen/porepy/blob/437db014d542f45d0a24214a434024c60a72e75c/src/porepy/utils/grid_refinement.py#L47-L55

keileg:
I agree with @haakon-e

keileg:
Btw, that will require modifications to the install instructions, correct?

keileg:
We will most likely go with #367 instead of this one. Let's put it on hold for a little while.

keileg:
Closed, but the same effect should be achieved by #367

PR #367, opened 2020-03-30
WIP: gmsh python api

haakon-e:
# Overview
Use the gmsh python api to generate meshes.

[The gmsh source code](https://gitlab.onelab.info/gmsh/gmsh) features application interfaces (APIs) to numerous languages. Fortunately for us, one of these is python. Using their API ensures access to most, if not all features of gmsh, without having to "backport" by calling the terminal console to execute gmsh commands.

The API is defined [here](https://gitlab.onelab.info/gmsh/gmsh/-/blob/master/api/gmsh.py). However, this may be a bit difficult to read, so I recommend taking a look at their [demos](https://gitlab.onelab.info/gmsh/gmsh/-/tree/master/demos%2Fapi), where they showcase gmsh code using the APIs.

## Requirements note
Implementing this PR requires a change in how gmsh is installed on users' systems. This is reflected in the updated Install instructions, which says:
* You need the gmsh sdk version, which I have downloaded from source. Alternatively (but I haven't tested this) there is a python package that does this for you, for a particular version of gmsh sdk.

### TODO
Allow for `**kwargs` to be passed to the new `run_gmsh` function.
@keileg do you have any examples of what `**kwargs` are currently used in the code base?

# Footnote
I made a separate PR so that no code conflicts with #365 occur.


keileg:
No examples of kwargs, this is old code. I think we can drop this, and use gmsh directly if we want more advanced functionality.

keileg:
Another matter: For this to work properly, we will need to update install instructions? and that further requires changes to .travis.yml? If correct, I can help out with this part.

keileg:
@alessiofumagalli @haakon-e Same comment as on #360 : I will take this from here, and decide whether we go with this or #365

haakon-e:
Sounds good. I updated the PR text to better reflect the opportunities of using the gmsh api directly.

haakon-e:
## Comment on the most recent commits
* The new gmsh function has no return value. This will break any expected return values elsewhere in the codebase.
* `meshio` import is removed. I don't know if this is used elsewhere in the codebase. If not, this could possibly be removed from the `requirements.txt`.

keileg:
With my setup, and newest versions of gmsh (4.5.6), I do not manage to get gmsh.py to produce an error message, even when feeding it a .geo file with clear syntax errors. Importing the file into the gmsh gui gives the expected torrent of error messages.

I am very hesitant to go further with this - we need to raise errors if meshing fails. It may be that my setup is wrong, but stupidity is also a practical limitation.

PR #371, opened 2020-04-06
Python 38

keileg:
Update various packages. Include python 3.8 in testing.


keileg:
Update: There is no vtk version compatible with py3.8. It seems this will be resolved by the upcoming release of vtk v9. This PR is put on hold until then.

keileg:
Delete this for now. vtk v9 is still not available for linux, while the rest of the changes in this PR have been introduced, mainly in #417.

PR #385, opened 2020-04-23
Fix tests

rbe051:
Fixed broke tutorials


src/porepy/viz/exporter.py
```diff
@@ -205,10 +205,13 @@ def __init__(self, grid, file_name, folder_name=None, **kwargs):
...
         self.file_name = file_name
         self.folder_name = folder_name
-        self.fixed_grid = kwargs.get("fixed_grid", True)
-        self.binary = kwargs.get("binary", True)
-        self.simplicial = kwargs.get("simplicial", False)
-
+        self.fixed_grid = kwargs.pop("fixed_grid", True)
+        self.binary = kwargs.pop("binary", True)
+        self.simplicial = kwargs.pop("simplicial", False)
+        if kwargs:
```
rbe051, line 211: We should consider doing this other places in the code as well

tutorials/data/flow_benchmark_2d_geiger_setup.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
```
rbe051, line 1: Moved file from test to avoid having to change python paths to run the tutorials

keileg:
I will close this, due to incompatbile versions of black (we will update this soon), but the important commits are preserved in #386 .

PR #387, opened 2020-04-23
Periodic bc

rbe051:
# Overview
Add periodic boundary conditions

# Proposed changes
The BoundaryCondition class has two new attributes, is_per and per_map, which defines the periodic faces and the mapping between them. In addition, BoundaryCondition is given the function set_periodic_map which set the attribute per_map (is_map is set at construction equivalently to is_dir, is_rob and is_neu).

Periodic boundaries are implemented for tpfa. The other discretizations throws an error if they are given a periodic boundary. 

The grid has to be matching to assign periodic boundaries (if not, you have to use mortars).

# Tasks:
@alessiofumagalli can you check that VEM throws an error at the correct places.
@keileg is BoundaryConditionNode supposed to have died? If so, I think it has reappeared when I rebased the branch. Then, I will delete it. 

# References to code
https://github.com/pmgbergen/porepy/blob/6321aeb8cc7fff9cd83b62af321a0479396186de/src/porepy/params/bc.py#L210

https://github.com/pmgbergen/porepy/blob/6321aeb8cc7fff9cd83b62af321a0479396186de/src/porepy/numerics/fv/tpfa.py#L137-L168


keileg:
@rbe051 Immediate answer: Yes BoundaryConditionNode is dead, so please do not resurrect it.

I'll try to find the time for the review tomorrow, or early next week.

src/porepy/numerics/fv/tpfa.py
```diff
@@ -133,8 +133,39 @@ def discretize(self, g, data):
...
+        right_ci = right_ci[I_right]
+
+        # Add periodic boundaries to subface cell mapping
+        face_idx = np.hstack((fi, left, right))
+        cell_idx = np.hstack((ci, right_ci, left_ci))
+        sgn_face = np.hstack((sgn, -left_sgn, -right_sgn))
+
         # Create flux matrix
-        flux = sps.coo_matrix((t[fi] * sgn, (fi, ci))).tocsr()
+        flux = sps.coo_matrix((t[face_idx] * sgn_face, (face_idx, cell_idx))).tocsr()
```
keileg, line 168: How about vector sources? I know this is not a primary concern for you, but I am worried about fragmentation of the code. I think the extension should be straightforward - would you care to have a short look and see what it would take?

src/porepy/params/bc.py
```diff
@@ -57,7 +59,13 @@ class BoundaryCondition(AbstractBoundaryCondition):
             face i has been assigned a Neumann condition.
         is_rob (np.ndarray, boolean, size g.num_faces): Element i is true if
             face i has been assigned a Robin condition.
-
+        is_per (np.ndarray, boolean, size g.num_faces): Element i is true if
+            face i has been assigned a periodic boundary condition. Note that
+            periodic boundary conditions are non-standard and might not work
+            for all discretizations. See also attribute per_map
+        per_map (np.ndarray, int, 2 x # periodic faces): Defines the periodic
```
keileg, line 66: Is there an assumption that the paired faces matches geometrically (perfect overlap)?

src/porepy/params/bc.py
```diff
@@ -174,9 +202,124 @@ def __repr__(self) -> str:
...
+            raise ValueError(
+                """Periodic map has wrong size. Given array size is: {},
+                              but should be: {}""".format(
+                    per_map.shape, map_shape
+                )
+            )
+        self.per_map = per_map
+
+
+class BoundaryConditionNode(AbstractBoundaryCondition):
```
keileg, line 232: Please kill this again.

src/porepy/numerics/fv/tpfa.py
rbe051: I had a look at vector sources and I believe they are included now. But I could not find any unittests for vector sources so it is a bit hard to know for sure. Maybe you can point me to the correct test?

src/porepy/params/bc.py
rbe051: Both yes and no. The BoundaryCondition class does not know anything about the geometry of the faces, it only stores which faces should be coupled. So it could couple two faces that do not matches. It does have a limitation that it can only couple exactly two faces. 

But the discretization assumes that the faces are matching. This is the only thing that makes sense for the TPFA discretization if the periodic faces are supposed to be treated exactly the same as an internal face (the face between two cells is assumed to be unique).

I think adding a mortar variable is the way to go if we want non-matching faces. In that case, we would have to rethink the per_map in the BoundaryCondition class.

src/porepy/params/bc.py
rbe051: done

src/porepy/numerics/fv/tpfa.py
keileg: unit/test_mpfa_gravity.py. If you have a suggestion for a better name for the file, I am listening.

src/porepy/params/bc.py
keileg: Let's leave it like this for now, then.

src/porepy/numerics/fv/tpfa.py
rbe051: Maybe test_fv_vector_source.py ?

keileg:
@rbe051 : Do you consider yourself done here?

rbe051:
> @rbe051 : Do you consider yourself done here?

Yes, will merge

PR #388, opened 2020-04-27
Fixed logical error and typo in Exporter.

rbe051:
The logical error was wrong condition in the if statement checking if time step is given.


alessiofumagalli:
It's also good to me @rbe051 :+1:

PR #401, opened 2020-05-12
Mixed schemes gravity

alessiofumagalli:
The branch introduces the vector source term for both the rt0 and mvem discretizations. Moreover, the operator that created the reconstruction of the Darcy velocity is now a matrix and can be used in a more flexible way. Type hinting are added for the classes RT0, MVEM, and DualElliptic. Tests are added to check the vector source term.


src/porepy/numerics/fem/rt0.py
```diff
@@ -15,13 +15,13 @@
 
 
 class RT0(pp.numerics.vem.dual_elliptic.DualElliptic):
-    def __init__(self, keyword):
+    def __init__(self, keyword: str):
```
keileg, line 18:
```suggestion
    def __init__(self, keyword: str) -> None:
```

src/porepy/numerics/fem/rt0.py
```diff
@@ -15,13 +15,13 @@
...
         super(RT0, self).__init__(keyword, "RT0")
         # variable name to store the structure that map a cell to the opposite nodes
         # of the local faces
         self.cell_face_to_opposite_node = "rt0_class_cell_face_to_opposite_node"
 
-    def discretize(self, g, data):
+    def discretize(self, g: pp.Grid, data: dict):
```
keileg, line 24:
```suggestion
    def discretize(self, g: pp.Grid, data: Dict) -> None:
```

keileg: You will also need from typing import Dict.

src/porepy/numerics/fem/rt0.py
```diff
@@ -57,9 +57,9 @@ def discretize(self, g, data):
         matrix_dictionary = data[pp.DISCRETIZATION_MATRICES][self.keyword]
         # If a 0-d grid is given then we return an identity matrix
         if g.dim == 0:
-            mass = sps.dia_matrix(([1], 0), (g.num_faces, g.num_faces))
-            matrix_dictionary["mass"] = mass
-            matrix_dictionary["div"] = sps.csr_matrix((g.num_faces, g.num_cells))
+            matrix_dictionary[self.mass_matrix_key] = sps.dia_matrix(([1], 0), (1, 1))
```
keileg, line 60: Very good!

src/porepy/numerics/fem/rt0.py
```diff
@@ -74,9 +74,10 @@ def discretize(self, g, data):
         # Map the domain to a reference geometry (i.e. equivalent to compute
         # surface coordinates in 1d and 2d)
         deviation_from_plane_tol = data.get("deviation_from_plane_tol", 1e-5)
```
keileg, line 76: As a general comment, I think this is a rather loose tolerance for deviations; in practice we need to be much stricter not to get problems in the meshing. But that probably is an argument that the test is not much needed.

Another issues is that it is stupid not to have a relative tolerance here, but that is common to all computational geometry in pp, and it is my bad.

src/porepy/numerics/vem/dual_elliptic.py
```diff
@@ -66,11 +67,23 @@ class DualElliptic(
 
     """
 
-    def __init__(self, keyword, name):
+    def __init__(self, keyword: str, name: str):
```
keileg, line 70:
```suggestion
    def __init__(self, keyword: str, name: str) -> None:
```

src/porepy/numerics/vem/mvem.py
```diff
@@ -22,10 +23,10 @@ class MVEM(pp.numerics.vem.dual_elliptic.DualElliptic):
 
     """
 
-    def __init__(self, keyword):
+    def __init__(self, keyword: str):
         super(MVEM, self).__init__(keyword, "MVEM")
 
-    def discretize(self, g, data):
+    def discretize(self, g: pp.Grid, data: dict):
```
keileg, line 29:
```suggestion
    def discretize(self, g: pp.Grid, data: Dict) -> None:
```

src/porepy/numerics/vem/mvem.py
```diff
@@ -22,10 +23,10 @@ class MVEM(pp.numerics.vem.dual_elliptic.DualElliptic):
 
     """
 
-    def __init__(self, keyword):
+    def __init__(self, keyword: str):
```
keileg, line 26:
```suggestion
    def __init__(self, keyword: str) -> None:
```

src/porepy/numerics/fem/rt0.py
alessiofumagalli: yes, I agree with you..

PR #403, opened 2020-05-13
Zero-dimensional grid should not have faces or nodes

rbe051:
This pull-request set  the number of faces and nodes from the PointGrid to zero as nodes and faces does not make sense in a 0d grid. Previously ,they where just given the identity mapping to the cell-center. However, this does not work when we have a 0d grid with several (disconnected) cells. 

This is a more conceptual choice, so you @keileg  gets to voice your opinion.

@alessiofumagalli : can you check the lines:
https://github.com/pmgbergen/porepy/blob/4b41d1dca984837b833a68705b4598c0c44ad2b2/src/porepy/numerics/vem/dual_elliptic.py#L155-L160

The 0d grids gives a mass matrix of dimension 0x0 which makes sps.linalg.norm crash. Have I made reasonable choice in setting the norm to 1?

### DIFF
--- a/src/porepy/grids/partition.py
+++ b/src/porepy/grids/partition.py
@@ -484,7 +484,13 @@ def __extract_submatrix(mat, ind):
     cols = sub_mat.indptr
     data = sub_mat.data
     unique_rows, rows_sub = np.unique(sub_mat.indices, return_inverse=True)
-    return sps.csc_matrix((data, rows_sub, cols)), unique_rows
+    shape = (unique_rows.size, cols.size - 1)
+    if shape[0]!=0:
+        A = sps.csc_matrix((data, rows_sub, cols))
+        if np.any(A.shape != shape):
+            import pdb; pdb.set_trace()
+    
+    return sps.csc_matrix((data, rows_sub, cols), shape), unique_rows
 
 
 def __extract_cells_from_faces(g, f, is_planar):
--- a/src/porepy/grids/partition.py
+++ b/src/porepy/grids/partition.py
@@ -485,11 +485,6 @@ def __extract_submatrix(mat, ind):
     data = sub_mat.data
     unique_rows, rows_sub = np.unique(sub_mat.indices, return_inverse=True)
     shape = (unique_rows.size, cols.size - 1)
-    if shape[0]!=0:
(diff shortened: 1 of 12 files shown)


alessiofumagalli:
the 0d should be an interface between the intersection of 1d grids. since the latter are linear, how can you have a 0d grid with multiple cells?

rbe051:
> the 0d should be an interface between the intersection of 1d grids. since the latter are linear, how can you have a 0d grid with multiple cells?

By having grid cells that are not connected. The same is true in 2d (imagine two separated domains that are gridded by one Grid). To be concrete:  The way I ended up with multiple cells was by merging all grids of the same dimension to one grid. So all 0d grids are represented by one node (and thus one grid with multiple cells) in the grid bucket.

keileg:
I have not yet looked at the code, the comments on runtime were more interesting. I guess there are two questions here: First the immediate one on whether the 0d grids should have points or not. Here I see no particular issues, we just have to be careful when writing the discretizations. So no particular objections there.

More generally on the merging of grids: This will only work for a subset of problems; as an immediate example, mpfa and mpsa will fail if applied to a non-planar grid. So it is not a general solution to the performance problems @rbe051 points to, but it is nevertheless a useful option to be aware of for people who know what they are doing. 

I'll do the formal review now.

src/porepy/grids/partition.py
```diff
@@ -484,7 +484,13 @@ def __extract_submatrix(mat, ind):
     cols = sub_mat.indptr
     data = sub_mat.data
     unique_rows, rows_sub = np.unique(sub_mat.indices, return_inverse=True)
-    return sps.csc_matrix((data, rows_sub, cols)), unique_rows
+    shape = (unique_rows.size, cols.size - 1)
+    if shape[0]!=0:
+        A = sps.csc_matrix((data, rows_sub, cols))
+        if np.any(A.shape != shape):
+            import pdb; pdb.set_trace()
```
keileg, line 491: Something like this, although the question marks should be replaced by an informative message. Also please add a comment on whath the outer if does.
```suggestion
            raise ValueError("???")
```

rbe051: Both the inner and outer loop was for debugging only. I removed both

PR #404, opened 2020-05-15
add a function to export to vtk 2d fracture networks

alessiofumagalli:
not much to add, not sure if it's possible to add some test since on travis we don't check vtk as far as I remember


alessiofumagalli:
@keileg I know that we have planned a refactoring of the vtk export but we should include also this part. let me know what do you think about the pr.

PR #409, opened 2020-05-21
fixes and a feature to pp.GridBucket methods

haakon-e:
# Overview
This PR primarily contains 1 fix and 1 new feature. Some minor stylistic changes have also been made.

Commit 53ac1ef7 fixes an issue for `add_node_props` and `add_edge_props` where if a string is passed to the methods, it would add each individual character to the data. This due to strings being iterable and the line: `keys=list(keys)`.

Commit 7c27ff44 adds the argument `overwrite` to `add_node_props` and `add_edge_props`. By default, this argument is set to `True`, so existing behaviour remains the same. However, if you set it to `False`, if the any of the keys to be added already exists on any of the grids or edges, we leave its data unchanged.
I have also made some stylistic simplifications to these methods, which improves code readability (especially for edges).

The other commits are summarized as follows:
* 40c7de6c, d471b304, 799ad70c: Update documentation and/or typing
* 7ddf4794: Various stylistic changes to the code

### DIFF
--- a/src/porepy/grids/grid_bucket.py
+++ b/src/porepy/grids/grid_bucket.py
@@ -29,8 +29,8 @@ T = TypeVar("T")
 
 class GridBucket(Generic[T]):
     """
-    Container for the hiererchy of grids formed by fractures and their
-    intersection.
+    Container for the hierarchy of grids formed by fractures and their
+    intersections.
 
     The mixed-dimensional grid can be considered a graph, with the fixed-dimensional grids
     defining nodes, and interfaces edges. The terminology in the implementation,
@@ -40,16 +40,16 @@ class GridBucket(Generic[T]):
     and the the lowest-dimensional neighboring grid.
 
     Attributes:
-        _nodes (dictionary): For all fixed-dimensional grids, a dictionary with
-            grids as keys, and data associated with the grid as node.
-        _edges (dictionary): For all fixed-dimensional grids, a dictionary with
-            the edges as keys, and data associated with the grid as node.
+        _nodes (dictionary): For all mixed-dimensional grids, a dictionary with
+            grids as keys, and data associated with the grid as values.
+        _edges (dictionary): For all mixed-dimensional grids, a dictionary with
+            the edges as keys, and data associated with the grid as values.
 
     """
 
     def __init__(self) -> None:
-        self._nodes = {}
-        self._edges = {}
+        self._nodes: Dict[pp.Grid, Dict] = {}
+        self._edges: Dict[Tuple[pp.Grid, pp.Grid], Dict] = {}
         self.name = "grid bucket"
 
     def __contains__(self, key: Any) -> bool:
@@ -80,12 +80,11 @@ class GridBucket(Generic[T]):
     # --------- Iterators -------------------------
 
     def __iter__(self) -> Generator[Tuple[pp.Grid, Dict], None, None]:
-        """
-        Iterator over the nodes in the GridBucket.
+        """ Iterator over the nodes in the GridBucket.
 
         Yields:
-            core.grid.nodes: The grid associated with a node.
-            data: The dictionary storing all information in this node.
+            g (pp.Grid): The grid associated with a node.
+            data (dict): The dictionary storing all information in this node.
 
         """
         for g, data in self._nodes.items():
@@ -98,8 +97,8 @@ class GridBucket(Generic[T]):
         with self.edges()
 
         Yields:
-            g: Grid associated with the current node.
-            data: The dictionary storing all information in this node.
+            g (pp.Grid): Grid associated with the current node.
+            data (dict): The dictionary storing all information in this node.
 
         """
         for g, data in self._nodes.items():
@@ -110,8 +109,10 @@ class GridBucket(Generic[T]):
         Iterator over the edges in the GridBucket
 
         Yields:
-            e: Grid pair associated with the current edge.
-            data: The dictionary storing all information in this edge..
+            e (Tuple[pp.Grid, pp.Grid]):
+                Grid pair associated with the current edge.
+            data (dict):
+                The dictionary storing all information in this edge..
 
         """
         for e, data in self._edges.items():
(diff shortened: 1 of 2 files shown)


src/porepy/grids/grid_bucket.py
```diff
@@ -940,11 +941,11 @@ def _find_shared_face(self, g0: pp.Grid, g1: pp.Grid, g_l: pp.Grid) -> np.ndarra
             cells 1 and 2 of g1 (second dimension of the array).
         """
         # Sort nodes according to node_number
-        g0, g1 = self.nodes_of_edge([g0, g1])
+        g0, g1 = self.nodes_of_edge((g0, g1))
```
alessiofumagalli, line 944: why this change?

haakon-e: This is because `nodes_of_edge` expects a  `Tuple[pp.Grid, pp.Grid]`. 
However, this is of course a completely arbitrary choice, and you could equivalently extend the expected type of `nodes_of_edge` to include `List[pp.Grid, pp.Grid]`.
However, since the norm is that we represent edges as tuples, not lists, I made this change.

(on a practical level; I spotted this simply because my IDE, PyCharm, notified me of the inconsistency).

haakon-e:
I added two more commits. 

Commit 14950f4 essentially mirrors the stylistic changes seen in 7c27ff4 and fixes the `keys=list(keys)` bug (similar to 53ac1ef).

Commit 491b002 mostly updates the documentation stylisticly, and corrects some typos, etc. I've also added typing to `_nodes` and `_edges` in `__init__`. I think this makes those constructs clearer for newcomers.

------
I also contemplated introducing two custom types, namely
```
Node = pp.Grid
Edge = Tuple[pp.Grid, pp.Grid]
```
and use these as aliases for all typing that refers to these constructs. However, decided against making these changes since I was not sure whether they improve clarity of the code. Do let me know whether you think something along those lines is a good idea though.

src/porepy/grids/grid_bucket.py
```diff
@@ -40,16 +40,16 @@ class GridBucket(Generic[T]):
     and the the lowest-dimensional neighboring grid.
 
     Attributes:
-        _nodes (dictionary): For all fixed-dimensional grids, a dictionary with
-            grids as keys, and data associated with the grid as node.
-        _edges (dictionary): For all fixed-dimensional grids, a dictionary with
-            the edges as keys, and data associated with the grid as node.
+        _nodes (dictionary): For all mixed-dimensional grids, a dictionary with
```
IvarStefansson, line 43:
```suggestion
        _nodes (dictionary): For each subdomain grids, a dictionary with
```

src/porepy/grids/grid_bucket.py
```diff
@@ -40,16 +40,16 @@ class GridBucket(Generic[T]):
...
+            grids as keys, and data associated with the grid as values.
+        _edges (dictionary): For all mixed-dimensional grids, a dictionary with
```
IvarStefansson, line 45:
```suggestion
        _edges (dictionary): For each interface grid, a dictionary with
```

src/porepy/grids/grid_bucket.py
```diff
@@ -59,7 +59,7 @@ def __contains__(self, key: Any) -> bool:
             key (object): Object to be tested.
 
         Return:
-            True if etiher key is a pp.Grid, and key is among the nodes of the graph
+            True if either key is a pp.Grid, and key is among the nodes of the graph
                 representation of this md-grid, *or* key is a 2-tuple, with both items
                 in self._nodes.
```
IvarStefansson, line 64: @keileg should all 2-tuples of grids be included, or only those which are exactly/at most one dimension apart?

src/porepy/grids/grid_bucket.py
```diff
@@ -308,70 +312,64 @@ def add_node_props(
...
+                if overwrite or key not in data:
+                    data[key] = None
 
     # ------------ Getters for node and edge properties
 
-    def has_nodes_prop(self, grids: Iterable[pp.Grid], key: Any) -> Tuple[Any]:
-        """
-        Test if a key exists for a node property of the bucket, for several nodes.
+    def has_nodes_prop(self, grids: Iterable[pp.Grid], key: Any) -> Tuple[bool]:
+        """ Test if a key exists for a node property of the bucket, for several nodes.
```
IvarStefansson, line 366:
```suggestion
        """ Test if a key exists in the data dictionary related to each of a list of nodes.
```

src/porepy/grids/grid_bucket.py
keileg, line 64:
```suggestion
                representation of this md-grid, *or* key is a 2-tuple, with both items
                in self._nodes and the tuple is in self._edges.
```

src/porepy/grids/grid_bucket.py
```diff
@@ -508,20 +502,19 @@ def remove_node_props(
...
-                    if h in g:
-                        del d[key]
+            for grid in g:
+                data = self._nodes[grid]
+                if key in data:
+                    del data[key]
 
     def remove_edge_props(
-        self, keys: Union[Any, List[Any]], e: List[Tuple[pp.Grid, pp.Grid]] = None
+        self, keys: Union[Any, List[Any]], edges: List[Tuple[pp.Grid, pp.Grid]] = None
```
keileg, line 517: This is really picky, but if you want edges instead of e, should it not also be grids instead of g in the corresponding method for nodes?

src/porepy/grids/grid_bucket.py
keileg: Agreed, edges are tuples.

src/porepy/grids/grid_bucket.py
haakon-e: Generally, I think we should avoid using one-letter variables in the _public_ API as it should be as intuitive as possible.
However this PR wasn't intended as a full overhaul of `GridBucket`, so I haven't by any means been fully consistent with my changes.

src/porepy/grids/grid_bucket.py
haakon-e:
```suggestion
        _nodes (dictionary): For each subdomain grid, a dictionary with
```

PR #417, opened 2020-06-15
build: PorePy for Windows! and transition to modern dependency management

haakon-e:
# Overview
This PR has the main goal of running porepy on Windows OS.

In the process, two main contributions have been made.
1. Relax version restrictions for nearly all dependencies.
2. Use [poetry](https://python-poetry.org/) as dependency manager instead of the plain `requirements.txt` file.

I have tested this sofar as running a script that 
- creates and meshes an unstructured fractured grid
- sets up and runs a biot problem with contact mechanics
- ~~exports the solution to vtk.~~ (not verified yet. Current test failed on an unrelated issue)

# Specifics
## PorePy on Windows
If you try to install porepy using Windows OS, a number of conflicts occur. The relaxation of versions of certain dependencies have solved most of these issues.
Cython depends on a C compiler. I have installed [Visual Studio 2019](https://visualstudio.microsoft.com/vs/) with the [necessary addons](https://visualstudio.microsoft.com/vs/features/cplusplus/) to allow cython to work properly.
According to the [Cython documentation](http://docs.cython.org/en/latest/src/quickstart/install.html), the [MinGW](http://www.mingw.org/) is another possibility.
## What is [poetry](https://python-poetry.org/)?
Poetry is essentially a python package that helps python projects manage consistent dependency versions across all your dependencies. 
Instead of a `requirements.txt` (and `...-dev.txt`), you'd have a `pyproject.toml` that lists dependencies (and dev dependencies).
To develop the project, simply navigate to it in your terminal and type `poetry install`, which creates a regular virtual environment local to that project. To activate the venv: `poetry shell` or `poetry run python`.
The package-specific stuff is stored in a `poetry.lock` file; which is generated from the `pyproject.toml` file.

* robust_point_in_polyhedron
To make this package work with poetry, I created a folder called robust_point_in_polyhedron somewhere and put the `robust_point_in_polyhedron.py` file inside. Then, in navigate to the folder in terminal and do: `poetry init` (press enter to get through the options, and say no to interactively defining packages). Then, do `poetry build`. This builds wheel files etc.
To add this as a dependency to porepy, type: `poetry add ../path/to/robust_point_in_polyhedron`

### Failing tests
Currently the following tests fails:
```
FAILED test/integration/test_2d_benchmark_geiger.py::TestVEMOnBenchmark::test_vem_blocking_coarse - ImportError: DLL load failed: Den angitte modulen ble ikke funnet.
FAILED test/integration/test_2d_benchmark_geiger.py::TestVEMOnBenchmark::test_vem_permeable_coarse - ImportError: DLL load failed: Den angitte modulen ble ikke funnet.
FAILED test/unit/test_fracture_importer.py::TestImport3dElliptic::test_create_fracture - TypeError: object of type <class 'numpy.float64'> cannot be safely interpreted as an integer.
FAILED test/unit/test_mcolon.py::TestMColon::test_type_conversion - AssertionError: False is not true
FAILED test/unit/test_vtk.py::BasicsTest::test_gb_1 - AssertionError: False is not true
FAILED test/unit/test_vtk.py::BasicsTest::test_gb_2 - AssertionError: False is not true
FAILED test/unit/test_vtk.py::BasicsTest::test_single_grid_1d - AssertionError: False is not true
FAILED test/unit/test_vtk.py::BasicsTest::test_single_grid_2d_cart - AssertionError: False is not true
FAILED test/unit/test_vtk.py::BasicsTest::test_single_grid_2d_polytop - AssertionError: False is not true
FAILED test/unit/test_vtk.py::BasicsTest::test_single_grid_2d_simplex - AssertionError: False is not true
FAILED test/unit/test_vtk.py::BasicsTest::test_single_grid_3d_cart - AssertionError: False is not true
FAILED test/unit/test_vtk.py::BasicsTest::test_single_grid_3d_polytop - AssertionError: False is not true
FAILED test/unit/test_vtk.py::BasicsTest::test_single_grid_3d_simplex - AssertionError: False is not true
```

* The first two tests relate to import of a certain scipy package. This is possibly due to an inconsistency between 32-bit and 64-bit installations on my local computer:
```
>   from scipy.optimize import minpack2
E   ImportError: DLL load failed: Den angitte modulen ble ikke funnet.
```
* The next two are type conversion tests for numpy.
* The rest are assertion errors for VTK. This is possible due to the installation of v 9.0.0, which is available on Windows, but not Linux.
    * However, by manual inspection (of the first test), I couldn't find any obvious differences between the generated file and the validation case, so these might be fine.

# Final note: Why is this PR relveant?
I took the MAT254 course fall 2019 and can attest to the many struggles the students had to set up  a Virtual Machine on their laptop to install porepy. For some, this process took weeks.
If we can get a stable Windows version (at the least as a branch!), I think we can save lots of trouble for upcoming students (who after all are our potential future research collaborators!!)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,6 @@ jobs:
         pip install shapely
         pip install shapely[vectorized]
         pip freeze
-        pip install black flake8 mypy
 
     - name: black
       if: ${{ always() }}
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,0 +1,716 @@
+[[package]]
+category = "dev"
+description = "Atomic file writes."
+marker = "sys_platform == \"win32\""
+name = "atomicwrites"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+version = "1.4.0"
+
+[[package]]
+category = "dev"
+description = "Classes Without Boilerplate"
+name = "attrs"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+version = "19.3.0"
+
+[package.extras]
+azure-pipelines = ["coverage", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "zope.interface", "pytest-azurepipelines"]
+dev = ["coverage", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "zope.interface", "sphinx", "pre-commit"]
+docs = ["sphinx", "zope.interface"]
+tests = ["coverage", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "zope.interface"]
+
+[[package]]
+category = "dev"
+description = "Cross-platform colored terminal text."
+marker = "sys_platform == \"win32\""
+name = "colorama"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+version = "0.4.3"
+
+[[package]]
+category = "dev"
+description = "Code coverage measurement for Python"
+name = "coverage"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4"
+version = "5.1"
+
+[package.extras]
+toml = ["toml"]
+
+[[package]]
+category = "main"
+description = "Composable style cycles"
+name = "cycler"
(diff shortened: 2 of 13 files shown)


keileg:
Looks interesting - I'll have a closer look within a few days. For version of dependencies, I completely agree, see (the stalled) #371. Same with poetry - basically what we have now was the recommended solution some years ago, and then nobody cared once it worked.

When it comes to Windows, support for this will not be an explicit policy; simply because we (in practice me) do not have the capacity to follow up on this. That said, it is of course interesting that you got it to work - for what it is worth, I have had similar experiences while playing with Windows.

PR #427, opened 2020-06-28
Mixed dim assembler filters

haakon-e:
# Overview
This PR implements additional logic to assembler filters.
In addition, general improvements to variable naming, documentation and typing is done.

## Extend `variable_filter` and `term_filter`
Previously, variable and term filters are constructed by passing a list of variables and terms, respectively, which restricts which operators will be discretized. The filters are "strict" in the sense that only terms contained in the respective filters will pass.

This PR extends this functionality such that you can pass a negative filter. That is: "discretize everything except variables/terms in the respective lists". In this proposal, the negative filters are indicated by prefixing each list item by "!" (means 'not'). 
For example:
`variable_filter = ["!temp"]`.
See also the documentation of `Assembler.discretize`, where I added a few examples.

## Filter by list of grids
Previously, you could restrict `Assembler.discretize()` to a single grid.

### Nodes
This PR extends the functionality to accept a list of grids on discretization of nodes.

### Edges and edge couplings
Edge discretizations and edge coupling discretizations have all previously been discretized, irrespective of whether any grid is passed to the grid filter.

This PR intends to allow grid filters to limit which edges/edge couplings are discretized.
Two options are possible. Either,
1) either the slave or master grid has to pass the grid filter, or
2) both the slave and master grid have to pass the grid filter.

### DIFF
--- a/src/porepy/numerics/mixed_dim/assembler.py
+++ b/src/porepy/numerics/mixed_dim/assembler.py
@@ -205,42 +205,60 @@ class Assembler:
         in a advection-diffusion system).
 
         The filters can be combined to select specified terms for specified equations.
+        See examples 1 and 2.
 
-        Example (discretization internal to a node or edge:
+        -- Experimental feature --
+        If you pass a filter where each item is prefixed by "!", this is interpreted as
+        exclusion. Then every variable/term will be discretized except those present in
+        the filter. See example 3 and 4.
+
+        Parameters
+        ----------
+        variable_filter : List[str] (optional)
+            List of variables to be discretized. If None (default), all active
+            variables are discretized.
+        term_filter : List[str] (optional)
+            List of terms to be discretized. If None (default), all terms for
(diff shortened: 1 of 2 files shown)


src/porepy/numerics/mixed_dim/assembler.py
```diff
@@ -248,22 +269,53 @@ def discretize(
...
 
-        Implemented actions are discretizaiton and assembly.
+        Implemented actions are discretization and assembly.
 
         """
+        def make_filter(var_term_list: List[str] = None) -> Callable[[str], bool]:
+            """ Construct a filter for variable and terms
+
+            The result is a callable which takes one argument (a string).
+            I
```
IvarStefansson, line 287: This looks a bit strange

haakon-e:
Grid filters ended up not being implemented in this PR. #428 will revisit it.

PR #428, opened 2020-06-29
Feat(DRAFT): mixed-dimensional assembler with generic filters

haakon-e:
# Overview
This PR proposes a new way to handle discretization of multi-physics systems.

The objects `NodeDiscretization`, `EdgeDiscretization` and `CouplingDiscretization` define the terms in your multi-physics system.
For example, the data contained in `data[pp.DISCRETIZATION]` for some node in a Grid Bucket, is a list of `NodeDiscretization` instances.

For simulation of non-linear, multi-physics system, it is often desirable to re-discretize particular terms on particular grids at certain times (for example before each Newton iteration). Flexible methods are needed to efficiently choose which terms you need to target for re-discretization, and which you want to leave as-is.
This PR introduces the class `DiscretizationFilter`, which you should subclass, and define the logic for which terms you want the assembler to discretize.

#### Todo:
- [ ] Discuss the overall idea after the summer.
- [ ] Accommodate other dependent code to this implementation
- [ ] Update existing tests and write new tests
- [ ] Implement things from discussions

# Details
### Equation terms
The core contribution of this is the implementation of [pydantic](https://pydantic-docs.helpmanual.io/) classes for terms.
This contribution achieves two things, 1) the deep and convoluted dictionary layers of `data[pp.DISCRETIZATION]` and `data[pp.COUPLING_DISCRETIZATION]` are flattened, and 2) by using [pydantic](https://pydantic-docs.helpmanual.io/), each term can be validated for consistency as they are created; avoiding unexpected errors upon discretization.

The assembler code relevant to discretization is re-written to accomodate these new objects, removing several for-loop layers compared to the previous code. All features of the existing code (including terms defined on edges that are not coupling terms, and edge coupling terms pointing only to a master or slave grid) is preserved.

### Discretization Filter
A generic `DiscretizationFilter` class is implemented. No restrictions are imposed by modularizing filtering as all relevant data is passed to the appropriate methods. 
For example, for each term to be discretized on a grid, the assembler sends the `NodeDiscretization` instance and the grid to the `node_filter`, which decides whether the term is to be discretized.

You can pass a discretization filter when initializing the Assembler class. Otherwise, the default filter is to discretize everything. Optionally, you can directly pass a discretization filter when running `assembler.discretize`, which will use the filter you passed, but not overwrite the filter from initialization.

### Tests
No thorough testing of this setup has been done yet as this was really just a quick write-up. We'll discuss the usefulness of this proposal over the summer.

# Comments for discussion
* I have not removed the list of `active_variables`, which is passed to the Assembler upon initialization. In principle this could also be incorporated into the filter method.
    * Another reason for not touching this at the moment, is that it is used by the method `_identify_dofs`, which I haven't really touched.
* A fairly straight-forward and possibly useful extension is to let the `DiscretizationFilter` directly return a list of the terms that are to be discretized. Then, the discretization process could be parallelized.
* It could be useful to implement some commonly used discretization filters.
* Following this pattern, the data in `pp.PRIMARY_VARIABLES` could also be parameterized using pydantic models (this is relevant to the point on `active_variables`).
* Currently, no filtering (except `active_variables`) is performed on assembly, but changing a few lines of code can implement filtering on assembly.

## End notes
Although my name is on all the commits, this wouldn't have been possible without the feedback from @keileg, and thorough discussions with @IvarStefansson, who both are interested in a flexible, and suistainable way to handle discretization filtering of multi-physics systems. This is one (of possibly many) ways to implement that.

This PR essentially overwrites the #427, which implements a light, non-intrusive extension of the current filters (exclusion filters).

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,5 +1,13 @@
 [[package]]
 category = "dev"
+description = "A small Python module for determining appropriate platform-specific dirs, e.g. a \"user data dir\"."
+name = "appdirs"
+optional = false
+python-versions = "*"
+version = "1.4.4"
+
+[[package]]
+category = "dev"
 description = "Atomic file writes."
 marker = "sys_platform == \"win32\""
 name = "atomicwrites"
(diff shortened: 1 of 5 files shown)


keileg:
Some of the filter suggestions have been included. As for the access to discretization objects, and how to write equations in PorePy, I am working on an extension which will substantially improve this, together with a lot of other stuff. To be continued.

I will close this now, let me know if you have comments or suggestions.

PR #435, opened 2020-07-15
build: Dockerfile for PorePy with Intelpython base

haakon-e:
# Overview
Add Dockerfile for PorePy using Intelpython as base image.
This allows very efficient use of numpy, scipy, numba, etc. compiled by Intel.
Install pypardiso, which is a large sparse solver for linear systems.

passes all porepy tests


keileg:
@haakon-e : This has been dormant forever, my bad. Any reason not to simply merge it, once the tests are done?

PR #436, opened 2020-07-16
Turn off contact mechanics feature 

IvarStefansson:
@keileg introduced a feature preventing fractures with zero tangential traction from being discretised as gliding. This has prevented convergence of the Newton scheme for me. This PR deactivates the feature. 

In terms of code, only a few lines are changed: 

- The actual check for zero tangential traction is removed.
- Change the order of checking values and computing coefficients for gliding cells to avoid dividing by zero. @haakon-e, could you check my logic here? I.e. that the resulting coefficients are not changed in any of the cases.

### DIFF
--- a/src/porepy/numerics/contact_mechanics/contact_conditions.py
+++ b/src/porepy/numerics/contact_mechanics/contact_conditions.py
@@ -33,7 +33,7 @@ import numpy as np
 
 import porepy as pp
 import logging
-from typing import Dict
+from typing import Dict, Tuple
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


src/porepy/numerics/contact_mechanics/contact_conditions.py
```diff
@@ -544,16 +548,18 @@ def _sliding_coefficients(
...
-
-        if alpha < 0:
-            beta = 1 / (1 - alpha * delta)
-        else:
-            beta = 1
+        # alpha, beta, delta. In degenerate cases, use
+        beta = 1
+        # Avoid division by zero:
+        l2_Tt = self._l2(-Tt)
+        if not np.isclose(l2_Tt, 0):
```
haakon-e, line 555: Vil denne sjekken tilsvare det du fjerner lengre oppe?
Burde du sjekke om denne er nærme self.tol eller stemmer det med 0?

IvarStefansson: Nei, den skal bare unngå å dele på null lokalt i beregningen av beta.
self.tol er et godt spørsmål. Svaret er nok sikkert ja, for konsistens.

IvarStefansson, line 555:
```suggestion
        if l2_Tt > self.tol:
```

PR #442, opened 2020-07-30
Error normalization

IvarStefansson:
There are far too many options for how to normalize errors! Instead of trying to anticipate them, I introduced "normalizations" to the error computation code, so that the user may specify how to normalize. Defaults to old version, i.e. norm of reference solution.

Also caught a bug and generalized to 0d intersections in the grid matching.

### DIFF
--- a/src/porepy/grids/match_grids.py
+++ b/src/porepy/grids/match_grids.py
@@ -450,7 +450,7 @@ def gb_refinement(
         else:
             raise NotImplementedError("Unknown refinement mode")
 
-        gb.set_node_prop(g=g, key="coarse_fine_cell_mapping", val=mapping)
+        gb.set_node_prop(grid=g, key="coarse_fine_cell_mapping", val=mapping)
 
 
 def structured_refinement(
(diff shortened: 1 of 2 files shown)


src/porepy/grids/match_grids.py
```diff
@@ -509,7 +509,9 @@ def structured_refinement(
     5. Assemble the mapping.
 
     """
-
+    if g.dim == 0:
+        mapping = sps.csc_matrix((np.ones(1), (np.array([0]), np.array([0]))))
```
haakon-e, line 513:
```suggestion
        mapping = sps.csc_matrix((np.ones(1), ([0], [0])))
```
I think this is equivalent, and (maybe?) easier to read.

src/porepy/utils/error.py
```diff
@@ -42,6 +45,10 @@ def grid_error(
         variable = [variable]
     if not isinstance(variable_dof, list):
         variable_dof = [variable_dof]
+    if normalizations is None:
+        normalizations = [lambda sol: np.linalg.norm(sol_ref, axis=0)] * len(variable)
+    elif not isinstance(normalizations, list):
+        normalizations = [normalizations]
```
haakon-e, line 51: Is this intended as a helper for the case where you only consider one variable?
So, if you have multiple variables and wish to apply the same normalization for all of them, you are yourself responsible to do `[my_normalization] * len(variable)`?

IvarStefansson: Indeed

PR #458, opened 2020-09-15
Axis for viz

rbe051:
The visualization routines should take an (optional) axes as argument and plot in this. This pull request only implements this for the fracture_visualization.py module. If no axes is given, the legacy behavior is obtained, except, an axes is returned (instead of None).

### DIFF
--- a/src/porepy/viz/fracture_visualization.py
+++ b/src/porepy/viz/fracture_visualization.py
@@ -36,9 +36,9 @@ def plot_fractures(pts, edges, domain=None, colortag=None, ax=None, **kwargs):
         domain = pp.bounding_box.from_points(pts)
 
     if ax is None:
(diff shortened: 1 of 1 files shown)


src/porepy/viz/fracture_visualization.py
```diff
@@ -25,12 +25,27 @@ def plot_fractures(pts, edges, domain=None, colortag=None, **kwargs):
             (e.g. by fracture family). If provided, different colors will be
             asign to the different families. Defaults to all fractures being
             black.
+        ax (matplotlib.axes.Axes, optional): If not given an axis, an axis will be created.
```
alessiofumagalli, line 28: only one comment, why not to put ax in the kwargs? for the rest okay

rbe051: I did really have any specific reason.

PR #459, opened 2020-09-23
Fix output bug from shapely

rbe051:
If the line does not cross the polygon shapely returns a line of length 0. This is only on my desktop computer and not my laptop. The difference must be due to version numbers, even if I have tried to get the same versions. Anyway, I don't think the extra check should hurt


rbe051:
Ups, merged wrong branch

PR #460, opened 2020-09-23
Handle lines of length 0 that are returned from shapey

rbe051:
If the line does not cross the polygon shapely returns a line of length 0. This is only on my desktop computer and not my laptop. The difference must be due to version numbers, even if I have tried to get the same versions. Anyway, I don't think the extra check should hurt (This is the correct branch from pull request #459 )


keileg:
I agree, this should not hurt.

PR #461, opened 2020-09-25
Speedup of gmsh reading

rbe051:
The tagging of fracture faces took an insane amount of time when the number of fracture faces was large. Rewrote the algorithm to make it faster and now the timing is quite good. However, I'm not quite sure why we have to do this at all, as the 3d version does not create these tags. Are these tags used anywhere?


test/integration/test_simplex.py
```diff
@@ -0,0 +1,58 @@
+import unittest
```
keileg, line 1: Can this be merged with test/unit/test_tag_gmsh.py? Alternatively with unit/test_tag (those two files should also have been merged, but that is another matter).

test/integration/test_simplex.py
```diff
@@ -0,0 +1,58 @@
...
+import numpy as np
+
+import porepy as pp
+
+
+class TestFaceTags(unittest.TestCase):
+    def test_face_tags_from_gmsh(self):
```
keileg, line 8: Is this different from test/unit/test_tag_gmsh? Please explain what is tested.

test/integration/test_simplex.py
```diff
@@ -0,0 +1,58 @@
...
+                    f_cc0 = g.face_centers[:, g.tags["fracture_0_faces"]]
+                    f_cc1 = g.face_centers[:, g.tags["fracture_1_faces"]]
+
+                    self.assertTrue(np.all(np.abs(f_cc0[1] - 0.5) < 1e-10))
+                    self.assertTrue(np.all(np.abs(f_cc1[0] - 0.5) < 1e-10))
+
+                db_cc = g.face_centers[:, g.tags["domain_boundary_faces"]]
+
+                self.assertTrue(np.all(g.tags["tip_faces"] == 0))
+                if g.num_faces == 0:
```
keileg, line 48: This would be 0d grids? Should be commented, but is it not better to explicitly test for `g.dim == 0` at the start of the inner loop?

test/integration/test_simplex.py
```diff
@@ -0,0 +1,58 @@
...
+                    continue
+
+                self.assertTrue(
```
keileg, line 51: Please comment what is the assertion really testing?

test/integration/test_simplex.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+        pp.meshing._tag_faces(grid_list, False)
+        for grid_of_dim in grid_list:
+            for g in grid_of_dim:
+                g.compute_geometry()
+                if g.dim == 2:
+                    f_cc0 = g.face_centers[:, g.tags["fracture_0_faces"]]
+                    f_cc1 = g.face_centers[:, g.tags["fracture_1_faces"]]
+
+                    self.assertTrue(np.all(np.abs(f_cc0[1] - 0.5) < 1e-10))
```
keileg, line 42: What is being tested here?

keileg (commented):
Main change looks good. See comments on testing, including moving it.

keileg:
.. and sorry it took me so long.

haakon-e:
Quick comment: Does it make sense to standardize whether tests should be written with pytest or unittest? Thoughts, @keileg?

keileg:
@haakon-e : Not sure, I have been thinking about the same. However, I don't think this is the weakest links for the tests right now; I would prioritize test documentation and consolidation before setting a test framework. 

But of course, we can say that new tests should be one of the format - and I presume your recommendation would be pytest then?

haakon-e:
I have mostly worked with pytest. Given that it supports everything unittest does (afaik) and that the tests are more transparent in my opinion, I think it's the way to go.
Although I do agree with you that it doesn't necessarily have to be prioritized at this instant.

test/integration/test_simplex.py
rbe051: Yes, I didn't see the unit/test_tag_gmsh.py as I was expecting to find it in integration tests. There is a difference between the two. The new one tests the tags before the splitting of the grids, while the unit/test_tag_gmsh.py tests the tags after splitting of the grid. However, I merged the two and moved them to integration tests. 

I also think integration/test_face_tag.py is exactly the same as unit/test_tag.py? I do not think these should be merged with the gmsh stuff because, well, they do not call gmsh. Anyway, I did not touch these files.

test/integration/test_simplex.py
rbe051: see previous comment

test/integration/test_simplex.py
rbe051: Not at the beginning of the loop because we want to do some tests on the zero-dim grids. I did re-organize the code to make it more clear

test/integration/test_simplex.py
rbe051: done

test/integration/test_simplex.py
rbe051: done

test/integration/test_simplex.py
keileg: Well, consolidation of the tests is a major task. I think we'll just go with this. At some point, someone will need to spend two days on sorting out all the tests, and then ideally two weeks deleting tests and writing new ones. Sigh.

PR #472, opened 2020-10-29
Replace master/slave by high/low for the whole distribution

jhabriel:
This PR introduces several changes related to the replacement of _master_ and _slave_ by _high_ and _low_, respectively. 
The changes were made from the top-level and below, affecting files inside `src` and `test`. This affects attributes, variables, methods, and documentation in general. However, the functionality of PorePy remains strictly unchanged.

In particular, the following methods from the  `MortarGrid` class will be renamed:
`master_to_mortar_avg` by `high_to_mortar_avg`
`master_to_mortar_int` by `high_to_mortar_int`
`slave_to_mortar_avg` by `low_to_mortar_avg`
`slave_to_mortar_int` by `low_to_mortar_int`
`mortar_to_master_avg` by `mortar_to_high_avg`
`mortar_to_master_int` by `mortar_to_high_int`
`mortar_to_slave_avg` by `mortar_to_low_avg`
`mortar_to_slave_int` by `mortar_to_low_int`
`update_master` by `update_high`
`update_slave` by `update_low`

FYI: I've used an IDE called IntelliJ IDEA to _relatively_ painlessly do the find&replace procedure. It might be useful for the future.

### DIFF
--- a/src/porepy/grids/coarsening.py
+++ b/src/porepy/grids/coarsening.py
@@ -352,7 +352,7 @@ def generate_seeds(gb):
         index = np.in1d(faces, tips).nonzero()[0]
         cells = np.unique(cells[index])
 
-        # recover the mapping between the slave and the master grid
+        # recover the mapping between the low and the master grid
         mg = gb._edges[(g_h, g)]["mortar_grid"]
         m2m = mg.high_to_mortar_int()
         l2m = mg.low_to_mortar_int()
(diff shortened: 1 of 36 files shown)


keileg:
Sorry about this, but I just got second thoughts regarding naming: In the case of equi-dimensional couplings (see #472), high and low may not be optimal. A different option is primary and secondary. What do you think?

jhabriel:
I agree, high/low won't be optimal for equidimensional couplings. There are other options:

- primary/secondary
- primary/replica
- leader/follower
- master/puppet
- master/minion
- parent/child

jhabriel:
Parent/child is a good option. However, they introduce the same inconsistency with equidimensional couplings and may generate ambiguity with classes. I guess my vote goes for high/ low, despite the mentioned flaw.

jhabriel:
I will close this PR, and create another one with primary/secondary instead of high/low. Possibly today.

PR #483, opened 2020-11-12
Vtk exporter update

alessiofumagalli:
We remove the vtk dependency from porepy, meshio is used instead which is already a dependency.

I add many reviewers cause this merge will affect all of you.
Please, check carefully if the new interface works as the old one.
The only thing you should modify is to write instead of ".write_vtk(" only ".write(", the rest is the same.

Please, @haakon-e and @keileg have a look also to the installation scripts. I tried to prune the vtk from them, not sure if I succeed.


alessiofumagalli:
@rbe051 and @IvarStefansson please check also speed-wise if it is okay

keileg:
So, some division of labor: I will take care of everything with install scripts, including purging the vtk package, no need for @haakon-e to do anything there. I will also use this on some simulations, including ones with evolving fracture geometries. Once I have some idea of the PR, I may assign some additional reviewers.

Regarding the scope of the PR: The ambition now is to purge the vtk package. A more thorough refactoring of the exporter is still on the cards, but that will have to wait for another day.

keileg:
Comments:
1. When I use the new version to export a 2d GridBucket, I get an error message from paraview when opening the 1d and mortar grid (but not 2d). It appears an offset error is given as double, not int. Not sure why this happens, but I am pretty sure this is not from meshio. The code needed to produce the error is:
``` python
import porepy as pp
import porepy_mesh_factory as pmf  # you need to download this repo and put in range of PYTHONPATH, but that you probably want to do anyhow
gb = pmf.main.generate(case="flow_benchmark_2d_case_3")
exp = pp.Exporter(gb, "/tmp/tmp.vtu")
exp.write_vtu()
```
I had a look at the exporter, but it was not clear to me what the reason is.

2. From what I understand, you still do the sorting of nodes in 2d and 3d grids, even if the grid type is known to be simplex or Cartesian. If correct, this will not give us the speedup in the exporter that I had hoped for. Is it possible to do something about this? And if you already have, sorry..

3. I have added typing to the exporter. This uncovered a host of minor bugs and inconsistencies; the only reason why the code actually worked was that it had been hacked to the point where the inconsistencies never kicked in in practice. The code should be much better now, but please review what I have done, and remove any changes.

keileg:
Also, when we both agree that the code should be merged, I will do an interactive rebase to cut the number of commits by an order of magnitude.

alessiofumagalli:
Hey, a couple of comments: 
- no need to specify the file format, it's actually considered as part of the file name
- I don't get any error from paraview (5.8.1) by opening these files
- you're  right there's an inconsistency in the data for the offset, I'll check

alessiofumagalli:
@keileg, I anyway forced the code to have int as connectivity data.

For your main question about the speedup, my main concern is related to the fact that a user has to specify the fact that a grid is composed by simplicial elements or quads. Otherwise I cannot know a-priori. How should we proceed? Do it now or leave it as next step?

Apart from this last point I think the branch is ready to be merged.

keileg:
I've changed my mind regarding tailoring to grid types - let's postpone this. I will merge now.

PR #484, opened 2020-11-18
Fixed memory bug in MPFA: Sparse matrix was converted to dense

rbe051:
Was there a reason to convert the matrices to dense matrices, @keileg ? For large grids the conversion is unfeasible (e.g. 200 000 cells -> 64 GB memory)


keileg:
Well, that was ironic: memory blow-up in a function meant to estimate peak memory. I'll have a look if the same thing happens in mpsa.

Thanks for pointing this out. I'll merge once I figure out what happened to the tests.

keileg:
for the record: Mpsa did not have the same problem.

PR #485, opened 2020-11-19
Removed recursive loop in normal computation and made it more robust

rbe051:
I think everyone agrees with removing the recursion. However, I am not sure about the change to using the mean for one of the points. I guess this may fail for some rare cases? The old way was no good on Cartesian grids, since points then are ordered by rows (if the grid is large enough we reach the maximum level of recursion). After removing the recursive loop, it will work on Cartesian grids without the mean, but you have to loop over the first row of cells, and end up with a normal and tangential vector that are almost parallel. 

Any thoughts ?

### DIFF
--- a/src/porepy/geometry/map_geometry.py
+++ b/src/porepy/geometry/map_geometry.py
@@ -332,13 +332,17 @@ def compute_normal(pts, check=True):
     normal: np.array, 1x3, the normal.
 
     """
-
     assert pts.shape[1] > 2
-    normal = np.cross(pts[:, 0] - pts[:, 1], pts[:, 2] - pts[:, 1])
-    if check and np.allclose(normal, np.zeros(3)):
-        return compute_normal(pts[:, 1:])
-    else:
-        return normal / np.linalg.norm(normal)
+    normal = np.zeros(3)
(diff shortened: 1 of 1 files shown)


keileg:
@rbe051 The recursion clearly has to go, and a change in the direction you're indicating is needed for Cartesian grids.

I cannot think of a case where the suggested change will not work. The other approach that I can think of would be to pick random points in the cloud, and use these as a basis. But in that way, it will take a long while to detect essentially collinear points.

What do you think, should we just go with this, if the tests work?

rbe051:
Apparently it broke some unittests, I will look into it

rbe051:
There are some unittests that are dependent on the normal vector as calculated by the compute_normal() (If the normal flip, so does the mortar_grid, which again changes the indexing of the mortar variables. If the indexing of the mortar variables are hard-coded the tests fails). If it was not Friday I might have had the motivation to update the tests, but I took the easy solution and flipped the normal vector instead.

rbe051:
To your question about if we should just go for this, I say yes. If it ever turns out to be a problem we can implement a fallback in the compute_normal() to use the old method.

PR #486, opened 2020-11-19
Fix bug in pip editable install mode

jhabriel:
I'm not entirely sure what's causing the error. It seems to be related either to `setuptools` or to `poetry`, or maybe both. Thus, this is more or less a brute force fix. I've found it here: https://github.com/python-poetry/poetry/issues/3001

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,5 +29,5 @@ mypy = "^0.790"
 flake8 = "^3.8.3"
 
 [build-system]
-requires = ["poetry>=0.12"]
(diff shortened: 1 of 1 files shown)


keileg:
@alessiofumagalli: You can confirm that this indeed fixed the issue you encountered?

If yes, I think we will merge. I will then do an extra check of this when I get around to testing python 3.9, but that is some steps down the road.

alessiofumagalli:
ready to merge for me, thanks @jhabriel

PR #495, opened 2020-12-15
Extend extrusion

rbe051:
The changes to fractures_2d is a small bugfix that does not really belong in this pullrequest, but it got included here anyway


keileg:
Looks reasonable. This is relevant for T-intersections? I may add a test for that before I merge.

keileg (approved):
Looks good,  but I may add a test before merging.

rbe051:
> Looks reasonable. This is relevant for T-intersections? I may add a test for that before I merge.

I guess it will also be relevant for T-intersections. Also fractures that are on domain boundary faces or curved fractures approximated by line segments

PR #498, opened 2020-12-16
MortarGrid sides are identified by Enum

keileg:
This leaves a more robust implementation than previously, when we relied on accessing module-level constants outside the module.

Also, added tests to check pickling of `Grid`, `MortarGrid` and `GridBucket`.

### DIFF
--- a/src/porepy/grids/mortar_grid.py
+++ b/src/porepy/grids/mortar_grid.py
@@ -1,6 +1,7 @@
 """ Module containing the class for the mortar grid.
 """
 import warnings
+from enum import Enum
 from typing import Dict, Generator, List, Optional, Tuple, Union
 
 import numpy as np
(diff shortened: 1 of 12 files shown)


src/porepy/grids/mortar_grid.py
```diff
@@ -1,19 +1,20 @@
...
-NONE_SIDE = 0
-LEFT_SIDE = 1
-RIGHT_SIDE = 2
-WHOLE_SIDE = np.iinfo(type(NONE_SIDE)).max
+
+class MortarSides(Enum):
+    # Enum of constants used to identify the grids on each side of the mortar
+    NONE_SIDE = 0
+    LEFT_SIDE = 1
+    RIGHT_SIDE = 2
```
alessiofumagalli, line 17: why did you remove the WHOLE_SIDE? We probably never used it, but maybe you've a specific reason

alessiofumagalli (approved):
It looks good, thanks @keileg

src/porepy/grids/mortar_grid.py
keileg: 1) I did not understand when we need it, so we'll introduce it later if necessary. 
2) If we reintroduce it, it should get a standard number, not some strange trick to get a random number.

PR #500, opened 2020-12-18
refactor: docs, typing, refactor code

haakon-e:
Various refactoring of the code

Pay particular attention to the suggestion that block matrix indices are set in the `__init__` functions of the respective parent classes, instead of explicitly writing 0, 1, 2.


keileg:
This one was left hanging for too long (we never really had the capacity to incorporate the relevant parts of the proposed changes), and in the meantime many of the proposed changes have been incorporated independently. 

Resurrecting the branch at this stage will not be worth it at this stage, so I'll close it.

PR #505, opened 2021-01-21
ENH: Export labels to networkx in FractureNetwork2D

rbe051:
Should there be an option to not export the tags?
I used this to be able to identify which edges in networkx belongs to which fractures in the network


keileg (approved):
Looks useful. I will merge.

PR #507, opened 2021-02-03
Add infinite colors to fracture viz

rbe051:
Just rotate on 20 colors if more than twelve fractures are given

### DIFF
--- a/src/porepy/viz/fracture_visualization.py
+++ b/src/porepy/viz/fracture_visualization.py
@@ -80,7 +80,7 @@ def plot_fractures(pts, edges, domain=None, colortag=None, ax=None, **kwargs):
                 (0, 0, 0.5),
             ]
         else:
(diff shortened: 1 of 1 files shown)


rbe051:
Pytest in 3.7 breaks because cython is not installed. Why mypy breaks in 3.7 and 3.8 I don't know. But it can not be because of the 1 line I changed?
I suggest merging this PR and figure out why 3.7 and 3.8 breaks in an other PR

keileg:
It likely breaks because the tests are run with numpy 1.20. I will see if I can make this work in a separate branch. Until then, we either enforce np < 1.20, or put this on hold; we will not merge before the tests are running.

keileg:
Support for 3.6 will also be dropped, since np no longer supports it. just as well.

keileg (approved):
I will take care of the merging once  I have dealt with numpy 1.20 et al

PR #508, opened 2021-02-03
[Feature] Added new functions to AD classes

rbe051:
including <, > len, min, max


keileg (approved):
Looks reasonable. I will take care of the merge.

PR #521, opened 2021-03-11
Fix bug when using multiple active Ad variables

jwboth:
The previous capability for solving for subsystems did only allow solving for a single field. The update allows for combining solution for multiple fields, e.g., given pressure, and two saturations, the current approach ultimately allows for solving for both saturations simultaneously.

Note: This PR is basis for the PR "Provide variant of distribute_variable of DofManager.".


keileg:
@jwboth It seems this PR har several commits in common with #522, at least the same parts of the code seems to be changed. Can we do one at a time?

jwboth:
@keileg This should also be deleted.

PR #523, opened 2021-03-12
Add treatment of scalar sources in the AD framework

jhabriel:
Now, it is possible to use scalar sources while defining AD equations. See, in particular, test_ad_equations.py for simple usage.


keileg (approved):
Looks good.  It breaks with the other discretizations, in that it does not use _wrap_discretization, etc., but that should be okay.

We should clean up the source implementation a some point (I don't see the need for the matrix term there etc), but that will have to be for another day.

Feel free to merge as you see fit.

PR #527, opened 2021-03-15
Grad p rhs

IvarStefansson:
If the reference scalar variable is nonzero, a rhs contribution arises for the stress. This is of particular relevance for thermal stresses, i.e.
sigma = stiffness : grad u - thermal_expansion (T-T_reference) I.
This PR:
- Makes GradP return a rhs.
- Introduces the same contribution in ContactMechanicsInterfaceLaws.
- Has the ContactMechanicsBiot and THM models assign default values of p_reference=T_reference=0.
- Tests the enhancement.
- Fixes a few minor bugs.


test/integration/test_contact_mechanics_biot.py
```diff
@@ -30,17 +30,84 @@ def _solve(self, setup):
...
+        m.p_reference = 1
+
+        u, p = self._solve(m)
+        # The initial and boundary conditions for pressure are 0. The absence of
+        # the positive reference pressure results in contraction of the medium,
+        # which is fixed at the bottom.
+        u_x = u[0::2]
+        u_y = u[1::2]
+        self.assertTrue(np.all(u_y < -m.zero_tol))
+        # Mirror indices left and right half of domain
```
keileg, line 63: This will break at some point, e.g. with an update to `gmsh`. These are cell indices? So if we did the test on a Cartesian grid, symmetri is to be expected as well?

test/regression/check_tutorials.py
```diff
@@ -21,8 +21,7 @@ def run_all():
             print("********************\n")
             failed = True
             failed_files.append(file)
-        cmd_delete = "rm " + new_file
-        os.system(cmd_delete)
+        os.remove(new_file)
```
keileg, line 24: Thanks!

keileg (commented):
My only concern is the hard coding of cell indices in two of the tests; experience shows that this will break at some point. Is it an option to do these tests on Cartesian grids - and if yes, will it be much work to change? I can contribute as well.

Beoynd that, a few minor suggestions.

keileg:
@IvarStefansson: let me know what you think about the comments on the tests. 

Once we have sorted this out, I will take over and do the update regarding Ad.

test/integration/test_contact_mechanics_biot.py
keileg: I understand that the grids are Cartesian, so disregard this comment.

PR #534, opened 2021-04-08
ENH: DFN's in 2d does not need auxillary points

rbe051:
When a fracture tip is almost touching another fracture an auxiliary point is inserted. For DFN's these auxillary points does not have
a purpose as there is no need to refine the mesh in these areas. 

This Pullrequest therefore make the FractureNetwork2D.prepare_for_gmsh( mesh_args, dfn=True) skip the insertion of auxillary points. As far as I understand this should be OK for DFN's, can you confirm @keileg ?

Note for the future: The pp.fracs.tools.determine_mesh_size(...) should probably change name (it does much more than determining the mesh size).

### DIFF
--- a/src/porepy/fracs/fractures_2d.py
+++ b/src/porepy/fracs/fractures_2d.py
@@ -545,7 +545,7 @@ class FractureNetwork2d(object):
         mesh_size_frac: Optional[float] = None,
         mesh_size_bound: Optional[float] = None,
         mesh_size_min: Optional[float] = None,
-    ):
+    ): -> None
         """
         Set the "Vailla" mesh size to points. No attemts at automatically
         determine the mesh size is done and no auxillary points are inserted.
--- a/src/porepy/fracs/fractures_2d.py
+++ b/src/porepy/fracs/fractures_2d.py
@@ -349,8 +349,16 @@ class FractureNetwork2d(object):
             self._snap_to_boundary()
 
         self._find_and_split_intersections(constraints)
-        self._insert_auxiliary_points(**mesh_args)
 
+        # Insert auxiliary points and determine mesh size.
+        # _insert_auxiliary_points(..) does both.
+        # _set_mesh_size_withouth_auxiliary_points() sets the mesh size
+        # to the existing points. This is only done for DFNs, but could
+        # also be used for any grid if that is desired.
+        if not dfn:
+            self._insert_auxiliary_points(**mesh_args)
+        else:
+            self._set_mesh_size_withouth_auxiliary_points(**mesh_args)
         # Transfer data to the format expected by the gmsh interface
         decomp = self._decomposition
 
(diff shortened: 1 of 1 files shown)


src/porepy/fracs/fractures_2d.py
```diff
@@ -531,6 +533,31 @@ def _insert_auxiliary_points(
         self._decomposition["edges"] = lines
         self._decomposition["mesh_size"] = mesh_size
 
+    @pp.time_logger(sections=module_sections)
+    def _set_dfn_mesh_size(
```
rbe051, line 537: This is needed because the _insert_auxillary_points also defines the mesh size. This is a bit of a hack, and I think a better approach would be to refactor the _insert_auxillary_points into a function for inserting the points and one for determining the mesh size. However, the mesh size determination is so interwoven with the insertion of the points in pp.fracs.tools.determine_mesh_size(..) that I was not brave enough to tackle this challenge.

I will however, add some comments about this

rbe051: Also do you have a better name for it? Could we maybe call them:
_set_mesh_size_with_auxiliary_points()
and
_set_mesh_size_without_auxilary_points() ?

src/porepy/fracs/fractures_2d.py
```diff
@@ -531,6 +533,31 @@ def _insert_auxiliary_points(
...
+        self,
+        mesh_size_frac: Optional[float]=None,
+        mesh_size_bound: Optional[float]=None,
+        mesh_size_min: Optional[float]=None,
```
rbe051, line 541: Is mesh_size_min outdated? I can not find that it is ever used (i.e., in pp.fracs.tools.determine_mesh_size(..))

src/porepy/fracs/fractures_2d.py
rbe051: I decided to change _set_dfn_mesh_size -> set_mesh_size_without_auxiliary_points().  I will wait for @keileg 's feedback before renaming  _insert_auxiliary_points

src/porepy/fracs/fractures_2d.py
keileg: It is used, L48 in pp.fracs.tools.

rbe051: It is assigned to val_min, but val_min is not used as far as I can tell

src/porepy/fracs/fractures_2d.py
keileg: Let's keep it like it is now. The idea is that the meshing should do two things:
1) Split all intersections, so that gmsh does not complain
2) Attempt to set reasonable mesh sizes with a minimal effort from the user. 
If one wants something more, the best option is to interact with gmsh directly, and I don't think it is a good idea to expose too many options on the PorePy side - that will soon be complex to maintain.

The dfn case you are working on deserves some special treatment, and if we ever get better way to interact with the gmsh python api, we may expand the new option to non-dfn cases as well, but until we see more of a need, I'm happy with the current structure.

keileg:
@rbe051: Can we merge this one now? I think I'm satisfied for the time being.

rbe051:
@keileg yes lets merge it. I completely agree with your comments. Hopefully nobody has to see these functions anyway.

I will merge this now, but maybe @alessiofumagalli wants to have a quick look since he also has been working a bit with DFN's

PR #537, opened 2021-04-23
Enhancement of flux-based upwind and addition of arithmetic average

jhabriel:


keileg (approved):
I think this is okay for now. To me, the whole construction of Ad discretizations are 'under construction', and will do with some cleanups in due course, but for now, this looks good.

jwboth (approved):
Looks OK to me.

PR #541, opened 2021-04-30
Freeze meshio version: >=4.3.5, <4.4.0

jhabriel:
Temporal workaround to pass vtk unit tests. Due to a new release of meshio (4.4.1), some tests failed during Github Actions, making (unrelated) new PRs to also fail the checks.

This requirement freezes the version to be strictly lower than 4.4.0, where there are no apparent problems.


keileg (approved):
Very good.

PR #545, opened 2021-05-31
Tensor frac

rbe051:
Same as pp.meshing.cart_grid(...) but with a pp.TensorGrid instead


keileg:
@rbe051 : From the first glance this looks good to me - it is useful, and different from the cart_grid option. Is there anything in particular I should have a look at?

rbe051:
No, nothing in particular. This was a straight forward extension that is quite handy if you want local refinement around wells and fractures, but still use a Cartesian grid.

If you are happy you can merge

PR #546, opened 2021-06-15
Make it possible to export cells with different number of faces/nodes

rbe051:
If you have a mix of cell types (e.g., tetrahedrons and cubes) the entries in the faces_loc will have a different sizes.
numpy tires to vstack the list. Instead we hstack the list before calling unique.

### DIFF
--- a/src/porepy/viz/exporter.py
+++ b/src/porepy/viz/exporter.py
@@ -755,7 +755,7 @@ class Exporter:
                     f_counter += 1
 
                 # collect all the nodes for the cell
-                nodes_loc = np.unique(faces_loc).astype(int)
+                nodes_loc = np.unique(np.hstack(faces_loc)).astype(int)
 
                 # define the type of cell we are currently saving
                 cell_type = "polyhedron" + str(nodes_loc.size)


keileg:
This has been on our todo list for a while. There may be a problem with polygons, which are currently (meshio < v4.4) tagged with their number of nodes - something similar will likely apply to polyhedra. For the moment, the supported meshio version is frozen, and I have postponed the PorePy upgrade necessary for now.

In this connection (note to self, mainly), it may also be appropriate to change the exporter interface so that, when the grid is known to be standard cells only (simplex or logical Cartesian), we use this format instead of going through a polygon / polyhedron. That will save a lot of time and disc space, in particular in 3d, and cover 99% of all grids that are exported.

My suggestion is that I take responsibility for this PR, see it in conjunction with the above changes, and adapt and merge when ready. Does that sound reasonable to you @rbe051?

rbe051:
Yes, please adopt this to your own changes as you find best. This PR only solved an issue when exporting general hexahedron grid (ugly cornerpoint grids)

PR #551, opened 2021-09-06
Feat/wells

IvarStefansson:
Main changes:

- Geometric representation of wells 
- Discretization for coupling wells to fractures

There may be need for improvements to this functionality. Will be added as demand arises.

Wells are represented as polylines and collected in a WellNetwork, analogously to the FractureNetworks. The wells are meshed and added to an existing grid bucket. Piecewise linear grids are constructed for each well between pair of "physical" points (corresponding to a boundary, internal tip or intersection with a fracture). Meshing is done independently for each grid.

The coupling discretization is a simple implementation of the Peaceman well model.

PR #552, opened 2021-09-15
Added functionality for extracting parameter arrays from edges dictionaries

jhabriel:
This new functionality is meant to work as `pp.ad.ParameterArray`, but instead of picking up values from subdomain dictionaries, now it is also possible to extract parameter arrays from edge dictionaries. The inputs are the parameter keyword, the array keyword, and then the edge list or the grid bucket.

### DIFF
--- a/src/porepy/numerics/ad/grid_operators.py
+++ b/src/porepy/numerics/ad/grid_operators.py
@@ -18,6 +18,7 @@ __all__ = [
     "Trace",
     "SubdomainProjections",
     "ParameterArray",
+    "EdgeParameterArray",
 ]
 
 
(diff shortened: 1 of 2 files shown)


IvarStefansson:
The `edges` parameter of `ParameterArray` was just added in #551. Does it cover this functionality?

jhabriel:
It certainly does.  I did not realize that the functionality was extended. And indeed, it makes much more sense to have only one class taking care of extracting the parameters of both nodes and edges. From what I noticed, there is a tendency to avoid passing the `GridBucket` and stick to passing the `grids_list`and `edges_list` instead. Is there any particular reason for this? Also, is there a way to catching an error if both `grids_list` and `edges_list` are passed at the same time?

IvarStefansson:
No, I don't think so. One could reasonably wish to use some parameter for both grids and edges.

jhabriel:
I agree. I did not contemplate that possibility. Very well then, I will close this PR. Thanks, Ivar.

PR #553, opened 2021-09-16
Readable error messages from Ad parsing

keileg:
This PR attempts to provide more informative messages when errors (typically dimension mismatches) are raised during parsing of Ad equations. The format of the error messages, including which information to provide, should be improved, but for that we need more experience.

Also, the PR contains some minor maintenance of the Ad framework.

### DIFF
--- a/src/porepy/numerics/ad/grid_operators.py
+++ b/src/porepy/numerics/ad/grid_operators.py
@@ -421,6 +421,7 @@ class Divergence(Operator):
         grids: Optional[List[pp.Grid]] = None,
         gb: Optional[pp.GridBucket] = None,
         dim: int = 1,
+        name: Optional[str] = None,
     ):
         """Construct divergence operators for a set of subdomains.
 
@@ -439,7 +440,7 @@ class Divergence(Operator):
             dim (int, optional): Dimension of vector field. Defaults to 1.
 
         """
-
+        super().__init__(name=name)
         self._g: List[pp.Grid] = _grid_list(grids, gb)
 
         self.dim: int = dim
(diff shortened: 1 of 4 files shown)


src/porepy/numerics/ad/grid_operators.py
```diff
@@ -294,32 +298,52 @@ def __init__(
...
         # Vertical stacking of the projections
-        self.primary_to_mortar_int = bmat([[m] for m in primary_to_mortar_int])
-        self.primary_to_mortar_avg = bmat([[m] for m in primary_to_mortar_avg])
-        self.secondary_to_mortar_int = bmat([[m] for m in secondary_to_mortar_int])
-        self.secondary_to_mortar_avg = bmat([[m] for m in secondary_to_mortar_avg])
+        self.primary_to_mortar_int = bmat(
+            [[m] for m in primary_to_mortar_int], name="PrimaryToMortarInt"
+        )
+        self.primary_to_mortar_avg = bmat(
+            [[m] for m in primary_to_mortar_avg], name="PrimarytoMortarAvg"
```
IvarStefansson, line 325:
```suggestion
            [[m] for m in primary_to_mortar_avg], name="PrimaryToMortarAvg"
```

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -412,6 +438,12 @@ def __repr__(self) -> str:
         )
         return s
 
+    def __str__(self) -> str:
```
IvarStefansson, line 441: We are introducing a lot of very similar __str__ methods. Would it be possible to do this by inheritance from Operator, using a method that converts the class type to string?

IvarStefansson (commented):
Looks good to me. This will be enormously useful!

src/porepy/numerics/ad/grid_operators.py
keileg: There are subtle, but important, nuances between the different string methods. For instance, the close relation to the class name applies to many, but far from all, string methods. I think it is better to spell it out for each class; at least as long as the whole Ad concept is under heavy development.

PR #554, opened 2021-09-16
BUG: Subtraction and addition of Ad arrays with numpy arrays did not commute

keileg:
Problem: For a `pp.Ad_array` a and a numpy array `b` (wrapped as a `pp.ad.Array`), `a+b` and `b+a `gave different results; specifically `b+a` was parsed wrongly. This PR is a minor fix to resolve the issue.


keileg:
Wrong base, will open new PR.

PR #566, opened 2021-09-29
Dep/ad expression

IvarStefansson:
This PR:
1. Removes the ad.Expression. Now, the operator tree that used to be converted to an expression is passed directly to an EquationManager. The parsing method Expression.to_ad(gb) is replaced by Operator.evaluate(dof_manager).
2. Unifies treatment of grids and edges in the ad package. Each Operator has any combination of the attribute lists grids and edges (including none of the two). For Operators which may sometimes have the attribute, an empty list is assigned if None is passed on initialization.
3. Significant refractoring of the ad module, including collecting utility functions in _ad_utils. This is henceforth the naming convention, i.e. utility modules in package x should be named _x_utils.

@keileg, I have left some comments tagged with "EK", please have a look. Also, I expect that at least one iteration is needed before the PR can be merged.


keileg (commented):
Looks good. I will push some more changes, mainly comments, then merge.

PR #568, opened 2021-10-02
DEP: Remove the k_inv_scaling option from RobinCoupling

IvarStefansson:
The PR removes the k-inv scaling option, which was deemed to cause more confusion than it was worth. 

Now, the Robin condition \lambda = - \kappa * (p_l - p_h) is discretized in the standard FV way by multiplying the pressure contributions by a matrix with diagonal entries -\kappa * mg.cell_volumes.

### DIFF
--- a/src/porepy/models/contact_mechanics_biot_model.py
+++ b/src/porepy/models/contact_mechanics_biot_model.py
@@ -780,19 +780,28 @@ class ContactMechanicsBiot(pp.ContactMechanics):
                 * accumulation_primary
                 + subdomain_proj_scalar.cell_prolongation(g_frac) * accumulation_fracs
             )
-
-            interface_flow_eq = robin_ad.mortar_scaling * (
-                mortar_proj_scalar.primary_to_mortar_avg
+            # Interface equation: \lambda = -\kappa (p_l - p_h)
+            # Robin_ad.mortar_discr represents -\kappa. The involved term is
+            # reconstruction of p_h on internal boundary, which has contributions
+            # from cell center pressure, external boundary and interface flux
+            # on internal boundaries (including those corresponding to "other"
+            # fractures).
+            interface_flow_eq = (
+                robin_ad.mortar_discr
+                * mortar_proj_scalar.primary_to_mortar_avg
                 * mpfa_ad.bound_pressure_cell
                 * p
-                + mortar_proj_scalar.primary_to_mortar_avg
+                + robin_ad.mortar_discr
+                * mortar_proj_scalar.primary_to_mortar_avg
                 * mpfa_ad.bound_pressure_face
-                * (
-                    mortar_proj_scalar.mortar_to_primary_int * mortar_flux
-                    + bc_val_scalar
-                )
-                - mortar_proj_scalar.secondary_to_mortar_avg * p
-                + robin_ad.mortar_discr * mortar_flux
+                * mortar_proj_scalar.mortar_to_primary_int
+                * mortar_flux
+                + robin_ad.mortar_discr
+                * mortar_proj_scalar.primary_to_mortar_avg
+                * mpfa_ad.bound_pressure_face
+                * bc_val_scalar
+                - robin_ad.mortar_discr * mortar_proj_scalar.secondary_to_mortar_avg * p
+                + mortar_flux
             )
 
             eq_manager.equations += [
--- a/src/porepy/models/contact_mechanics_biot_model.py
+++ b/src/porepy/models/contact_mechanics_biot_model.py
@@ -786,21 +786,20 @@ class ContactMechanicsBiot(pp.ContactMechanics):
             # from cell center pressure, external boundary and interface flux
             # on internal boundaries (including those corresponding to "other"
             # fractures).
-            interface_flow_eq = (
-                robin_ad.mortar_discr
-                * mortar_proj_scalar.primary_to_mortar_avg
-                * mpfa_ad.bound_pressure_cell
-                * p
-                + robin_ad.mortar_discr
-                * mortar_proj_scalar.primary_to_mortar_avg
-                * mpfa_ad.bound_pressure_face
+            p_primary = (
+                mpfa_ad.bound_pressure_cell * p
+                + mpfa_ad.bound_pressure_face
                 * mortar_proj_scalar.mortar_to_primary_int
                 * mortar_flux
-                + robin_ad.mortar_discr
-                * mortar_proj_scalar.primary_to_mortar_avg
-                * mpfa_ad.bound_pressure_face
-                * bc_val_scalar
-                - robin_ad.mortar_discr * mortar_proj_scalar.secondary_to_mortar_avg * p
+                + mpfa_ad.bound_pressure_face * bc_val_scalar
+            )
+            # Project the two pressures to the interface and equate with \lambda
+            interface_flow_eq = (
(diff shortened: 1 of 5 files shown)


src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -22,7 +22,10 @@ class RobinCoupling(
 ):
     """A condition with resistance to flow between subdomains. Implementation
     of the model studied (though not originally proposed) by Martin et
-    al 2005.
+    al 2005. The equation reads
+        \lambda = -\int{\kappa_n [p_l - p_h +  a/2 g \cdot n]} dV,
+    where the last gravity type term is zero by default (controlled by the vector_source
+    parameter).
```
keileg, line 28:
```suggestion
    parameter). That is, \lambda is an extensive quantity.
```

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -241,9 +212,10 @@ def assemble_matrix(
                     g[ambient_dimension]= -G * rho.
             matrix: original discretization
 
-            The discretization matrices must be included since they will be
-            changed by the imposition of Neumann boundary conditions on the
-            internal boundary in some numerical methods (Read: VEM, RT0)
+        The discretization matrices must be included since they will be changed
+        by the imposition of Neumann boundary conditions on the internal boundary
+        in some numerical methods (Read: VEM, RT0)
```
keileg, line 217:
```suggestion
        in some numerical methods (Read: VEM, RT0).
```

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -344,6 +321,7 @@ def _assemble(
         """Actual implementation of assembly. May skip matrix and rhs if specified."""
 
         matrix_dictionary_edge = data_edge[pp.DISCRETIZATION_MATRICES][self.keyword]
```
keileg, line 323:
```suggestion
        matrix_dictionary_edge: Dict[str, sps.spmatrix] = data_edge[pp.DISCRETIZATION_MATRICES][self.keyword]
```

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -429,15 +404,16 @@ def _assemble(
...
-            for block in range(cc.shape[1]):
-                # Scale the pressure blocks in the mortar problem
-                cc[2, block] = (
-                    matrix_dictionary_edge[self.mortar_scaling_matrix_key]
-                    * cc[2, block]
-                )
+            for block in range(3):
+                # Scale the pressure contributions in the mortar equation by
+                # normal diffusivity.
+                cc[2, block] = diffusivity_discr * cc[2, block]
```
keileg, line 413: diffusivity_discr is diagonal scaling with volume and normal diffusivity? If so, it should not hit `cc[2, 2]`?

keileg: .. I see below that you have not yet added it. Still, I would prefer if we let block go in `range(2)`, and add a comment that the mortar block should not be scaled.

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -525,23 +501,25 @@ def assemble_edge_coupling_via_high_dim(
             ):
                 proj_flux = mg_secondary.mortar_to_secondary_int()
             # Assemble contribution between higher dimensions.
+            # import pdb
+            # pdb.set_trace()
```
keileg, line 505:
```suggestion
```

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -525,23 +501,25 @@ def assemble_edge_coupling_via_high_dim(
...
             self.discr_primary.assemble_int_bound_pressure_trace_between_interfaces(
                 g, data_grid, proj_pressure, proj_flux, cc, matrix, rhs
             )
             # Scale the equations (this will modify from K^-1 to K scaling if relevant)
-            for block in range(cc.shape[1]):
+
+            for block in range(3):
                 # Scale the pressure blocks in the row of the primary mortar problem.
```
keileg, line 512: primary -> secondary?

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -920,7 +898,7 @@ def __init__(
 
         # Keys used to identify the discretization matrices of this discretization
         self.well_discr_matrix_key = "well_mortar_discr"
-        self.mortar_vector_source_matrix_key = "well_vector_source_discr"
+        self.well_vector_source_matrix_key = "well_vector_source_discr"
```
keileg, line 901:
```suggestion
        self.well_vector_source_matrix_key: str = "well_vector_source_discr"
```

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
```diff
@@ -920,7 +898,7 @@ def __init__(
 
         # Keys used to identify the discretization matrices of this discretization
         self.well_discr_matrix_key = "well_mortar_discr"
```
keileg, line 900:
```suggestion
        self.well_discr_matrix_key: str = "well_mortar_discr"
```

src/porepy/models/contact_mechanics_biot_model.py
```diff
@@ -780,19 +780,28 @@ def _assign_discretizations(self) -> None:
...
-                - mortar_proj_scalar.secondary_to_mortar_avg * p
-                + robin_ad.mortar_discr * mortar_flux
+                * mortar_proj_scalar.mortar_to_primary_int
+                * mortar_flux
+                + robin_ad.mortar_discr
+                * mortar_proj_scalar.primary_to_mortar_avg
+                * mpfa_ad.bound_pressure_face
+                * bc_val_scalar
+                - robin_ad.mortar_discr * mortar_proj_scalar.secondary_to_mortar_avg * p
+                + mortar_flux
```
keileg, line 804: Can we collect terms and use parentheses here?

keileg (commented):
Minor comments. Feel free to merge when incorporated.

src/porepy/models/contact_mechanics_biot_model.py
IvarStefansson: Probably. I'll see if ad allows it

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
IvarStefansson: If only. But block 2 has to be hit prior to addition of "own" mortar flux due to contributions from interface fluxes to trace of primary pressure.

src/porepy/numerics/interface_laws/elliptic_interface_laws.py
IvarStefansson: This is the row corresponding to the mortar variable to which the contributions from the flux of block 2 is being added. I think the comment (probably written by you) refers to the two fluxes as "primary" in block 1 and "secondary" in block 2. I'll elaborate in the comment.

PR #569, opened 2021-10-04
BUG: Error while trying to access the previous iteration of a MergedVariable

jhabriel:
Small bug while returning the previous iteration state of a `MergedVariable`.  The returned `Variable` was created with the parameter `grids=self.grids` for `self._is_edge_var = True`, and vice versa. This was causing a critical error while trying to create a `merged_variable.previous_iteration()`.


IvarStefansson (approved):
Thanks for catching this!

PR #572, opened 2021-10-14
Updates to dof_manager

keileg:
Main changes:
* Fixed several bugs and issues in `DofManager`. The method distribute_variable should now be used to set the variable state in `pp.STATE` and `pp.ITERATE`
* Changed names of some methods (`DofManager.dof_ind -`> `grid_and_variable_to_dofs()`)
* Added new methods for mapping between grid-variable combinations and degrees of freedom. Aimed at debugging
* Added tests directly for the `DofManager`, this was only indirectly tested previously.
* Updated tutorials to use some of the new `DofManager` functionality. 

One change that should be done is to convert the attributes `block_dof` and `full_dof` in the `DofManager` to private, and thereby force clients to use the new methods. This will requrie major changes throughout the code base, and should be done in as a separate project.

Also various other minor changes to the rest of the code.

### DIFF
--- a/src/porepy/numerics/ad/equation_manager.py
+++ b/src/porepy/numerics/ad/equation_manager.py
@@ -99,7 +99,8 @@ class EquationManager:
         secondary_grid_name = [
             (var._g, var._name) for var in self._variables_as_list(secondary_variables)
         ]
-        # Do the same for all variables, but make this a map to the atomic (non-merged) variable.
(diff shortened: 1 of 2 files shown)


keileg:
@IvarStefansson : Turns out typing.Literal is not supported in Python 3.7, which therefore fails. I will take care of, but I think you can do the code review independently of this. No rush though.

PR #591, opened 2021-11-11
Delay numba compilation of setmembership function

keileg:
No functional changes, just moved some code around. By delaying jit compilation until the function is used, the time for import can be substantially reduced (>50%). 

Also fixed a typo in documentation.


IvarStefansson (approved):
Thanks for the clear PR description!

PR #595, opened 2021-11-25
bug fix in the compute_well_rock_matrix_intersections: use the key as…

alessiofumagalli:
… cell index and not the ad node id, the bug was present when the cells input parameter was used


keileg (approved):
I trust that this is correct. Thanks for the fix. Will merge later today.

PR #598, opened 2021-11-29
Safer distribute_variable method

IvarStefansson:
While pythonic, the += version of additive STATE/ITERATE updating has proven to lead to errors (related to initialization of the values without using deep copies). This PR changes to the slightly uglier and less efficient but far safer d[pp.STATE] = d[pp.STATE] + vals.


keileg (approved):
Agree. This is much safer.

PR #599, opened 2021-12-02
Fetching external content

OmarDuran:
In regards to the file polyhedron.py, this implementation is intended to smooth the installation process. Downloading, renaming, and placing of polyhedron.py are performed in a single command via `python fetch_content.py`. This implementation is platform-independent.


IvarStefansson (commented):
Looks good to me. Could it be run automatically as part of the actual installation of PorePy? Is that desireable, @keileg?

keileg:
@OmarDuran I suppose I should wait while you consider alternative approaches as we discussed today?

OmarDuran:
> @OmarDuran I suppose I should wait while you consider alternative approaches as we discussed today?

We can include it. In the meantime, I will work on a new implementation for point_in_polyhedron and hopefully get rid of polyhedron.py

keileg (approved):

PR #609, opened 2022-01-26
Allow empty lists of lower-dimensional grids and interfaces in Ad framework. Resolves #607

### LINKED ISSUE #607: Ad framework should not break when called with an empty edge list
If `MergedVariables `(and possibly other Ad operators) are called with an empty list of edges, the code currently gives an error message. Thus it is necessary to handle the case of no interfaces (no fractures) with special code.

The same probably applies if an empty list of subdomains are passed, although that is a less used option.

keileg:
This PR resolves the issue of the Ad framework dealing with variables, and some other operators, defined on empty lists of grids or interfaces. This also led to several other corrections being needed throughout the Ad framework, mainly of a rather technical sort.

After this PR, the model classes ContactMechanics and ContactMechanicsBiot can be run in Ad modus without modifications (this used not to be the case). Tests are added to verify this.

Also some minor improvements and maintenance along the way.

### DIFF
--- a/src/porepy/models/contact_mechanics_biot_model.py
+++ b/src/porepy/models/contact_mechanics_biot_model.py
@@ -534,17 +534,21 @@ class ContactMechanicsBiot(pp.ContactMechanics):
 
             tangential_normal_proj_list = []
             normal_proj_list = []
-            for gf in g_frac:
-                proj = gb.node_props(gf, "tangential_normal_projection")
-                tangential_normal_proj_list.append(
-                    proj.project_tangential_normal(gf.num_cells)
+            if len(g_frac) > 0:
+                for gf in g_frac:
+                    proj = gb.node_props(gf, "tangential_normal_projection")
+                    tangential_normal_proj_list.append(
+                        proj.project_tangential_normal(gf.num_cells)
+                    )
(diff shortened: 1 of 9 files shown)


src/porepy/models/contact_mechanics_biot_model.py
```diff
@@ -774,7 +791,10 @@ def _assign_discretizations(self) -> None:
             )
 
             # Accumulation term on the fractures.
-            frac_vol = np.hstack([g.cell_volumes for g in g_frac])
+            if len(g_frac) > 0:
+                frac_vol = np.hstack([g.cell_volumes for g in g_frac])
+            else:
+                frac_vol = np.array([])
```
IvarStefansson, line 797: Could this be a MassMatrix?

IvarStefansson (commented):
Looks good!

IvarStefansson (approved):

src/porepy/models/contact_mechanics_biot_model.py
keileg: Yes, I think you are right. We should take of that when reworking that model, similar to what you do with pure ContactMechanics for the moment.

PR #614, opened 2022-02-02
Test incompressible flow

IvarStefansson:


OmarDuran:
Working alright!

src/porepy/models/incompressible_flow_model.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, params: Optional[Dict] = None) -> None:
         self.variable: str = "p"
         self.mortar_variable: str = "mortar_" + self.variable
         self.parameter_key: str = "flow"
+        self._use_ad = True
```
keileg, line 54: Is this included for consistency only? Just for curiosity.

test/unit/test_incompressible_flow.py
```diff
@@ -0,0 +1,141 @@
...
+class FlowModel(pp.IncompressibleFlow):
+    def create_grid(self):
+        if self.params["n_fracs"] == 1:
+            self.gb, self.box = pp.grid_buckets_2d.single_horizontal()
+        elif self.params["n_fracs"] == 2:
+            self.gb, self.box = pp.grid_buckets_2d.two_intersecting()
+        pp.contact_conditions.set_projections(self.gb)
+
+def test_incompressible_flow_model_no_modification():
+    """Test that the raw contact mechanics model with no modifications can be run with
```
keileg, line 42:
```suggestion
    """Test that the raw incompressible flow model with no modifications can be run with
```

keileg (approved):
Control question: Is this unit or integration test? Cf placement of similar tests for contact_mechanics etc. Those tests also cover functionality (in that respect, the present module is a unit test), but we should also limit the number of test files.

I have no strong opinions, and let you decide this. Merge when you feel ready.

IvarStefansson:
> Control question: Is this unit or integration test? Cf placement of similar tests for contact_mechanics etc. Those tests also cover functionality (in that respect, the present module is a unit test), but we should also limit the number of test files.
> 
> I have no strong opinions, and let you decide this. Merge when you feel ready.

Yes, I actually think of this as a unit test

src/porepy/models/incompressible_flow_model.py
IvarStefansson: What do you mean?

keileg: The module is Ad only, so no Assembler, I suppose? That is of course how it should be, and it probably is okay to have the attribute to be consistent with the older models.

IvarStefansson: Yes, I think it is needed for now for safe inheritance from AbstractModel

PR #616, opened 2022-02-08
Use optimized storage format for projection matrices

keileg:
# Summary
This PR takes care of two issues that were detected while using the Ad framework to solve single phase flow on the field case in the 3d flow benchmark (though using different grid parameters than in the official benchmark setup):

1. The memory consumption in the Ad representation of mortar projections was unexpectedly heavy on memory (several GB).
2. Construction of the same projections took much longer than expected (34s on a reasonably fast computer, for reference the grid has 52 grids, and the 3d grid ended up with ~450K cells).

# Diagnostics
The problem turned out to be the storage format used for the matrix representation of the projections: Depending on the direction of the projection, these can have many more columns than rows (or vice versa) and should be assigned CSC storage (or oposite). 

# Solution
I ended up making a method, `pp.matrix_operations.optimized_compressed_storage()`, that converts a sparse matrix to CSC or CSR, depending on the number of rows and columns. I then used this to force the storage format for all projection matrices in the `MortarGrid` class and the `MortarProjections` in Ad. These changes reduced the construction time to 4s (which is still unexpectedly high, but at least it is better), and essentially removed the memory bottleneck.

I also ensure that the projection matrices in pp.ad.SubdomainProjections use the most suitable storage formate.

# Extended usage?
The storage for sparse matrices is generally not optimized in PorePy, and there sure are cases where a different format would be preferrable. The main impact is however expected for matrices with few elements and substantial differences in the number of rows and columns, while for the moment, there is no reason to reformat matrices that do not fit with this description. As an illustration that these are non-trivial matters, see https://github.com/pmgbergen/porepy/blob/c6e4504f56c09565334de003a5c05e097e33bce9/src/porepy/numerics/ad/grid_operators.py#L346

### DIFF
--- a/src/porepy/numerics/ad/grid_operators.py
+++ b/src/porepy/numerics/ad/grid_operators.py
@@ -87,7 +87,7 @@ class SubdomainProjections(Operator):
             if len(grids) > 0:
                 # A key error will be raised if a grid in g is not known to
                 # self._cell_projection
-                # IMPLEMENTATION NOTE: Use csc format, since the number of rows can
+                # IMPLEMENTATION NOTE: Use csr format, since the number of rows can
                 # be much less than the number of columns.
                 mat = sps.bmat([[self._cell_projection[g].T] for g in grids]).tocsr()
             else:
@@ -153,7 +153,7 @@ class SubdomainProjections(Operator):
             if len(grids) > 0:
                 # A key error will be raised if a grid in grids is not known to
                 # self._face_projection
-                # IMPLEMENTATION NOTE: Use csc format, since the number of rows can
+                # IMPLEMENTATION NOTE: Use csr format, since the number of rows can
                 # be much less than the number of columns.
                 mat = sps.bmat([[self._face_projection[g].T] for g in grids]).tocsr()
             else:
--- a/src/porepy/numerics/ad/grid_operators.py
+++ b/src/porepy/numerics/ad/grid_operators.py
@@ -92,7 +92,6 @@ class SubdomainProjections(Operator):
                 # If the grid list is empty, we project from the full set of cells to
                 # nothing.
                 mat = sps.csr_matrix((0, self._tot_num_cells * self._nd))
-
             return pp.ad.Matrix(
                 mat,
                 name="CellRestriction",
@@ -123,7 +122,6 @@ class SubdomainProjections(Operator):
                 # If the grid list is empty, we project from nothing to the full set of
                 # cells
                 mat = sps.csr_matrix((self._tot_num_cells * self._nd, 0))
-
             return pp.ad.Matrix(
                 mat,
                 name="CellProlongation",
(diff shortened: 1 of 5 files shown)


src/porepy/numerics/ad/grid_operators.py
```diff
@@ -87,12 +87,13 @@ def cell_restriction(self, grids: List[pp.Grid]) -> Matrix:
             if len(grids) > 0:
                 # A key error will be raised if a grid in g is not known to
                 # self._cell_projection
+                # IMPLEMENTATION NOTE: Use csc format, since the number of rows can
```
IvarStefansson, line 90: This comment and those below: The note has been copy-pasted and is not consistent with the code in the cases where you have used .tocsr()

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -87,12 +87,13 @@ def cell_restriction(self, grids: List[pp.Grid]) -> Matrix:
...
+                # be much less than the number of columns.
                 mat = sps.bmat([[self._cell_projection[g].T] for g in grids]).tocsr()
```
IvarStefansson, line 92: Why not use the new function here?

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -839,10 +869,12 @@ def _subgrid_projections(
         )
         cell_sz = g.num_cells * nd
 
+        # Create matrix and convert to csc format, since the number of rows is (much)
+        # higher than the number of columns.
         cell_projection[g] = sps.coo_matrix(
             (np.ones(cell_sz), (cell_ind, np.arange(cell_sz))),
             shape=(tot_num_cells, cell_sz),
-        ).tocsr()
+        ).tocsc()
```
IvarStefansson, line 877: Again, wouldn't it make sense to use the new function?

tests/unit/test_grid_refinement.py
```diff
@@ -915,56 +915,12 @@ def test_mortar_grid_2d(self):
...
-                    4,
-                    4,
-                    4,
-                    5,
-                    6,
-                    7,
-                    8,
-                ]
-            )
+            indptr_known = np.array([0, 1, 2, 3, 4, 5, 6, 7, 8])
```
IvarStefansson, line 923: Am I right that the size difference illustrates the use of this PRs fix?

src/porepy/numerics/ad/grid_operators.py
keileg: Because explicit is better than implicit. In this case the shape of the matrix is clear, it is a single matrix (no product, thus clean code) and the format can be set directly.  For the cases further up in the file, the size of the matrix can be inferred, but it requires reasoning, and calling the function overall gives the best readability. In the mortar grid class, the size cannot be inferred in the case of non-matching grids, and the function is by far the better option.

src/porepy/numerics/ad/grid_operators.py
keileg: This is by purpose, I wrote explanation below.

IvarStefansson: I don't get it

keileg: Sorry, my response was completely wrong. Fixed it.

IvarStefansson (approved):

PR #617, opened 2022-02-09
BUG: Ensure LinearSolver for Models store solution.

keileg:
Depending on how the Model's after_newton_iteration() and
after_newton_convergence() methods are implemented, it may be necessary
to call both (in that order) for the computed solution to be stored in
pp.STATE.

Specifically, the ContactMechanics model need this fix to work as
intended.


IvarStefansson (approved):
Great that you documented this somewhat awkward fix properly!

PR #618, opened 2022-02-09
Update match_grids.py

enricoballini:


keileg:
@enricoballini thanks a lot. I'll try to process this on Friday, but it may stretch into next week.

@IvarStefansson I'll take main responsibility for this one. I'll see if I can do some maintenance on the module at the same time, from a brief glance yesterday, that is sorely needed.

keileg:
@enricoballini: The tests are failing, something's not right with the mortar updates. I'm not sure whether the problem is the tests or your fix, but it may very well be the former. I'll look into it and get back to you.

Since I will most likely do some more fixes here, can you please allow maintainers to push changes to your fork, if you have not done so, see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

keileg:
Short update: Upon further debugging and testing, it seems this could be a problem that applies to the corresponding 1d grid as well. I will pursue this further when I have the time, but it will not be this week.

keileg:
@enricoballini I ended up doing significant maintenance based on your initial suggestion. Adding this to a new PR was simpler than including it in your branch, so your changes are now included in #637. So, even though I close this one, your commits will make it into the code base.

PR #619, opened 2022-02-10
Fixed missing type transformation in subtraction operator overload of ad.Operator superclass

vlipovac:
The subtraction of np.ndarrays and ad.Operator child instances should work now without having to transform them explicitely to ad.Array.

The changed line is now same as in all other operator overloads. Seems like it was simply overlooked earlier.


keileg (approved):
Nice catch.

Issue #565, opened 2021-09-29
Exporter does not export edge variables


jhabriel:
The following MWE demonstrates the scenario where the variable `var` is exported successfully for the subdomains grids. However (even though is correctly defined), `var` is not exported in the `vtu` file corresponding to the mortar grid.

Note that `pp.Exporter` does not raise any errors, but rather the field is missing when inspected the `vtu` file.

```python
import porepy as pp
import numpy as np

# Create a grid bucket
fracs = np.array([[0.5, 0.5], [0, 1]])
nx = np.array([2, 1])
physdims = np.array([1, 1])
gb = pp.meshing.cart_grid([fracs], nx=nx, physdims=physdims)

# Set states
var_name = "var"
for g, d in gb:
    pp.set_state(d, state={var_name: g.dim * np.ones(g.num_cells)})
for e, d in gb.edges():
    mg = d["mortar_grid"]
    pp.set_state(d, state={var_name: np.zeros(mg.num_cells)})

#%% Export
save = pp.Exporter(gb, file_name="out")
save.write_vtu([var_name])
```

PR #620, opened 2022-02-15
Fix export variables on edges

jwboth:
Fix handles export of variables on edges.

Fix to issue #565

### DIFF
--- a/src/porepy/viz/exporter.py
+++ b/src/porepy/viz/exporter.py
@@ -21,29 +21,23 @@ import porepy as pp
 # Module-wide logger
 logger = logging.getLogger(__name__)
 
-module_sections = ["visualization"]
-# ------------------------------------------------------------------------------#
-
 
 class Field:
     """
     Internal class to store information for the data to export.
     """
 
-    @pp.time_logger(sections=module_sections)
     def __init__(self, name: str, values: Optional[np.ndarray] = None) -> None:
         # name of the field
         self.name = name
         self.values = values
 
-    @pp.time_logger(sections=module_sections)
     def __repr__(self) -> str:
         """
         Repr function
         """
         return self.name + " - values: " + str(self.values)
 
-    @pp.time_logger(sections=module_sections)
     def check(self, values: Optional[np.ndarray], g: pp.Grid) -> None:
         """
         Consistency checks making sure the field self.name is filled and has
@@ -56,13 +50,9 @@ class Field:
         if np.atleast_2d(values).shape[1] != g.num_cells:
             raise ValueError("Field " + str(self.name) + " has wrong dimension.")
 
-    @pp.time_logger(sections=module_sections)
     def _check_values(self) -> None:
         if self.values is None:
             raise ValueError("Field " + str(self.name) + " values not valid")
-
-
-# ------------------------------------------------------------------------------#
 
 
 class Fields:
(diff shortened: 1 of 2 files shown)


src/porepy/viz/exporter.py
```diff
@@ -366,21 +366,37 @@ def _export_g(self, data: Dict[str, np.ndarray], time_step):
...
-        if len(data) > 0:
-            fields.extend([Field(d) for d in data])
+        # Extract data which is contained in nodes (and not edges).
+        # To make the unit test 'test_vtk' successfull, keep the sorting from data.
+        node_data = list()
+        for key in data:
+            for _, d in self.gb.nodes():
+                if pp.STATE in d and key in d[pp.STATE]:
+                    node_data.append(key)
+                    continue
```
keileg, line 389: The rational here is that you have found this key, added it to `node_data`, and don't want to add it once more? If so, should it not be `break `rather than `continue`? 

Anyhow, please document the functionality, this is the future me may wonder about.

src/porepy/viz/exporter.py
```diff
@@ -408,15 +424,30 @@ def _export_gb(self, data: List[str], time_step: float):
                 field.values = np.hstack(values)
 
             if self.meshio_geom[dim] is not None:
-                self._write(fields, file_name, self.meshio_geom[dim])
+                self._write(node_fields, file_name, self.meshio_geom[dim])
 
         self.gb.remove_node_props(extra_node_names)
 
+        # Extract data which is contained in nodes (and not edges).
+        edge_data = list()
```
keileg, line 432:
```suggestion
        # IMPLEMENTATION NOTE: See the above loop to construct node_data for an explanation of this
        # elaborate construction of `edge_data`
        edge_data = list()
```

keileg (commented):
Mainly looks good to me, I have left a few suggestions and one request for better documentation.

Parts of the code is somewhat hard to read, with variable names that makes little sense etc., but fixing that would require a more thorough refactoring of the module, and that is not for this PR.

keileg:
@jwboth: I took the liberty to remove all logging statements from the module and commit to this branch (as a general comment, feel free to delete those upon sight). I also deleted some more unnecessary breaks between the function. Hopefully I have not broken the code (I'm working via the editor in GitHub so no chance to do local testing), but I have may have introduced changes that Black or flake won't like. Could you take care of the cleanup, provided it is only minor things? If I have messed up, please let me know.

Also note that if you accept my proposed changes from the code review, you risk that these introduce longer lines than what flake appreciates - we have found no good way of making sure we stay within those bounds.

src/porepy/viz/exporter.py
```diff
@@ -408,15 +390,30 @@ def _export_gb(self, data: List[str], time_step: float):
                 field.values = np.hstack(values)
 
             if self.meshio_geom[dim] is not None:
-                self._write(fields, file_name, self.meshio_geom[dim])
+                self._write(node_fields, file_name, self.meshio_geom[dim])
 
         self.gb.remove_node_props(extra_node_names)
 
+        # Extract data which is contained in nodes (and not edges).
```
IvarStefansson, line 397:
```suggestion
        # Extract data which is contained in edges (and not nodes).
```

IvarStefansson (commented):
Nothing to add beyond the one comment and that I agree with Eirik, this part of the code could do with a general overhaul at a later time.

keileg (approved):
Thanks.

PR #622, opened 2022-02-21
MAINT: Refactor SlightlyCompressibleFlow

OmarDuran:
This PR is intended to refactor SlightlyCompressibleFlow.

- The implementation matches a semianalytical 1D solution (see figure below). The numerical setting was 2D. 
- A subsequent PR will be issued for testing the model.

@keileg and @IvarStefansson you were automatically added as reviewers. 

Screen Shot 2022-02-22 at 12 37 50 AM


src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -68,15 +67,14 @@ def _set_parameters(self) -> None:
...
-
             pp.initialize_data(
                 g,
                 d,
                 self.parameter_key,
                 {
-                    "mass_weight": mass_weight
-                },
+                    "mass_weight": self._compressibility(g),
+                    'time_step': self.time_step
```
IvarStefansson, line 76: I don't think the time step is needed

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -109,63 +95,27 @@ def _assign_discretizations(self) -> None:
...
-        p = self._eq_manager.merge_variables([(g, self.variable) for g in grid_list])
-
-        # Update equations
-        subdomain_flow_eq = self._eq_manager.equations["subdomain_flow"]
-        interface_flow_eq = self._eq_manager.equations["interface_flow"]
-
-        mass = (
-            mass_mat * (p - p.previous_timestep()) / self.time_step
+        
+        if not hasattr(self, "dof_manager"):
```
vlipovac, line 99: I think this is not necessary, since the super call to the parent class method asserts this attribute is available

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -109,63 +95,27 @@ def _assign_discretizations(self) -> None:
...
+        # Collection of subdomains
+        subdomains: List[pp.Grid] = [g for g, _ in self.gb]
+        
+        # AD representation of the mass operator
+        accumulation_term = pp.ad.MassMatrixAd(self.parameter_key, subdomains)
+        
+        # Need to find a way / refactor to access the superclass "and pressure" instead of create a new one
+        # If for some reason the mesh ordering change, this will be inconsistent
+        p = self._eq_manager.merge_variables([(g, self.variable) for g in subdomains])
+        time_step_ad = pp.ad.Scalar(self.time_step, "time step")
```
vlipovac, line 111: Basic operations with floats and ints should work in combination with pp.ad.Operator (see Operator._parse_other)
No need for type conversion

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -109,63 +95,27 @@ def _assign_discretizations(self) -> None:
...
+            self.dof_manager = pp.DofManager(self.gb)
+        
+        # Collection of subdomains
+        subdomains: List[pp.Grid] = [g for g, _ in self.gb]
+        
+        # AD representation of the mass operator
+        accumulation_term = pp.ad.MassMatrixAd(self.parameter_key, subdomains)
+        
+        # Need to find a way / refactor to access the superclass "and pressure" instead of create a new one
+        # If for some reason the mesh ordering change, this will be inconsistent
```
vlipovac, line 109: @keileg your solution in branch [multiphase_flow_reactive_base](https://github.com/pmgbergen/porepy/blob/multiphase_reactive_base/src/porepy/models/multiphase_reactive_transport.py) with the AdVariables dataclass might help. Can we generalize that class and include it in the abstract model ( @IvarStefansson ), so that every model has access to the MergedVariable instances (global variables) defined on the whole domain?

vlipovac (commented):
See comments of mine

src/porepy/models/slightly_compressible_flow_model.py
vlipovac: The MergedVariable instances which are defined in e.g. a parent class method

src/porepy/models/slightly_compressible_flow_model.py
IvarStefansson: There actually is a reason for this: We want to be able to change the time step through updating time_step_ad._value

src/porepy/models/slightly_compressible_flow_model.py
IvarStefansson: Something along that line is indeed the plan

src/porepy/models/slightly_compressible_flow_model.py
OmarDuran: Sure this is not necessary.

src/porepy/models/slightly_compressible_flow_model.py
OmarDuran: Sure.

src/porepy/models/slightly_compressible_flow_model.py
OmarDuran: Instead of creating a dual storage class, we can use a dictionary to store AD variables. The key is the already defined string representation for the variable e.g. "self.variable". I have updated the branch with this idea.

keileg: @vlipovac as @IvarStefansson  said, we'll get there, but the concept should be clearer before we promote it to the base class.

@OmarDuran I vote for a class in this case, since that makes documentation of attributes more natural. Simpler syntax is also a plus.

OmarDuran: Alright, I have created a dual storage class for AD variables.

vlipovac (approved):

PR #626, opened 2022-02-24
Slightly compressible flow

OmarDuran:
This PR introduces the Slightly compressible flow, develop by @vlipovac with some minor changes. This model is an extension of the incompressible flow model. It only supports AD calculations.


src/porepy/models/incompressible_flow_model.py
```diff
@@ -13,6 +15,11 @@
 logger = logging.getLogger(__name__)
 
 
+class AdVariables:
```
keileg, line 18: Do you think we should mark this as a private class, so `_AdVariables`?

src/porepy/models/incompressible_flow_model.py
```diff
@@ -224,6 +237,23 @@ def _assign_variables(self) -> None:
                     self.mortar_variable: {"cells": 1},
                 }
 
+    def _create_dof_and_eq_manager(self) -> None:
+        self.dof_manager = pp.DofManager(self.gb)
+        self._eq_manager = pp.ad.EquationManager(self.gb, self.dof_manager)
+
+    def _create_assembler(self) -> None:
```
keileg, line 244: If we only use Ad, we don't need the `Assembler`.

src/porepy/models/incompressible_flow_model.py
```diff
@@ -224,6 +237,23 @@ def _assign_variables(self) -> None:
...
+        self.assembler = pp.Assembler(self.gb, self.dof_manager)
+
+    def _create_ad_variables(self) -> None:
+        grid_list = [g for g, _ in self.gb.nodes()]
+        edge_list = [e for e, d in self.gb.edges() if d["mortar_grid"].codim < 2]
```
keileg, line 249:
```suggestion
        # Make a list of interfaces, but only for couplings one dimension apart (e.g. not for
        # couplings that involve wells)
        edge_list = [e for e, d in self.gb.edges() if d["mortar_grid"].codim < 2]
```

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -0,0 +1,101 @@
...
+    def _assign_discretizations(self) -> None:
+        """Define equations through discretizations.
+
+        Uses the Mpfa discretization of the parent class for the elliptic part
+        and same boundary conditions coupling on internal boundaries.
+
+        Implements an Explicit Euler discretization in time.
+        """
+
+        super()._assign_discretizations()
```
keileg, line 84: You sure this is explicit Euler? The call to `super()._assign_discretizations()` will give an implicit Euler, I believe.

vldebug.py
```diff
@@ -0,0 +1,2 @@
+import porepy as pp
```
keileg, line 1: Do we need to add this file?

src/porepy/params/tensor.py
```diff
@@ -15,7 +15,8 @@
 
 
 class SecondOrderTensor(object):
-    """Cell-wise permeability represented by (3 ,3 ,Nc)-matrix.
+    """Cell-wise permeability represented by (3 ,3 ,Nc)-matrix, where Nc
```
keileg, line 18:
```suggestion
    """Cell-wise permeability represented by (3, 3, Nc)-array, where Nc
```

keileg (commented):
Minor comments only. Nice work.

src/porepy/models/slightly_compressible_flow_model.py
OmarDuran: Yes indeed, It is implicit.

vldebug.py
OmarDuran: Not at all.

src/porepy/models/incompressible_flow_model.py
OmarDuran: Incompressible_flow_model.py supports the case without AD. For instance, see `after_newton_convergence` therein.

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -0,0 +1,101 @@
...
+    Attributes:
+        end_time (float): Upper limit of considered time interval
+        time_step (float): time step size
+        time_index (int): number of time loops passed.
+        time (float): simulation time
+    """
+
+    def __init__(self, params: Optional[Dict] = None) -> None:
+        """
+        Parameters:
```
IvarStefansson, line 39: The only parameter is params. Please update.

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -0,0 +1,101 @@
...
+        Starting time set to 0, end time and timestep size optional paramters
+        2. _set_parameters: compressibility added
+
+    New methods:
+        1. _compressibility: constant compressibility per cell
+
+    Attributes:
+        end_time (float): Upper limit of considered time interval
+        time_step (float): time step size
+        time_index (int): number of time loops passed.
```
IvarStefansson, line 33:
```suggestion
        time_index (int): number of time steps passed.
```

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def _compressibility(self, g: pp.Grid) -> np.ndarray:
+        """Unitary compressibility.
+
+        Units: Pa^(-1)
+        """
+        return np.ones(g.num_cells)
+
+    def _assign_discretizations(self) -> None:
+        """Define equations through discretizations.
```
IvarStefansson, line 76: This description is more fitting for non-ad equation definition. Now, they are written explicitly in the ad framework

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -0,0 +1,101 @@
...
+            end_time (optional float): specifices end time for simulation.
+                Simulation stops when first time value bigger than end_time is
+                reached.
+            time_step (optional float ): Timestep size. Currently only
+                uniform timestepping is supported.
+        """
+        super().__init__(params)
+
+        # attributes
+        self.end_time = float(0)
```
IvarStefansson, line 49:
```suggestion
        self.end_time = self.params.get("end_time", float(1))
```

IvarStefansson (commented):
Good work!!

src/porepy/models/incompressible_flow_model.py
```diff
@@ -224,6 +237,25 @@ def _assign_variables(self) -> None:
                     self.mortar_variable: {"cells": 1},
                 }
 
+    def _create_dof_and_eq_manager(self) -> None:
```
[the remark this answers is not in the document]
OmarDuran: I have tried and the non-AD case is not supported, my bad. I thought non-AD was supported since in the `after_newton_convergence` there existed an instance of the assembler object being used. Now any instance of the assembler object has been removed. The methods are now documented.

src/porepy/models/slightly_compressible_flow_model.py
OmarDuran: Agree, it is done.

src/porepy/models/slightly_compressible_flow_model.py
OmarDuran: Done.

src/porepy/models/slightly_compressible_flow_model.py
OmarDuran: Agree, it is done.

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -0,0 +1,101 @@
...
+        Uses the Mpfa discretization of the parent class for the elliptic part
+        and same boundary conditions coupling on internal boundaries.
+
+        Implements an Implicit Euler discretization in time.
+        """
+
+        super()._assign_discretizations()
+
+        # Collection of subdomains
+        subdomains: List[pp.Grid] = [g for g, _ in self.gb]
```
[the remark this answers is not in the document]
OmarDuran: It seems that the object `subdomains` is about 88 bytes and building time is just a few milliseconds. So, I believe objects with these characteristics should be computed on-demand. Also, it does not fit with the AD storage class. 

Do you think there is a situation where building `subdomains` can spend several seconds?

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class SlightlyCompressibleFlow(pp.models.incompressible_flow_model.IncompressibleFlow):
+    """This class extends the Incompressible flow model by including a
+    cummulative term expressed through pressure and a constant compressibility
+    coefficient. For a full documenation refer to the parent class.
+
+    The simulation starts at time t=0.
+
+    Overwritten methods include:
```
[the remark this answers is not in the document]
OmarDuran: Agree, it is updated.

src/porepy/models/slightly_compressible_flow_model.py
```diff
@@ -0,0 +1,101 @@
...
+
+        # Collection of subdomains
+        subdomains: List[pp.Grid] = [g for g, _ in self.gb]
+
+        # AD representation of the mass operator
+        accumulation_term = pp.ad.MassMatrixAd(self.parameter_key, subdomains)
+
+        # Access to pressure ad variable
+        p = self._ad.pressure
+        time_step_ad = pp.ad.Scalar(self.time_step, "time step")
```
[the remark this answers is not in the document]
OmarDuran: Agree, it is done.

src/porepy/models/incompressible_flow_model.py
```diff
@@ -13,6 +15,11 @@
 logger = logging.getLogger(__name__)
 
 
+class _AdVariables:
```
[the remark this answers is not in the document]
OmarDuran: Agree, but out of the scope for this PR.

src/porepy/models/slightly_compressible_flow_model.py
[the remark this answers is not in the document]
IvarStefansson: No, you're probably right. But I think that in principle, it fits with the AD storage class (as I envision it).

src/porepy/models/incompressible_flow_model.py
OmarDuran: My bad non-AD is not supported on this model. The assembler object was deleted and code was adapted accordingly.

IvarStefansson (approved):

PR #628, opened 2022-02-28
Fix two typos in sort_points

jwboth:
Fix two small typos.

### DIFF
--- a/src/porepy/utils/sort_points.py
+++ b/src/porepy/utils/sort_points.py
@@ -12,7 +12,6 @@ module_sections = ["utils"]
 def sort_point_pairs(
     lines: np.ndarray,
     check_circular: Optional[bool] = True,
-    ordering: Optional[bool] = False,
     is_circular: Optional[bool] = True,
 ) -> Union[Tuple[np.ndarray, np.ndarray], Tuple[np.ndarray, np.ndarray, np.ndarray]]:
     """Sort pairs of numbers to form a chain.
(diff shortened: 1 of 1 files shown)


src/porepy/utils/sort_points.py
```diff
@@ -22,21 +21,21 @@ def sort_point_pairs(
...
-    surely be imporved.
+    surely be improved.
 
     Parameters:
     lines: np.ndarray, 2xn, the line pairs. If lines has more than 2 rows, we assume
         that the points are stored in the first two rows.
     check_circular: Verify that the sorted polyline form a circle.
-                    Defaluts to true.
+                    Defaults to true.
     ordering: np.array, return in the original order if a line is flipped or not
```
IvarStefansson, line 31: The description of "ordering" is not intuitive to me. Could you improve it if you understand it?

IvarStefansson (approved):

keileg (approved):
Thanks. As indicated, I believe @IvarStefansson's question about the ordering parameter can be solved by deletion of the relevant code. Could you please have a quick look at this before merging?

src/porepy/utils/sort_points.py
jwboth: The parameter `ordering` has still an effect. If set `True`, a boolean array is returned specifying which of the columns in `lines` had to be flipped in order to construct a chain. If we want to keep this option, I suggest changing the name to something like `return_flip_info` as well as changing the corresponding boolean array `is_ordered` to something like `is_flipped`. On the other hand, this option is actually nowhere used in PP core. 

Any opinion on whether to keep this option? If we want to remove it, we can also remove anything related to `is_ordered`.

keileg: I think I know where it is used (in an external package of mine). Let's kill it, then I will deal with the consequences.

PR #641, opened 2022-04-19
in the case of wells do not consider their contribution in the comput…

alessiofumagalli:
…ation of the interpolated P0 velocity

### DIFF
--- a/src/porepy/numerics/vem/dual_elliptic.py
+++ b/src/porepy/numerics/vem/dual_elliptic.py
@@ -45,6 +45,8 @@ def project_flux(gb, discr, flux, P0_flux, mortar_key="mortar_solution"):
 
             for _, d_e in gb.edges_of_node(g):
                 g_m = d_e["mortar_grid"]
+                # do not consider the contribution from the mortar grid when the latter is
+                # of the same dimension or codim > 1 (e.g., wells)
                 if g_m.dim == g.dim or g_m.codim > 1:
                     continue
                 # project the mortar variable back to the higher dimensional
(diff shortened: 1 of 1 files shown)


src/porepy/numerics/vem/dual_elliptic.py
```diff
@@ -43,9 +43,10 @@ def project_flux(gb, discr, flux, P0_flux, mortar_key="mortar_solution"):
             _, indices = np.unique(g.cell_faces.indices, return_index=True)
             sign = sps.diags(g.cell_faces.data[indices], 0)
 
-            for _, d_e in gb.edges_of_node(g):
+            for e, d_e in gb.edges_of_node(g):
                 g_m = d_e["mortar_grid"]
-                if g_m.dim == g.dim:
+                g1, _ = gb.nodes_of_edge(e)
+                if g_m.dim == g.dim or g1.well_num > -1:
```
IvarStefansson, line 49: Should this only catch the cases when the lower-dimensional grid is a well? If not, would a check of the codimension be more appropriate?

alessiofumagalli: so far my problem was with a well (second condition), but yes checking on the codimension includes this case. I'll make the change and propose it

keileg:
Hi @alessiofumagalli: I have no idea which issue this PR resolves. Can you please provide background information here (this should really be a part of any PR), and document the code so that it is clear what happens?

alessiofumagalli:
> Hi @alessiofumagalli: I have no idea which issue this PR resolves. Can you please provide background information here (this should really be a part of any PR), and document the code so that it is clear what happens?

You are right, @keileg . In the case with only fractures, the reconstruction of the velocity as P0 for each element is influenced by the mortar variable (flux) if a cell is connected to a 1-codimensional object. In the case of well, this does not apply and the contribution should be skip. In the general case for object of codimension >1 this contribution should not be considered. If you need any more details do not hesitate to ask.

keileg:
Okay, that makes sense. 

BTW, is this part of the code covered by tests? The reason for asking is mainly #643, but more generally, since this code is not on the radar of any of the core developers, it is a weak point.

keileg (approved):

PR #644, opened 2022-04-20
BUG: Fix uncovered case in division with Ad operators.

keileg:
## Proposed changes

Fix an issue in evaluation of Ad Operator trees, when the parsing ends up with dividing a numpy array by an Ad_array (this is not possible, but the fix introduces workarounds for what seemed the most relevant cases).

## Types of changes

## Checklist

- [] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation 
- [X] I have checked that I have not duplicated existing functionality

## Further comments
There may be other closely related cases not covered by the current fix. These are not included in the current PR, but are indicated in the comments of the new implementation. Moreover, the new implementation should raise error messages if such uncovered cases are encountered (previously, such cases were silently ignored).

### DIFF
--- a/src/porepy/numerics/ad/operators.py
+++ b/src/porepy/numerics/ad/operators.py
@@ -422,7 +422,37 @@ class Operator:
             return results[0].apply(*results[1:])
 
         elif tree.op == Operation.div:
-            return results[0] / results[1]
+            # Some care is needed here, to account for cases where item in the results
+            # array is a numpy array
+            if isinstance(results[0], pp.ad.Ad_array):
+                # If the first item is an Ad array, the implementation of the forward
+                # mode should take care of everything.
+                return results[0] / results[1]
+            elif isinstance(results[0], np.ndarray):
+                # The first array is a numpy array, and numpy's implementation of
+                # division will be invoked.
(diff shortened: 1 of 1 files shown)


IvarStefansson:
I cannot see that there are any new tests. This might be fine, but I would prefer for the box in the PR description to be unchecked

keileg:
Fair point, unticked. 

This part of the code is not explicitly tested. That's a big fixme, what we will need is a systematic check of all combinations of different operators, operations (+,-,etc.), and combinations with numpy arrays, scalars (floats, ints, etc.). This should make the code more robust, and also put us in a position to remove the spurious errors that we sometimes get in evaluation of Ad operators (the current PR is an example). However, this will be a major undertaking that I have never had the time and energy to start working on, and I'm afraid it will take some more time, but we need to get there.

IvarStefansson:
I think setting up the tests might not be too much work if done cleverly. But fixing the bugs we might identify could require substantial work, yes. Should we make this an issue?

IvarStefansson (approved):

keileg:
Created #645. Let's add to the list if we see more needs.

PR #647, opened 2022-04-22
MAINT:

vlipovac:
- Removal of artifact AD-operations 'localeval' and 'apply'.
- Refactoring of error parsing. Added paranthesis to error message for binary operations
- Refactoring of internal function 'get_shape'
- Minor spelling and typing fixes

# Additional comments

- Probably a computationally insignificant change, but a cleaner implementation is to define 'get_shape' and 'error_message' outside of the scope of '_parse_operator'.
- The parsing of the error message contained inconsistent calls to _parse_readable ( sometimes with arguments, sometimes without).
- The parsing itself had a bug since 'operator_str' was referenced in the second if-clause, without necessarily being defined in the first if-clause.
- There were minor TODOs in the comments, I took the liberty and finished them.

### DIFF
--- a/src/porepy/numerics/ad/operators.py
+++ b/src/porepy/numerics/ad/operators.py
@@ -32,8 +32,18 @@ GridLike = Union[pp.Grid, Edge]
 
 # Abstract representations of mathematical operations supported by the Ad framework.
 Operation = Enum(
-    "Operation", ["void", "add", "sub", "mul", "evaluate", "div", "localeval", "apply"]
+    "Operation", ["void",
+                  "add", "sub", "mul", "div",
+                  "evaluate"]
 )
+
+
+def _get_shape(mat):
+    """ Get shape of a numpy.ndarray or the Jacobian of Ad_array"""
+    if isinstance(mat, (pp.ad.Ad_array, pp.ad.forward_mode.Ad_array)):
+        return mat.jac.shape
+    else:
+        return mat.shape
 
 
 class Operator:
@@ -218,7 +228,7 @@ class Operator:
         the tree of this operator, using data from gb.
 
         IMPLEMENTATION NOTE: The discretizations was identified at initialization of
-        Expression - it is now done here to accomodate updates (?) and
+        Expression - it is now done here to accommodate updates (?) and
 
         """
         unique_discretizations: Dict[
@@ -238,7 +248,7 @@ class Operator:
     def set_name(self, name: str) -> None:
         self._name = name
 
-    def parse(self, gb) -> Any:
+    def parse(self, gb: pp.GridBucket) -> Any:
         """Translate the operator into a numerical expression.
         Subclasses that represent atomic operators (leaves in a tree-representation of
         an operator) should override this method to retutrn e.g. a number, an array or a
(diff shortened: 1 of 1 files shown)


src/porepy/numerics/ad/operators.py
```diff
@@ -454,31 +420,63 @@ def error_message(operation):
...
+        Make a human-readable error message related to a parsing error.
+        NOTE: The exact formatting should be considered work in progress,
+        in particular when it comes to function evaluation.
+        """
 
         # There are three cases to consider: Either the operator is a leaf,
         # it is a composite operator with a name, or it is a general composite
         # operator.
-        if op.is_leaf():
+        if self.is_leaf():
```
keileg, line 465: I agree, changing to `self`, and calling this function on `tree.children `(above) makes more sense.

src/porepy/numerics/ad/operators.py
```diff
@@ -488,24 +486,20 @@ def _parse_readable(self, op: "Operator") -> str:
             operator_str = "*"
         elif tree.op == Operation.div:
             operator_str = "/"
-        elif tree.op in [Operation.evaluate, Operation.apply]:
-            # TODO: This has not really been tested.
+        elif tree.op == Operation.evaluate:
```
keileg, line 489: I'm not sure I agree we should delete the last else, covering unknown operations (it should raise an error rather than print a statement, though). Two reasons:
1. Someone may try to hack the Ad framework, and somehow make tree with a not recognized error message. Due to the complexity of this code, I prefer to raise explicit errors, rather than letting the hacker suffer the consquences.
2.  More likely than 1., we will implement new operations, but forget to add them to this type of parsing. If so, a reminder in the form of an error would be good.

src/porepy/numerics/ad/operators.py
```diff
@@ -488,24 +486,20 @@ def _parse_readable(self, op: "Operator") -> str:
...
-            print("Have not implemented string parsing of this operator")
 
         if is_func:
-            # TODO: Not sure what to write here
-            msg = f"{child_str[0]} evaluated on ("
-            for child in range(1, len(child_str)):
-                msg += f"{child_str[child]}, "
-
+            msg = f"{child_str[0]}("
+            msg += ", ".join([f"{child}" for child in child_str[1:]])
```
keileg, line 494: Much better!

src/porepy/numerics/ad/operators.py
keileg: Ah, you rather let operator_str be None, and give a special message below? If so, please document shortly.

keileg (commented):
A few minor comments. Good work, thanks.

src/porepy/numerics/ad/operators.py
```diff
@@ -488,24 +487,25 @@ def _parse_readable(self, op: "Operator") -> str:
...
             msg += ")"
             return msg
+        # if operation is unknown, a new error will be raised to raise awareness
+        elif operator_str is None:
+            msg = "UNKNOWN parsing of operation on: "
+            msg += ", ".join([f"{child}" for child in child_str])
+            raise NotImplementedError(msg)
```
vlipovac, line 505: Hi @keileg ,
I have supplemented the docs now and introduced this NotImplementedError here for unknown operations.
This will give error message of the following form:
'During handling of the above exception, another exception occurred:' (raising errors in an except block)
So the actual error with the operation should not be covered up.

keileg (approved):

keileg:
I'm happy with the update, but I guess @jwboth  should have a short look to on the removal of the local evaluation?

jwboth (approved):
Looks all very good to me. I in particular okay the deletion of `apply`, `localeval`, and all corresponding supplements. Thanks for taking care of this.

Issue #645, opened 2022-04-21
Poor test coverage of the operator part of the Ad framework


keileg:
The part of the Ad framework that deals with the pp.ad.Operator formulation (that is, not the forward mode part), is currently mainly tested in integration tests. The following components are missing (list to be expanded):
* Unit testing of the Operator evaluation framework, where all Operators are exposed to various arithmetic operations (+, - etc.), and combined with each other, and with related data structures which it is natural the operators can be combined with (e.g., numpy arrays, ints, floats etc.). Fixing the bugs that are bound to pop up is likely to be time consuming, but should also remove issues that show up because of special combinations that have not yet been implemented.

PR #651, opened 2022-05-04
BUG: Ad_array divide and power functions special cases.

keileg:
Fix issues related to taking `Ad_array`s to negative integer powers. 

## Proposed changes
Raising numpy arrays to negative integer powers is not allowed (no idea why). Fixed this by selective casting to floats when needed.

Also significantly improved test coverage for the relevant functions. As an observation, it took some effort to realize which special cases must be covered etc., indicating that fully dealing with #645 will be a major undertaking just to cover the forward ad part.

## Types of changes

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/porepy/numerics/ad/forward_mode.py
+++ b/src/porepy/numerics/ad/forward_mode.py
@@ -91,24 +91,84 @@ class Ad_array:
 
     def __pow__(self, other):
         if not isinstance(other, Ad_array):
-            val = self.val**other
-            jac = self.diagvec_mul_jac(other * self.val ** (other - 1))
+            if isinstance(other, int) or isinstance(other, np.integer):
+                # Standard ints and numpy scalars of integer format can be converted to
+                # float in a standard way
+                val = self.val ** float(other)
+                jac = self.diagvec_mul_jac(other * self.val ** float(other - 1))
+            elif isinstance(other, np.ndarray) and np.issubdtype(
+                other.dtype, np.integer
+            ):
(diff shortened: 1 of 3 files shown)


src/porepy/numerics/ad/forward_mode.py
```diff
@@ -91,24 +91,84 @@ def __rmul__(self, other):
...
+                jac = self.diagvec_mul_jac(
+                    other * self.val ** (other.astype(float) - 1)
+                )
+            else:
+                # Other should be a float, or have float data type; raising to the power
+                # of other should anyhow be fine. If there are more special cases not
+                # yet hit upon, an error message will be given here.
+                val = self.val**other
+                jac = self.diagvec_mul_jac(other * self.val ** (other - 1))
+
```
IvarStefansson, line 113: Couldn't case 1 and 3 be handled as one (with the solution for 1, which doesn't do anything for 3)? Don't know if it's better

IvarStefansson (approved):
Very good!

src/porepy/numerics/ad/forward_mode.py
keileg: Unfortunately no . `other` can be a np.array, in which case `float(other)` gives an error (it turned out). One could let the second condition find all `numpy `arrays (and do an unnecessary conversion if `dtype=float`), but the third else would still be needed for `floats`, and other data types for which the operation works out of the box. It also turned out, there are more special cases in the form of data types than what I would have thought.

PR #652, opened 2022-05-05
Approximate Jacobian Operator

vlipovac:
# New AD Operation `approximate`

An abstract base class for blackbox function evaluations and approximation of the respective Jacobian was created.
Usage is the same as for existing `ad.Function` (by inheritence), only this time values are provided by feeding the blackbox function (vector or cell-wise) with values and the Jacobian has to be provided by provided by overwritting an abstract method.

For this reason a new operation `approximate` had to be introduced (compare with `evaluate`).

Additionally to the abstract base class `ApproxJacFunction`, which provides the above described functionality, a simple `LJacFunction` class is introduced, which approximates the Jacobian with an constant times identity (per argument variable)

Finally, a respective parametrizable test case was added, where derived approximate operators can be tested by
1. solving a simple non-linear problem
2. checking evaluated values and consistency of dimensions (of the Jacobian) in the case of a mD-domain

# Remarks
- @OmarDuran noted that a test for error propagation in nested, approximate operators might be necessary in the future
- The test for dimension-consistency tests only variables on subdomains (mortar variables on mortar grids is equivalent). Tests for calls with both, mortar and subdomain variables are not tested. Is there a use case for that?


keileg:
Hi @vlipovac. I'll have a close look at this as soon as possible, however, it may take some time due to several other commitments on my part over the coming days.

I'll get back to you.

vlipovac:
> @vlipovac, from my side, reviewing the PR will take me longer than I expected since the functionality we look for is to set the Jacobian matrix (with full entries). In addition, allowing to set the values for the quantity could be better instead to use lambda expressions. 
> 
> The user should handle tailored approximated Jacobians. He will be in charge of estimating the Jacobian based on the L-scheme for another form.

Hi @OmarDuran ,

I am not sure I can follow. The lambda functions are only used to define a test function for the unit test, not the Jacobian.
I am open for implementing new class with Approximate Jacobians, let us discuss that during a meeting in person.

vlipovac:
Closing PR for implementing discussed abstraction of evaluating functions

PR #654, opened 2022-05-10
fix the exporter for the polygon

alessiofumagalli:
## Proposed changes

Bug fix, the current version of pp fails the exporting to vtk in presence of polygonal 2d cells with different number of cell to node relation.

## Types of changes

## Checklist

- [X] I have added necessary documentation 
- [X] I have checked that I have not duplicated existing functionality

## Further comments

Not sure if the exporter is really tested in pp, which is the reason why I didn't put the first X in the checklist

### DIFF
--- a/src/porepy/viz/exporter.py
+++ b/src/porepy/viz/exporter.py
@@ -628,9 +628,8 @@ class Exporter:
                 # define the type of cell we are currently saving
                 cell_type_raw = "polygon" + str(nodes_loc.shape[1])
 
-                default_type = "polygon"
                 # Map to triangle/quad if relevant, or simple polygon for general cells.
-                cell_type = polygon_map.get(cell_type_raw, default_type)
+                cell_type = polygon_map.get(cell_type_raw, cell_type_raw)
                 # if the cell type is not present, then add it
                 if cell_type not in cell_to_nodes:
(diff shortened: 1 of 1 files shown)


IvarStefansson (approved):
Looks good to me

keileg:
Looks good to me; I trust that the new code is correct (Paraview will tell you otherwise, I suppose)

@alessiofumagalli: Meshio used to differ between polygons with different number of vertexes, so this must have changed at some point, but I could not find it in the log. Do you have any idea when it was introduced, and specifically, whether we should update the version number in the requirements lists?

@jwboth: Are you aware of this fix? It will be relevant for your update of the exporter.

keileg (approved):

jwboth:
> ...
> @jwboth: Are you aware of this fix? It will be relevant for your update of the exporter.

The update of exporter has indeed already incorporated this fix. I don't recall why I changed it, most certainly I also observed some issue with meshio after updating its version.

alessiofumagalli:
@keileg not sure when it has changed since I've not using polygonal elements since some months :/

alessiofumagalli:
(not sure I'm the one able to do the merge)

keileg:
Then we'll not limit the meshio version before needed. 

Thanks a lot @alessiofumagalli

PR #656, opened 2022-05-18
Refactor Exporter

jwboth:
## Proposed changes

This PR contains an overhaul of the Exporter. Main changes are the way data can be prescribed from the user-perspective, there is a distinction between constant and non-constant data to be exported to different files, and the creation of mesh-related data has been improved in terms of performance as well as the covering of special geometries (hexahedra have not been treated in a special manner before). A new tutorial has been added to demonstrate how to use the Exporter.

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)
- [X ] Testing (contribution related mainly to testing of existing functionality)
- [X ] Documentation Update (if none of the other choices apply)

## Checklist

- [X ] I have added tests that prove my fix is effective or that my feature works
- [X ] I have added necessary documentation 
- [X ] I have checked that I have not duplicated existing functionality

## Further comments

This is relatively large change. Thus, I expect not eveyone being pleased with all changes. I am happy to hear feedback and adaopt the code. 

Some things to consider are:
- Meshio only treates unstructured grids. How often do users use large hexahedral grids? Writting a dedictaed exporter for structured meshes could be an option.
- The VTU file format is actually not perfectly useful for time series since the mesh data has to eb exported at every time step. The corresponding memory footprint can be significant. XDMF format should be considered for this issue.
- Is there someone who uses change_name? Or is this routine obsolete?
- Currently, explicit data can be defined from the run script for any geometrical dimension. However, the user has to make sure that data is provided for all grids of that dimension. Would it be of interest to dedicate a single file for each part of a subdomain?  Then one could also allow for extra data for parts of grids also. At the moment this option has not been chosen.
- Note, that vectorial states and explicit data is interpreted internallly as vectorial data and does not have to be padded with zeros. This has the advantage of less memory use (in certain situations), but when wanting to warp data by vector in ParaView, one will have to pad the zero (using the Calculator) in the image analysis.


src/porepy/viz/exporter.py
```diff
@@ -520,546 +1064,789 @@ def _export_grid(
...
+        }
 
-        # points
+        # Data structure for storing node coordinates of all 1d grids.
         num_pts = np.sum([g.num_nodes for g in gs])
         meshio_pts = np.empty((num_pts, 3))  # type: ignore
-        pts_pos = 0
 
-        # loop on all the 1d grids
+        # Initialize offset. Required taking into account multiple 1d grids.
```
IvarStefansson, line 1097: I don't understand this comment

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1064,789 @@ def _export_grid(
...
+        return Meshio_Geom(meshio_pts, meshio_cells, meshio_cell_id)
+
+    def _simplex_cell_to_nodes(
+        self, n: int, g: pp.Grid, cells: Optional[np.ndarray] = None
+    ) -> np.ndarray:
+        """
+        Determine cell to node connectivity for a general n-simplex mesh.
+
+        Parameters:
+            n (int): order of the simplices in the grid.
```
IvarStefansson, line 1139: "order" is what we usually refer to as "dimension", right? If so, could we use that description to avoid confusion (order could also be related to e.g. the orientation/node ordering in this context)

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1064,789 @@ def _export_grid(
...
+                    #  /\          |  |
+                    # x--x         x--x .
+                    # Therefore, use both g.cell_faces and g.face_nodes to make use of face
+                    # information and sort those to retrieve the correct connectivity.
+
+                    # Strategy: Collect all cell nodes including their connectivity in a
+                    # matrix of double num cell size. The goal will be to gather starting
+                    # and end points for all faces of each cell, sort those faces, such that
+                    # they form a circlular graph, and then choose the resulting starting
+                    # points of all faces to define the connectivity.
```
IvarStefansson, line 1264: Excellent documentation! Should we consider extracting a subfunction?

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1064,789 @@ def _export_grid(
...
 
+        # For each cell_type store the connectivity pattern cell_to_nodes for
+        # the corresponding cells with ids from cell_id.
         for block, (cell_type, cell_block) in enumerate(cell_to_nodes.items()):
-            # remove the number of nodes associated to the polygon since meshio requires only
-            # the keyword "polygon" for polygons
-            cell_type_ = "polygon" if "polygon" in cell_type else cell_type
-            meshio_cells[block] = meshio.CellBlock(cell_type_, cell_block.astype(int))
+
+            # Meshio requires the keyword "polgon" for general polygons.
```
IvarStefansson, line 1318:
```suggestion
            # Meshio requires the keyword "polygon" for general polygons.
```

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1064,789 @@ def _export_grid(
...
+            meshio_cells[block] = meshio.CellBlock(
+                cell_type_meshio_format, cell_block.astype(int)
+            )
             meshio_cell_id[block] = np.array(cell_id[cell_type])
 
-        return meshio_pts, meshio_cells, meshio_cell_id
+        # Return final meshio data: points, cell (connectivity), cell ids
+        return Meshio_Geom(meshio_pts, meshio_cells, meshio_cell_id)
+
+    def _sort_point_pairs_numba(self, lines: np.ndarray) -> np.ndarray:
```
IvarStefansson, line 1329: No chance of avoiding code duplication? If no, could we colocate the two functions? It seems self is not accessed.

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1064,789 @@ def _export_grid(
...
+
+        Returns:
+            Meshio_Geom: Points, 3d cells (storing the connectivity), and
+            cell ids in correct meshio format.
+        """
+
+        # Three different cell types will be distinguished: Tetrahedra, hexahedra,
+        # and general polyhedra. meshio does not allow for a mix of cell types
+        # in 3d within a single grid (and we export all 3d grids at once). Thus,
+        # if the 3d grids contains cells of varying types, all cells will be casted
```
IvarStefansson, line 1441:
```suggestion
        # if the 3d grids contains cells of varying types, all cells will be cast
```

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1064,789 @@ def _export_grid(
...
+                # Store short cuts to cell-face and face-node information
+                cf_indptr = g.cell_faces.indptr
+                cf_indices = g.cell_faces.indices
+                fn_indptr = g.face_nodes.indptr
+                fn_indices = g.face_nodes.indices
+
+                # Determine the cell-face connectivity (with faces described by their
+                # nodes ordered such that they form a chain and are identified by the
+                # face boundary. The final data format is a List[List[np.ndarray]].
+                # The outer list, loops over all cells. Each cell entry contains a
```
IvarStefansson, line 1711:
```suggestion
                # The outer list loops over all cells. Each cell entry contains a
```

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1064,789 @@ def _export_grid(
...
-        # we need to split the data for each group of geometrically uniform cells
-        cell_id = meshio_geom[2]
-        num_block = cell_id.size
-
+        # Split the data for each group of geometrically uniform cells
+        # Utilize meshio_geom for this.
         for field in fields:
+
+            # TODO RM? as implemented now, field.values should never be None.
             if field.values is None:
```
IvarStefansson, line 1767: In that case, I suggest
```suggestion
           assert field.values is not None
```

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1064,789 @@ def _export_grid(
...
+            folder_name (str, optional): name of the folder
+            name (str): prefix of the name of the files to be written
+
+        Returns:
+            str: complete path to the files to be written.
+        """
 
-    def _make_folder(self, folder_name: Optional[str] = None, name: str = "") -> str:
+        # If no folder_name is prescribed, the files will be stored in the
+        # same folder.
```
IvarStefansson, line 1807: Does "same folder" mean "working directory"?

tests/unit/test_vtk.py
```diff
@@ -157,6 +205,7 @@ def test_gb_1(self):
             content = self.sliceout(content_file.read())
         self.assertTrue(content == self._gb_1_mortar_grid_vtu())
 
+    # TODO Do we need this test if we have the subsequent one?
```
IvarStefansson, line 208: I think not

tests/unit/test_vtk.py
```diff
@@ -197,6 +252,88 @@ def test_gb_2(self):
...
+        with open(self.folder + self.file_name + "_2.vtu", "r") as content_file:
+            content = self.sliceout(content_file.read())
+        self.assertTrue(content == self._nonconstant_data_grid_vtu())
+
+        with open(self.folder + self.file_name + "_constant_2.vtu", "r") as content_file:
+            content = self.sliceout(content_file.read())
+        self.assertTrue(content == self._constant_data_grid_vtu())
+
+    # NOTE: This test is not directly testing Exporter, but the capability to export networks.
+    # Is this a used capability? If not, suggest removing the test and the corresponding capability.
```
IvarStefansson, line 336: I think this can be useful when working with complex grids (difficult or expensive to mesh)

IvarStefansson (commented):
Terrific work!
I get an error when running the tutorial. Is this a local problem?

src/porepy/viz/exporter.py
```diff
@@ -290,191 +409,563 @@ def write_pvd(
...
+            return value
 
-    def _export_gb(self, data: List[str], time_step: float) -> None:
-        """Export the entire GridBucket and additional data to vtu.
+        # Auxiliary functions for type checking
 
-        Parameters:
-            data (List[str]): Data to be exported in addition to default GridBucket data.
-            time_step (float) : Time step, to be appended at the vtu output file.
+        def isinstance_interface(e: Interface) -> bool:
```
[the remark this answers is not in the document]
jwboth: I do it here - my understanding is that you were asking what I have been wanting to achieve and how I did it:

My main goal was to perform the python routine isinstance on user-defined types, which mainly are of tuple type (e.g., `Interface`). However, isinstance, does not work on types defined using typing. It can only recognize objects that have been created using an constructor of an object. When however defining inline a tuple `a=("tmp", 2)` and the type `customTuple=Tuple[str,int]`, checking for `isinstance(a, customTuple)` is not possible.

A rather short solution to this is asking for the type of each argument of `a`, using `list(map(type, a))` which returns `[str,int]` in this case. 

When checking whether an argument is a list of grids, an additional check has be performed, as a list of grids will in the first place simply be identified as `list`. Thus, each element of this list is checked as well.

src/porepy/viz/exporter.py
jwboth: @keileg , what do you think, as I believe you have been using the original routine. `_sort_point_pairs_numba` as defined here has the simplification that it assumes circular polygons and on the other hand allows for a list of them instead of only considering a single one. I stripped the routine down to its essentials to allow for numba - this was the main motivation.  Moving it to utils could be indeed an idea.

src/porepy/viz/exporter.py
```diff
@@ -10,247 +10,343 @@
...
-    def __iter__(self) -> Generator[Field, None, None]:
-        """
-        Iterator on all the fields.
-        """
-        for f in self._fields:
-            yield f
+    Optional arguments in kwargs:
+    fixed_grid: (optional) in a time dependent simulation specify if the
+        grid changes in time or not. The default is True.
+    binary: (optional) export in binary format, default is True.
```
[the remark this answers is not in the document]
jwboth: I added a sentence describing the alternative case, i.e., setting 'binary' False, which should implicitly also explain the case when binary=True. Is this sufficient? One could add that the data is base64-encoded, but as the storage procedure is performed by meshio, I would suggest not adding more details here.

src/porepy/viz/exporter.py
```diff
@@ -10,247 +10,343 @@
...
+    If you need to export the state with key "pressure" on a single grid:
+    save = Exporter(g, "solution", folder_name="results")
+    save.write_vtu(["pressure"])
+
+    In a time loop, if you need to export states with keys "pressure" and
+    "displacement" stored in a grid bucket.
+    save = Exporter(gb, "solution", folder_name="results")
+    while time:
+        save.write_vtu(["pressure", "displacement"], time_step=i)
+    save.write_pvd(times)
```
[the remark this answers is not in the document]
jwboth: I made the change as proposed.

src/porepy/viz/exporter.py
```diff
@@ -10,247 +10,343 @@
...
+    where times, is a list of actual times (not time steps), associated
+    to the previously exported time steps.
+
+    In general, pvd files gather data exported in separate files, including
+    data on differently dimensioned grids, constant data, and finally time steps.
+
+    In the case of different keywords, change the file name with
+    "change_name".
+
+    NOTE: the following names are reserved for data exporting: grid_dim,
```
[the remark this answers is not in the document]
jwboth: Added a comment indicating that the constant data associated to these keywords will be overwritten otherwise.

src/porepy/viz/exporter.py
```diff
@@ -290,191 +409,563 @@ def write_pvd(
...
+                            )
+                        )
 
         o_file.write("</Collection>\n" + "</VTKFile>")
         o_file.close()
 
-    def _export_g(self, data: Dict[str, np.ndarray], time_step):
+    # Some auxiliary routines used in write_vtu()
+
+    def _sort_and_unify_data(
```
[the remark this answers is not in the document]
jwboth: I tried to resort the code, and hopefully it is more readible now. I however see that this routine may be a bit long. Depending on the amount of allowed data formats for the input data, the longer this routine...

With the new clear separation of auxiliary routines and main body of the routine, I hope the overall strategy is clearer and readible.

src/porepy/viz/exporter.py
```diff
@@ -486,28 +977,81 @@ def _export_pvd_gb(self, file_name: str, time_step: float) -> None:
...
+
         o_file.write("</Collection>\n" + "</VTKFile>")
         o_file.close()
 
-    def _export_grid(
-        self, gs: Iterable[pp.Grid], dim: int
-    ) -> Union[None, Tuple[np.ndarray, np.ndarray, np.ndarray]]:
+    def _update_meshio_geom(self) -> None:
+        """
+        Auxiliary method upating internal copies of the grids. Merely asks for
```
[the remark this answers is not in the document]
jwboth: Udpated, please check again.

src/porepy/viz/exporter.py
jwboth: updated, please check again.

tutorials/exporter.ipynb
```diff
@@ -0,0 +1,379 @@
...
+    "interface_states = set(interface_states) - set([pp.ITERATE])\n",
+    "print(\"Keys of the states defined on interfaces:\", set(interface_states))   "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Defining the exporter\n",
+    "The definition of an object of type pp.Exporter, merely a grid bucket as well as a target is required for the output."
```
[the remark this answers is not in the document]
jwboth: Changed it, please review again.

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1064,789 @@ def _export_grid(
...
+            cell_id += (cells + cell_offset).tolist()
+
+            # Determine local connectivity for all cells. Simplices are invariant
+            # under permuatations. Thus, no specific ordering of nodes is required.
+
+            # After all, the procedure is fairly similar to the treatment of
+            # tetra cells. Most of the following code is analogous to
+            # _simplex_cell_to_nodes(). However, for hexaedron cells the order
+            # of the nodes is important and has to be adjusted. Based on the
+            # specific definition of TensorGrids however, the node numbering
```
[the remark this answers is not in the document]
jwboth: An internal test (using numba, checking all cells) has been added, checking whether the hard-coded numbering really defines a hex.

src/porepy/viz/exporter.py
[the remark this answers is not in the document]
IvarStefansson: I can't see the added sentence. Sorry if this is just me beeing stupid/blind!

src/porepy/viz/exporter.py
```diff
@@ -10,247 +10,343 @@
...
+            data (Union[DataInput, List[DataInput]], optional): node and
+                edge data, prescribed through strings, or tuples of
+                grids/edges, keys and values. If not provided only
+                geometical infos are exported.
+
+                NOTE: The user has to make sure that each unique key has to
+                associated data values for all or no grids of each specific
+                dimension.
+        """
+        # Identify change in constant data. Has the effect that
```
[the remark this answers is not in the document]
IvarStefansson: I think my comment was to the documentation, which says "identify constant data". That reads to me like some sort of check is being performed.

src/porepy/viz/exporter.py
IvarStefansson: As in "required for/when taking into..."?

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1177,871 @@ def _export_grid(
...
+        # There is only one cell type.
+        meshio_cells = np.empty(1, dtype=object)
+        meshio_cell_id = np.empty(1, dtype=object)
+        meshio_cells[0] = meshio.CellBlock("hexahedron", cell_to_nodes.astype(int))
+        meshio_cell_id[0] = np.array(cell_id)
+
+        # Return final meshio data: points, cell (connectivity), cell ids
+        return Meshio_Geom(meshio_pts, meshio_cells, meshio_cell_id)
+
+    def _test_hex_meshio_format(self, nodes, cn_indices) -> bool:
```
jwboth, line 1756: This method is the source of problems in the current build test. It is only a safety check which does not affect the funcitonality of the code. I suggest ignoring it for the moment. I will try to fix this, while the rest of the code still can be reviewed.

jwboth:
> Terrific work! I get an error when running the tutorial. Is this a local problem?

The tutorial should work now.

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1178,889 @@ def _export_grid(
...
+                # array is not the same troubling numba
+                # global_ind_0 = cn_indices[i, 0:4]
+                # global_ind_1 = cn_indices[i, 4:8]
+                global_ind_0 = np.array(
+                    [
+                        cn_indices[i, 0],
+                        cn_indices[i, 1],
+                        cn_indices[i, 2],
+                        cn_indices[i, 3],
+                    ]
```
jwboth, line 1795: For the moment the build tests are successful, but the fix is not nice...

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1178,889 @@ def _export_grid(
...
+            nodes_offset += g.num_nodes
 
-        # construct the meshio data structure
-        num_block = len(cell_to_nodes)
-        meshio_cells = np.empty(num_block, dtype=object)
-        meshio_cell_id = np.empty(num_block, dtype=object)
+        # Construct the meshio data structure
+        num_blocks = len(cell_to_nodes)
+        meshio_cells = np.empty(num_blocks, dtype=object)
+        meshio_cell_id = np.empty(num_blocks, dtype=object)
```
keileg, line 1237: My clear preference in these cases is to not use np.array as general purpose containers, but rather use List. From what I can tell, this should be no problem from the meshio side, or have I missed something?

src/porepy/viz/exporter.py
keileg: No need for a subfunction until we need the ordering somewhere else. I hope that will not be the case for a while.

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1178,889 @@ def _export_grid(
...
+            nodes_offset += g.num_nodes
+            cell_offset += g.num_cells
 
-        # construct the meshio data structure
-        num_block = len(cell_to_nodes)
-        meshio_cells = np.empty(num_block, dtype=object)
-        meshio_cell_id = np.empty(num_block, dtype=object)
+        # Construct the meshio data structure
+        num_blocks = len(cell_to_nodes)
+        meshio_cells = np.empty(num_blocks, dtype=object)
```
keileg, line 1433: Again, I would prefer list rather than np.ndarray as containers.

src/porepy/viz/exporter.py
keileg: Moving to utils.point_pairs makes sense to me. As @jwboth says, the two functions make different assumptions, thus a unified function will get some keywords, and then an if-else that essentially calls one of the two methods. That may or may not be an improvement, and the perspective may be different from the user and maintainer side.

I suggest we first move it, and then see what it looks like.

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1178,889 @@ def _export_grid(
...
+
+        Returns:
+            np.ndarray: Sorted version of lines, where for each chain, the collection
+                of line segments has been potentially flipped and sorted.
+        """
+
+        import numba
+
+        @numba.jit(nopython=True)
+        def _function_to_compile(lines):
```
keileg, line 1475: It should be possible to specify the data types of in- and output. That should speed up numba significantly. @jwboth, let's discuss this.

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1178,889 @@ def _export_grid(
...
+            cell_to_nodes = np.vstack((cell_to_nodes, cn_indices + nodes_offset))
+
+            # Update offset
+            nodes_offset += g.num_nodes
+            cell_offset += g.num_cells
+
+        # Initialize the meshio data structure for the connectivity and cell ids.
+        # There is only one cell type.
+        meshio_cells = np.empty(1, dtype=object)
+        meshio_cell_id = np.empty(1, dtype=object)
```
keileg, line 1651: See previous comments on containers.

src/porepy/viz/exporter.py
```diff
@@ -520,546 +1178,889 @@ def _export_grid(
...
+
+        # Return final meshio data: points, cell (connectivity), cell ids
+        return Meshio_Geom(meshio_pts, meshio_cells, meshio_cell_id)
+
+    def _test_hex_meshio_format(self, nodes, cn_indices) -> bool:
+
+        import numba
+
+        @numba.jit(nopython=True)
+        def _function_to_compile(nodes, cn_indices) -> bool:
```
keileg, line 1762: Again, if this is time consuming, informing numba on data types could help a lot.

keileg (commented):
Excellent work! Some comments, mainly small things. We should talk about 1) sort_point_ind, 2) (possibly) numba.

src/porepy/viz/exporter.py
```diff
@@ -10,247 +10,343 @@
...
+        # Parameter to be used in several occasions for adding time stamps.
+        self.padding = 6
+
+        # Require numba
+        if "numba" not in sys.modules:
+            raise NotImplementedError(
+                "The sorting algorithm requires numba to be installed."
+            )
 
+    # TODO in use by anyone?
```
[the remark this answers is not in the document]
jwboth: I remove it then.

src/porepy/viz/exporter.py
jwboth: Done as @keileg suggested.

src/porepy/viz/exporter.py
```diff
@@ -10,275 +10,409 @@
...
+            self._constant_interface_data[key_i] = value.copy()
+
     def write_vtu(
         self,
-        data: Optional[Union[Dict, List[str]]] = None,
-        time_dependent: bool = False,
-        time_step: int = None,
-        grid: Optional[Union[pp.Grid, pp.GridBucket]] = None,
+        data: Optional[Union[DataInput, List[DataInput]]] = None,
+        time_dependent: Optional[bool] = False,
```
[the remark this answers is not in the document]
jwboth: Thanks for this valuable comment!

src/porepy/viz/exporter.py
jwboth: Done, also regarding similar comments.

tests/unit/test_vtk.py
jwboth: @keileg: Any opinion? If not sure, we can keep it until a possible unit-test-overhaul afer the summer.

src/porepy/viz/exporter.py
jwboth: Included now.

tests/unit/test_vtk.py
jwboth: TBC

keileg (approved):

PR #658, opened 2022-05-23
Extension of discretization matrices

taj004:
In this PR, I modified several discretization matrices (tpfa, trace, interface fluxes) to account for several components. I also slightly changed the upwind discretization for 0D


keileg:
By agreement with @taj004, this will not be included in the main branch of PorePy.

PR #661, opened 2022-05-31
Ad operator functions

vlipovac:
## Proposed changes

Abstraction of Operation.evaluate used in ad.Function into an AbstractFunction object, s.t. general functions on AD operators can be implemented. The abstraction involves the evaluation of a callable as well as approximating the Jacobian of the callable with respect to passed operator arguments.
The original ad.Function was incorporated in this new hierarchy in a backwards-compatible way.

Added a decorator class which can be used to decorate bound and unbound (numerical) functions in Python, turning them effectively into AD operator functions.

An example test for AD operator functions was also added.

## Types of changes

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

## Further comments

Changes implemented here are result of the discussion at UiB and the withdrawn PR for `blackbox` functions (here generalized into AbstractFunction).


vlipovac:
@keileg mypy is throwing errors for things I did not program and which are already on the develop branch...

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+        The returned numpy array will be be set as 'jac' argument for the AD array representing
+        this instance.
+
+        NOTE: the necessary dimensions for the jacobian can be extracted from the dimensions
+        of the jacobian of passed Ad_array instances.
+        """
+        pass
+
+
+class AbstractJacFunction(AbstractFunction):
```
keileg, line 168: I'm not sure about the naming convention here, we have been moving towards full names (so, Jacobian rather than Jac).

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+    ):
+        super().__init__(func, name, is_vector_func, True)
+
+    def get_values(self, *args: "Ad_array") -> "np.ndarray":
+        result = self.func(*args)
+        return result.val
+
+    def get_jacobian(self, *args: "Ad_array") -> "np.ndarray":
+        result = self.func(*args)
+        return result.jac
```
keileg, line 231: I don't like this signature at all. If `self.func` is costly (and somtimes it will be), we will have twice the computational cost for no clear reason.

It seems to me the need to split this comes from forcing `pp.ad.Function` to be on the same inheritance level at the wrapping of an Ad_array with known value and Jacobian. It is not clear this is the right design choice (my instinct is it is wrong).

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+
+
+class LJacFunction(AbstractJacFunction):
```
keileg, line 234: The naming here is misleading (the L-scheme is just a special case of quasi-Newton method with a fixed Jacobian with a certain structure, it has nothing to do with approximation of derivatives per se), it should be something like DiagonalJacFunction. More important, I would rather have a function with a preset sparse matrix as the Jacobian and value. Although assigning diagonal matrices may be an important special case, it is not at all clear this deserves a special class.

I have the feeling the superclasses may be intended to cover the case of more general approximated Jacobian matrices - if so, see comments elsewhere about naming conventions and the class hierarchy.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+        func: Callable,
+        name: str,
+        is_vector_func: Optional[bool] = False,
+    ):
+        """Constructor.
+
+        The L-multiplier for the L-scheme can be passed for every argument specifically
+        using a list.
+        The order in the list has to mach the order of arguments when calling
+        this instance.
```
keileg, line 252: Documentation is unclear, which arguments are being referred to?

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+        self,
+        L: Union[List[float], float],
+        func: Callable,
+        name: str,
+        is_vector_func: Optional[bool] = False,
+    ):
+        """Constructor.
+
+        The L-multiplier for the L-scheme can be passed for every argument specifically
+        using a list.
```
keileg, line 250: Again, the L-method may be a motivation, but there is no reason to make a class dedicated to this.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+        # the Jacobian of a (Merged) Variable is already a properly sized block identity
+        if len(args) >= 1:
+            jac = args[0].jac * self._L[0]
+
+            # summing identity blocks for each dependency
+            if len(args) > 1:
+                # TODO think about exception handling in case not enough
+                # L-values were provided initially
+                for arg, L in zip(args[1:], self._L[1:]):
+                    jac += arg.jac * L
```
keileg, line 280: I may understand how this is supposed to work, but I'm not sure - and I'm pretty sure a client will misunderstand. Also, be aware that for a problem with multiple variables, the full jac of arg (not only the block related to 'itself') - is this indended?

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+
+
+### FUNCTION DECORATOR ------------------------------------------------------------------------
+
+
+class ADmethod:
+    """
+    Automatic-Differentiation method/function.
+
+    (Decorator) Class for methods representing e.g., physical properties.
```
keileg, line 296: functions I suppose, not methods?

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+                + str(self._func.__qualname__)
+            )
+            self._op_kwargs.update({"name": name})
+
+        # calling the operator
+        wrapping_operator = self._ad_func_type(func=operator_func, **self._op_kwargs)
+
+        return wrapping_operator(*args, **kwargs)
+
+    def __get__(self, binding_instance: object, binding_type: type):
```
keileg, line 436: I don't understand this at all. A much simpler documentation (less technical) documentation is needed.

keileg (changes requested):
Several comments (there may be more from my side eventually, but this will have to do for now). I did not go through the files linearly, so you may want to pass through all comments before studying in detail. 

Most important remarks:
1. I do not agree with the class structure you suggest for the functions. There may be aspects that I don't see, but the changes you had to introduce to pp.ad.Function indicate the design is wrong. I suggest we discuss this on a blackboard.
2. Tests must be improved significantly, both in structure, but also in what is being tested and how.
3. I think the AdMethod may be more useful and important than what I could infer now, so we should discuss this at some point also.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+
+    The abstraction intends to provide means for approximating operators, where values are
+    e.g. interpolated and the Jacobian is approximated using FD.
+    """
+
+    def __init__(
+        self,
+        func: Callable,
+        name: str,
+        is_vector_func: Optional[bool] = False,
```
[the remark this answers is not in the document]
vlipovac: How about `is_array_func`?

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+        Important NOTE: flagging this instance with `is_adarray_func` will lead to a direct
+        evaluation of the callable using Ad_array instances, effectively bypassing
+        the abstract methods (but still leading to NotImplementedError if not implemented)
+        Might not be elegant, but we can put the old `Function` class this way in the same
+        category/hierarchy.
+
+        :param func: callable Python object representing the function
+        :type func: callable
+        :param name: Name of this operator instance
+        :type name: str
```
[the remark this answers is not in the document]
vlipovac: I will rework it with the doc proposal for the AD submodule, as we agreed last time.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+        func: Callable,
+        name: str,
+        is_vector_func: Optional[bool] = False,
+        is_adarray_func: Optional[bool] = False,
+    ):
+        """Constructor. Saves information about the passed callable.
+        The passed callable is expected to take numerical information in some form and
+        return numerical information in some form.
+
+        Important NOTE: flagging this instance with `is_adarray_func` will lead to a direct
```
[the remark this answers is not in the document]
vlipovac: `is_adarray_func` was added to fit the original `Function` class in the hierarchy, i.e. if marked `True` the class will attempt  to feed the `Ad_array`s of the children directly as arguments for the callable. Let's talk at the blackboard as you've suggested above.

src/porepy/numerics/ad/operator_functions.py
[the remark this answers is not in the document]
vlipovac: check!

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+        of the jacobian of passed Ad_array instances.
+        """
+        pass
+
+
+class AbstractJacFunction(AbstractFunction):
+    """'Half'-abstract base class, for which a direct evaluation of the passed callable is
+    implemented. Only the Jacobian has to be concretized / approximated."""
+
+    def get_values(self, *args: "Ad_array") -> "np.ndarray":
```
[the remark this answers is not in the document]
vlipovac: I got this one already fixed by mypy (as well as the whole `operator_functions.py` file)
It explicitly told me to annotate it this way. I used 'Tuple' before and mypy was unhappy.

tests/unit/test_ad.py
```diff
@@ -521,6 +520,174 @@ def test_variable_combinations(grids, variables):
...
+            iterations = i
+            break
+
+        dz = sps.linalg.spsolve(A, b)
+
+        Z += dz
+        dof_manager.distribute_variable(Z)
+        dof_manager.distribute_variable(Z, to_iterate=True)
+
+    assert iterations != "NOT CONVERGED"
```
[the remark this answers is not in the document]
vlipovac: check

tests/unit/test_ad.py
```diff
@@ -521,6 +520,174 @@ def test_variable_combinations(grids, variables):
...
+    circle_a_ = operator_cls(
+        **constructor_kwargs[1], func=circle_, name="circle_a", is_vector_func=False
+    )(x, y)
+    circle_a = circle_a_ - radii**2
+
+    eq_manager.equations = {"line": line_a, "circle": circle_a}
+    iterations = "NOT CONVERGED"
+    res_norm = None
+
+    for i in range(4000):
```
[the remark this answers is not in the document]
vlipovac: This number is pretty random. 
Original tests shows a number of up to 150 necessary iterations.
This depends on the tested function class, we might need a better parametrization maybe?

tests/unit/test_ad.py
```diff
@@ -521,6 +520,174 @@ def test_variable_combinations(grids, variables):
...
+
+        dz = sps.linalg.spsolve(A, b)
+
+        Z += dz
+        dof_manager.distribute_variable(Z)
+        dof_manager.distribute_variable(Z, to_iterate=True)
+
+    assert iterations != "NOT CONVERGED"
+
+    ############ mD domain for test of dimensions
```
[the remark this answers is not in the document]
vlipovac: check

tests/unit/test_ad.py
```diff
@@ -521,6 +520,174 @@ def test_variable_combinations(grids, variables):
...
+    # this grid has in total 200 cell DOFs
+    # 88 cells in the 2D domain,
+    # 4 cells in the 1D domain,
+    # 8x2 cells on the Mortar grid
+    coord_point = np.array([[0.2, 0.8], [0.5, 0.5]])
+    indices_point = np.array([[0], [1]])
+    domain = {"xmin": 0, "xmax": 1, "ymin": 0, "ymax": 1}
```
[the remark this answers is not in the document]
vlipovac: check

tests/unit/test_ad.py
```diff
@@ -521,6 +520,174 @@ def test_variable_combinations(grids, variables):
...
+    fracture_network = pp.FractureNetwork2d(coord_point, indices_point, domain)
+    mesh_args = {"mesh_size_frac": 0.3}
+    gbmd = fracture_network.mesh(mesh_args)
```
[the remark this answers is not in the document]
vlipovac: Switched to `mdg`

tests/unit/test_ad.py
```diff
@@ -828,10 +995,10 @@ def test_div_advar_advar(self):
...
 
     def test_full_jac(self):
-        J1 = sps.csc_matrix(
-            np.array([[1, 3, 5], [1, 5, 1], [6, 2, 4], [2, 4, 1], [6, 2, 1]])
-        )
-        J2 = sps.csc_matrix(np.array([[1, 2], [2, 5], [6, 0], [9, 9], [45, 2]]))
+        # J1 = sps.csc_matrix(
+        #     np.array([[1, 3, 5], [1, 5, 1], [6, 2, 4], [2, 4, 1], [6, 2, 1]])
+        # )
+        # J2 = sps.csc_matrix(np.array([[1, 2], [2, 5], [6, 0], [9, 9], [45, 2]]))
```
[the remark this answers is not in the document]
vlipovac: Since it was not my code, I was not sure if I am allowed to delete it (at that point)
But I had to comment it out due to mypy complaining.

tests/unit/test_ad.py
```diff
@@ -521,6 +520,174 @@ def test_variable_combinations(grids, variables):
...
+    x = eq_manager.merge_variables([(g, "x") for g, _ in gbmd])
+    y = eq_manager.merge_variables([(g, "y") for g, _ in gbmd])
+    # mortar_x = eq_manager.merge_variables([(e, "mortar_x") for e, _ in gbmd.edges()])
+    # mortar_y = eq_manager.merge_variables([(e, "mortar_y") for e, _ in gbmd.edges()])
+
+    initial_values = np.ones(dof_manager.num_dofs())
+    dof_manager.distribute_variable(initial_values)
+
+    multiply = lambda x, y: x * y
+    op = operator_cls(**constructor_kwargs[0], func=multiply, name="bb-multip")(x, y)
```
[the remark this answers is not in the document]
vlipovac: The problem with the general operator functions is, that the validity of the analytical implementation for the Jac in `Ad_array` is overwritten with something arbitrary (from a programming point of view).
Any non-exact function can return anything as the Jacobian, leading to errors somewhere else in the whole framework.
So the motivation of this test is to avoid exactly that, since we know how the AD framework with operators work and how the Jacobians are supposed to be combined in the end.

tests/unit/test_ad.py
```diff
@@ -521,6 +520,174 @@ def test_variable_combinations(grids, variables):
...
+    [
+        ("LJacFunction", ({"L": [-1.0, 0.0]}, {"L": [0.0, 2.0 * 1.816]})),
+        ("Function", ({}, {})),
+    ],
+)
+def test_operator_functions(operator_class, constructor_kwargs):
+    """
+    1. Test: test of if a certain operator function can solve the following nonlinear problem
+
+    Find x, y s.t.
```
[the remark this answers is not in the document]
vlipovac: This test was suggested by Omar as a demonstration at first. We can design a more complex, but still simple flow problem as a test maybe? Considering your comment above about splitting the two tests, I will do so and put the first test under tests/integration

tests/unit/test_ad.py
```diff
@@ -521,6 +520,174 @@ def test_variable_combinations(grids, variables):
...
+        y-x = 0
+        y**2 + x**2 - r = 0
+    for r in [1., 1.25, 1.5, 1.75]
+    with initial guess
+        x_0 = y_0 = r / (1 + 0.2**2) + 0.1
+
+    2. Test: test of dimensions compatibility for a mixed-dimensional domain with mortar grids
+    but without mortar variables
```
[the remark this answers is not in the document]
vlipovac: It does not add any additional value to the test, if I am not mistaken.
If we add mortar grids, we have respective lines with zeros in the global Jac and variable vector for the dimensionality test.
Defining mortar vars and equations would not change the outcome.

src/porepy/numerics/ad/operator_functions.py
vlipovac: The `get_values` and `get_jacobian` methods of this re-designed `pp.ad.Function` are never called by the framework
(see file `operators.py` line 422)

The idea to put them together is motivated by the concept of a function (and the possibility to unify it) and to save code.
Otherwise we would quite some duplicate code in `Function` and `AbstractFunction`. Also in the `Operator` class, we would have to different "evaluate" operations.

What we can fall back on is the unfinished "approximate" operation, which was not completely implemented by Jakub.
Then we would keep the concept of an analytic `Function` and introduce the concept of approximating `AbstractFunction`.
But since they are so similar, it is worth a shot to unify them somehow. My first attempt was this hierarchy and the argument `is_adarray_func`.

src/porepy/numerics/ad/operator_functions.py
vlipovac: This class is a first working example. I agree that it's wider usability is not given.

src/porepy/numerics/ad/operator_functions.py
vlipovac: I am not sure I understand properly
The Jacobian constructed here is exact for the following function:
L_1 * x_1 + L_2 * x_2 + L_3 * x_3 +....
where x_i is a variable passed during the function call (e.g. `density(pressure, temperature)` density -function, pressure and temp - variables)
The multipliers L_i are the multipliers passed at instantiation.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+
+    def get_jacobian(self, *args: "Ad_array") -> "sps.spmatrix":
+        """The approximate jacobian is identity times L.
+
+        Where the respective blocks appears,
+        depends on the total DOFs and the order of arguments passed during the
+        call to this instance.
+        """
+        # the Jacobian of a (Merged) Variable is already a properly sized block identity
+        if len(args) >= 1:
```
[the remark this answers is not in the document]
vlipovac: Do you mean whether we know that the Jacobian of a MergedVariable is as stated in my comment or not?

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,463 @@
...
+        class IdealGas:
+
+            @ADmethod(ad_operator=pp.ad.LJacFunction, operators_args={"L"=[1,1]})
+            def density(self, p: float, T: float) -> float:
+                return p/T
+
+        # decorating function
+        @ADmethod(ad_operator=pp.ad.Function)
+        def dummy_rel_perm(s):
+            return s**2
```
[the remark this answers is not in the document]
vlipovac: Your example will throw an error because the argument `name` for `Function` is not provided by this syntax.
And that is the first purpose of this wrapper, parametrization of the used Function class.

Also one has to be careful when using a class as a decorator with or without arguments.
Depending on that, the reference to the function `dummy_rel_perm` will be fed to either `__init__` or `__call__`

And finally, one has two make very important differences between decorating functions and class methods, because of the `self` argument for class methods. Both cases are covered by `ADmethod`.

src/porepy/numerics/ad/operator_functions.py
vlipovac: Both. See my comment in above conversation.

src/porepy/numerics/ad/operator_functions.py
vlipovac: Unfortunately, this is the best I can do and it is a complex issue.
This is all due to the problem, that I want this class to be able to decorate also class methods, not just plain functions.
See my comment two conversations above.

keileg:
@vlipovac: I have had a look, pushed a few very minor changes. Mostly I am very happy with this, the added functionality will be very useful.

I have some comments and remarks, but it may be better to discuss this in person (or on a call). Shall we try to meet at some time?

keileg:
Also, you may want to have a quick look at the changes I added: I believe they are correct, but I may be missing something.

vlipovac:
@keileg the remaining unresolved issue is your remark regarding the way I included the (original) Function class into this hierarchy.
One alternative would be to pick up the ApproximateFunction ansatz which was hinted by @jwboth, effectively splitting again Operation.evaluate into the Operation.evaluate and Operation.approximate.
This new hierarchy would then serve Operation.approximate and the Function directly operating with Ad_array would remain for Operation.evaluate.

We can keep Function in this hierarchy and just change its operation, but I don't know if this is a clean way or to just leave them  cleanly separated.
Evaluation and approximation are in the end two different approaches.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,617 @@
...
+
+    """
+
+    def __init__(
+        self,
+        func: Callable,
+        name: str,
+        array_compatible: bool = False,
+        ad_compatible: bool = False,
+    ):
```
vlipovac, line 58: @keileg I have changed the signature and added extensive documentation as discussed.
I also resolved the outdated conversation on the issue with the hierarchy. We can use this thread now, since we boiled down the problem to proper documentation of intended use.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -0,0 +1,617 @@
...
+        Numerical operations with a zero always works with any numeric formats in
+        numpy, scipy and PorePy's AD framework.
+        Since the constant function (most likely) gets no arguments passed,
+        we have no way of knowing the necessary SHAPE for a zero matrix. Hence scalar.
+
+        Returns:
+            The trivial derivative of a constant.
+
+        """
+        return 0.0
```
vlipovac, line 299: @keileg  This unfortunately will have to stay, as explained in the implementation note.

keileg: I figured that might be a problem. Let's keep this for now, but mypy will likely force us to reconsider at some point.

src/porepy/numerics/ad/operator_functions.py
keileg: I think this is fair. The whole thing is still a bit opaque to me, but I believe that just reflects the inner workings of the AD framework in general, and of the functions in particular.

keileg (approved):
I'm okay with this now. We will likely revisit in the months to come, as we get more experience in how this really works.

@vlipovac: I don't know if you have any last comment to make, if not, let's mark the two final conversations resolved, and I will merge the PR.

PR #665, opened 2022-06-08
Bugfix AD arithmetics

vlipovac:
## Issue description

right-subtraction for ad.Operator class was implemented wrongly...
**The implications of this bug might be quite serious and far-reaching!**

Division operator did not allow for division by scalars or division of sparse matrices by something else
Both options were added.

## Types of changes

## Checklist


keileg:
I'll have a look. In the meantime, could you set up tests that verify that the new code is correct. I know this part of the code is not well covered, but we have to start somewhere.

src/porepy/numerics/ad/operators.py
```diff
@@ -575,7 +588,11 @@ def __radd__(self, other):
         return self.__add__(other)
 
     def __rsub__(self, other):
-        return self.__sub__(other)
```
IvarStefansson, line 578: Would other.__sub__(self) work as an alternative fix?

vlipovac: This depends on what the other is and if it supports that syntax.
My proposal makes that question obsolete since other is wrapped into the AD framework using `_parse_other`, which tells us what is reasonable, i.e. what it supports.

keileg (approved):
Nice catch.

We really need a testing framework for these operations, but that will have to be another PR.

PR #667, opened 2022-06-09
DEL: Purged old cython function for block inversion.

keileg:
## Proposed changes
Removed outdated cython function to invert block diagonal matrices. The code was never used (or tested) since the corresponding numba function was more efficient.

## Types of changes

- [x] General maintenance of the code.
## Checklist
Only deleted code.

- [ ] I have added necessary documentation 
- [ ] I have checked that I have not duplicated existing functionality


keileg:
No changes to functionality, so I'll merge.

PR #668, opened 2022-06-10
Speed improvements for MPFA

keileg:
## Proposed changes
This PR contains several improvements that speed up MPFA discretization (will have a positive effect on MPSA as well). There are no changes to code functionality. 

To summarize:
* Inversion of block matrices, at the core of MPXA discretizations, now uses parallelized numba. This leads to significant speedup of discretization.
* In Mpfa, slight reorganizations of the code (mainly, avoiding to compute the same matrix-matrix products over again) lead to further savings. NOTE: Similar changes have already been done in MPSA, but we might want to revisit that code and see if there are more easy pickings.
* In grid partitioning, a bottleneck in preprocessing before calling metis was essentially removed.
* Improved and modernized other parts of the code that gave minor performance enhancements.

To quantify the changes, anecdotally, I see a >50% reduction of discretization time on the 3d flow benchmark with a regular geometry on my system (this is likely dependent on a host of other factors, but the gains are real and substantial). My expectation is that quite a bit of this will carry over to MPSA.

## Types of changes

- [x] General maintenance / improvements of the code

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works : NOT RELEVANT

## Further comments
* A significant part of the remaining MPFA cost (~30%) relates to the vector source discretization. We should consider to make it possible to opt out of this.
* I also tested various ways of further speeding up the block inversion, but got no gains. It seems this code is as efficient as we can get it now, the main cost is in the actual inversion as implemented by lapack, which essentially is irreducible.
* Under testing, I observed that the block inversion can speed up significantly (again >50%, perhaps much more) by using numpy compiled with MKL, rather than the precompiled version from pip (conda install may be much better here, since it is already compiled with MKL).

### DIFF
--- a/src/porepy/grids/partition.py
+++ b/src/porepy/grids/partition.py
@@ -43,6 +43,8 @@ def partition_metis(g: pp.Grid, num_part: int) -> np.ndarray:
 
     # Connection map between cells
     c2c = g.cell_connection_map()
+    # Enforce csr. This should be the case anyhow, but better safe than sorry.
+    c2c = c2c.tocsr()
 
     # Convert the cells into the format required by pymetis
     indptr = c2c.indptr
--- a/src/porepy/grids/partition.py
+++ b/src/porepy/grids/partition.py
@@ -45,7 +45,10 @@ def partition_metis(g: pp.Grid, num_part: int) -> np.ndarray:
     c2c = g.cell_connection_map()
 
     # Convert the cells into the format required by pymetis
-    adjacency_list = [list(c2c.getrow(i).indices) for i in range(c2c.shape[0])]
+    indptr = c2c.indptr
+    adjacency_list = [
+        c2c.indices[indptr[i] : indptr[i + 1]].tolist() for i in range(c2c.shape[0])
+    ]
     # Call pymetis
     # It seems it is important that num_part is an int, not an int.
     part = pymetis.part_graph(int(num_part), adjacency=adjacency_list)
--- a/src/porepy/numerics/fv/fvutils.py
+++ b/src/porepy/numerics/fv/fvutils.py
@@ -847,7 +847,7 @@ class ExcludeBoundaries(object):
             [[0, 1, 0, 0],
               [0, 0, 0, 1]]
         """
-        col = np.argwhere([not it for it in ids])
+        col = np.argwhere(np.logical_not(ids))
         row = np.arange(col.size)
         return sps.coo_matrix(
             (np.ones(row.size, dtype=bool), (row, col.ravel("C"))),
--- a/src/porepy/numerics/fv/mpfa.py
+++ b/src/porepy/numerics/fv/mpfa.py
@@ -921,7 +921,10 @@ class Mpfa(pp.FVElliptic):
         # unknown in the discretized sytsem to the right hand side.
         # The negative in front of pr_trace_cell comes from the grad_eqs
         rhs_cells = -sps.vstack([nk_cell, -pr_trace_cell, pr_cont_cell])
-        flux = darcy * igrad * rhs_cells
+        # Compute matrix-matrix product, this is used in several steps below
+        darcy_igrad = darcy * igrad
+
+        flux = darcy_igrad * rhs_cells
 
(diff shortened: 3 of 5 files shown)


src/porepy/grids/partition.py
```diff
@@ -45,7 +45,10 @@ def partition_metis(g: pp.Grid, num_part: int) -> np.ndarray:
     c2c = g.cell_connection_map()
 
     # Convert the cells into the format required by pymetis
-    adjacency_list = [list(c2c.getrow(i).indices) for i in range(c2c.shape[0])]
+    indptr = c2c.indptr
```
IvarStefansson, line 48: Would it be wise to do a type check/conversion here to ensure no problems with csr vs csc?

src/porepy/numerics/fv/fvutils.py
```diff
@@ -847,7 +847,7 @@ def _exclude_matrix(self, ids):
             [[0, 1, 0, 0],
               [0, 0, 0, 1]]
         """
-        col = np.argwhere([not it for it in ids])
+        col = np.argwhere(np.logical_not(ids))
```
IvarStefansson, line 850: Good!

src/porepy/numerics/fv/mpfa.py
```diff
@@ -921,7 +921,10 @@ def _flux_discretization(
         # unknown in the discretized sytsem to the right hand side.
         # The negative in front of pr_trace_cell comes from the grad_eqs
         rhs_cells = -sps.vstack([nk_cell, -pr_trace_cell, pr_cont_cell])
-        flux = darcy * igrad * rhs_cells
+        # Compute matrix-matrix product, this is used in several steps below
+        darcy_igrad = darcy * igrad
+
+        flux = darcy_igrad * rhs_cells
```
IvarStefansson, line 927: Could this decomposition help us in differentiable tpfa? It seems that you've found an improved factorisation, which might be exactly what we want.

src/porepy/numerics/fv/mpfa.py
```diff
@@ -1013,9 +1016,11 @@ def _discretize_vector_source(
                 etc.
             bound_exclusion: Object that can eliminate faces related to boundary
                 conditions.
-            darcy: discretized Darcy's law
-            igrad: inverse gradient
-            nk_grad_all: nK products on a sub-cell level
+            darcy_igrad: Product of the discrete Darcy law with the basis functions
+                (igrad)
+            dist_igrad: Product of
+            nk_grad_all: Product between nK products on a sub-cell level and the basis
```
IvarStefansson, line 1022: Should we rename? nk_igrad or nk_igrad_all?

src/porepy/numerics/fv/mpfa.py
```diff
@@ -1013,9 +1016,11 @@ def _discretize_vector_source(
...
+                functions (igrad)
             nk_grad_paired: nK products after pairing
```
IvarStefansson, line 1024: Same as above

src/porepy/numerics/linalg/matrix_operations.py
```diff
@@ -608,48 +608,61 @@ def invert_diagonal_blocks_numba(a: sps.spmatrix, size: np.ndarray) -> np.ndarra
...
 
-            It may be possible to restructure the code to further help numba,
-            this has not been investigated.
+            IMPLEMENTATION NOTES BELOW
+
+            The code consists of a loop over the blocks. For each block, a local square
+            matrix is formed, the inverse is computed using numpy (which again will
+            invoke LAPACK), and the inverse is stored in an (raveled) array. The most
+            complex part of the code is the formation of the local matrix: Since the
+            original matrix is be sparse, there may be zero elements in the blocks
```
IvarStefansson, line 624:
```suggestion
            original matrix is sparse, there may be zero elements in the blocks
```

IvarStefansson (commented):
Looking forward to benefitting from this!

keileg:
I merge this after clarification with @IvarStefansson

PR #673, opened 2022-06-30
Overhaul mixed-dimensional grid, deprecate Assembler 

keileg:
# Overview
This PR introduces two major changes:
1. The `GridBucket` class has been renamed `MixedDimensionalGrid` and has undergone a substantial overhaul, including modifying names and functionality of many of its central methods.
2. The `Assembler` has been marked as deprecated, as has many of the functions and classes that are designed to make discretization methods adhere to the `Assembler`.

See below for a more detailed description of each of these changes.

In addition, maintenance has been preformed throughout the code, with increased used of typing and improved and updated documentation.

# The mixed-dimensional grid structure
Since it was introduced more than 5 years ago, the old `GridBucket` has undergone few major updates. The result was a class with high entropy, many methods that were used sparingly, and a prevalence of inconsistent and non-intuitive names (starting with the name `GridBucket` itself). The main changes can be summarized as follows:
1. The class `GridBucket` has been renamed `MixedDimensionalGrid.`
2. The previous terminology reflecting the  the graph representation of the mixed-dimensional grid has been replaced with concepts from domain decomposition. Specifically, what was previously referred to as `nodes` in the `GridBucket` has become subdomains in the `MixedDimensionalGrid`, and similarly, edges have become `interfaces`.
3. The `interfaces` are now represented by their `MortarGrid`, compared to the previous representation as a pair of `nodes`.
4. Many little used methods (like getters and setters) have been deleted.

### Subdomains
To iterate over subdomains (note that the old style `for g, d in gb` no longer works, removing the asymmetry between subdomains and interfaces):
```python
for sd in mdg.subdomains():
    # Get hold of the data dictionary for this subdomain
    data: dict = mdg.subdomain_data(sd)
```
Note that the standard notation for grids in such loops has changed; they are referred to as `sd` rather than `g`. Similarly, objects of the class `MixedDimensionalGrid` are denoted `mdg`, and the data dictionary is denoted `data` (making debugging with pdb a whole lot easier).

To iterate over subdomains and the associated data dictionary:
```python
for sd, data in mdg.subdomains(return_data=True):
    # do something
```
To iterate over subdomains of a specified dimension:
```python
for sd in mdg.subdomains(dim=2):
    # do something
```
The method `subdomains()` returns a list, rather than being a generator as the old method `nodes()`. Thereby, the behavior of the  deleted method `grids_of_dimension()` is replicated by simply writing
```python
sd_list: list[pp.Grid] = mdg.subdomains(2)
```

### Interfaces
Interfaces are accessed via the method `mdg.interfaces()`, which has the same options as `mdg.subdomains()`. For instance:
```python
for intf, data in mdg.interfaces(return_data=True):
    # Note that intf is a pp.MortarGrid.
    # Get hold of the neighboring subdomains:
    sd_primary, sd_secondary = mdg.interface_to_subdomain_pair(intf)
```
The method `mdg.interface_to_subdomain_pair()` is designed so that the first subdomain in the pair has the higher dimension (this is different from the previous representation of edges).

To get the `MortarGrid` representing an interface from a subdomain pair, use
```python
intf: pp.MortarGrid = mdg.subdomain_pair_to_interface((sd_primary, sd_secondary))
```

For more information on the new `MixedDimensionalGrid`, see the (much improved) documentation of that class.

# Removal of the Assembler
While the `Assembler` class was an important steps for making PorePy capable of solving general mixed-dimensional multiphysics problems, it also suffers from several severe issues that have become aparent over the last few years:
1. The way equations must be specified is cumbersome, and leads to a lack of transparency that makes debugging challenging, in particular for complex problems (see `thm_model` ). Non-linearities makes the problem even worse.
2. For a discretization to comply with the `Assembler` paradigm, it must implement several specialized methods (see subclasses of `EllipticDiscretization`, methods `assemble_pressure_int_bound...` etc.), and in some cases dedicated classes to couple discretizations between dimensions (see the classes in the module `contact_mechanics_interface_laws`).
3. The `Assembler's` performance scales amazingly poorly for problems with many subdomains. This boils down to for-loops being slow. 
4. The implementation is extremely complex and difficult to maintain. This applies both to the `Assembler` itself, and the required extra methods and classes for coupling discretizations between dimensions.

For these reasons, it has been decided to remove the `Assembler` and the classes and methods designed to adhere to the `Assembler` paradigm. This will remove litterally thousands of lines of code that are hard to maintain, and also a large number of tests, The change is scheduled for the fall 2022.   

The recommended replacement for the `Assembler` is preferrably to use some of the model classes if these cover or can be adapted to the governing equations at hand. Alternatively, the algorithmic differentiation framework can be used to define equations in a much more intuitive way than the `Assembler.`


IvarStefansson (approved):
Credit for this effort goes to @keileg, @OmarDuran, @jwboth and myself. Great work!

PR #683, opened 2022-07-27
Add _storativity and update _volume_change methods in ContactMechanicsBiot

jhabriel:
## Proposed changes

This PR introduces the method `_storativity(self, sd: pp.Grid) -> np.ndarray` to account for compressibility effects in the `ContactMechanicsBiot` model class.  This is done in a similar fashion as in the `SlightlyCompressibleFlow` model class. In the _volume_change method, a bug was identified. The fix is achieved by using the _displacement_jump method instead of constructing the jumps locally.

## Types of changes

## Checklist


src/porepy/models/contact_mechanics_biot_model.py
```diff
@@ -211,8 +211,8 @@ def _set_mechanics_parameters(self) -> None:
 
     def _set_scalar_parameters(self) -> None:
         tensor_scale = self.scalar_scale / self.length_scale**2
-        mass_weight = 1 * self.scalar_scale
         for sd, data in self.mdg.subdomains(return_data=True):
+            mass_weight = self._compressibility(sd) * self.scalar_scale
```
IvarStefansson, line 215: Shouldn't this be the biot modulus?

src/porepy/models/contact_mechanics_biot_model.py
```diff
@@ -412,7 +412,35 @@ def _reference_scalar(self, sd: pp.Grid) -> np.ndarray:
...
+        Args:
+            sd: Subdomain grid.
+
+        Returns:
+            Unitary compressibility. If AD is used, an np.ndarray of ones of shape
+                (sd.num_cells, ) is returned. Otherwise, 1.0 is returned.
+
+        """
+
+        if self._use_ad:
```
IvarStefansson, line 427: I don't think this is needed. The biot coefficient is a special case for some annoying reason which I don't remember atm. It should be safe to return the array both with and without ad.

src/porepy/models/contact_mechanics_biot_model.py
jhabriel: I don't think so. This is the factor multiplying $\frac{\partial p}{\partial t}$, or after time discretization, the one multiplying $\left(p^{n} - p^{n-1}\right)$.

This is used in: https://github.com/pmgbergen/porepy/blob/5ba4ce0fa93d33c6cad332663c628db1e249c1a7/src/porepy/models/contact_mechanics_biot_model.py#L845

src/porepy/models/contact_mechanics_biot_model.py
jhabriel: Ok. I will refactor this to return only an `np.ndarray`.

src/porepy/models/contact_mechanics_biot_model.py
jhabriel: Method was renamed to `_storativity` instead of `_compressibility`.

jhabriel:
This PR is ready (at least from my side). Feel free to review it when you can @IvarStefansson @keileg .

src/porepy/models/contact_mechanics_biot_model.py
```diff
@@ -412,7 +413,31 @@ def _reference_scalar(self, sd: pp.Grid) -> np.ndarray:
         """
         return np.zeros(sd.num_cells)
 
+    def _storativity(self, sd: pp.Grid) -> np.ndarray:
+        """Set unitary storativity (also called Biot modulus or storage coefficient).
```
IvarStefansson, line 417:
```suggestion
        """Set unitary storativity.
        
         The storativity is also called Biot modulus or storage coefficient.
```

tests/unit/test_model_setup.py
```diff
@@ -0,0 +1,918 @@
...
+For each model, two main methods are constructed: one that checks default parameters and
+another that checks non-default parameters. For the latter, three different sets of
+non-default parameters for a mixed-dimensional Cartesian grid containing a single vertical
+fracture have been written.
+
+Moreover, for each model, the parameters are checked at two stages, namely before running
+the model and after running the model.
+
+Note that only parameters belonging to the specific class are tested. This means that
+inhereted parameters are not tested as this would be too cumbersome (and redundant).
```
IvarStefansson, line 14:
```suggestion
inherited parameters are not tested as this would be too cumbersome (and redundant).
```

tests/unit/test_model_setup.py
```diff
@@ -0,0 +1,918 @@
+"""
```
IvarStefansson, line 1: Excellent summary!

tests/unit/test_model_setup.py
```diff
@@ -0,0 +1,918 @@
...
+another that checks non-default parameters. For the latter, three different sets of
+non-default parameters for a mixed-dimensional Cartesian grid containing a single vertical
+fracture have been written.
+
+Moreover, for each model, the parameters are checked at two stages, namely before running
+the model and after running the model.
+
+Note that only parameters belonging to the specific class are tested. This means that
+inhereted parameters are not tested as this would be too cumbersome (and redundant).
+However, this naturally requires writing tests for all present models.ß
```
IvarStefansson, line 15:
```suggestion
However, this naturally requires writing tests for all present models.
```

tests/unit/test_model_setup.py
```diff
@@ -0,0 +1,918 @@
...
+
+import porepy as pp
+import pytest
+import numpy as np
+
+from typing import Union
+
+
+class TestContactMechanicsBiot:
+    """The following tests check the correct setup of the ContactMechanicsBiot classß."""
```
IvarStefansson, line 29:
```suggestion
    """The following tests check the correct setup of the ContactMechanicsBiot class."""
```

tests/unit/test_model_setup.py
```diff
@@ -0,0 +1,918 @@
...
+    # ----------> Setting up parameters corresponding to the test of non-default parameters
+
+    # -----> Mixed-dimensional grid
+    mdg, box = pp.md_grids_2d.single_vertical(mesh_args=np.array([2, 1]), simplex=False)
+    sd_prim = mdg.subdomains(dim=2)[0]
+    data_prim = mdg.subdomain_data(sd_prim)
+    sd_sec = mdg.subdomains(dim=1)[0]
+    data_sec = mdg.subdomain_data(sd_sec)
+    intf = mdg.interfaces(dim=1)[0]
+    data_inf = mdg.interface_data(intf)
```
IvarStefansson, line 138: Interface abbreviation should be intf, not inf

tests/unit/test_model_setup.py
```diff
@@ -0,0 +1,918 @@
...
+
+    # -----> Time parameters
+    time_steps = [2.0, 5.0, 100.0]
+    end_times = [2.0, 10.0, 300.0]
+
+    # -----> Scalar bc types
+    all_bf = sd_prim.get_boundary_faces()
+    tol = 1e-10
+    east = sd_prim.face_centers[0] > box["xmax"] - tol
```
IvarStefansson, line 147: These can be obtained by self._domain_boundary_sides

tests/unit/test_model_setup.py
```diff
@@ -0,0 +1,918 @@
...
+
+    bc_mech_type_primary = [bc_type_prim_0, bc_type_prim_1, bc_type_prim_2]
+
+    # -----> Mechanics bc values
+    ones = np.ones(sd_prim.num_faces)
+
+    # Case 0
+    bc_val_prim_0 = np.array([0 * ones, 0 * ones])
+    bc_val_prim_0[0, east] = 0.01
+    bc_val_prim_0[1, east] = 0
```
IvarStefansson, line 329: Explicit setting of zeros is redundant

tests/unit/test_model_setup.py
```diff
@@ -0,0 +1,918 @@
...
+                "bc_mech_values_primary": bc_mech_values_primary[case],
+            }
+            model_params.append(model_param)
+
+    @pytest.mark.parametrize("model_params", model_params)
+    def test_non_default_parameters_contact_mechanics_biot(self, model_params):
+        """Test model with non-default scalar sources"""
+
+        # Construct the class
+        class NonDefaultContactMechanicsBiot(pp.ContactMechanicsBiot):
```
IvarStefansson, line 410: Wouldn't it be cleaner to define the class outside the test?

tests/unit/test_model_setup.py
```diff
@@ -0,0 +1,918 @@
...
+
+            def create_grid(self) -> None:
+                """Modify default grid.
+
+                Create a Cartesian grid in a unit square with two cells in the horizontal
+                direction and one cell in the vertical direction. A single vertical fracture
+                is included in the middle of the domain.
+
+                """
+                self.mdg = self.params["mdg"]
```
IvarStefansson, line 424: Where is the mdg created?

tests/unit/test_model_setup.py
jhabriel: Yes. Initially, I tried to do what you suggest, i.e., define the `NoDefaultContactMechanicsBiot` class outside the test and parametrize the models instead of the models' parameters. 

To my surprise, when I tried to do this, all models had the same parameters inside their respective data dictionaries. AFAIK, the reason for this behavior is that  when several `NoDefaultContactMechanicsBiot` instances are created outside the test (as it is done by the `pytest` parametrize decorator), the `data` dictionaries mutate along process, since they all point to the same class.

tests/unit/test_model_setup.py
jhabriel: Yes, but at this point, I don't have access to the model object yet. That's why I don't use the `self._domain_boundary_sides()` method.

tests/unit/test_model_setup.py
jhabriel: `mdg` is created outside the test and then passed as a model parameter to the `NonDefaultContactMechanicsBiot` class. This is to avoid having to create the same `mdg` every time the class is instantiated.

src/porepy/models/contact_mechanics_model.py
```diff
@@ -405,7 +416,7 @@ def prepare_simulation(self) -> None:
             self.assembler: pp.Assembler = pp.Assembler(self.mdg, self.dof_manager)
 
         self._discretize()
-        self._initialize_linear_solver()
+        self._initialize_linear_solver()  # TODO: Purge method from standard models?
```
keileg, line 419:
```suggestion
        self._initialize_linear_solver()
```
My expectation is that we will finally make progress on this part in the coming months, so I suggest we keep the method, and hopefully see it grow.

tests/unit/test_model_setup.py
```diff
@@ -0,0 +1,904 @@
...
+"""
+
+import porepy as pp
+import pytest
+import numpy as np
+
+from typing import Union
+
+
+class TestContactMechanicsBiot:
```
keileg, line 28: I really appreciate the effort to test the models, but I don't think the approach taken here is a good one: The code is so long that it begs for a better solution, in particular considering that this is actually not our complex model.

I have some (vague) ideas on how this can be realized in a better manner; I suggest we discuss this @jhabriel

jhabriel: Point taken. I also underestimated how long turned out to be to test the _non-default_ parameters of a model. It becomes ugly very fast.

IvarStefansson: Should we remove the new test and try to merge this branch?

jhabriel: I'm OK with this. Let me create an issue so we don't forget to incorporate the test in the (near) future.

jhabriel: Are you OK with merging this PR without the tests for the models @keileg? I've created a separate issue for this, see #685.

keileg (approved):

src/porepy/models/contact_mechanics_model.py
IvarStefansson: In that case, should we move to AbstractModel?

PR #698, opened 2022-08-23
Update of test-vtk

jwboth:
## Proposed changes

The following issues are identified with test-vtk, which are aimed at to be improved (not all of them for now):
- make folder `test_vtk` temporal
- do not use operating system dependent checks
- do not use string comparison to evaluate success of the exporter

The following approach is suggested, only applied to a single test in `test-vtk.py` for now:
1. Move vtu file to external folder `test-vtk-reference`.
2. Use meshio and deepdiff to compare test and reference vtu files.

## Types of changes

- [ ] Other: 

## Checklist

||Yes|No|NA|
|---|---|---|---|
|The update is covered by the  test suite |_|_|X|
|The documentation is up-to-date |_|X|_|
|I have not duplicated existing functionality |X|_|_|

## Further comments

Issues, which require discussion: 
1. Using deepdiff means adding an extra dependency. This is the first time, I am using deepdiff - it allows for recursive comparison of dictionaries, containing all sorts of types. It can ignore keys, ignore typing (int32 vs int64) and much more.
2. Is this approach after all not too depending on meshio doing what we expect it would do?
3. In the current update, the exporter actually exports in binary format, while the reference files are stored in ascii format. The latter could be changed now.


IvarStefansson (commented):
My initial impression is that this is a very good solution and that it is worth the extra dependency. Also, thanks for a good PR description! I took the liberty of converting to draft PR since you said it's still WIP.

keileg:
I agree with @IvarStefansson that both the code and the PR look very good. I have no problems with the extra dependency, indeed, there must be other tests beyond VTK that would benefit from a similar approach. If we can test the binary functionality, that would be great, as you say, this is what we really need.

As for our reliance on meshio, this is pretty heavy by now. On the other hand, meshio has an extensive test suite, and although the API has been prone to abrubt changes in the past, the implementation seems to have stabilized by now. Also, even through these previous changes, the underlying functionality has stayed fixed or expanded. All this considered, I am comfortable with trusting meshio to do its things, and let our tests verify only that the meshio-PorePy interface works as intended. If that makes any sense.

To conclude: Thumbs up from me, good work!

jwboth:
Important "milestone": The procedure of how to compare the binary output of the exporter with the original ascii vtk outputs has been applied to all tests in `test_vtk.py`. All tests are successful. The next step is to replace all reference vtk files and store them as binary and not ascii. In this step, the test is not adapted, only the reference files. By this two-step approach and intermediate build check, the changes can be (hopefully) trusted.

jwboth:
The proposed changes have been applied now.

Potentially missing: 
+ More documentation; however, at the moment I find the names of the tests and the few routines descriptive enough. Any other opinions?
+ Removing redundant tests - not too critical at the moment.

jwboth:
The latest update utilizes the almost identical structure of the previous tests `test_single_subdomain*`. Using some sort of parametrization, one can gather these tests in one. Will this make trouble if once a tests fails, and the test has to be debugged?

IvarStefansson:
> The latest update utilizes the almost identical structure of the previous tests `test_single_subdomain*`. Using some sort of parametrization, one can gather these tests in one. Will this make trouble if once a tests fails, and the test has to be debugged?

No, this is exactly the way to go. It should be possible to identify which parameters causes the failure.

jwboth:
> > The latest update utilizes the almost identical structure of the previous tests `test_single_subdomain*`. Using some sort of parametrization, one can gather these tests in one. Will this make trouble if once a tests fails, and the test has to be debugged?
> 
> No, this is exactly the way to go. It should be possible to identify which parameters causes the failure.

It is not clear to me whether you had a look at the actual implementation of the latest commit. I failed to use `pytest.mark.parametrize` for this. Would be happy to use it though, as I believe the current implementation would not allow for that.

jwboth:
> > The latest update utilizes the almost identical structure of the previous tests `test_single_subdomain*`. Using some sort of parametrization, one can gather these tests in one. Will this make trouble if once a tests fails, and the test has to be debugged?
> 
> No, this is exactly the way to go. It should be possible to identify which parameters causes the failure.

Now `test_vtk` is converted from unittest to pytest, and the parametrization functionality of pytest is used to follow the idea as you problably have intended.

IvarStefansson:
Is the PR finished from your side, @jwboth?

jwboth:
> Is the PR finished from your side, @jwboth?

I would say yes. The transition to use parametrization of `test_single_subdomains` unfortunately resulted in a less tidy organization of the document, in my opinion. Whoever reviews, please be critical regarding that as well. I tried to keep some of the previous style, but maybe it is simpler to not have a ExporterTestSetup but store all paths as `folder_reference`.

tests/unit/test_vtk.py
```diff
@@ -1,9112 +1,409 @@
...
+    save.write_vtu([("dummy_scalar", dummy_scalar), ("dummy_vector", dummy_vector)])
+
+    # Check that exported vtu file and reference file are the same
+    assert _compare_vtu_files(
+        f"{setup.folder}/{setup.file_name}_{sd.dim}.vtu",
+        f"{subdomain.ref_vtu_file}",
+    )
+
+
+# NOTE: Suggest removing this test.
```
keileg, line 174: Is there a specific reason you want to remove this?

tests/unit/test_vtk.py
```diff
@@ -1,9112 +1,409 @@
...
+    if not os.path.exists(setup.folder):
+        os.makedirs(setup.folder)
+
+    # Export data
+    network_2d.to_file(
+        setup.folder + setup.file_name + ".vtu",
+        data=data,
+    )
+
+    # Check that exported vtu file and reference file are the same
```
keileg, line 371: So, this test tacitly assumes that Gmsh returns the same grid every time? If so, I think we should do ourselves a favor and add a comment that if this test breaks, but not those above, one should rerun the test with an earlier version of gmsh and, if that works, the change is quite likely due to external factors. Or something like that.

tests/unit/test_vtk.py
```diff
@@ -1,9112 +1,409 @@
...
+    if not os.path.exists(setup.folder):
+        os.makedirs(setup.folder)
+
+    # Export data
+    network_3d.to_file(
+        setup.folder + setup.file_name + ".vtu",
+        data=data,
+    )
+
+    # Check that exported vtu file and reference file are the same
```
keileg, line 405: Same comment as above regarding documentincg the gmsh dependency.

keileg (commented):
I have left a few comments, but mostly I'm very happy with this. I have no problems with the setup to parametrization being a bit messy - to me it gives a separation between setup and actual testing. This may be an aquired taste.

tests/unit/test_vtk.py
jwboth: I was only thinking, that `test_mdg_2` covers a more complex test compared to `test_mdg_1`. The latter has a slightly different mesh setup (two intersecting fractures instead of a single one), and it only tests the export of subdomain data, and not interface data. If assuming that `test_mdg_*` merely tests mdg-specific functionality of Exporter, `test_mdg_1` can be seen as an underset of `test_mdg_2`.

I have not problem with keeping both. I just thought, incase we want to remove (slightly) redundayt tests, I make a suggestion.

tests/unit/test_vtk.py
```diff
@@ -1,9112 +1,409 @@
...
+
+    # Setup
+    setup = ExporterTestSetup()
+    yield setup
+
+    # Teardown
+    shutil.rmtree(
+        os.getcwd() + "/" + os.path.basename(os.path.normpath(setup.folder)),
+        ignore_errors=True,
+    )
```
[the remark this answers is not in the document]
keileg: The point is that with Path objects / is interpreted independent of the operating system.

tests/unit/test_vtk.py
jwboth: For all instances, it is updated. Please check.

tests/unit/test_vtk.py
jwboth: See above.

tests/unit/test_vtk.py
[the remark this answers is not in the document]
jwboth: On linux this works (modulo some simple syntax changes). @IvarStefansson could you check whether this raises trouble on Windows?

tests/unit/test_vtk.py
keileg: I think you are right, let's delete it. In some cases it is beneficial to have such subsets since taking the differential between the super and subset gives information of what fails, but it is hard to imagine such a case here. So I agree, let's reduce the number of tests.

Feel free to resolve the conversation when the test is deleted.

tests/unit/test_vtk.py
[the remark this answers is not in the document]
keileg: I'm pretty sure it will not work because windows wants backslash instead of slash. That is the point of pathlib.Path, it takes care of OS-specific cases, and will for instance translate the / (note: not a string) to the right separator.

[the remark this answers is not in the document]
jwboth: My mistake. "this" meant the proposed changes by @keileg and the original. See line 52.

tests/unit/test_vtk.py
```diff
@@ -1,9112 +1,390 @@
...
+    # Teardown: remove temporary directory for vtu files
+    full_path = Path.cwd() / Path.resolve(Path(setup.folder)).name
+    shutil.rmtree(full_path)
+
+
+def _compare_vtu_files(test_file: str, reference_file: str) -> bool:
+    # Helper method to determine whether two vtu files, accessed by their
+    # paths, are identical. Returns True if both files are identified as the
+    # same, False otherwise. This is the main auxiliary routine used to compare
+    # down below wheter the Exporter produces identical outputs as stored
```
IvarStefansson, line 60:
```suggestion
    # down below whether the Exporter produces identical outputs as stored
```

tests/unit/test_vtk.py
```diff
@@ -1,9112 +1,390 @@
...
+    # reference files.
+
+    # Trust meshio to read the vtu files
+    test_data = meshio.read(test_file)
+    reference_data = meshio.read(reference_file)
+
+    # Determine the difference between the two meshio objects.
+    # Ignore differences in the data type if values close.
```
IvarStefansson, line 68:
```suggestion
    # Ignore differences in the data type if values are close.
```

tests/unit/test_vtk.py
```diff
@@ -1,9112 +1,390 @@
...
+    save.write_vtu([("dummy_scalar", dummy_scalar), ("dummy_vector", dummy_vector)])
+
+    # Check that exported vtu file and reference file are the same
+    assert _compare_vtu_files(
+        f"{setup.folder}/{setup.file_name}_{sd.dim}.vtu",
+        f"{subdomain.ref_vtu_file}",
+    )
+
+
+def test_mdg_2(setup):
```
IvarStefansson, line 172: Maybe change name to test_mdg_2d or something else. The "2" is currently ambiguous

tests/unit/test_vtk.py
```diff
@@ -1,9112 +1,390 @@
...
+    # rerun with an older version of Gmsh to test for failure due to
+    # external reasons.
+    for appendix in ["2", "constant_2"]:
+        assert _compare_vtu_files(
+            f"{setup.folder}/{setup.file_name}_{appendix}.vtu",
+            f"{setup.folder_reference}/constant_data_test_grid_{appendix}.vtu",
+        )
+
+
+def test_fractures_2d(setup):
```
IvarStefansson, line 320:
```suggestion
def test_fracture_network_2d(setup):
```

tests/unit/test_vtk.py
```diff
@@ -1,9112 +1,390 @@
...
+    # for the reference grid; thus, if this test fails, it should be
+    # rerun with an older version of Gmsh to test for failure due to
+    # external reasons.
+    assert _compare_vtu_files(
+        f"{setup.folder}/{setup.file_name}.vtu",
+        f"{setup.folder_reference}/fractures_2d.vtu",
+    )
+
+
+def test_fractures_3d(setup):
```
IvarStefansson, line 355:
```suggestion
def test_fracture_network_3d(setup):
```

tests/unit/test_vtk.py
```diff
@@ -1,9112 +1,390 @@
...
+        os.makedirs(setup.folder)
+
+    # Export data
+    network_3d.to_file(
+        setup.folder + setup.file_name + ".vtu",
+        data=data,
+    )
+
+    # Check that exported vtu file and reference file are the same
+    # NOTE: It is implicitly assumed that Gmsh returns the same grid as
```
IvarStefansson, line 383: Perhaps move this NOTE inside _compare_vtu_files instead of repeating it each time the function is called?

tests/unit/test_vtk_reference/constant_data_test_grid_constant_2.vtu
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0"?>
```
IvarStefansson, line 1: It's surely not important, but I don't understand the name of this file.

tests/unit/test_vtk.py
keileg: I think I disagree with this - I like having such messages to the future self right where they will be needed, in this case next to the place the test will fail.

tests/unit/test_vtk_reference/constant_data_test_grid_constant_2.vtu
IvarStefansson: I'm perfectly fine with leaving the name as is!

tests/unit/test_vtk.py
IvarStefansson: I thought about that (and half expected a disagreement). But isn't the obvious place to check if that assertion breaks inside the function being called? I mean, it shouldn't be difficult to find the note inside the function in this case. Just a suggestion, though, no strong opinions

tests/unit/test_vtk.py
jwboth: Renamed both test_mdg_2 (dropped the 2). And test_mdg_3 (tried to give a more descriptive name).

tests/unit/test_vtk.py
keileg: Maybe you are right, but if we both agree, I suggest we leave it to @jwboth to decide.

IvarStefansson: Sure!

jwboth: No super strong feeling. But in the end, I actually also lean towards mentioning it only once in the auxiliary method.

IvarStefansson (approved):

keileg (approved):

PR #706, opened 2022-09-05
Ad geometry

IvarStefansson:
## Proposed changes

Instead of having to wrap all current discretizations into ad classes, the proposed idea is to provide ad.Geometry as a wrapper of (Mortar)Grids giving access to cell volumes, face centers etc. Thus, simple discretizations s.a. MassMatrix may be replaced, i.e.

`Geometry.cell_volumes * compressibility * pressure`
instead of
`MassMatrix.mass * pressure`

Complex discretization s.a. mpXa will remain.

Testing of grid_operators has been extended and refactored.

## Types of changes

- [ ] Other: 

## Checklist

||Yes|No|NA|
|---|---|---|---|
|The update is covered by the test suite |&check;|_|_|
|The documentation is up-to-date |&check;|_|_|


src/porepy/numerics/ad/grid_operators.py
```diff
@@ -47,6 +50,9 @@ def __init__(
             subdomains (List of pp.Grid): List of subdomains. The order of the subdomains in
                 the list will establish the ordering of the subdomain projections.
             nd (int, optional): Dimension of the quantities to be projected.
+            FIXME: Is dim a better argument and attribute name? It's the dimension of the
```
keileg, line 53: I agree, dim is better.

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -557,6 +563,86 @@ def __str__(self) -> str:
         return s
 
 
+class Geometry(Operator):
+    """Wrapper class for Ad representations of grids.
+
+
```
keileg, line 569:
```suggestion
```

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -557,6 +563,86 @@ def __str__(self) -> str:
...
+    Attributes:
+        cell_volumes (pp.ad.Matrix): Diagonal ad matrix of cell volumes.
+        face_areas (pp.ad.Matrix):  Diagonal ad matrix of face areas.
+        nd (int): Ambient/highest dimension of the mixed-dimensional grid.
+
+    FIXME: Implement parse??
+    """
+
+    def __init__(self, subdomains: List[pp.Grid], nd: int, name: Optional[str] = None):
```
keileg, line 578:
```suggestion
    def __init__(self, subdomains: list[pp.Grid], nd: int, name: Optional[str] = None):
```

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -557,6 +563,86 @@ def __str__(self) -> str:
...
+            """Expand matrix of size [N x M] to [nd*N x M].
+
+            When left multiplied to a matrix A, each row of A is
+            repeated nd times.
+
+            Usage example: Scaling from traction to force is
+
+                    force = (scalar_to_nd_face * face_areas) * traction
+
+            FIXME: Refactor?
```
keileg, line 618: Come to think of it, I believe this is very similar to what we do here https://github.com/pmgbergen/porepy/blob/83c2e4af5094cf3e445d2e3db2274493a1a3c14e/src/porepy/numerics/fv/fvutils.py#L433 (or perhaps in the function below that one), although the present function of course has an Ad wrapping.

Suggested solution: Keep this for now, revisit at a later point (while refactoring parts of fvutils into array_operations).

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -557,6 +563,86 @@ def __str__(self) -> str:
...
+                shape=(size * self.nd, size),
+            )
+            return pp.ad.Matrix(mat)
+
+        self.scalar_to_nd_cell = scalar_to_nd(self.num_cells)
+        self.scalar_to_nd_face = scalar_to_nd(self.num_faces)
+
+    def __repr__(self) -> str:
+        s = (
+            f"Geometry operator for {self._num_grids} grids.\n"
```
keileg, line 634: I'm not sure 'Geometry operator' is the best word for this, but I have no better suggestions. Perhaps it is mainly a question of getting used to it.

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -557,6 +563,86 @@ def __str__(self) -> str:
...
+            Usage example: Scaling from traction to force is
+
+                    force = (scalar_to_nd_face * face_areas) * traction
+
+            FIXME: Refactor?
+            """
+            rows: np.ndarray = np.arange(size * self.nd)
+            cols: np.ndarray = np.kron(np.arange(size), np.ones(self.nd))
+            data: np.ndarray = np.ones(size * self.nd)
+            mat = sps.csr_matrix(
```
keileg, line 623: I think this should be csc (there are more rows than columns).

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -875,7 +960,7 @@ def parse(self, mdg: pp.MixedDimensionalGrid) -> np.ndarray:
             data = mdg.interface_data(intf)
             val.append(data[pp.PARAMETERS][self.param_keyword][self.array_keyword])
         if len(val) > 0:
-            return sps.diags(np.hstack([v for v in val]))
+            return sps.diags(np.hstack([v for v in val]), format="csr")
```
keileg, line 963: It may not matter here, but I think `diag` matrices are considerably faster to operate with than `csr,` even if the csc represents a diagonal matrix.

keileg (commented):
Mainly minor comments.

src/porepy/numerics/ad/grid_operators.py
IvarStefansson: Do you mean a different format with the same constructur method or a different constructor?

keileg (approved):

Issue #709, opened 2022-09-08
Docker image cannot be build since `black==22.8.0` is not yet available through conda-forge


jhabriel:
The docker image porepy/dev cannot currently be build since `black==22.8.0` is not available through conda-forge. The Build Test, on the other hand passes, since it relies on pip not on conda-forge.

PR #710, opened 2022-09-09
BLD: separate reqs and reqs-dev

OmarDuran:
## Proposed changes

By separating production and development requirements, this PR avoids duplicate entries in those files. As a result, the workflow "ci.yml" checks requirements.txt and requirements-dev.txt separately.

## Types of changes

- [x] Other: Minor refactor on PorePy building

## Checklist

||Yes|No|NA|
|---|---|---|---|
|The update is covered by the test suite |&check;|_|_|
|The documentation is up-to-date |_|_|&check;|

## Further comments

Pip installation is performed as in https://github.com/pmgbergen/porepy/blob/develop/Readme.md

Conda installation can be performed for:
Production: conda install --file requirements.txt 
Development: conda install --file requirements.txt  --file requirements-dev.txt

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,7 @@ jobs:
         pip install -U pip
         # Install requirements, using eager updates to avoid stalled dependencies due to caching
         # https://blog.allenai.org/python-caching-in-github-actions-e9452698e98d
+        pip install --upgrade --upgrade-strategy eager -r requirements.txt
         pip install --upgrade --upgrade-strategy eager -r requirements-dev.txt
         pip install .
         pip freeze
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,15 +69,21 @@ jobs:
         sudo apt-get install libglu1-mesa libgeos-dev
         export LD_LIBRARY_PATH=/usr/local/lib64/:$LD_LIBRARY_PATH
 
-    - name: Install PorePy
+    - name: Install requirements
       run: |
         pip install -U pip
         # Install requirements, using eager updates to avoid stalled dependencies due to caching
         # https://blog.allenai.org/python-caching-in-github-actions-e9452698e98d
         pip install --upgrade --upgrade-strategy eager -r requirements.txt
-        pip install --upgrade --upgrade-strategy eager -r requirements-dev.txt
+
+    - name: Install requirements-dev
+      run: |
+        pip install --upgrade --upgrade-strategy eager -r requirements-dev.txt        
+
+    - name: Install PorePy
+      run: |
         pip install .
-        pip freeze
+        pip freeze       
 
     - name: black
       if: ${{ always() }}
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,22 +1,8 @@
-meshio >= 5.0
-networkx
-numpy
-scipy
-sympy
-matplotlib
-future
-six
-shapely
-shapely[vectorized]
+deepdiff
 pytest >= 4.6
 pytest-cov
 pytest-runner
-numba >= 0.53.0, < 0.56.1
-gmsh
 flake8
 mypy
-black == 22.8.0
+black
 isort
-typing_extensions
-requests
-deepdiff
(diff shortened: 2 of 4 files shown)


requirements-dev.txt
```diff
@@ -1,22 +1,8 @@
...
-shapely[vectorized]
+deepdiff
 pytest >= 4.6
 pytest-cov
 pytest-runner
-numba >= 0.53.0, < 0.56.1
-gmsh
 flake8
 mypy
 black == 22.8.0
```
jhabriel, line 7:
```suggestion
black == 22.*
```

jhabriel: Starting from this year, all versions of black are stable (see https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy). In short, this means that any version with a tag starting with `22` should result in the same formatted code. This motivates passing as a requirement, the less restrictive `22.*` instead of freezing a particular `22.x.x` version. 

As a bonus, this also solves  #709. Since a conda installation will currently use `22.6.0` whereas a pip installation will use `22.8.0`, although both installations will have black behaving exactly the same.

Next year we can adopt `black=23.*`, but dependabot should let us know about this automatically.

What do you think @keileg?

.github/workflows/ci.yml
```diff
@@ -74,6 +74,7 @@ jobs:
         pip install -U pip
         # Install requirements, using eager updates to avoid stalled dependencies due to caching
         # https://blog.allenai.org/python-caching-in-github-actions-e9452698e98d
+        pip install --upgrade --upgrade-strategy eager -r requirements.txt
```
jhabriel, line 77: We might want to split the `Install Porepy` stage into three parts: (1) one where the porepy dependencies from `requirements.txt` are installed, (2) another where the porepy dependencies from `requirements_dev.txt` are installed, and (3) another where porepy is actually installed.

This will allow us to isolate build failures which are strictly related to those steps, i.e., install of the requirements, install of the dev requirements, and install of porepy itself.

What are your thoughts about this suggestion @OmarDuran @keileg?

jhabriel (changes requested):
Excellent initiative @OmarDuran. 

I suggested some changes that I think will make the installation process a bit more robust. However, you should wait @keileg to read my suggestions before you incorporate any changes in your PR.

When we settled on the final versions of `requirements.txt` and `requirements_dev.txt` in this PR, I suggest we sort the dependencies in alphabetical order. This of course non-critical at all, but it greatly helps when inspecting the dependencies; for example, it should make very hard to introduce repeated dependencies.

As a final comment, I'm almost certain that if this PR is approved, the docker build of porepy/base will brake. We should check this before merging. I understand that we can trigger build actions also manually.

.github/workflows/ci.yml
keileg: @jhabriel: How is that different from lines 77-79 now?

requirements-dev.txt
keileg: I agree, that should work.

.github/workflows/ci.yml
jhabriel: The  only difference is that Actions will treat the 3 processes separately. So, if there is an issue installing some package from `requirements.txt`, it will show that this action failed (rather than the whole porepy installation), which IMO will allow us to isolate potential dependencies issues more easily.

In practice is no different from what we have now in lines 77-79.

requirements-dev.txt
OmarDuran: I support adding more flexibility to formatters by adding black=23.*. It strikes a balance between strong and weak formatting constraints.

.github/workflows/ci.yml
keileg: Ah, I see the point. Fair enough, let's split it.

OmarDuran: The installation has been split PorePy installation into three actions.

OmarDuran:
@jhabriel I have tested building the base docker image and no issues were found.

jhabriel (approved):

keileg (approved):

PR #716, opened 2022-09-15
FEAT: Split assemble_and_solve_linear_system

IvarStefansson:
## Proposed changes

I have split assemble_and_solve_linear_system into assemble_linear_system and linear_solve. The system is stored as a model attribute as `self.linear_system = (A, b)` in the former method. I removed the unused tolerance parameter and suggest to use something like `self.params["linear_solve_tolerance"]` if the need arises.

Solves #715 

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other: 

## Checklist

||Yes|No|NA|
|---|---|---|---|
|The update is covered by the  test suite |&check;|_|_|
|The documentation is up-to-date |&check;|_|_|

### DIFF
--- a/src/porepy/models/abstract_model.py
+++ b/src/porepy/models/abstract_model.py
@@ -235,14 +235,10 @@ class AbstractModel:
         if solver not in ["scipy_sparse", "pypardiso", "umfpack"]:
             raise ValueError(f"Unknown linear solver {solver}")
 
-    def assemble_linear_system(self):
+    def assemble_linear_system(self) -> None:
         """Assemble the linearized system.
 
         The linear system is defined by the current state of the model.
-
-
-
-
 
         Attributes:
             linear_system is assigned.
@@ -256,7 +252,7 @@ class AbstractModel:
         self.linear_system = (A, b)
         logger.debug(f"Assembled linear system in {t_0-time.time():.2e} seconds.")
 
-    def linear_solve(self) -> np.ndarray:
+    def solve_linear_system(self) -> np.ndarray:
         """Solve linear system.
 
         Default method is a direct solver. The linear solver is chosen in the
--- a/src/porepy/models/abstract_model.py
+++ b/src/porepy/models/abstract_model.py
@@ -50,7 +50,7 @@ class AbstractModel:
             "folder_name": "visualization",
             "file_name": "data",
             "use_ad": False,
-            # Set the default linear solver to Pardiso; this can be overriden by
+            # Set the default linear solver to Pardiso; this can be overridden by
             # user choices. If Pardiso is not available, backup solvers will automatically be
             # invoked.
             "linear_solver": "pypardiso",
@@ -72,6 +72,7 @@ class AbstractModel:
         self.dof_manager: pp.DofManager
         self.mdg: pp.MixedDimensionalGrid
         self.box: dict
+        self.linear_system: tuple[sps.spmatrix, np.ndarray]
 
     def create_grid(self) -> None:
         """Create the grid bucket.
@@ -220,10 +221,10 @@ class AbstractModel:
 
         Raises:
             ValueError if the chosen solver is not among the three currently
-            supported, see assemble_and_solve_linear_system.
+            supported, see linear_solve.
 
             To use a custom solver in a model, override this method (and
-            assemble_and_solve_linear_system).
+            linear_solve).
 
         """
         solver = self.params["linear_solver"]
@@ -232,20 +233,17 @@ class AbstractModel:
         if solver not in ["scipy_sparse", "pypardiso", "umfpack"]:
             raise ValueError(f"Unknown linear solver {solver}")
 
-    def assemble_and_solve_linear_system(self, tol: float) -> np.ndarray:
+    def assemble_linear_system(self):
         """Assemble the linearized system, solve and return the new solution vector.
 
         The linear system is defined by the current state of the model.
 
-        The linear solver is set according to the parameter 'linear_solver' set under
-        initialization of this model.
 
-        Parameters:
-            tol (double): Target tolerance for the linear solver. May be used for
-                inexact approaches.
 
-        Returns:
-            np.ndarray: Solution vector.
+
+
+        Attributes:
+            linear_system is assigned.
 
         """
         t_0 = time.time()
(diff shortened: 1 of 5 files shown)


src/porepy/models/abstract_model.py
```diff
@@ -232,28 +233,42 @@ def _initialize_linear_solver(self) -> None:
...
         t_0 = time.time()
         if self._use_ad:
             A, b = self._eq_manager.assemble()
         else:
             A, b = self.assembler.assemble_matrix_rhs()  # type: ignore
-        t_1 = time.time()
+        self.linear_system = (A, b)
+        logger.debug(f"Assembled linear system in {t_0-time.time():.2e} seconds.")
+
+    def linear_solve(self):
```
jhabriel, line 259: @pmgbergen/porepy-maint  Although this is largely a matter of taste, I would suggest the more descriptive name `solver_linear_system(self)` instead of `linear_solve(self)`. The suggested name also goes more in line with `assemble_linear_system(self)`.

src/porepy/models/abstract_model.py
```diff
@@ -232,28 +235,42 @@ def _initialize_linear_solver(self) -> None:
...
```
jwboth, line 259: This is of course a nice default implementation of a linear solver. However, I actually think I would almost always overwrite this method (when considering a time-dependent problem at least). Instead I would try to precompute the LU factorization in a preparing step (e.g. already within `assemble_linear_system`). In `linear_solve`, I would then only call the sparse solver based on the factorization. Even when considering only a single time step, there should be not difference in terms of performance by splitting the sparse solve into the actual components which will be called internally by the solver library. Of course by allowing for different libraries, this introduces a bit more code than just one more line.

If this is no issue, then it is also OK with me.

jwboth (changes requested):
I really appreciate the split. As this is absolutely natural. I would even go one step further, see my comment on precomputing the lu factorization. This is at least how I always overwrite `assemble_and_solver_linear_system` for transient simulations. And I think there is actually no reason to not do it here as well. You may have another opinion though.

src/porepy/models/abstract_model.py
jwboth, line 259:
```suggestion
    def linear_solve(self) -> None:
```

jwboth (changes requested):

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -71,14 +66,15 @@ def solve(self, model):
 
         return error_norm, is_converged, iteration_counter
 
-    def iteration(self, model, lin_tol):
+    def iteration(self, model):
```
jwboth, line 69:
```suggestion
    def iteration(self, model) -> np.ndarray:
```

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -71,14 +66,15 @@ def solve(self, model):
...
         """A single Newton iteration.
 
         Right now, this is a single line, however, we keep it as a separate function
         to prepare for possible future introduction of more advanced schemes.
         """
```
jwboth, line 74: Please add description of arguments and returned value.

jwboth (dismissed):

src/porepy/models/abstract_model.py
OmarDuran: In support of @jwboth comment, I would add a third step like `factorize_lhs()` to precompute LU or LLT factors and keep `assemble_linear_system()` as it is. However, we don't have a separate method for RHS like `assemble_rhs()`, and having a factorization step will be not beneficial because LHS is always computed when you evaluate AD equations. 
In the way I see this PR, the proposed changes are a good starting point and methods like `factorize_lhs()` should be included in a separate pull request dedicated to  implement`assemble_rhs()`.

OmarDuran (commented):
Thank you @IvarStefansson I am glad to see this split happening.

src/porepy/models/abstract_model.py
```diff
@@ -232,28 +235,42 @@ def _initialize_linear_solver(self) -> None:
         if solver not in ["scipy_sparse", "pypardiso", "umfpack"]:
             raise ValueError(f"Unknown linear solver {solver}")
 
-    def assemble_and_solve_linear_system(self, tol: float) -> np.ndarray:
+    def assemble_linear_system(self):
```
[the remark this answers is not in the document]
IvarStefansson: I don't think your suggestion is in line with the style guide. Please correct me if I'm wrong, @jhabriel

src/porepy/models/abstract_model.py
IvarStefansson: I agree with Omar's conclusion. I see your point, Jakub, but suggest you for now consider factorization part of "assemble" and adjust that method accordingly.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
IvarStefansson: While in principle always in favour of updating documentation, I respectfully don't think it's the PR creator's responsibility to update documenation in all files touched in that PR.

src/porepy/models/abstract_model.py
IvarStefansson: @keileg, could you should weigh in on the naming, please?

keileg (approved):
A few comments. I say we wait with rhs/lhs etc., as well as factorizaiton, all this seems to me premature.

Also please incorporate the typing of @jwboth .

IvarStefansson:
> A few comments. I say we wait with rhs/lhs etc., as well as factorizaiton, all this seems to me premature.
> 
> Also please incorporate the typing of @jwboth .

You mean those I applied in [this commit](https://github.com/pmgbergen/porepy/pull/716/commits/be41fc9d3e804ca12024bc68f18998112331ff08), right?

src/porepy/models/abstract_model.py
jwboth: Fine for me.

src/porepy/models/abstract_model.py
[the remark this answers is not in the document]
jwboth: I believe you.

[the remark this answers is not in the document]
IvarStefansson: Seems you're right, Jakub. Explicit is encouraged, as implicit (i.e. leaving blank) is interpreted as an un-annotated method return and is not checked. This is "official" pep 484, see https://peps.python.org/pep-0484/#the-meaning-of-annotations

PR #718, opened 2022-09-15
BUG: Use meshio's convention to define cell types.

jwboth:
## Proposed changes

While working on an importer (via meshio), I noticed a bug in the exporter. Meshio defines cell types by the number of nodes and not number of faces per cell. The latter has been used in the current exporter (I believe I introduced this error - at some other location of the code in the exporter, still faces are used to distinguish tets, hex's and polyhedra, and in my opinion this is still OK, as this information does not propagate any further). 

Note, that `test_vtk.py` did not notice this inconsistency as meshio actually overwrites cell types, and the only critical case (for 3d polyhedra) is so easy that there has been never any trouble.

Now, a consistent use of cell types is suggested with this PR. 

## Types of changes

- [ ] Other: 

## Checklist

||Yes|No|NA|
|---|---|---|---|
|The update is covered by the  test suite |X|_|_|
|The documentation is up-to-date |X|_|_|
|I have not duplicated existing functionality |X|_|_|

### DIFF
--- a/src/porepy/viz/exporter.py
+++ b/src/porepy/viz/exporter.py
@@ -181,8 +181,7 @@ class Exporter:
         self._padding = 6
 
     def add_constant_data(
-        self,
-        data: Optional[Union[DataInput, list[DataInput]]] = None,
+        self, data: Optional[Union[DataInput, list[DataInput]]] = None
     ) -> None:
         """
         Collect user-defined constant-in-time data, associated with grids,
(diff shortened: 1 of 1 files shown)


jwboth:
Side note: When this PR should be accepted, the importer will follow. For complete information, the tests for the importer actually detected the inconsistency. These tests import from existing vtu files for polyhedral 3d grids, and export again. As meshio overwrites the cell types, an internal check fails whether the grid corresponding to the vtu file and the current grid of a running simulation are the same - the cell types are the reason. With this PR, the tests will all pass.

keileg (approved):
Looks good to me.

PR #720, opened 2022-09-15
Add possibility to use constant time step in the `TimeSteppingControl` class

jhabriel:
## Proposed changes

This PR extends the capability of the class `TimeSteppingControl` to handle a constant time step. This is done by an optional boolean parameter `constant_dt` which is False by default.

In essence, the whole routine is bypassed such that the method `next_time_step(iterations=int_var, recompute_solution=bool_var)` always returns the same time step. Note this occurs even if `recompute_solution=True` is passed to the method. 

## Types of changes

- [x] Documentation update (if none of the other choices apply)
- [ ] Other: 

## Checklist

||Yes|No|NA|
|---|---|---|---|
|The update is covered by the  test suite |x|_|_|
|The documentation is up-to-date |x|_|_|
|I have not duplicated existing functionality |x|_|_|

## Further comments

Although the algorithm itself is largely bypassed when a constant time step is used, the user can still pass a `schedule` that is compatible with the given time step. The immediate usage of this feature will be, for example, to export the results corresponding only to the scheduled times. If the `dt_init` and `schedule` are not compatible, an error is raised.


jhabriel:
@vlipovac: Feel free to commit the changes to the tests when you're ready.

jhabriel:
Note to self: Add public attributes to the class docstrings.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
+        time_init (float): Initial simulation time.
+    """
 
     def __init__(
         self,
         schedule: list,
         dt_init: float,
-        dt_min_max: Tuple[float, float],
+        constant_dt: bool = False,
+        dt_min_max: Optional[Tuple[float, float]] = None,
```
keileg, line 92:
```suggestion
        dt_min_max: Optional[tuple[float, float]] = None,
```
Please also update in other parts of this module.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
+                [0, 1*pp.HOUR, 3*pp.HOUR].
+            Examples of INVALID inputs are:
+                [1]
+                [1, 0]
+                [0, 1, 1, 2]
+        dt_init: Initial time step. If `constant_dt` is True, then `dt_init` is taken as
+            the constant time step.
+        constant_dt: Whether to treat the time step as constant or not. If True, then
+            the time-stepping control algorithm is effectively bypassed. To be
+            precise, the algorithm won't adapt the time step in any situation.
```
keileg, line 29: To be sure: If I set a constant time step of 2, and a schedule of [0, 5, 10], the 5 will be ignored? I think this fair, I just want to understand.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
 
 __all__ = ["TimeSteppingControl"]
 
 
 class TimeSteppingControl:
-    """Parent class for iteration-based time stepping control routine."""
+    """Parent class for iteration-based time stepping control routine.
+
+    Parameters:
+        schedule: List containing the target times for the simulation.
```
keileg, line 12: We can imagine combining this with a scheduler which determines necessary target times due to visualization, changes in well rates or boundary conditions etc. No need to adjust anything here, but let's keep this in mind.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
+                )
+            elif dt_init > dt_min_max[1]:
+                raise ValueError(
+                    "Initial time step cannot be larger than maximum time step."
+                )
+
+            # NOTE: The above checks guarantee that minimum time step <= maximum time step
+
+            # Sanity checks for maximum number of iterations.
+            # Note that 0 is a possibility. This will imply that the solver reaches
```
keileg, line 140: Maybe it is just me, but this corresponds to 1 iteration, not 0.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
+
+        # More sanity checks below. Note that all the remaining sanity checks (but one) are
+        # only needed when constant_dt = False. Thus, to save time when constant_dt = True,
+        # we use this rather ugly if-statement from below.
+
+        if not constant_dt:
+
+            # Sanity checks for dt_min and dt_max
+            if dt_init < dt_min_max[0]:
+                raise ValueError(
```
keileg, line 129:
```suggestion
                # NOTE: If dt_min_max was not set on initialization, a value is assigned besed on the total
                # simulation time.  This may unwittingly trigger the below error.
                raise ValueError(
```
I suggest adding a comment like this to give users something to work with if they unwittingly trigger this error. Someone will appreciate it.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
+                    "Initial time step cannot be smaller than minimum time step."
+                )
+            elif dt_init > dt_min_max[1]:
+                raise ValueError(
+                    "Initial time step cannot be larger than maximum time step."
+                )
```
keileg, line 135:
```suggestion
            elif dt_init > dt_min_max[1]:
                # NOTE: If dt_min_max was not set on initialization, a value is assigned besed on the total
                # simulation time.  This may unwittingly trigger the below error.            
                raise ValueError(
                    "Initial time step cannot be larger than maximum time step."
                )
```

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
+            # Note that 0 is a possibility. This will imply that the solver reaches
+            # convergence directly, e.g., as in linear solvers.
+            if iter_max < 0:
+                raise ValueError("Maximum number of iterations must be non-negative.")
+
+            # Sanity checks for optimal iteration range
+            if iter_optimal_range[0] > iter_optimal_range[1]:
+                s = "Lower optimal iteration range cannot be larger than"
+                s += " upper optimal iteration range."
+                raise ValueError(s)
```
keileg, line 149: In such cases, I like it when the error message prints out the values given, it makes debugging easier, but I'm never sure if it is worth the added code complexity (small, yes, but so is the debugging effort). Any opinions?

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
+    Attributes:
+        dt (float): Time step.
+        dt_init (float): Intial time step.
+        dt_max (float): Maximum time step.
+        dt_min (float): Minimum time step.
+        is_constant (bool): Constant time step flag.
+        iter_low (int): Lower bound of the optimal iteration range.
+        iter_low_factor (float): Lower multiplication factor. Strictly greater than one.
+        iter_max (int): Maximum number of iterations.
+        iter_upp (int): Upper bound of the optimal iteration range.
```
keileg, line 77: A lower factor that is >1 and an upper factor < 1 is confusing. Is it meaningful to write out what the update formula is somewhere?

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
+
+            # Sanity checks for lower and upper multiplication factors
+            if iter_lowupp_factor[0] <= 1:
+                raise ValueError("Expected lower multiplication factor > 1.")
+            elif iter_lowupp_factor[1] >= 1:
+                raise ValueError("Expected upper multiplication factor < 1.")
+
+            # Checks for sensible combinations of iter_optimal_range and iter_lowupp_factor
+            if dt_min_max[0] * iter_lowupp_factor[0] > dt_min_max[1]:
+                raise ValueError("Encountered dt_min * iter_low_factor > dt_max.")
```
keileg, line 166: I understand the logic here, but I did not expected such requirements on the parameters in addition to what was described above (in particular since the documentation is so good, well done!). My suggestion is to give the update formula in the class documentation, and then write short comments just above these error messages so that a user would understand what went wrong. You could also document this in the class documentation, but that would lead to  so many conditions that it is easy to overlook something.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
+            if recomp_max <= 0:
+                raise ValueError("Number of recomputation attempts must be > 0.")
+
+        else:
+
+            # If the time step is constant, check that the scheduled times and the time
+            # step are compatible.
+            # E.g. dt=2 works with schedule = [0,2,4,8], but dt=3 does not.
+            sim_times = np.arange(schedule[0], schedule[-1] + dt_init, dt_init)
+            intersect = np.intersect1d(sim_times, schedule)
```
keileg, line 184: I see, here is the answer to a question above. Please document this above.

src/porepy/numerics/time_step_control.py
```diff
@@ -137,10 +198,14 @@ def __init__(
         # Minimum and maximum allowable time steps
         self.dt_min, self.dt_max = dt_min_max
 
-        # Maximum amount of iterations
+        # Maximum number of iterations
+        # TODO: This is really a property of the nonlinear solver. A full integration will
```
keileg, line 202: Agree. This is a natural next step, after we have integrated this into the models.

src/porepy/numerics/time_step_control.py
```diff
@@ -218,25 +284,32 @@ def next_time_step(
...
+        # choice was made to give more flexibility, in the sense that we are no limiting
+        # the recomputation criteria to _only_ reaching the maximum number of iterations,
+        # even though that is the primary intended usage.
         if recompute_solution and self._recomp_num < self.recomp_max:
-            self.time -= self.dt
-            self.dt *= self.recomp_factor
-            self._recomp_num += 1
+            self.time -= self.dt  # (S1)
+            self.dt *= self.recomp_factor  # (S2)
+            self._recomp_num += 1  # (S3)
```
keileg, line 307: Nice documentation.

src/porepy/numerics/time_step_control.py
```diff
@@ -218,25 +284,32 @@ def next_time_step(
...
-        # Note that iterations is not really used here. So, as long as the passed
-        # recompute_solution==True and recomputation_attempts < max_recomp_attempts, the user
-        # can pass any number of iterations. This, in principle, allows for more flexibility,
-        # in the sense that we are no restricting the recomputation criteria to only reaching
-        # the maximum number of iterations, even though that is the primary intended usage.
+        # If time step is constant, always return that value
+        if self.is_constant:
+            return self.dt_init
+
+        # If the solution did not convergence and we are allow to recompute it, then:
```
keileg, line 291:
```suggestion
        # If the solution did not convergence and we are allowed to recompute it, then:
```

src/porepy/numerics/time_step_control.py
```diff
@@ -251,39 +324,49 @@ def next_time_step(
...
-        # i.e., "factor_low". If the number of iterations is greater than the upper optimal
-        # iteration range "iter_upp", we have to decrease the time step by multiplying by an
-        # upper multiplication factor smaller than 1, i.e., "factor_upp". If neither of these
-        # situations occur, then the number iterations lies between the optimal iteration
-        # range and the time step remains unchanged.
-        if iterations <= self.iter_low:
+        # following criteria:
+        #     (C1) If iters is less than the lower optimal iteration range "iter_low",
+        #     we can relax the time step, and multiply by a lower multiplication factor
+        #     greater than 1, i.e., "factor_low".
```
keileg, line 330: again, the terminology is confusing - in my mind, low maps to lowering of the time step, rather than elevating it (that is, the time step is too low). Not sure about what is the best solution here.

keileg (commented):
One major remark: The sanity checks effectively impose additional and undocumented constraints on permissible parameter values. I have some suggestions on how to fix this, but @IvarStefansson may have other thoughts.

Besides this, the comments are mainly on terminology and similar.

jhabriel:
Thanks for the thorough review @keileg. I will address the comments/suggestions tomorrow.

@vlipovac: You can wait until we sort things out with the class to start working on the tests (just to avoid having to rewrite them).

src/porepy/numerics/time_step_control.py
jhabriel: Humm, good question. I think this is a matter of convention. To me, the first iteration is the second time a linear system is solved. That's why I use iteration = 0 to refer to the case when a linear system is solved exactly (directly). I have no strong preferences, but we have to use the same convention. 

@IvarStefansson @OmarDuran @jwboth What is your preference regarding this point?

src/porepy/numerics/time_step_control.py
jhabriel: Done.

src/porepy/numerics/time_step_control.py
jhabriel: Thanks!

src/porepy/numerics/time_step_control.py
jhabriel: You are right, I did not consider this possibility. If the user passes any `dt_min < 0.001 schedule[-1]`, and don't pass `dt_min_max`, then this error will be triggered.

I will raise a costume error for this case, suggesting to pass `dt_min_max` to avoid getting an error triggered by `dt_min < dt_min_max[0]`.

src/porepy/numerics/time_step_control.py
jhabriel: As you mentioned, the only drawback here is more code complexity + slightly less readable code. I'm in favour of adding the actual value in the error message in these cases, considering the number of parameters that the user has to deal with.

src/porepy/numerics/time_step_control.py
jhabriel: Done.

src/porepy/numerics/time_step_control.py
jhabriel: I'm quite sure that in neither of these cases the algorithm will crash. In the first case, it will simply select `dt_max`, and in the second case `dt_min`.

For the record, I've added these checks because when such choices of `dt_min_max` and `iter_lowupp_factor` are made, the algorithm can't really do much (except selecting `dt_max` and `dt_min`). Nevertheless, this does not qualify as an error, but as a (very bad) decision from the user side. And we cannot, of course, cover the whole set of bad choices (believe me, it's a huge set). 

That being said, and to your point, I would suggest demoting this error to a (more meaningful) warning, provided the algorithm does not fail. This, we will have to make sure with unit tests. If it turns out that the algorithm actually fail, we'll go with your suggestion, since, again, this is really a more or less edge case. How does that sound?

jhabriel:
@keileg : I believe I answer all your questions. There are some things that we have to decide before introducing more changes to the class. When these conversations are resolved, I will incorporate the changes.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,209 @@
...
+        # More sanity checks below. Note that all the remaining sanity checks (but one) are
+        # only needed when constant_dt = False. Thus, to save time when constant_dt = True,
+        # we use this rather ugly if-statement from below.
+
+        if not constant_dt:
+
+            # Sanity checks for dt_min and dt_max
+            mssg_dtmin = "Initial time step cannot be smaller than minimum time step. "
+            mssg_dtmax = "Initial time step cannot be larger than maximum time step. "
+            mssg_unset = "This error was raised since `dt_min_max` was not set on "
```
IvarStefansson, line 138: As a matter of taste, I would use a multiline string here

src/porepy/numerics/time_step_control.py
IvarStefansson: I agree with Eirik.

src/porepy/numerics/time_step_control.py
IvarStefansson: I agree. The code is not really more complex in this case using string formatting.

src/porepy/numerics/time_step_control.py
IvarStefansson: First impression is that I agree on the warning suggestion. But could this kind of incompatible values lead to bad behaviour? E.g. first relaxing the time step, and then restricting because the restriction rendered it larger than dt_max?

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,209 @@
...
+
+            # Checks for sensible combinations of iter_optimal_range and iter_lowupp_factor
+            if dt_min_max[0] * iter_lowupp_factor[0] > dt_min_max[1]:
+                raise ValueError("Encountered dt_min * iter_low_factor > dt_max.")
+            elif dt_min_max[1] * iter_lowupp_factor[1] < dt_min_max[0]:
+                raise ValueError("Encountered dt_max * iter_upp_factor < dt_min.")
+
+            # Sanity check for recomputation factor
+            if recomp_factor >= 1:
+                raise ValueError("Expected recomputation multiplication factor < 1.")
```
IvarStefansson, line 190: Again, warning

src/porepy/numerics/time_step_control.py
```diff
@@ -192,22 +278,20 @@ def __repr__(self) -> str:
...
-            step will be tuned accordingly.
+        If convergence was not achieved, then the time step is reduced by recomp_factor. The
+        time-stepping control routine will recompute the solution recomp_max times.
+        Otherwise, an error will be raised and the simulation stopped.
 
-        Returns
-        dt (float or None):  Next time step if time < final_simulation time. None otherwise.
+        Parameters:
+            iterations: Number of non-linear iterations. In time-dependent simulations,
+                this tipically represent the number of iterations for a given time step.
```
IvarStefansson, line 289:
```suggestion
                this typically represents the number of iterations for a given time step.
```

src/porepy/numerics/time_step_control.py
```diff
@@ -218,25 +302,32 @@ def next_time_step(
...
+        # recompute_solution=True and recomputation_attempts < max_recomp_attempts,
+        # the method is entirely agnostic to the number of iterations passed. This design
+        # choice was made to give more flexibility, in the sense that we are no limiting
+        # the recomputation criteria to _only_ reaching the maximum number of iterations,
+        # even though that is the primary intended usage.
         if recompute_solution and self._recomp_num < self.recomp_max:
-            self.time -= self.dt
-            self.dt *= self.recomp_factor
-            self._recomp_num += 1
+            self.time -= self.dt  # (S1)
```
IvarStefansson, line 323: Nice way of documenting the algorithm!

src/porepy/numerics/time_step_control.py
jhabriel: Indeed, this oscillatory behavior that you describe might be a possibility for really bad choices of parameters. But in principle, it won't break the algorithm. It will most likely jump from `dt_min` to `dt_max` between iteration steps. 

We have to decide whether we want to allow this or not.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
+            Nevertheless, the attributes (such as scheduled times) will still be accesible.
+        dt_min_max: Minimum and maximum permissible time steps. If None, then the
+            minimum time step is set to 0.1% of the final simulation time and the
+            maximum time step is set to 10% of the final simulation time. If given, then
+            the first and second elements of the tuple corresponds to the minimum and
+            maximum time steps, respectively.
+        iter_max: Maximum number of iterations.
+        iter_optimal_range: Optimal iteration range. The first and second elements of the
+            tuple corresponds to the lower and upper bounds of the optimal iteration range.
+        iter_lowupp_factor: Lower and upper multiplication factors. The first and second
```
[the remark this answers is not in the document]
jhabriel: I have proposed an alternative for `iter_lowupp_factor`. Please, refer to https://github.com/pmgbergen/porepy/pull/720#discussion_r974158778

src/porepy/numerics/time_step_control.py
jhabriel: Noted. I will replace it.

src/porepy/numerics/time_step_control.py
jhabriel: In this case, we have to raise an error. Otherwise, the algorithm will (very easily) break.

src/porepy/numerics/time_step_control.py
jhabriel: Thanks :)

src/porepy/numerics/time_step_control.py
jwboth: Same for me. Iteration counts for me are unrelated to coding languages and thereby start with 1, in the very moment the respective iterative algorithm is initiated.

jhabriel: Fair enough. I will modify this then.

OmarDuran: When the system is linear, a newton process performs 1  iteration, so I agree with Eirik.

src/porepy/numerics/time_step_control.py
jhabriel: Ok. I will add the actual values to the error messages.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,191 @@
...
 __all__ = ["TimeSteppingControl"]
 
 
 class TimeSteppingControl:
-    """Parent class for iteration-based time stepping control routine."""
+    """Parent class for iteration-based time stepping control routine.
+
+    Parameters:
+        schedule: List containing the target times for the simulation.
+            The time-stepping algorithm will adapt the time step so that the target
```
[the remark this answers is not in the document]
jhabriel: Suggestion incorporated.

src/porepy/numerics/time_step_control.py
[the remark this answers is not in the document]
jhabriel: Let's move the discussion over there.

src/porepy/numerics/time_step_control.py
OmarDuran: I agree with @jhabriel  suggestion. Nevertheless, it is necessary to write the actual time step updating formula. For clarity, I suggest splitting next_time_step into simple functions.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,210 @@
...
         schedule: list,
         dt_init: float,
-        dt_min_max: Tuple[float, float],
+        constant_dt: bool = False,
+        dt_min_max: Optional[tuple[float, float]] = None,
         iter_max: int = 15,
-        iter_optimal_range: Tuple[int, int] = (4, 7),
-        iter_lowupp_factor: Tuple[float, float] = (1.3, 0.7),
+        iter_optimal_range: tuple[int, int] = (4, 7),
+        iter_lowupp_factor: tuple[float, float] = (1.3, 0.7),
```
OmarDuran, line 105: This could be a matter of taste, but some defaults in academic reservoir simulators are:
* A factor of  0.5 for decreasing the time step
* A factor of 1.25 for increasing the time step
* For constant time step just setting up `iter_lowupp_factor = (1.0,1.0)` and delete the variable constant_dt.
This can be found in "Principles of Applied Reservoir Simulation" John R. Fnachi.

OmarDuran (commented):
This is a good starting point to formalize time-stepping methods. I left a few comments.

src/porepy/numerics/time_step_control.py
keileg: I'm fine with the suggested naming.

src/porepy/numerics/time_step_control.py
keileg: This is surely a matter of taste. 

As for the constant, a compromise could be to keep constant_dt as a user input, but implement it as setting both relaxation factors to 1. What do you think @jhabriel?

src/porepy/numerics/time_step_control.py
keileg: My opinion:
1) We want to check for this, users (including myself) can easily mix this up, and it is hard to debug.
2) Add sufficient documentation so that the user does not get an error message from what seems to be acceptable parameter values.

keileg (commented):
Some more comments. And I would still like to see the update formula in the documentation.

src/porepy/numerics/time_step_control.py
jhabriel: Ok. I will rename the parameter and the attributes. I agree with your suggestion Omar, the `next_time_step()` method got too long and has to be split in smaller parts.

src/porepy/numerics/time_step_control.py
jhabriel: Ok. Then it stays as an error, but with a more complete error message.

src/porepy/numerics/time_step_control.py
jhabriel: Thanks Omar. I did not know about the standards in reservoir simulation (since I primarily worked with unsaturated systems). The default parameters can be easily changed upon consensus, and I don't have strong opinions about them. 

As for the constant time step being implemented using unitary relaxation factors, I do have strong opinions on why this is not a good idea.

Bypassing the algorithm and always returning the same time step is (1) cleaner, (2) arguably faster, (3) easier to debug, and (4) more scalable. One then may argue: Why incorporate constant dt into the adaptive time step strategy? As a matter of fact, the only reason is the possibility to access the schedule, which I think is really important.

Another way to see this is: users that are only interested in linear transient models should not be required to read through the whole documentation when debugging their models, i.e.: to check what are these relaxation factors. 

As for scalability, in the future, we might want to incorporate other time-stepping control algorithms, this will most likely require creating an abstract time control class. One can then think of the constant time step class as the most basic type of time control, and keeping implementations separated will easy future implementations.

jhabriel:
@keileg : I'm currently exploring the best way to incorporate the update formula into the documentation.

src/porepy/numerics/time_step_control.py
jhabriel: Omar: I've added a TODO for your suggestion of splitting the next_time_step method. Since I would like to prioritize other aspects of the PR right now. With that, I will resolve this conversation.

src/porepy/numerics/time_step_control.py
jhabriel: I've added extensive documentation.

jhabriel:
The first code review really improved the class, thanks. In this new version, you will see improvements related to:
- Documentation: better documentation of parameters and attributes in general. In addition, I've included an algorithm overview and the full algorithm in pseudocode as part of the class documentation.
- Naming: Better and less confusing names of parameters and attributes.
- Structure: The `next_time_step()` method was split into smaller and more readable parts.
- Errors and warnings: Error and warning messages are more descriptive, making (hopefully) debugging easier.
- Requirements: Requirements on parameters and combination of parameters are explicitly communicated in the class.

The class is now ready for re-review @keileg @IvarStefansson @OmarDuran @jwboth. Some unit tests will have to be updated and some will have to be added. Since this is directly linked to the "final" version of the class, you don't have to review them (yet).

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,235 @@
...
+    Parameters:
+        schedule: List containing the target times for the simulation.
+            Unless a constant time step is prescribed, the time-stepping algorithm will adapt
+            the time step so that the scheduled times are guaranteed to be hit.
+
+            The `schedule` list must contain minimally two items, corresponding to the
+            initial and final simulation times. Lists of length > 2 must contain strictly
+            increasing times. Examples of VALID inputs are: [0, 1], [0, 10, 30, 50], and
+            [0, 1*pp.HOUR, 3*pp.HOUR]. Examples of INVALID inputs are: [1], [1, 0], and
+            [0, 1, 1, 2].
```
jwboth, line 23: Not crucial: Yet, why not internally providing simple mechanisms which unify and sort the items? This would relax the requirements. Typical use case would be if one wants to add another prescribed time to a long list of times. Then searching for the right position is doable, but maybe not necessary.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,235 @@
...
+
+            If a constant time step is used (`constant_dt = True`), then the time step
+            (`dt_init`) is required to be compatible with the scheduled times in `schedule`.
+            Otherwise, an error will be raised. Examples of VALID inputs for `constant_dt =
+            True` and `dt_init = 2` are: [0, 2] and [0, 4, 6, 10]. Examples of INVALID inputs
+            for `constant_dt = True` and `dt_init = 2` are [0, 3] and [0, 4, 5, 10].
```
jwboth, line 29: Sounds like trouble to me. Why not ignoring the schedule when choosing `constant_dt=True`? The `TimeSteppingControl` anyhow will visit all of the scheduled times, you are currently requesting.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,235 @@
...
+            The first and second elements of the tuple corresponds to the under-relaxation and
+            over-relaxation factors, respectively. We require the under-relaxation factor
+            to be strictly lower than one, whereas the over-relaxation factor is required to
+            be strictly greater than one.
+
+            To avoid oscilations and ensure a stable time step adaptation in combination with
+            the minimum and maximum allowable time steps, we further require:
+                dt_min_max[0] * iter_relax_factors[1] < dt_min_max[1], and
+                dt_min_max[1] * iter_relax_factors[0] > dt_min_max[0].
+            Note that in practical applications, these conditions are ussually met.
```
jwboth, line 61:
```suggestion
            Note that in practical applications, these conditions are usually met.
```

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,235 @@
...
+        recomp_factor: Failed-to-converge solution recomputation factor.
+            Factor by which the time step will be multiplied in case the solution must be
+            recomputed. We require `recomp_factor` to be strictly less than one.
+        recomp_max: Failed-to-converge maximum recomputation attempts. The maximum allowable
+            number of consecutive recomputation attempts.
+        print_info. Wheter to print on-the-fly time-stepping information or not.
```
jwboth, line 67:
```suggestion
        print_info. Whether to print on-the-fly time-stepping information or not.
```

src/porepy/numerics/time_step_control.py
jwboth: The only requirement which one could stick to is that max time - min time is a multiple of `dt_init`.

src/porepy/numerics/time_step_control.py
```diff
@@ -180,110 +291,233 @@ def __repr__(self) -> str:
...
+            `recomp_factor` with the hope of achieving convergence in the next time level. To
+            avoid an infinite loop, an error will be raised if the method is called more than
+            `recomp_max` consecutive times with the flag `recompute_solution = True`.
+
+            Now that the algorithm has determined a new time step, it has to ensure three more
+            conditions, (1) the calculated time step cannot be smaller than `dt_min`,
+            (2) the calculated time step cannot be larger than `dt_max`, and (3) the time
+            step cannot be too large such that the next time will exceed a scheduled
+            time. These three conditions are implemented in this order of precedence and
+            will overrride any of the previous calculated time steps.
```
jwboth, line 340:
```suggestion
            will override any of the previous calculated time steps.
```

src/porepy/numerics/time_step_control.py
```diff
@@ -180,110 +291,233 @@ def __repr__(self) -> str:
...
+            self._adaptation_based_on_recomputation()
+
+        # Correct time step
+        self._correction_based_on_dt_min()
+        self._correction_based_on_dt_max()
+        self._correction_based_on_schedule()
+
+        return self.dt
+
+    def _adaptation_based_on_iterations(self, iterations) -> None:
```
jwboth:
```suggestion
    def _adaptation_based_on_iterations(self, iterations: int) -> None:
```

jwboth (commented):
Very dedicated comments! I added some minor comments.

src/porepy/numerics/time_step_control.py
jhabriel: Interesting feature. This should doable, yes. Do we want something like this @keileg?

src/porepy/numerics/time_step_control.py
jhabriel: I understand the logic, yes. Sounds like a better and less confusing way to handle constant time steps. So, essentially, we will only need to check if `mod((final_time - initial_time), dt_init) == 0`. Right?

src/porepy/numerics/time_step_control.py
jhabriel: Indeed, mypy actually complained about this.

IvarStefansson:
> The first code review really improved the class, thanks. In this new version, you will see improvements related to:
> 
>     * Documentation: better documentation of parameters and attributes in general. In addition, I've included an algorithm overview and the full algorithm in pseudocode as part of the class documentation.
> 
>     * Naming: Better and less confusing names of parameters and attributes.
> 
>     * Structure: The `next_time_step()` method was split into smaller and more readable parts.
> 
>     * Errors and warnings: Error and warning messages are more descriptive, making (hopefully) debugging easier.
> 
>     * Requirements: Requirements on parameters and combination of parameters are explicitly communicated in the class.
> 
> 
> The class is now ready for re-review @keileg @IvarStefansson @OmarDuran @jwboth. Some unit tests will have to be updated and some will have to be added. Since this is directly linked to the "final" version of the class, you don't have to review them (yet).

You ask for second review, yet the PR is still wip and not "ready for review". Sorry, but I'm a bit confused.

jhabriel:
Yes, is not ready for review because the tests are not ready. That's why I explicitly wrote that the class is ready for re-review and not the full PR, which in this case encompasses changes in the class and the tests. 

Hope this clarifies your confusion.

src/porepy/numerics/time_step_control.py
jhabriel: We decided to keep it as it is, meaning , to require strictly increasing times.

src/porepy/numerics/time_step_control.py
jhabriel: We decided to keep it as it is, meaning , to require strictly increasing times.

src/porepy/numerics/time_step_control.py
jhabriel: We decided to keep it as it is for the meantime. If, in practice, this becomes too restrictive, we'll go for this alternative.

jhabriel:
After discussing with some core devs today, we agree on some changes. I will incorporate them and also update the tests.

tests/unit/test_time_step_control.py
```diff
@@ -8,85 +8,72 @@
...
-    #     assert tsc.dt_max == 1
-    #     assert tsc.iter_low_factor == 1
-    #     assert tsc.iter_upp_factor == 1
-    #     assert tsc.recomp_factor == 1
-    #     assert tsc.is_constant
+    @pytest.mark.parametrize(
+        "schedule",
+        [(0, 0.5, 1), [0, 0.5, 1], np.array([0, 0.5, 1])]
+    )
+    def test_schedule_argument_type_compatibility(self, schedule):
```
jhabriel, line 31: Great! Just for the sake of completeness. Could you add an assert that compares the given schedule (converted into an np.ndarray) with tsc.schedule?

tests/unit/test_time_step_control.py
```diff
@@ -204,57 +181,52 @@ def test_lower_iter_endpoint_greater_than_or_equal_to_zero(self):
             Ts(
                 schedule=[0, 1],
                 dt_init=0.1,
-                dt_min_max=(0.1, 0.5),
                 iter_max=5,
                 iter_optimal_range=iter_optimal_range,
             )
         assert msg in str(excinfo.value)
 
     # TODO: Remove dt_min_max from parameters
```
jhabriel, line 189: Remove TODO

tests/unit/test_time_step_control.py
```diff
@@ -340,22 +305,20 @@ class TestTimeControl:
...
         assert dt is None
         # Now, assume we are above the final time
-        tsc = Ts(schedule=[0, 1], dt_init=0.1, dt_min_max=(0.1, 0.5))
+        tsc = Ts(schedule=[0, 1], dt_init=0.1)
         tsc.time = 2
         dt = tsc.next_time_step(iterations=0, recompute_solution=False)
         assert dt is None
 
-    # TODO: Purge dt_min_max from parameters
     def test_constant_time_step(self):
```
jhabriel, line 322: Excellent!

tests/unit/test_time_step_control.py
```diff
@@ -411,36 +372,32 @@ def test_recomputed_solutions(self):
...
-    # TODO: Purge dt_min_max from parameters
     def test_recomputed_solution_with_calculated_dt_less_than_dt_min(self):
         """Test when a solution is recomputed and the calculated time step is less than
         the minimum allowable time step, the time step is indeed the minimum time step"""
-        tsc = Ts([0, 100], 2, dt_min_max=(0.6, 10), recomp_factor=0.5)
+        tsc = Ts([0, 100], 0.15, recomp_factor=0.5)
         # Emulate the scenario where the solution must be recomputed
         tsc.time = 5
-        tsc.dt = 1
+        #tsc.dt = 1
```
jhabriel, line 381: Remove commented line.

jhabriel (commented):
Great jo @vlipovac. I've requested only minor changes.

vlipovac:
Hei @jhabriel , I have implemented your comments on  my last push and marked respective conversations resolved, if that is alright.

tests/unit/test_time_step_control.py
```diff
@@ -1,3 +1,4 @@
+import numpy as np
```
IvarStefansson, line 1: Could we have a summary of the tests here at the top of the file, please?

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,321 @@
-from typing import Tuple, Union
+"""
+This module contains a routine for iteration-based time-stepping control.
+
+The algorithm is heavily inspired on [1], which was later used on [2].
```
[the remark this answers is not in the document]
jhabriel: Done.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,321 @@
...
+    """Parent class for iteration-based time-stepping control routine.
+
+    Parameters:
+        schedule: Array-like object containing the target times for the simulation.
+            Unless a constant time step is prescribed, the time-stepping algorithm will adapt
+            the time step so that the scheduled times are guaranteed to be hit.
+
+            The `schedule` must contain minimally two elements, corresponding to the
+            initial and final simulation times. Schedules of size > 2 must contain strictly
+            increasing times. Examples of VALID inputs are: [0, 1], np.array([0, 10, 30,50]),
```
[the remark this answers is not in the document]
jhabriel: Done.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,321 @@
...
+        iter_optimal_range: Optimal iteration range.
+            The first and second elements of the tuple correspond to the lower and upper
+            endpoints of the optimal iteration range.
+        iter_relax_factors: Relaxation factors.
+            The first and second elements of the tuple corresponds to the under-relaxation and
+            over-relaxation factors, respectively. We require the under-relaxation factor
+            to be strictly lower than one, whereas the over-relaxation factor is required to
+            be strictly greater than one.
+
+            To avoid oscilations and ensure a stable time step adaptation in combination with
```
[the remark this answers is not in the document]
jhabriel: Done.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,321 @@
...
+            the minimum and maximum allowable time steps, we further require:
+                dt_min_max[0] * iter_relax_factors[1] < dt_min_max[1], and
+                dt_min_max[1] * iter_relax_factors[0] > dt_min_max[0].
+            Note that in practical applications, these conditions are usually met.
```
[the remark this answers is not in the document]
jhabriel: Done.

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,321 @@
...
 
-        # Checks for schedule
-        if len(schedule) < 2:
-            raise ValueError("Expected schedule with at least two items.")
+        schedule = np.array(schedule)
+        # Sanity checks for schedule
+        if np.size(schedule) < 2:
+            raise ValueError("Expected schedule with at least two elements.")
         elif any(time < 0 for time in schedule):
             raise ValueError("Encountered at least one negative time in schedule.")
```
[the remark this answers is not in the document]
jhabriel: Not a problem from a computation/implementation standpoint. But from a physical perspective, I assume many things can go wrong.

On the other hand, we could allow for negative times and let the user deal with the consequences.

I'm OK with either option. If you have strong opinions about this and want me to remove this sanity check, let me know and I will update the code (and the tests).

src/porepy/numerics/time_step_control.py
```diff
@@ -1,130 +1,321 @@
...
         elif not self._is_strictly_increasing(schedule):
             raise ValueError("Schedule must contain strictly increasing times.")
 
-        # Checks for initial time step
+        # Sanity checks for initial time step
         if dt_init <= 0:
             raise ValueError("Initial time step must be positive.")
```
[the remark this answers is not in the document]
jhabriel: Same reason as before. Unless there is very good reason, I think allowing for negative or zero time step is simply a very bad idea. 

In addition, some sanity checks rely on the fact that time step is positive.

src/porepy/numerics/time_step_control.py
```diff
@@ -170,120 +366,173 @@ def __init__(
...
+            )
+            raise ValueError(msg)
 
-        # Check if the calculated time step is greater than the maximum allowable time step
-        if self.dt > self.dt_max:
-            self.dt = self.dt_max
+    def _correction_based_on_dt_min(self) -> None:
+        """Correct time step if dt < dt_min."""
+        if self.dt < self.dt_min_max[0]:
+            self.dt = self.dt_min_max[0]
```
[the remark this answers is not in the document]
jhabriel: Indeed, this is a very good idea. In fact, it will save the user several hopeless iteration attempts.

I will, however, incorporate a different error message for this case, just to make debugging easier for the user.

src/porepy/numerics/time_step_control.py
jhabriel: I agree, this is will be a good enhancement. However, I believe the PR is already too long and I suggest dealing with this feature request in a different PR. Is that OK with you?

tests/unit/test_time_step_control.py
jhabriel: Sure. I have included a brief summary of the tests as requested.

src/porepy/numerics/time_step_control.py
```diff
@@ -170,120 +366,173 @@ def __init__(
...
+            recompute_solution: Whether the solution needs to be recomputed or not. If True,
+                then the time step is multiplied by `recomp_factor`. If False, then the time
+                step will be tuned accordingly.
+
+        Returns: Next time step if time < final_time. None, otherwise.
 
         """
 
         # For bookkeeping reasons, save recomputation flag
+        # TODO: Is this really needed?
```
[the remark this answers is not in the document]
jhabriel: Done.

src/porepy/numerics/time_step_control.py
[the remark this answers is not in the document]
jhabriel: I included this check, with a slight different logic, but with same result. Basically, instead of manually setting self._recomp_num = self.recomp_max, I've added an extra check where an error is raised if adaptation based on recomputation is attempted when dt = dt_min.

src/porepy/numerics/time_step_control.py
IvarStefansson: It is! Should we make an issue?

jhabriel: Sure, I can create one.

IvarStefansson (approved):

PR #724, opened 2022-09-16
Add importer capabilities to the exporter

jwboth:
## Proposed changes
The exporter becomes more an IO tool. Meshio is used to read from vtu files to data dictionaries. Specific keywords can be addressed to read only parts of data, and also a list of files can be provided allowing to read similar data on multiple subdomains for different dimension at the same time. Also the importing routine makes sanity checks whether the external vtu file is based on a comptabile grid, compared to the mixed-dimensional grid which is the basis for the simulation. 

Some possible application areas:
- Initialization of simulations at later stages.
- Returning to old time steps, in case of propagating errors in inexact solver routines.
- Analysis of methods which may fail at some particular iteration/time step. So instead of waiting for when a method fails over the course of the run, one can jump straight away to this stage from the start of the simulation.

## Types of changes

- [ ]  Minor change (e.g., dependency bumps, broken links, etc).
- [ ]  Bugfix (non-breaking change which fixes an issue).
- [X]  New feature (non-breaking change which adds functionality).
- [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ]  Testing (contribution related to testing of existing or new functionality).
- [X]  Documentation (contribution related to adding, improving, or fixing documentation).
- [ ]  Maintenance (e.g., improve logic and performance, remove obsolete code, etc).
- [ ]  Other:

## Checklist

## Further comments

- This PR does not interact with time stepping or models; it merely provides the importing capabilities. The integration into these will be part of a future PR.
- One could consider renaming the directory `viz` to `io`. And the Exporter to DataIO.


jwboth:
Interestingly the unit tests actually did all pass (not as mentioned in one of the commit messages) even before merging develop into this branch. Locally, one of the new import tests was depending on the right addressing of cell types, as corrected in #718.

tests/unit/test_vtk.py
```diff
@@ -173,6 +173,45 @@ def test_single_subdomains(setup, subdomain):
...
+
+    # Define exporter
+    save = pp.Exporter(
+        sd,
+        setup.file_name,
+        setup.folder,
+        export_constants_separately=False,
+    )
+
+    # Define keys (correpsonds to all data stored in the vtu file).
```
jwboth, line 193:
```suggestion
    # Define keys (here corresponding to all data stored in the vtu file to pass the test).
```

tests/unit/test_vtk.py
```diff
@@ -220,6 +259,48 @@ def test_mdg(setup):
...
+
+    # Define exporter
+    save = pp.Exporter(
+        mdg,
+        setup.file_name,
+        setup.folder,
+        export_constants_separately=False,
+    )
+
+    # Define keys (correpsonds to all data stored in the vtu file).
```
jwboth, line 279:
```suggestion
    # Define keys (here corresponding to all data stored in the vtu file to pass the test).
```

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+                are_subdomain_data (bool or list of bool): comparible with file_names; list
+                    of indicators whether the corresponding grid is a subdomain grid;
+                    default is True
+
+        Raises:
+            ValueError if some of the data is not compatible with the supposedly corresponding
+                grid
+        """
+
+        # Aux. method: Inversoe of _toVectorFormat, used to define vector-ranged values
```
jwboth, line 209:
```suggestion
        # Aux. method: Inverse of _toVectorFormat, used to define vector-ranged values
```

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
         # Parameter to be used in several occasions for adding time stamps.
         self._padding = 6
 
+    def import_from_vtu(
+        self, keys: Union[str, list[str]], file_names: Union[str, list[str]], **kwargs
+    ) -> None:
+        """
+        Import state variables from vtu file.
```
keileg, line 189: Perhaps comment that it is assumed the file is generated by PorePy, follows pp naming conventions etc.? I'm not sure exactly what I mean, perhaps I want to say that kick-starting a simulation from outside PorePy using this function is for the brave.

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+                    file_name_pieces[dim_pos - 1] == "mortar"
+                    or file_name_pieces[dim_pos - 1] == "constant"
+                    and file_name_pieces[dim_pos - 2] == "mortar"
+                )
+
+            else:
+                # Read from keyword arguments
+
+                # Dimensionality of the grid
+                if "dims" not in kwargs:
```
keileg, line 278: I think this effectively means dims should not be a kwargs, but a standard argument.

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+                    raise ValueError(
+                        """Provide dimensionality of the grids associated
+                        to the vtu files."""
+                    )
+                dims = kwargs.pop("dims")
+                assert isinstance(dims, list) or isinstance(dims, int)
+                dim = dims[i] if isinstance(dims, list) else dims
+
+                # Whether the grid is a subdomain or interface
+                if "are_subdomain_data" not in kwargs:
```
keileg, line 288: Any reason not to make `are_subdomain_data` a standard argument.

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
         # Parameter to be used in several occasions for adding time stamps.
         self._padding = 6
 
+    def import_from_vtu(
```
keileg, line 185: So the idea is that if I have a `MixedDimensionalGrid`, I can create an `Exporter` and use this function to populare the `mdg` state? I like the idea, and although we can certainly discuss whether this is the right placement of such functionality, I can live with a two-stage approach.

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+                    if is_subdomain_data:
+                        for sd, sd_data in self._mdg.subdomains(
+                            dim=dim, return_data=True
+                        ):
+                            if pp.STATE not in sd_data:
+                                sd_data[pp.STATE] = {}
+                            sd_data[pp.STATE][key] = _toScalarFormat(
+                                value[offset : offset + sd.num_cells], sd
+                            )
+                            offset += sd.num_cells
```
keileg, line 350: This assumes that the ordering of the state in the current `mdg` is the same as that used to generate the vtu file. I guess that will always be true, but it will also be incredibly hard to detect if this is not the case. I'm not sure we can do anything about it, though.

keileg (commented):
Various comments.

keileg:
I think this is interesting. The `Exporter,` in particular after it was reconfigured some months ago, really is two things: A translation of PorePy grids and data to the `meshio` format, and a second layer that uses this translation to write simulation data (time dependent and independent) to `vtu.` This PR is the next step, to use `meshio` to import back from vtu to PorePy (note that this is different from the `meshio` import during mesh construction, the latter is prior to faces and notes being split).

It should also be possible to split the Exporter to take care of, respectively, I/O proper and visualization. That is for me a next step, I'm fine with doing this part first. 

Opinions, @IvarStefansson?

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+        """
+        Import state variables from vtu file.
+
+        Args:
+            keys (string or list of strings): keywords addressing cell data to be transferred
+            file_names (string or list of strings): list of vtu files to be considered
+            kwargs: Optional keyword arguments:
+                automatic (boolean): controlling whether dimensionality of the grids and
+                    whether it is a subdomain or interface grid is read automatically from
+                    the file names; default is True
```
jwboth, line 197:
```suggestion
                    the file names; default is True.
```

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+                dims (int or list of int): compatible with file_names; list of dimensions of
+                    the corresponding grids
```
jwboth, line 199:
```suggestion
                    the corresponding grids.
```

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+                are_subdomain_data (bool or list of bool): comparible with file_names; list
+                    of indicators whether the corresponding grid is a subdomain grid;
+                    default is True
```
jwboth, line 202:
```suggestion
                    default is True.
```

src/porepy/viz/exporter.py
jwboth: You are addressing two issues:
1. The two stage process: Yes, indeed, the `MixedDimensionalGrid` is assumed for now, i.e., no mesh creation from vtu or projection of data from non-matching grids is covered in this addition. For now, I assumed the simplest applications possible.
2. The placement of the importer functionality. Indeed, when starting developing, I started with a separate file. However, in order to double check whether a vtu file is complying with the current `MixedDimensionalGrid`, many of the already predetermined quantities of the `Exporter`  are required. Thereby, as you also commented somewhere else, the `Exporter` becomes more an two-way interface bewteen meshio and porepy.

src/porepy/viz/exporter.py
jwboth: Initially I was thinking of adding a map (dict) allowing to address names previously used in some vtu file, but different than intended to use in the current simulation. However, to better focus on the mere importer functionality, I decided to omit any distractions. For practical use, some discussion could be anyhow helpful, as I also see possibilities to interact with initial times etc. So many different components may come together here. But first we would have to decide what we would want to use this for.

src/porepy/viz/exporter.py
IvarStefansson: What do you mean by "ordering of the state"? Ordering of subdomains? If so, proper subdomain tags could go some way.

IvarStefansson (commented):
Nice! I also support your suggestion to rename from viz to io

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+                    of indicators whether the corresponding grid is a subdomain grid;
+                    default is True.
+
+        Raises:
+            ValueError if some of the data is not compatible with the supposedly corresponding
+                grid.
+        """
+
+        # Aux. method: Inverse of _toVectorFormat, used to define vector-ranged values
+        def _toScalarFormat(
```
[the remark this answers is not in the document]
jwboth: Why not.

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+            Args:
+                value (np.ndarray): input array to be converted
+                grid (pp.Grid or pp.MortarGrid): subdomain or interface to which value
+                    is associated to
+
+            Raises:
+                ValueError if the value array is not compatible with the grid
+            """
+            # Make some checks
+            if not value.size % grid.num_cells == 0:
```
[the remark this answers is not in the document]
jwboth: I added a comment.

You are absolutely right. However, implicitly there is currently no danger for this method to be run on e.g. node data, as the corresponding method is run only for data which has been stored as `cell_data` in a vtu file. When also considering node data, one would have then to extend the main routine and loop over all `node_data` in addition to `cell_data`.

I think this comment can be resolved. I just did want to share my comment first with you

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+                # This line will raise an error if node or face data is exported.
+                raise ValueError("The data array is not compatible with the grid.")
+
+            # Apply flatten with the order compatible with _toVectorFormat.
+            return value.flatten("F")
```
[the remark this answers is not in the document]
jwboth: Done.

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,183 @@ def __init__(
...
+                # one could utilize those, but this does not have to be the case.
+                # Thus, use naming convention of the Exporter for this.
+
+                # Remove ending '.vtu' from file_name, and decompose into pieces
+                # separated by '_'.
+                file_name_pieces = Path(file_name).stem.split("_")
+
+                # If the last two are numbers, the first one denotes the dimension.
+                assert file_name_pieces[-1].isnumeric()
+                dim_pos = -2 if file_name_pieces[-2].isnumeric() else -1
```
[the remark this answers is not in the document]
jwboth: You mean that you do not trust the approach here?

The idea here is that when requesting automatic deduction of dimensionality of the grid and its nature (mortar or subdomain), one can set `automatic=True`. Then these properties are deduced purely from the name of the vtu file, assuming the file uses the same naming convention as used for exporting. Of course, if this is not true, the simple deduction it not possible.

Considering the current naming convention for vtu files when exporting with porepy, grid dimensions are always attached as appendix, and when time dependent data is exported, also the time step is added. (Trouble could be introduced when starting to define own naming conventions including for instance iterations). 

Here it is assumed that the name of a vtu file is of type: name_dim.vtu or name_dim_timestep.vtu. Both dim and timestep are numeric (name can be whatever, it can also contain numbers etc) and this is checked.

When automatic deduction is not trusted, one should set `automatic=False` and provide all extra information through keyword arguments. This may be not the prettiest solution, however, my intention was to ease the process for the standard case.

Any other suggestions?

src/porepy/viz/exporter.py
jwboth: `dims` is only required when all required properties like dimensionality of the grid and nature (mortar or subdomain) corresponding to the vtu file are not deducted automatically. See my comment above on the automatic deduction.

Do you still think, I should promote `dims` to be a standard argument and set it to None by default?

BTW: I would be happy to hear your opinion on how to distribute arguments to standard, **args and **kwargs arguments!

src/porepy/viz/exporter.py
jwboth: Same as for `dims`.

src/porepy/viz/exporter.py
jwboth: Sure. (Note that I am not familiar with pickle). Can't the gmsh file be used to recreate a `MixedDimensionalGrid`. I only used this functionality for `MixedDimensionalGrids` with constraints and not true mortar grids. Or is there a difference between using pickle and reading from gmsh?

src/porepy/viz/exporter.py
jwboth: My impression is that @keileg thinks of the ordering of subdomains. Yes, some unique sorting of subdomains would help to lower the risk. The hope is the second step of the import already lowers the risk in some parts, by checking whether the grids comply to each other. For non-articifical scenarios (for which importing becomes more interesting) it would be a miracle that the multiple subdomains have the same number of cells e.g. I believe the second step would catch this error already.

jwboth:
@keileg  @IvarStefansson After/while working on the review: Do you maybe think some additional tests should be added testing some of the functionalities that are more critical for the import than for the export (e.g., order of subdomains, ...)? At the moment the tests are a mirror of a subset of tests for exporting data.

src/porepy/viz/exporter.py
keileg: Pickle may not be optimal, since it is very particular that the object is loaded into the exact same environment it was made. So if you pickle an `mdg`, pull an update from GH, and then load the pickled file, you risk compatibility errors (which is probably fair from pickle's side, even though we know the grid is the same). Gmsh should be safe, but then you need to do all the splitting again, and this takes time for large domains. I have some other ideas, but this is not the PR to do that.

I suggest we keep this for now, and then revisit import/export later on.

I'm fine with resolving this comment unless @IvarStefansson has other opinions.

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,185 @@ def __init__(
         # Parameter to be used in several occasions for adding time stamps.
         self._padding = 6
 
+    def import_from_vtu(
+        self, keys: Union[str, list[str]], file_names: Union[str, list[str]], **kwargs
+    ) -> None:
+        """
+        Import state variables from vtu file.
```
keileg, line 189:
```suggestion
        Import state variables from vtu file. It is assumed that the vtu file was created using PorePy,
        e.g., that the file names follow PorePy conventions, the mixed-dimensional grid is split
        in the usual way etc.
```

src/porepy/viz/exporter.py
```diff
@@ -180,6 +182,185 @@ def __init__(
...
+                        number_format_notation="e",
+                        ignore_numeric_type_changes=True,
+                    )
+                    == {}
+                )
+
+            # 3rd step: Transfer data. Consider each key separately.
+            for key in keys:
+
+                # Only continue if the key is present in the data
```
keileg, line 333:
```suggestion
                # Only continue if the key is present in the data
                # IMPLEMENTATION NOTE: To also consider node data, add an else below.
```

src/porepy/viz/exporter.py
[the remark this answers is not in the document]
keileg: Fair enough. I added a short comment where this method is called.

src/porepy/viz/exporter.py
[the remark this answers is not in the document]
keileg: I added a comment in the method documentation stating that standard formating of the files is assumed - effectively the files should be created from PorePy. This will be enough for 99% of the cases, and we can't accommodate the final 1%. 

I'm fine with resolving this comment.

src/porepy/viz/exporter.py
keileg: I understand, I'm fine with this solution.

keileg (commented):
Good answers. I added some suggested documentation, but besides, I think I'm fine with this.

jwboth:
A closer look (using a less trivial test with non-constant data) showed that there is something wrong with the conversion of vector valued data. Scalar valued data seems to work as expected.

jwboth:
@keileg @IvarStefansson now this PR became a bit old (even declared 'stale' in the meantime). As step towards restart capabilities I would however expect importer capabilities to play a role. And to not mix different aspects I think considering this PR is still relevant. The question would be what would be missing. I am a bit unsure whether the above issues of the discussions have been fully resolved.

jwboth:
Currently lacking before merge would be possible:
- The standard routines of the old but also updated PR template.
- Deserves a highlight: Updated documentation of the exporter.

src/porepy/viz/exporter.py
jwboth: The updated code does not address the discussed issues mentioned here. It is still further assumed, that the `MixedDimensionalGrid` is provided and one wishes to attach data to the grid read from a complying vtu file, to be checked for compatibility.

src/porepy/viz/exporter.py
jwboth: Note that the compatibility of the vtu grid and the considered `MixedDimensionalGrid` is checked. It is not a full guarantee, but one should be very unlucky if the compatibility succeeds by the ordering is different.

jwboth:
In order to resolve the conversations, I need your feedback @keileg  and @IvarStefansson . No changes have been added to the code to address any of the issues discussed. My sense would be that it is not required, although some extra capability would of course be nice to have. Do you see this differently?

IvarStefansson:
> In order to resolve the conversations, I need your feedback @keileg and @IvarStefansson . No changes have been added to the code to address any of the issues discussed. My sense would be that it is not required, although some extra capability would of course be nice to have. Do you see this differently?
> 
> Moreover, I have not changed neither the name of the folder to 'io', nor the class to `DataIO` or so. I thought, I leave these decision to the core developer group and focus on functionality only.

Could you point me to the specific questions you want my feedback on, please?

jwboth:
> > In order to resolve the conversations, I need your feedback @keileg and @IvarStefansson . No changes have been added to the code to address any of the issues discussed. My sense would be that it is not required, although some extra capability would of course be nice to have. Do you see this differently?
> > Moreover, I have not changed neither the name of the folder to 'io', nor the class to `DataIO` or so. I thought, I leave these decision to the core developer group and focus on functionality only.
> 
> Could you point me to the specific questions you want my feedback on, please?

It is the two conversations the two of you started in an earlier review. I was nto sure whether you have agreed upon each other's answers or whether you expect me to change something.

src/porepy/viz/exporter.py
keileg: Looking at this now, it is an interesting discussion, but off-topic for this PR. We should come back to the question of how to do grid import soon, but not here.

src/porepy/viz/exporter.py
```diff
@@ -131,56 +132,270 @@ def __init__(
...
+                    assert isinstance(are_subdomain_data, bool) or isinstance(
+                        are_subdomain_data, list
+                    )
+                    is_subdomain_data = (
+                        are_subdomain_data[i]
+                        if isinstance(are_subdomain_data, list)
+                        else are_subdomain_data
+                    )
+
+            # 2nd step: Make sure that the vtu file and the current grid are identical.
```
keileg, line 337: Please specify what is meant by identical: Should the number of cells, faces etc. be the same, or also coordinates.

keileg (commented):
I have two small comments regarding documentation. Besides this, I think this will become useful and that we can merge this quite soon.

src/porepy/viz/exporter.py
jwboth: I made it more concrete and detailed.

jwboth:
> I have two small comments regarding documentation. Besides this, I think this will become useful and that we can merge this quite soon.

Thanks for the feedback. I took care of the two doc issues.

I am then waiting for a reaction of the last open conversation.

keileg (approved):
Looks good, thanks. Feel free to merge when you are ready, or let me know if I should take care of it.

Issue #735, opened 2022-09-29
Partial integration of TimeSteppingControl into models


jhabriel:
Integrate the time-stepping class into the models in a non-invasive way. Meaning, that this will be only a partial integration, where all the models will behave exactly the same, but the time control will now be managed by the `tsc` object. 

The integration will aim at modularising the TSC class as much as possible. Therefore, the temporal attributes will be purged from all time-dependent models. A time-dependent model will  be in communication with the `tsc` object (i.e., to access the current time step) only via `run_time_dependent_model()`. 

Milestones:
- [x] Make `iterations` an optional parameter, since it is not really needed for linear systems or constant dt -> Unit tests.
- [x] Raise warning when `iterations` is given and `recompute_solution=True` -> Unit tests.
- [x] Add `time_index` as a public attribute of `tsc` object.
- [x] Add methods to increase time and time index -> Unit tests.
- [x] Purge time-dependent attributes from `SlightlyCompressibleFlow` and `ContactMechanicsBiot` model classes.
- [x] Integrate `tsc` into `run_time_dependent_model()` -> Integration tests
- [x] Update documentation of affected classes

IvarStefansson:
For now, information related to non-linear iterations (counters) will remain in the models, right?

jhabriel:
Yes. This will be most likely the next step.

PR #738, opened 2022-10-04
Partial integration of time-stepping control into models

jhabriel:
## Proposed changes

This PR proposes a non-breaking integration of the `TimeSteppingControl` class into time-dependent models. The PR preserves all previous functionality of the models, including natural values of initial time, final time, and time step when the `params` dictionary does not contain the key `"time_manager"` with value `tsc: pp.TimeSteppingControl`. 

Note that this PR does not incorporate any time adaptation into the models (thus the name partial). This will be the natural next step. This PR solves #735, where more information about the integration can be found.

Two main changes have been introduced:
1. Models no longer have time attributes. Instead, all time-related variables are centralised in `model.tsc`.
2.  `run_time_dependent_model()` and `_run_iterative_model()` from `run_models.py` now use `model.tsc` in the time loops. 

After code review, the following naming-related changes were incorporated to the PR:

- The class `TimeSteppingControl` was renamed to `TimeManager`
- The method `next_time_step()` was renamed to `compute_time_step()`
- The object `tsc` was renamed to `time_manager`

## Types of changes

- [ ] Other: 

## Checklist

### DIFF
--- a/src/porepy/models/contact_mechanics_biot_model.py
+++ b/src/porepy/models/contact_mechanics_biot_model.py
@@ -99,11 +99,19 @@ class ContactMechanicsBiot(pp.ContactMechanics):
     def __init__(self, params: Optional[Dict] = None) -> None:
         super().__init__(params)
 
+        # Time-step control
+        tsc = pp.TimeSteppingControl(
+            schedule=[0, 1],
+            dt_init=1,
+            constant_dt=True
+        )
+        self.tsc: pp.TimeSteppingControl = self.params.get("time_manager", tsc)
+
         # Time
-        self.time: float = 0
-        self.time_step: float = self.params.get("time_step", 1.0)
-        self.end_time: float = self.params.get("end_time", 1.0)
-        self.time_index: int = 0
+        # self.time: float = 0
+        # self.time_step: float = self.params.get("time_step", 1.0)
+        # self.end_time: float = self.params.get("end_time", 1.0)
+        # self.time_index: int = 0
 
         # Temperature
         self.scalar_variable: str = "p"
@@ -196,7 +204,7 @@ class ContactMechanicsBiot(pp.ContactMechanics):
                     {
                         "bc": self._bc_type_mechanics(sd),
                         "bc_values": self._bc_values_mechanics(sd),
-                        "time_step": self.time_step,
+                        "time_step": self.tsc.dt,
                         "biot_alpha": self._biot_alpha(sd),
                         "p_reference": self._reference_scalar(sd),
                     },
@@ -238,7 +246,7 @@ class ContactMechanicsBiot(pp.ContactMechanics):
                     "biot_alpha": alpha,
                     "source": self._source_scalar(sd),
                     "second_order_tensor": diffusivity,
-                    "time_step": self.time_step,
+                    "time_step": self.tsc.dt,
                     "vector_source": self._vector_source(sd),
                     "ambient_dimension": self.mdg.dim_max(),
                 },
@@ -716,7 +724,7 @@ class ContactMechanicsBiot(pp.ContactMechanics):
         ad.local_fracture_coord_transformation_normal = normal_proj
         # Facilitate updates of dt. self.time_step_ad.time_step._value must be updated
         # if time steps are changed.
-        ad.time_step = pp.ad.Scalar(self.time_step, "time step")
+        ad.time_step = pp.ad.Scalar(self.tsc.dt, "time step")
 
     def _force_balance_equation(
         self,
--- a/src/porepy/models/contact_mechanics_biot_model.py
+++ b/src/porepy/models/contact_mechanics_biot_model.py
@@ -96,8 +96,8 @@ class ContactMechanicsBiot(pp.ContactMechanics):
         super().__init__(params)
 
         # Time manager
-        tsc = pp.TimeSteppingControl(schedule=[0, 1], dt_init=1, constant_dt=True)
-        self.tsc: pp.TimeSteppingControl = self.params.get("time_manager", tsc)
+        tsc = pp.TimeManager(schedule=[0, 1], dt_init=1, constant_dt=True)
+        self.tsc: pp.TimeManager = self.params.get("time_manager", tsc)
 
         # Temperature
         self.scalar_variable: str = "p"
--- a/src/porepy/models/contact_mechanics_biot_model.py
+++ b/src/porepy/models/contact_mechanics_biot_model.py
@@ -95,15 +95,9 @@ class ContactMechanicsBiot(pp.ContactMechanics):
     def __init__(self, params: Optional[Dict] = None) -> None:
         super().__init__(params)
 
-        # Time-step control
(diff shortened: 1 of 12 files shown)


src/porepy/models/contact_mechanics_biot_model.py
```diff
@@ -52,11 +52,7 @@ class ContactMechanicsBiot(pp.ContactMechanics):
     adjustment needed is to specify the method create_grid().
 
     Attributes:
-        time (float): Current time.
-        time_step (float): Size of an individual time step
-        time_index (int): Index of current time step. Used/updated in
-            run_time_dependent_model.
-        end_time (float): Time at which the simulation should stop.
+        tsc: Time-stepping control manager.
```
vlipovac, line 55: Regarding documentation of attributes (though there was no discussion so far after my proposal), should we start commenting them in separate docstrings in the constructor?

src/porepy/numerics/time_step_control.py
```diff
@@ -435,16 +450,31 @@ def next_time_step(
 
         return self.dt
 
-    def _adaptation_based_on_iterations(self, iterations: int) -> None:
+    def increase_time(self) -> None:
```
vlipovac, line 453: @jhabriel why are increase_time and increase_time_index public?
Shouldn't only the tsc object call them and handle time increase during the request for the next time step ?

jhabriel: Good question. This was also the way I implanted first, i.e. to advance in time after computing the next time step as part of `next_time_step()`

However, I decided against that approach for two reasons: (1) in the time loop, the simulation time must be increased before solving the model and only then the next time step should be computed, and (2) marching in time should be independent of the way the time step is computed. 

Decoupling these steps could be also useful in situations where we are only interested in knowing what would the next time step be without taking any other action.

src/porepy/models/contact_mechanics_biot_model.py
jhabriel: Yes, I didn't do it since it wasn't settled. But it makes sense to agree on this so new code can already be properly documented.

src/porepy/numerics/time_step_control.py
vlipovac: Okay I understand.

My motivation to combine that comes from the fact that the tsc class has all the information it needs to be able to decide if time should be advanced or not. So the usage of tsc would be simplified to almost one method call.

But I understand that you want to be able to get an outlook on what the next dt would be.

src/porepy/models/run_models.py
```diff
@@ -21,71 +34,85 @@ def run_stationary_model(model, params):
...
-    while model.time < t_end:
-        model.time += model.time_step
-        model.time_index += 1
+
+    # Time loop
+    while model.tsc.time < model.tsc.time_final:
+        model.tsc.increase_time()
+        model.tsc.increase_time_index()
         logger.info(
             "\nTime step {} at time {:.1e} of {:.1e} with time step {:.1e}".format(
```
taj004, line 65: Wouldn't we lose the initial time and time index in the logger, by increasing them before storing the info?

jhabriel: Isn't the logger deprecated @IvarStefansson @keileg? But to your point @taj004, in this case the logger will not store the initial conditions, only the information starting from `time_index = 1`.  If we want to log the information related to the initial conditions without modifying the current implementation of the time loop, we'll have to add a logger.info() at initialisation.

IvarStefansson: I think it's still in use, but defer to @keileg on this one

src/porepy/numerics/time_step_control.py
```diff
@@ -393,7 +397,7 @@ def __repr__(self) -> str:
         return s
 
     def next_time_step(
```
IvarStefansson, line 399: I find the interpretation of "next_time_step" ambiguous at best. Wouldn't "compute_time_step" be more precise?

src/porepy/numerics/time_step_control.py
```diff
@@ -435,16 +450,31 @@ def next_time_step(
...
             iterations: Number of non-linear iterations needed to achieve convergence.
 
-        Raises: Warning if `iterations` > `max_iter`.
+        Raises:
+            ValueError if `iterations` is None.
+            Warning if `iterations` > `max_iter`.
 
         """
 
+        # Sanity check: Make sure iterations is give
```
IvarStefansson, line 473:
```suggestion
        # Sanity check: Make sure iterations is given
```

tests/integration/test_ad_equations.py
```diff
@@ -437,14 +437,16 @@ def _timestep_stepwise_newton_with_comparison(model_as, model_ad):
     model_ad.prepare_simulation()
     model_ad.before_newton_loop()
 
-    model_as.time_index = 0
-    model_ad.time_index = 0
-    t_end = model_as.end_time
+    #model_as.time_index = 0
+    #model_ad.time_index = 0
+    #t_end = model_as.end_time
```
IvarStefansson, line 442: Please remove all old code which has been commented

IvarStefansson (commented):
Looks good. Some comments about naming

src/porepy/models/contact_mechanics_biot_model.py
```diff
@@ -99,11 +95,9 @@ class ContactMechanicsBiot(pp.ContactMechanics):
...
         super().__init__(params)
 
-        # Time
-        self.time: float = 0
-        self.time_step: float = self.params.get("time_step", 1.0)
-        self.end_time: float = self.params.get("end_time", 1.0)
-        self.time_index: int = 0
+        # Time manager
+        tsc = pp.TimeSteppingControl(schedule=[0, 1], dt_init=1, constant_dt=True)
+        self.tsc: pp.TimeSteppingControl = self.params.get("time_manager", tsc)
```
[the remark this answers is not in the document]
jhabriel: I agree. I will rename `TimeSteppingControl` to `TimeManager` and the instantiated object to `time_manager`.

src/porepy/models/run_models.py
```diff
@@ -1,16 +1,29 @@
...
+def run_stationary_model(model, params: dict) -> None:
+    """
+    Run a stationary model.
+
+    Args:
+        model: Model class containing all information on parameters, variables,
+            discretization, geometry. Various methods such as those relating to solving
+            the system, see the appropriate model for documentation.
+        params: Parameters related to the solution procedure. # Why not just set these
+            as e.g. model.solution_parameters.
```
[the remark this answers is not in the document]
jhabriel: Ok!

src/porepy/numerics/time_step_control.py
jhabriel: Agree. I will rename the method.

tests/integration/test_ad_equations.py
jhabriel: Done.

IvarStefansson (approved):

PR #746, opened 2022-10-21
FEAT: Sorting of grids within mdgs

### LINKED ISSUE #629: The replace_grids method from the GridBucket class should preserve the node ordering of the graph
_This issue is rather technical, and you probably won't be affected if you never tried to replace grids_

The method `gb.replace_grids(g_map: Dict)` should preserve the grid ordering from the _old_ grid bucket in the _new_ grid bucket. Currently, this will only occur if _all_ nodes from the graph are replaced. However, if only _some_ nodes are replaced, the node ordering of the graph is lost. This does not qualify as a BUG, since the grids are replaced correctly,  but it might be extremely counterintuitive, in addition, of course, to being computationally expensive if this process has to be performed many times.

I will try to fix this in the future.

IvarStefansson:
## Proposed changes

Sort subdomain and interface grids within a mixed-dimensional grid. Might fix #629 
mdg._subdomain_data is expanded with an id and the mdg is assigned an node_id_counter (and same for interfaces).
Note: After creating the PR, I figured out there might be a more elegant solution using a class attribute counter for pp.(Mortar)Grid. Let's discuss after you've had a brief look, @keileg.

## Types of changes

- [ ] Other: 

## Checklist

- [x] Static typing is included in the update. Not quite. Awaiting some decisions after initial review.


IvarStefansson:
I have cleaned up following our decision to go for grid.id for sorting and fixed tests. @jwboth, could you have a look at changes in exporter and test_vtk? Particularly the hack in _compare_vtu_files. Feel free to skip anything else!

jwboth:
@IvarStefansson Currently, I am not sure, why `test_vtk.py` is failing. Locally, it works on my machine.

Motvated by (non-critical) parts of the error message on GH, I would suggest the following changes (I can also commit but it is so short, that you can also change these quickly):
image

However, I am afraid, this will not change the outcome of pytest. One possibility could be also to `print(diff)` in `_compare_vtu_files` to see what it is that is different. The output should be in general readable for humans.

jwboth:
Great effort to debug `test_vtk.py`! I will have a look at the relevant bits of this PR very soon.

IvarStefansson:
> Great effort to debug `test_vtk.py`! I will have a look at the relevant bits of this PR very soon.

Thanks! And sorry for any notification spam!

src/porepy/grids/md_grid.py
```diff
@@ -439,120 +413,86 @@ def remove_subdomain(self, sd: pp.Grid) -> None:
...
+        inds = self.argsort_grids(subdomains)
 
-        Args:
-            overwrite_existing (bool, optional): If True (default), any existing
-                subdomain ordering will be overwritten.
+        sorted_subdomains = [subdomains[i] for i in inds]
+        return tuple(sorted_subdomains)
 
-        """
+    def sort_interfaces(self, interfaces: List[pp.MortarGrid]) -> List[pp.MortarGrid]:
```
keileg, line 439:
```suggestion
    def sort_interfaces(self, interfaces: list[pp.MortarGrid]) -> list[pp.MortarGrid]:
```

src/porepy/grids/md_grid.py
```diff
@@ -439,120 +413,86 @@ def remove_subdomain(self, sd: pp.Grid) -> None:
...
-            sorted_subdomain_data: The same subdomains, sorted.
-
+            np.ndarray that sorts the grids, e.g.
+                inds = mdg.argsort_subdomains(subdomain_list)
+                sorted_subdomains = [subdomain_list[i] for i in inds]
         """
-        assert all(["node_number" in self._subdomain_data[sd] for sd in subdomains])
-
-        return sorted(subdomains, key=lambda n: self.node_number(n))
+        sort_inds_all: np.ndarray = np.array([], dtype=int)
```
keileg, line 473: Generally, I prefer to build such arrays by making a list of numpy arrays and then to a hstack at the end. This will be more efficient, but it may not be worth the effort to change it here.

src/porepy/grids/md_grid.py
```diff
@@ -659,68 +597,6 @@ def replace_subdomains_and_interfaces(
                 # subdomains by node number
                 del self._subdomain_data[sd_old]
 
-    # ----------- Apply functions to subdomains and interfaces
```
keileg, line 662: I agree, these can go.

src/porepy/grids/md_grid.py
```diff
@@ -745,22 +621,16 @@ def diameter(
...
     def dim_min(self) -> int:
         """Get minimal dimension represented in the mixed-dimensional grid.
 
         Returns:
             int: Minimum dimension of the grids present in the hierarchy.
 
         """
-        return np.min([sd.dim for sd in self.subdomains()])
+        # Access protected attribute instead of subdomains() to avoid sorting, which can
+        # lead to a recursion error.
```
keileg, line 632: Nice.

tests/integration/test_TpfaMonoCoupling.py
```diff
@@ -77,23 +77,16 @@ def generate_grids(self, n, xmax, ymax, split):
 
         g1.compute_geometry()
         g2.compute_geometry()
-        grids = [g2, g1]
+        grids = [g1, g2]
```
keileg, line 80: General comments on tests: I have not reviewed these in detail, let me know if there is something I should look at in particular. 

Generally speaking, I agree with your idea that this PR to some extent defines a new truth for some of these tests.

jwboth (commented):
To me the changes in the exporter and test_vtk look good. Data types changed from int64 to int32 for all data arrays. But due to the use of deepdiff, it should be any issue. The tests are running.

src/porepy/grids/match_grids.py
```diff
@@ -527,12 +527,7 @@ def mdg_refinement(
 
     for i in np.arange(len(subdomains)):
         sd, sd_ref = subdomains[i], subdomains_ref[i]
-
-        node_num, node_num_ref = (
-            mdg.node_number(sd),
-            mdg_ref.node_number(sd_ref),
-        )
-        assert node_num == node_num_ref, "Weakly check that grids refer to same domain."
+        assert sd.id == sd_ref, "Strongly check that grids refer to same domain."
```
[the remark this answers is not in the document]
IvarStefansson: I agree with your reasoning.
1. This has never been well tested, I think. 
2. I agree with the suggestion and the uneasyness.
1. + 2. => Should we move this out of the core? t (not necessarily in this PR)

src/porepy/grids/md_grid.py
```diff
@@ -186,30 +187,16 @@ def interface_to_subdomain_pair(
...
-            if subdomain_indexes[0] < subdomain_indexes[1]:
-                return subdomains[0], subdomains[1]
-            else:
-                return subdomains[1], subdomains[0]
-
-        elif subdomains[0].dim > subdomains[1].dim:
-            return subdomains[0], subdomains[1]
-        else:
-            return subdomains[1], subdomains[0]
+        subdomains: Tuple[pp.Grid, pp.Grid] = self._interface_to_subdomains[intf]
```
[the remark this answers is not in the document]
IvarStefansson: I'll replace all.

src/porepy/grids/md_grid.py
IvarStefansson: Would it be list of arrays or list of ints in this case? I'm inclined to think we should care about efficiency, as this method may be called relatively often.

src/porepy/grids/grid.py
```diff
@@ -128,6 +142,8 @@ def __init__(
         if not (dim >= 0 and dim <= 3):
             raise ValueError("A grid has to be 0, 1, 2, or 3.")
 
+        # Set global counter for all grids:
+
```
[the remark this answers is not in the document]
IvarStefansson: Yes, done

IvarStefansson:
> To me the changes in the exporter and test_vtk look good. Data types changed from int64 to int32 for all data arrays. But due to the use of deepdiff, it should be any issue. The tests are running.

Thanks!

src/porepy/grids/match_grids.py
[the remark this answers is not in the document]
keileg: Agreed, this should be moved somewhere else, but not now. I can make an issue - let's not close this conversion before it is done.

src/porepy/grids/md_grid.py
keileg: List of arrays. The time will be saved generating many small and one big array, instead of progressively larger arrays.

src/porepy/grids/match_grids.py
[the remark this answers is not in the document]
keileg: Done.

keileg (approved):

PR #756, opened 2022-11-03
Fix plotting vector fields

Yuriyzabegaev:
## Proposed changes

The function `pp.plot_grid` now doesn't work if you pass `mdg` and `vector_value: str`. The internal function `_quiver` expects the array of shape (3 x n), but the state contains 1D arrays. However, this visualization works if you pass not `mdg`, but `grid`, and `vector_value: ndarray`. In this case, the user ensures that he provides the array of shape (3 x n). So we have 2 cases of behaviour:

1. `mdg` and `vector_value: str` - fails due to an error. `mdg` and `vector_value: ndarray` is not allowed according to the docstring.
2. `grid` and `vector_value: ndarray` - works.

All the examples use the second case. I made a minor fix so the first case is not failing, and verified that it doesn't break the second case.

I provided the tests for both cases. Also, I provided a test for the function `save_img` as @jhabriel suggested. The tests cover three setups: simplex 2D grid, cartesian 2D grid and cartesian 3D grid. The vector variable is stored at the cell centres everywhere, but I hope it must not affect the behaviour of the plotting function significantly.

## Types of changes

- [ ] Other: 

## Checklist

### DIFF
--- a/src/porepy/viz/plot_grid.py
+++ b/src/porepy/viz/plot_grid.py
@@ -279,11 +279,19 @@ def plot_mdg(
 
         vector_value_array = sd_data.get(pp.STATE, {}).get(vector_value, None)
         if vector_value_array is not None:
-            # The further algorithm requires the vector_value array of shape (3 x n). But
-            # we have a 1D data array. Thus, we first reshape and then fill the remaining dimensions with zeros.
-            vector_value_array = vector_value_array.reshape(sd.dim, -1, order="F")
+            # The further algorithm requires the vector_value array of shape (3 x n).
+            # Now, we have a 1D array.
+            # Thus, we first reshape and then fill the remaining dimensions with zeros.
+            vector_value_array = vector_value_array.reshape(
+                (mdg.dim_max(), -1), order="F"
+            )
             vector_value_array = np.vstack(
-                [vector_value_array, np.zeros((3 - sd.dim, vector_value_array.shape[1]))]
+                [
+                    vector_value_array,
+                    np.zeros(
+                        (3 - vector_value_array.shape[0], vector_value_array.shape[1])
+                    ),
+                ]
             )
         _plot_sd_xd(
             sd,
--- a/src/porepy/viz/plot_grid.py
+++ b/src/porepy/viz/plot_grid.py
@@ -377,7 +377,7 @@ class _Arrow3D(FancyArrowPatch):
         """
         # Render the 3d coordinates of the arrows as preparation for plotting in the 2d plane.
         xs3d, ys3d, zs3d = self._verts3d
-        xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, renderer.M)
+        xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, self.axes.M)
         # Extract the rendered positions in the 2d plotting plane
         self.set_positions((xs[0], ys[0]), (xs[1], ys[1]))
         # Draw the arrows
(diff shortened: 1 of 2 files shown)


src/porepy/viz/plot_grid.py
```diff
@@ -276,10 +276,19 @@ def plot_mdg(
         # Adjust rgb colors depending on the subdomain ordering
         kwargs["rgb"] = np.divide(kwargs.get("rgb", [1, 0, 0]), sd.id + 1)
         # Plot the subdomain and data
+
+        vector_value_array = sd_data.get(pp.STATE, {}).get(vector_value, None)
+        if vector_value_array is not None:
+            # The further algorithm requires the vector_value array of shape (3 x n). but
+            # We have the 1D data array. Thus, we fill the remaining dimensions with zeros.
+            vector_value_array = vector_value_array.reshape(sd.dim, -1, order="F")
```
IvarStefansson, line 284: I have a feeling shape should be (mdg.dim_max(), -1) or perhaps rather (-1, sd.num_cells)

src/porepy/viz/plot_grid.py
```diff
@@ -276,10 +276,19 @@ def plot_mdg(
         # Adjust rgb colors depending on the subdomain ordering
         kwargs["rgb"] = np.divide(kwargs.get("rgb", [1, 0, 0]), sd.id + 1)
         # Plot the subdomain and data
+
+        vector_value_array = sd_data.get(pp.STATE, {}).get(vector_value, None)
+        if vector_value_array is not None:
+            # The further algorithm requires the vector_value array of shape (3 x n). but
```
IvarStefansson, line 282:
```suggestion
            # The further algorithm requires the vector_value array of shape (3 x n). But
```

tests/unit/test_plotting.py
```diff
@@ -0,0 +1,99 @@
...
+)
+def mdg(request):
+    """Helper for initialization of the parametrized mdgs"""
+    mdg_ = request.param[0]
+    _initialize_mdg(mdg_)
+    return mdg_
+
+
+def test_plot_grid_mdg(mdg):
+    """Tests that no error is risen if we plot mdg and provide variable names."""
```
IvarStefansson, line 32:
```suggestion
    """Tests that no error is raised if we plot mdg and provide variable names."""
```

IvarStefansson (commented):
Nice work! One substantial comment, rest is cosmetics.

src/porepy/viz/plot_grid.py
Yuriyzabegaev: (mdg.dim_max(), -1) - what if the domain is 3D, the subdomain and the data are 2D?
(-1, sd.num_cells) - the data can be defined at the edges.

Why don't you prefer my choice - (sd.dim, -1)?

IvarStefansson: The vector is usually nd. E.g. traction on a fracture is a nd-vector, even though the fracture is 2d.
Does the vector plotting work for faces/edges?

keileg: Alternatively, the dimension of the vector field can be made an optional input argument, with the default being (-1, sd.num_cells).

IvarStefansson: Yes, if there's a (not too exotic) use case for that.

Yuriyzabegaev: > The vector is usually nd. E.g. traction on a fracture is a nd-vector, even though the fracture is 2d. Does the vector plotting work for faces/edges?

@IvarStefansson I see your point. I believe, (mdg.dim_max(), -1) must cover all possible cases. The vector plotting was intended to work with faces/edges, see: https://github.com/pmgbergen/porepy/blob/d4cc3adcbecf5401a5b2c2dcc4398e586cf5d9ff/src/porepy/viz/plot_grid.py#L395

@keileg probably, that will be superfluous. Taking mdg.dim_max() will always allow us to understand if the vector is placed in cell centres or edges etc. Or maybe there is some exotic use case I don't see now.

IvarStefansson: I agree. If we ever need it we can always add a vector_dim-argument later.

PR #762, opened 2022-11-09
Fix errors related to mypy v0.99

keileg:
## Proposed changes
After an update, mypy (to v0.99) gave error mesasges at function and method parameters that had the default argument `None`, but were not marked as optional (e.g., `def foo(bar: list = None)` instead of `def(foo: Optional[list] = None)`. This PR resolves all such conflicts.

In addition, the tagging system in the interface between the `FractureNetwork` classes and `Gmsh` (the tags themselves are defined as an enum in the module `GmshInterface`) is used in a more principled manner: We now use numerical tags under geometry processing (this is very convenient when the geometry is defined as numpy arrays, for instance, fractures defined as an array of point indices that are connected), while Gmsh interface mainly uses the Enum form, which is closer to Gmsh, easier to read etc. There are some exceptions to the latter, where the Gmsh interface has to work with numerical tags, but this is now explicitly noted and should be less incomprehensible.

I also tried to introduce some spot-wise improvements in the documentation of how tagging works in `FractureNetwork2d`. I am fully aware that much more comments really are necessary, together with a refactoring of the code (the same applies even more to `FractureNetwork3d`), but do not have the capacity to take this on now.

## Types of changes

- [ ] Other: 

## Checklist

- [ ] This PR does not duplicated existing functionality: No new functionality introduced.
- [ ] The update is covered by the test suite (including tests added in the PR). No new functionality introduced.

### DIFF
--- a/src/porepy/fracs/fracture_network_2d.py
+++ b/src/porepy/fracs/fracture_network_2d.py
@@ -5,7 +5,7 @@ import copy
 import csv
 import logging
 import time
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Optional, Tuple, Union
 
 import meshio
 import numpy as np
@@ -14,12 +14,13 @@ import porepy as pp
 import porepy.fracs.simplex
 from porepy.fracs import tools
 
-from .gmsh_interface import GmshData2d, GmshWriter, Tags
+from .gmsh_interface import GmshData2d, GmshWriter
+from .gmsh_interface import Tags as GmshInterfaceTags
 
 logger = logging.getLogger(__name__)
 
 
-class FractureNetwork2d(object):
+class FractureNetwork2d:
     """Class representation of a set of fractures in a 2D domain.
 
     The fractures are represented by their endpoints. Poly-line fractures are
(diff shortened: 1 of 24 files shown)


src/porepy/fracs/fracture_network_2d.py
```diff
@@ -73,20 +87,23 @@ def __init__(self, pts=None, edges=None, domain=None, tol=1e-8):
...
             self.edges = edges
+
         self.domain = domain
+
         self.tol = tol
 
-        self.tags = {}
-        self.bounding_box_imposed = False
-        self._decomposition = {}
+        self.tags: dict = {}
```
IvarStefansson, line 100: If I understood the new doc style correctly, attributes assigned in init should be followed by a docstring.

src/porepy/fracs/fracture_network_2d.py
```diff
@@ -411,39 +439,65 @@ def prepare_for_gmsh(
...
-
-        # Only lines specified as fractures are defined as physical
-        phys_line_tags = {}
+        # Information about line types is found in the third row of edges
         edge_types = edges[2]
+
+        # Process information about lines that should be tagged as physical by Gmsh.
+        # These are fractures, domain boundaries and auxiliary (constraints).
+        # phys_line_tags is a mapping from line index to the Tag.
+        phys_line_tags: dict[int, GmshInterfaceTags] = {}
```
IvarStefansson, line 463: Isn't it a bit strange that auxiliary is "physical"?

src/porepy/fracs/fracture_network_3d.py
```diff
@@ -504,48 +512,53 @@ def prepare_for_gmsh(self, mesh_args, dfn=False, constraints=None) -> GmshData3d
...
         # used to tag information in the output from gmsh. They are kept as separate
         # variables since a user may want to modify the physical surfaces before
         # generating the .msh file.
-        physical_surfaces: Dict[int, Tags] = {}
-        polygon_tags = np.zeros(len(self._fractures), dtype=int)
+        physical_surfaces: dict[int, GmshInterfaceTags] = dict()
+        polygon_tags: dict[int, GmshInterfaceTags] = dict()
 
         for fi, _ in enumerate(self._fractures):
+            # Translate to from numerical tags to the GmshInterfaceTags system.
```
IvarStefansson, line 539:
```suggestion
            # Translate from numerical tags to the GmshInterfaceTags system.
```

IvarStefansson (approved):

src/porepy/fracs/fracture_network_2d.py
keileg: The physical is `gmsh` terminology; I am afraid we'll have to do translation between terminologies somewhere.

src/porepy/fracs/fracture_network_2d.py
keileg: Documentation updated (this line and other class attributes).

PR #763, opened 2022-11-09
Maint ad docs

vlipovac:
As discussed with @keileg , in order to establish the new workflow with regards to the docs branch, here an update to docstrings and type annotations of some AD classes and functions.
(src code synchronization with docs).

## Types of changes

- [ ] Minor change (e.g., dependency bumps, broken links, etc).
- [ ] Bugfix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] Testing (contribution related to testing of existing or new functionality).
- [X] Documentation (contribution related to adding, improving, or fixing documentation).
- [X] Maintenance (e.g., improve logic and performance, remove obsolete code, etc).
- [ ] Other: 

## Checklist

_Put an `x` in the boxes that apply or explain briefly why the box is not relevant._

- [X] The documentation is up-to-date.	
- [X] Static typing is included in the update.
- [X] This PR does not duplicated existing functionality.
- [ ] The update is covered by the test suite (including tests added in the PR).

PR #768, opened 2022-11-14
Synchronize docs develop

vlipovac:
Manual synchronization of docs with develop until workflow is established.


vlipovac (commented):
This synchronization involves

- a general update to the latest code in src/
- eliminating the discrepancy in return types which was noticed in #763 
- updating code ownership of docs/

IvarStefansson (approved):

Issue #774, opened 2022-11-25
Update requirements to include Sphinx


vlipovac:
For PorePy docs to be buildable using Sphinx, two packages are required

- sphinx
- sphinx-rtd-theme

The first one includes the builder and apidoc.
The second one the standard readthedocs theme template.
Developers only.

OmarDuran:
Is this issue related to a specific GitHub action or another automated process?

vlipovac:
As discussed internally, requirements for the docs will be handled separately.

PR #775, opened 2022-11-25
Update requirements for docs

vlipovac:
## Proposed changes

Include relevant packages for documentation building in `requirements-dev.txt`, see #774 .

## Types of changes

- [X] Minor change (e.g., dependency bumps, broken links, etc).
- [ ] Other: 

## Checklist

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,3 +6,5 @@ mypy >= 0.981
 pytest >= 4.6
 pytest-cov
 pytest-runner
+sphinx
+sphinx-rtd-theme
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,4 +7,3 @@ pytest >= 4.6
 pytest-cov
 pytest-runner
 sphinx
-sphinx-rtd-theme


OmarDuran (commented):
Thank your contribution. I think we should include those dependencies only if they are strictly required on automated builds of the documentation. Otherwise, we can relax on required those dependencies.

IvarStefansson:
> > Thank your contribution. I think we should include those dependencies only if they are strictly required on automated builds of the documentation. Otherwise, we can relax on required those dependencies.
> 
> Hi,
> 
> the automatized building of documentation will certainly need these packages, but they (like any package) can be accessed separately in the .yml files. I though `requirement-dev.txt` also serves as an overview for developer, telling them which packages they need for developing activities?

I agree the -dev should contain what is needed for developing. And documentation is a mandatory part of that.

OmarDuran (approved):

OmarDuran:
> > Thank your contribution. I think we should include those dependencies only if they are strictly required on automated builds of the documentation. Otherwise, we can relax on required those dependencies.
> 
> Hi,
> 
> the automatized building of documentation will certainly need these packages, but they (like any package) can be accessed separately in the .yml files. I though `requirement-dev.txt` also serves as an overview for developer, telling them which packages they need for developing activities?

vlipovac:
> @vlipovac I ran the following test with the local modification on `dockerfiles/base/Dockerfile`: `RUN wget https://raw.githubusercontent.com/pmgbergen/porepy/update-issue-774/requirements.txt && \ wget https://raw.githubusercontent.com/pmgbergen/porepy/update-issue-774/requirements-dev.txt` I have found that for `sphinx-rtd-theme` the base Docker image generation fails because the package is not available from current conda channels. Perhaps we just simply remove `sphinx-rtd-theme` from requirements-dev.txt. What is your opinion on it?

Hei @OmarDuran ,
Jhabriel told me yesterday that I should try that, but I did not have the time.

I agree, even more, let me abort this PR until the automatization is implemented.

jhabriel:
I believe the right move here is to create a `requirements_doc.txt`. Then, people who wants to generate the docs could easily pip install such requirements.

That being said, I suggest we isolate  the docker build from the docs generation, and keep it as it is, i.e., install only packages present in `requirements.txt` and `requirements_dev.txt`.

Note that this is the approach used for numpy, for example: https://github.com/numpy/numpy/blob/main/doc_requirements.txt

jhabriel:
Also, may I suggest to add the `docs` folder to `gitignore`. In this way, we won't have to deal with the dozens of files generated after `make html`.

OmarDuran:
Absolutely agree with all your points.

vlipovac:
I am aborting this PR after a discussion with Jhabriel and Omar. We will implement another solution for this (see Jhabriels comments).

PR #779, opened 2022-12-13
Update polygon constraint algorithm to work with shapely v2

keileg:
## Proposed changes
This is a minor update of the function `geometry.constrain_geometry.lines_by_polygon()`, which uses the dependency `shapely` to find the intersection between line segments and non-convex polygons. After a major upgrade to `shapely`, one of the functions we use changed, and this PR updates the code accordingly. 

The new code is compatible with both the old and new version of Shapely, instead of forcing users to update. The consequences of such an update would likely have been minor to none, but it can be hard to tell. The workaround is properly documented locally in the code.

## Types of changes

- [ ] Other: 

## Checklist

- [x] Static typing is included in the update: No changes needed
- [x] This PR does not duplicated existing functionality. Bugfix only
- [x] The update is covered by the test suite (including tests added in the PR). Bugfix only.

### DIFF
--- a/src/porepy/geometry/constrain_geometry.py
+++ b/src/porepy/geometry/constrain_geometry.py
@@ -73,9 +73,22 @@ def lines_by_polygon(
                 int_pts = np.c_[int_pts, np.array(int_lines.xy)]
                 edges_kept_aslist.append(ei)
         elif type(int_lines) is shapely_geometry.MultiLineString:
-            # consider the case of multiple intersections by avoiding considering
-            # lines on the boundary of the polygon
-            for int_line in int_lines:
+            # Consider the case of multiple intersections by avoiding considering
+            # lines on the boundary of the polygon.
+
+            # NOTE: After updating to shapely 2.0, iteration over the components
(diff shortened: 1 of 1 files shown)


src/porepy/geometry/constrain_geometry.py
```diff
@@ -73,9 +73,22 @@ def lines_by_polygon(
...
+            # Consider the case of multiple intersections by avoiding considering
+            # lines on the boundary of the polygon.
+
+            # NOTE: After updating to shapely 2.0, iteration over the components
+            # of a multiline should call the geoms attribute.
+            # We could have enforced an update for all users, but instead do a
+            # gentler version which should accommodate v1 and v2.
+            import shapely
+
+            if shapely.__version__[0] > "1":
```
IvarStefansson, line 85: This would break on v10, right?

keileg: Yes. However, shapely has moved from v1.6 through 1.8 to 2.0 over the last 6 years, so I think we should be safe.

IvarStefansson (approved):

PR #783, opened 2023-01-03
Docs update grids

vlipovac:
## Proposed changes

@IngridKJ @Yuriyzabegaev  @mariusnevland  and myself have reworked the docstrings of the grids subpackage to the latest style guid.

I have additionally updated all docstrings in the geometry subpackage to some changes we had to do in the porepy style, due to conflict with some sphinx extension. The main point is that we cannot use the brackets after an argument name in the Parameters directive, since they are modified by sphinx. Argument specifications like shape or default value have to be put after the colon.

The style guide will be updated respectively in the coming days.

## Types of changes

- [ ] Other: 

## Checklist


keileg:
Thanks. @IvarStefansson @OmarDuran, I will take care of this one.

keileg:
@vlipovac: I see there are some issues with mypy. I will try to fix them while reviewing the code, but will get in touch if something turns out to be difficult.

keileg:
@vlipovac I fixed the typing issues, and made quite a few minor edits myself. Most of these entailed reformatting comments so that they fit better with the prescribed line length, but I also did a few changes to the docstring, where the formatting had not been updated (most notably geometry.intersections.segments_2d). I am almost sure I did not mess up.

src/porepy/geometry/distances.py
```diff
@@ -607,21 +654,33 @@ def segment_overlap_segment_set(
...
     Returns:
         If ``return_indices==True``, we return a tuple of 2 elements.
 
-        bool:
+        :obj:`bool`:
             ``True`` if the segment overlaps any of the segments in the set, ``False``
             otherwise.
-        ndarray ``(shape=(num_overlapping,))``:
+        :obj:`~numpy.ndarray`: ``shape=(num_overlapping,)``
+
```
keileg, line 683: Have I understood it correctly that you need the blank between `:obj:...` and the description in cases where there is a shape argument, but not otherwise?

keileg (approved):
I left question, that can be addressed here or (perhaps better) in the updated docstring template. 

Besides, great work from everybody involved.

src/porepy/geometry/distances.py
vlipovac: Yes, shape, default or any other additional specifications. The blank/ linea break gives a readable format which distinguishes between the format specification and the content/arg description.

Without the blank I have noticed that Sphinx produces some not-so-nice-looking content containers, especially if the text after the specifications is short and no explicit line break is introduced.

keileg: Thanks.

PR #818, opened 2023-02-20
Update friction bound

mariusnevland:
This PR changes the friction bound from $b=-F\lambda_n$ to $b=-F(\lambda_n+\tilde{c}([[\mathbf{u}]]_n-g))$. The reason for this proposed change is that the latter definition is what is used in Hüeber’s thesis (Discretization techniques and efficient algorithms for contact problems, p.104) and in the paper by Stefansson et al. (A fully coupled numerical model of thermo-hydro-mechanical processes and fracture contact mechanics in porous media, p. 11). 

Note that this new formula for the friction bound is already being used in the normal complementary function in PorePy (equation (39) in Stefansson et al.), and so this change will only affect the definition of the tangential complementary function (equation (40) in Stefansson et al.).

I have done some simple test simulations using the `MomentumBalance` model, comparing the results when using the two friction bounds. My first impression is that for setups where the Newton solver converges quickly, the results are more or less unaffected by the updated friction bound. However, for setups where the Newton solver struggles to converge, the updated friction bound can have a large impact on the results. For instance, in one of my setups, the Newton solver converged (after about 100 iterations) when using the old friction bound, while the Newton solver did not converge even after 1000 iterations when using the new friction bound.

Ivar said that he will review this PR.

- [ ] Other: 

## Checklist

- [ ] The update is covered by the test suite (including tests added in the PR): "test_dual_hybrid_vem_2d_ani_simplex" fails on my computer, but all other tests pass. This test should however be unaffected by the updated friction bound, which I have "confirmed" by noting that this test also fails on the develop branch on my computer.

### DIFF
--- a/src/porepy/models/constitutive_laws.py
+++ b/src/porepy/models/constitutive_laws.py
@@ -2219,6 +2219,12 @@ class FrictionBound:
     This class is intended for use with fracture deformation models.
     """
 
+    displacement_jump: Callable[[list[pp.Grid]], pp.ad.Operator]
+    """Operator giving the displacement jump on fracture grids. Normally defined in a
+    mixin instance of :class:`~porepy.models.models.ModelGeometry`.
+    
+    """
+
     normal_component: Callable[[list[pp.Grid]], pp.ad.Matrix]
     """Operator giving the normal component of vectors. Normally defined in a mixin
     instance of :class:`~porepy.models.models.ModelGeometry`.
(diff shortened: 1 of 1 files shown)


IvarStefansson:
Am I getting this right: The new version is sometimes a factor 10 **worse** than the old one?

mariusnevland:
> Am I getting this right: The new version is sometimes a factor 10 **worse** than the old one?

This was indeed what I observed. However, it turns out that the specific setup that I was running gave rise to an ill-posed problem, with extremely high condition numbers (around 10^17), so this was a bad example on my part. In any case, it seems like we should study the friction bound more closely before making any changes, so I will close this PR.

Issue #835, opened 2023-03-19
Export scaled variables


IvarStefansson:
The DataSavingMixin should take care of scaling variables with its Units.

PR #837, opened 2023-03-21
FEAT: Back-scale variables before exporting

IvarStefansson:
## Proposed changes

This PR provides functionality for back-scaling quantities (including grid coordinates) before exporting them. Also introduces unit tags for primary variables.
Solves #835.

## Types of changes

- [ ] Other: 

## Checklist

- [ ] The update is covered by the test suite (including tests added in the PR). - Not really. I don't know if it's natural to include it.


src/porepy/viz/exporter.py
```diff
@@ -146,6 +147,12 @@ def __init__(
         )
         """Flag controlling whether constant data is exported to a separate file."""
 
+        self._length_scale: float = length_scale
+        """Length scale for the grid. All coordinates are multiplied by this value
+        before exporting.
+
+        """
```
jwboth, line 154: OK. Though, using a 'dim'-dimensional array wouldn't be such a big deal either, thinking of overly thin media for instance. But this can also be postponed until someone sees need for it.

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -46,19 +53,76 @@ def finalize_data_saving(self) -> None:
...
+                )
+
+        # We combine grids and mortar grids. This is supported by the exporter, but not
+        # by the type hints in the exporter module. Hence, we ignore the type hints.
+        return data  # type: ignore[return-value]
+
+    def _evaluate_and_scale(
+        self,
+        grid: Union[pp.Grid, pp.MortarGrid],
+        method_name: str,
```
jwboth, line 106: Intuitively, I would have chosen a name as `quantity` to link to the physical nature of the object, but `method_name` seems OK as well.

jwboth (commented):
Exporter and DataSavingMixin look good to me. Just two small non-crucial side comments.

jwboth:
General comment regarding tests. A natural test would be to run two scenarios: one with and one without rescaling. If the output is exported in (rescaled) SI units, it should be expected that the outputs of both scenarios are close, assuming no numerical challenges due to bad scaling.

src/porepy/viz/exporter.py
IvarStefansson: I think your suggestion is an interesting one. But I don't think it's relevant in the context of this PR. Doing this on a run script level with inheritance from Exporter should be very doable in something like 30 lines.

src/porepy/viz/data_saving_model_mixin.py
IvarStefansson: I see, but I also think it's important to convey that this name corresponds to a method (of self). I am open to changing it, though, if somebody has a brilliant suggestion (ideally covering both our conserns).

jhabriel (approved):
I pushed a commit with a more transparent explanation of the resulting units of `interface_darcy_flux_variable`. I think all the units of primary variables are physically correct.

We should tackle in the future the problem of misleading names of Darcy fluxes , since as we have it now, the aren't really fluxes. But that's a matter for a different PR.

Great work and thanks for this feature :)

src/porepy/models/energy_balance.py
```diff
@@ -468,18 +475,23 @@ def create_variables(self) -> None:
         self.equation_system.create_variables(
             self.temperature_variable,
             subdomains=self.mdg.subdomains(),
+            tags={"si_units": "K"},
```
keileg, line 478: Excellent! Let's try to gradually follow up.

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -46,19 +53,76 @@ def finalize_data_saving(self) -> None:
...
+                    )
+                data.append(
+                    (
+                        sd,
+                        "specific_volume",
+                        self._evaluate_and_scale(
+                            sd, "specific_volume", f"m^{self.nd-sd.dim}"
+                        ),
+                    )
+                )
```
keileg, line 97: This will be very useful.

src/porepy/viz/data_saving_model_mixin.py
keileg: No suggestion from my side, I think method name is least likely to cause confusion.

src/porepy/viz/exporter.py
```diff
@@ -112,6 +112,7 @@ def __init__(
         grid: Union[pp.Grid, pp.MixedDimensionalGrid],
         file_name: str,
         folder_name: Optional[str] = None,
+        length_scale: float = 1.0,
```
keileg, line 115: Maybe I have gotten lost in the mixins, but should we (is it possible to) initialize model.exporter with the length scale? I'm not sure if things are composed so that this is possible, e.g., if the scaling is set sufficiently early.

IvarStefansson: Yes, this happens in initialize_data_saving of DataSavingMixin

PR #839, opened 2023-03-23
Initialization of models from vtu/pvd

jwboth:
## Proposed changes

Restart capabilities are helpful when crashed or stopped simulations shall be resumed from a previous state. This mostly saves time for costly simulations. This PR has been developed in close collaboration with @enricoballini 

The proposes restart capabilities are light-weight and relatively basic. By providing a simple set of options via the standard parameters dictionary for models, a restart from pvd file can be invoked. The last time step exported to file will be used to initialize the states of a model. In addition, the physical time and last used time step size is read from file and communicated to the time manager allowing to revisit a relatively close simulation setup as before. In addition, the exporter is informed about the current time step index, allowing to resume writing to the same pvd file, also resuming the export of data in the same manner as performed before stopping the simulation previously.

For the proposed strategy to work, updated pvd files are exported at every application of export of vtu files. In addition physical times and time step sizes are exported to a json file.

Tutorial: In my opinion the integration test pretty much displays the entire functionality to the extent of interest of the user. I do not see a point in providing a tutorial for this functionality.

IMPORTANT NOTE: As part of the (integration) testing, it has become evident that it is important that the restart files are created within the same system as used. The ordering of the mesh nodes e.g. can vary (this happened when generating reference files locally and compare to the Exporter's meshio interpretation of the grid online). The integration test here therefore generated the files itself.

What is not included in this PR:
- Split of Exporter in two objects or renaming. The Exporter is currently two things: A translator from and to meshio format. And I/O tool for data (and grid, only output).
- Functionality to read geometry specification from vtu - only data I/O is fully supported. 
- Functionality to address specific time steps besides the last available - targeting for now restarts of crashed simulations.
- Functionality compatible with multi-step time discretizations.
- A foreseeing action to take care of failing time steps while keeping track of the history of times and time steps sizes. This should be revisited when resuming PR #752 . 
- An advanced test is lacking for restarting when not exporting at all time steps, yet the code is intended to work here as well.

## Types of changes

- [ ] Other: 

## Checklist


jwboth:
@keileg Pay extra attention to the changes in `SolutionStrategy`. The changes may lead to different simulation results in previous simulations of md-poromechanics.

@jhabriel To keep you posted, mild changes of the TimeManager have been added to keep track of the history. At the moment only the history of exported times is of interest, thus the update of the history is integrated in routines solely called by the `DataSavingMixin`. A simple test was added as well. As discussed, I would not be suprised if parts of this update should be updated when you resume PR #752

keileg:
Question: To the list of not-covered, should we add read a geometry specification from a set up vtu-files?

tests/integration/models/test_restart.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def create_fractured_setup(
+    solid_vals: dict, fluid_vals: dict, uy_north: float, restart: bool
+):
+    """Create a setup for a fractured domain.
+
+    This is an enhanced copy of .test_poromechanics.create_fractured_setup.
+    It enables multiple time steps, and the export of the solution.
+
```
keileg, line 33: Is there a way in which you can import and enhance, rather than copy? The latter more than doubles the maintenance workload and increases the chance of errors.

tests/integration/models/test_restart.py
```diff
@@ -0,0 +1,151 @@
...
+    reference_times = json.load(reference_times_json)
+    for key in ["time", "dt"]:
+        assert np.all(
+            np.isclose(np.array(restarted_times[key]), np.array(reference_times[key]))
+        )
+    restarted_times_json.close()
+    reference_times_json.close()
+
+    # Remove temporary visualization folder
+    shutil.rmtree("./visualization")
```
keileg, line 146: Thanks!

tests/integration/models/test_restart.py
```diff
@@ -0,0 +1,151 @@
...
+    # In order to use the data as restart and reference data, move it
+    # to a reference folder.
+    pvd_files = list(Path("./visualization").glob("*.pvd"))
+    vtu_files = list(Path("./visualization").glob("*.vtu"))
+    json_files = list(Path("./visualization").glob("*.json"))
+    for f in pvd_files + vtu_files + json_files:
+        dst = Path(current_dir) / Path("restart_reference") / Path(f.stem + f.suffix)
+        shutil.move(str(f), str(dst))
+
+    # Now use the reference data to restart the simulation.
```
keileg, line 106: Which time step will this restart from?

tests/integration/models/test_restart.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # To verify the restart capabilities, test whether:
+    # - the states have been correctly initialized at restart time;
+    # - the follow-up time step has been computed correctely;
+    # - the overall solution stored in a global pvd file is compiled correctly;
+    # - the logging of times and step sizes is correct.
+
+    for ending in ["000001", "000002"]:
+        for i in ["1", "2"]:
+            assert _compare_vtu_files(
```
keileg, line 118: What will a failure tell us? When will we need to recompute the reference files?

tests/unit/test_time_step_control.py
```diff
@@ -513,8 +516,8 @@ def test_warning_iteration_is_greater_than_max_iter(self, iterations):
         time_manager = pp.TimeManager([0, 1], 0.1, iter_max=10)
         warn_msg = (
             f"The given number of iterations '{iterations}' is larger than the maximum "
-            f"number of iterations '{time_manager.iter_max}'. This usually means that the solver "
-            f"did not converge, but since recompute_solution = False was given, the "
+            f"number of iterations '{time_manager.iter_max}'. This usually means that the "
+            f"solver did not converge, but since recompute_solution = False was given, the "
             f"algorithm will adapt the time step anyways."
```
keileg, line 521: Please restructure, no line should extend beyond 88 characters.

tests/unit/test_time_step_control.py
```diff
@@ -635,3 +638,32 @@ def test_update_time_index(self, schedule, dt_init, is_constant, time_index):
         time_manager.time_index = time_index
         time_manager.increase_time_index()
         assert time_manager.time_index == (time_index + 1)
+
+    def test_io_history(self):
+        """Check I/O functionality."""
```
keileg, line 643: Please add a short comment regarding what a failure here would signify.

tests/unit/test_vtk.py
```diff
@@ -299,6 +380,83 @@ def test_mdg(setup):
         )
 
 
+@pytest.mark.parametrize("case", np.arange(2))
+def test_restart_mdg(setup, case):
```
keileg, line 384: What does `case` control?

src/porepy/models/solution_strategy.py
```diff
@@ -173,6 +186,14 @@ def initial_condition(self) -> None:
         vals = np.zeros(self.equation_system.num_dofs())
         self.equation_system.set_variable_values(vals, to_iterate=True, to_state=True)
 
+        # Overwrite the initial condition in case of a restart
```
keileg, line 189: If a child method first calls this method through `super` and then does some modifications (see for instance contact forces in contact mechanics), one would have to overwrite `initial_condition` in the child method? I think this should be better documented.

src/porepy/models/solution_strategy.py
```diff
@@ -332,7 +354,7 @@ def after_nonlinear_failure(
 
     def after_simulation(self) -> None:
         """Run at the end of simulation. Can be used for cleanup etc."""
-        self.finalize_data_saving()
+        pass
```
keileg, line 357: Why was this removed?

keileg (commented):
Various comments. Also, I think we can move to using only `Path` to represent file paths.

keileg (changes requested):
Various comments. Also, I think we can move to using only `Path` to represent file paths.

tests/integration/models/test_poromechanics.py
```diff
@@ -50,51 +50,52 @@ def initial_condition(self):
...
-            to_iterate=True,
-        )
-        sd, sd_data = self.mdg.subdomains(return_data=True)[0]
-        # Initial displacement.
-        if len(self.mdg.subdomains()) > 1:
-
-            top_cells = sd.cell_centers[1] > 0.5
-            vals = np.zeros((self.nd, sd.num_cells))
-            vals[1, top_cells] = self.fluid.convert_units(0.042, "m")
+        if self.time_manager.is_init():
```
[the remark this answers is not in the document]
jwboth: I will have to put my answer ti this on hold. I had guessed that the answer to your question is "yes" and one could omit the conditional which I believe I added early in the test-driven development. However, when removing it (and this is directly connected or maybe even more reasoned by the use of super().initial_condition() which also invokes the restart. Since I have not the required overview of the implications what which state changes in the discretization etc. I will have to go in more details and debug again. I have a slightly bad feeling that there is a high risk in the current approach that one is not doing what intended if not testing the restart for each particular model, because it seems the effects can be quite hidden. I will report back.

src/porepy/models/solution_strategy.py
jwboth: After finishing answerign to your questions, I want to put this question on hold. I will have to debug some attempt to disconnect the restart from initial condition. Because I absolutely see that one has to be very careful what one is doing if not applying conditionals as time_manager.is_init() as in test_poromechanics. However, the switch from the combined use of initial condition and restart (despite the use of is_init) results in a different simulation result and I have aboslutely no clue why. So I will have to investigate and I will report further later. My feeling is though a cleaner separation for more robustness (instead of a comment which not every user will know of) will be the better solution.

jwboth (commented):
Thanks for your comment @keileg . I tred to answer to the best of my ability. Here and there I requested some updated review. I am leaving 2 comments for me to work on, because I am not understanding why changes are not working.

During the review, I felt test coverage for restarting is not perfect (restarting from mdg pvd files and vtu files is not tested in the integration test). I am suggesting we decide whether these functionalities should be supported or we just reduce to a single case (and I remove accordingly code and flags) if this simplifies the understanding. I was aiming at a quite general use.

I had trouble with github yesterday, so I am using a review to answer. Sorry for that.

src/porepy/models/solution_strategy.py
```diff
@@ -332,7 +354,7 @@ def after_nonlinear_failure(
 
     def after_simulation(self) -> None:
         """Run at the end of simulation. Can be used for cleanup etc."""
-        self.finalize_data_saving()
```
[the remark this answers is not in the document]
keileg: Thanks.

src/porepy/numerics/time_step_control.py
```diff
@@ -644,3 +649,71 @@ def is_schedule_in_simulated_times(
...
+        self.dt_history.append(
+            int(self.dt) if isinstance(self.dt, np.integer) else float(self.dt)
+        )
+
+        # Storing as json
+        out_file = open(path if path is not None else "visualization/times.json", "w")
+        json.dump({"time": self.time_history, "dt": self.dt_history}, out_file)
+        out_file.close()
+
+    def load(self, path: Optional[str] = None) -> None:
```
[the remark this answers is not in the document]
keileg: I like it

src/porepy/numerics/time_step_control.py
```diff
@@ -644,3 +649,71 @@ def is_schedule_in_simulated_times(
...
+        """Keep track of history of time and time step size and store.
+
+        Mirrors self.write().
+
+        Parameters:
+            path: specified path for retrieving time and dt.
```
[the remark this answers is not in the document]
keileg: Let's move to Path only. We will have to get used to this.

src/porepy/models/solution_strategy.py
```diff
@@ -173,6 +211,31 @@ def initial_condition(self) -> None:
         vals = np.zeros(self.equation_system.num_dofs())
         self.equation_system.set_variable_values(vals, to_iterate=True, to_state=True)
 
+        # Overwrite states from file if restart is enabled.
+        if self.restart_options.get("restart", False):
+            if "pvd_file" in self.restart_options:
```
keileg, line 216: I think the default `restart_option` will have `"pvd_file"` as a key, but with a `None` value.

src/porepy/models/solution_strategy.py
```diff
@@ -131,12 +132,49 @@ def __init__(self, params: Optional[dict] = None):
...
+                "times_file": None,
+                # Path to json file containing evolution of exported time steps and
+                # used time step size at that time. If 'None' a default value,
+                # defined in :class:`~porepy.numerics.time_step_control.TimeManager
+                # is internally used.
+                "time_index": -1,
+                # Index addressing history in times_file; only relevant if "vtu_files"
+                # is not 'None' or "is_mdg_pvd" is 'True'. The index corresponds to
+                # the single time step vtu/pvd files. If not provided, internally
+                # assumed to address the last time step in times_file.
```
keileg, line 162: Thanks for the thorough documentation.

src/porepy/models/solution_strategy.py
```diff
@@ -131,12 +132,49 @@ def __init__(self, params: Optional[dict] = None):
...
         """Time manager for the simulation."""
 
+        self.restart_options: dict = params.get(
+            "restart_options",
+            {
+                "restart": False,
+                # Boolean flag controlling whether restart is active. Internally
+                # assumed to be False.
+                "pvd_file": None,
+                # Path to pvd file collecting either multiple time steps (generated
```
keileg, line 142: Is this a `Path`?

src/porepy/numerics/time_step_control.py
```diff
@@ -644,3 +650,84 @@ def is_schedule_in_simulated_times(
             schedule[ss + 1], sim_times, rtol=rtol, atol=atol
         )
         return schedule.size == np.sum(in1d)
+
+    # I/O
+    def write_time_information(self, path: Optional[Union[Path, str]] = None) -> None:
```
keileg, line 655: As mentioned elsewhere (I believe), let's just use Path here.

src/porepy/viz/exporter.py
```diff
@@ -217,16 +220,108 @@ def __init__(
...
+            else:
+                subdomain_vtu_files.append(str(root / Path(vtu_file)))
+
+        # Import from vtu
+        self.import_from_vtu(subdomain_vtu_files, keys, are_subdomain_data=True)
+        self.import_from_vtu(interface_vtu_files, keys, are_subdomain_data=False)
+
+        return time_index
+
     def import_from_vtu(
```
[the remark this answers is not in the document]
keileg: Let's do the renaming.

src/porepy/models/solution_strategy.py
jwboth: You are right. I changed to a slightly different conditional, testing for both its presence (if the restart options are defined by the user wanting to restart from vtu files, it may happen that the keyword pvd_file is not provided, or set to None, which are both OK here.)

src/porepy/models/solution_strategy.py
```diff
@@ -131,12 +132,49 @@ def __init__(self, params: Optional[dict] = None):
...
+                # through pp.Exporter.write_pvd()); or a pvd file associated to a single
+                # time step (generated through pp.Exporter._export_mdg_pvd()).
+                "is_mdg_pvd": False,
+                # Boolean flag controlling whether prescribed pvd file is a mdg pvd
+                # file, i.e., created through Exporter._export_mdg_pvd(). Otherwise,
+                # it is assumed, the provided pvd file originates from
+                # Exporter.write_pvd(). If not provided, assumed to be False.
+                "vtu_files": None,
```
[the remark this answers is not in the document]
jwboth: It is meant to be a list of Paths or a single Path (useful only in the case of a equi-dimensional setting). `vtu_files` is only considered if `pvd_file` is not provided or is None. I am expecting the option to restart from vtu files to be less important. Therefore I simply chose None here. Would you instead wish to have something like `[Path(".")]` as default argument?

src/porepy/models/solution_strategy.py
jwboth: No, you are right. The main motivation to provide `pvd_file` here was to discuss the keywords. The default restart options hold anyhow an inactive state. Thus no keyword will be visited. Nevertheless, to comply with the expected typing, one could also use `Path(".")` here. Would this be better?

jwboth:
@keileg After some debugging of ideas and code, I realized something you most likely already indicated in your review.

The method `initial_condition` is dangereous and the restart should most likely not mixed up with setting the initial condition. Through the use of Mixin and cross-use of initial condition across many different models, it is not clear that the restart is applied at the correct time. To add more security, I therefore suggest separating the actual reset/restart from initial_condition. A new subroutine in solution strategy has been added: `reset_state_from_file` (I did not spent much time on finding a good name - it is more about the concept right now).

A simple separating is however also not sufficient - I am not super happy with that, but I assume a good discussion without someone knowing all models and how they setup initial conditions and their implications would be very helpful. This PR somewhat only tests the functionality for the poromechanics case.

One example where extra care is needed are initializing time-dependent boundary conditions for the poromechanics model. They need to be set, when the actual physical time has been loaded, which is part of the restart mechanism. Thus, such a routine has to be repeated after calling `reset_state_from_file`.

And I can think of similar examples, where it is not sufficient to just read the state, but in general parameters that lag in time (e.g. Darcy flux) and/or are time dependent (other sources).

Regarding the boundary conditions, my feeling is that it may be better to aim at finding a general philosphy how PP deals with time dependent boundary conditions and sources across all models instead of fixing such issues for specific classes right now.

tests/integration/models/test_poromechanics.py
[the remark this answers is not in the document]
jwboth: After separating restart from intial_condition, I agree that this was too hardcoded and should not be the foreseen practice.

src/porepy/models/solution_strategy.py
jwboth: With the most recent design for how to restart a simulation, this discussion becomes obsolete. I therefore resolve this conversation. And open a new one below.

src/porepy/models/solution_strategy.py
```diff
@@ -131,26 +137,64 @@ def __init__(self, params: Optional[dict] = None):
...
+        # but prior to data initialization.
+        self.initialize_data_saving()
 
         # Set variables, constitutive relations, discretizations and equations.
         # Order of operations is important here.
         self.set_equation_system_manager()
         self.set_materials()
         self.create_variables()
         self.initial_condition()
+        self.reset_state_from_file()
```
jwboth, line 190: @keileg I want to reopen a discussion I believe is closely related to some of your concerns you raised about the design. I want to stress the separation of initial_condition and restart. I believe in this way the user will not have to be careful in how to define initial conditions, though the user still needs to consider tailoring `reset_state_from_file` for his/her purposes.

src/porepy/models/poromechanics.py
```diff
@@ -190,6 +190,16 @@ def initial_condition(self) -> None:
...
+    def reset_state_from_file(self) -> None:
+        """Reset states but through a restart from file.
+
+        Add treatment of boundary conditions to the standard reset of states from file.
+
+        """
+        super().reset_state_from_file()
+
+        self.update_time_dependent_ad_arrays(initial=True)
+
```
jwboth, line 202: Tightly connected to the reopened discussion in another discussion. This is a demonstration of the need to tailor `reset_state_from_file` depending on the model. In this case, by separating time dependent boundary condition treatment from intial_condition, one could get rid of this case.

keileg: Who is `super` in this context?

src/porepy/models/solution_strategy.py
keileg: So, the idea is to always use the standard initialization, then always run the reset function, but leave it to the individual models (both standard and as changed by the user) to decide what to do? Then the standard models will presumably implement a reset method which does the natural things, and which can be activated by the reset parameters?

src/porepy/models/solution_strategy.py
keileg: I think `None` is fine, but the expected type should be mentioned in the documentation.

src/porepy/models/solution_strategy.py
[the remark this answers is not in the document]
keileg: `None` is fine, but please document the expected format.

src/porepy/models/solution_strategy.py
jwboth: I believe the answer here is yes, without anticipating everything possible.

src/porepy/models/poromechanics.py
jwboth: Of course my sight is limited. It is SolutionStrategy. One should think this certainly through and consider what happens in case of models of which multiple ones require a tailored reset routine. My current thinking is that resetting in principle mimicks initial_condition, and if initial_condition works out with all it's "super" calls. A similar setup will also work for resetting. After all, resetting requires: setting states, boundary conditions, parameters. Similar to initial_condition.

src/porepy/numerics/time_step_control.py
```diff
@@ -458,6 +459,10 @@ def increase_time_index(self) -> None:
         """Increase time index counter by one."""
         self.time_index += 1
 
+    def is_init(self) -> bool:
+        """Check whether time is equal to initial simulation time."""
+        return np.isclose(self.time, self.time_init)
```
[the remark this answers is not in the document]
jwboth: After separating initial_condition and restart, `is_init` disappeared from the code base. Only an integartion test is using (was using a simple "less than" and a tolerance before). I suggest I remove it again.

src/porepy/models/solution_strategy.py
```diff
@@ -215,6 +236,7 @@ def discretize(self) -> None:
         # discretization process can be achieved by exploiting knowledge of the equation
         # system (e.g., which terms are linear and need not be discretized at every
         # iteration).
+        self.set_discretization_parameters()
```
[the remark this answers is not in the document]
jwboth: #848 should overwrite this change, and it should have the same effect for the test considered here. Once #848 is merged in main, I will update this PR accordingly.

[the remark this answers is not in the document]
jwboth: With the recent changes in #848 this line is now called at every nonlinear time step before `rediscretize` is called. Thus, it can be safely removed.

keileg (approved):
Looks good. Let's merge once the tests pass, and then possibly revisit once we have some experience with how it works in practice.

PR #854, opened 2023-04-11
Stress_sign

### LINKED ISSUE #849: Negative defined mechanics submatrix
Hei,

I noticed that the matrix arising from the mechanics problem in PorePy is negatively defined. It has a dominant negative diagonal and gives a negative scalar when multiplied by a random vector from left and right. That's alright if we solve the mechanics block separately, but it might lead to suboptimal solver performance if we couple mechanics with something else. For example, the single-phase flow problem matrix is positively defined.

Providing the code to reproduce:

```
import matplotlib.pyplot as plt
import porepy as pp

model = pp.poromechanics.Poromechanics()
pp.run_stationary_model(model, {})
mat, rhs = model.linear_system
plt.matshow(mat.A)
plt.colorbar()
```
Result:
image

IvarStefansson:
## Proposed changes

Change sign of the surface term (stress) when assembled in MomentumBalance. Also improve treatment of fluid flux vector sources at interfaces.
Resolves #849.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/momentum_balance.py
+++ b/src/porepy/models/momentum_balance.py
@@ -151,7 +151,10 @@ class MomentumBalanceEquations(pp.BalanceEquation):
 
         """
         accumulation = self.inertia(subdomains)
-        stress = self.stress(subdomains)
+        # By the convention of positive tensile stress, the balance equation is
+        # accumulation - stress = body_force. The balance_equation method will *add* the
+        # surface term (stress), so we need to multiply by -1.
+        stress = pp.ad.Scalar(-1) * self.stress(subdomains)
         body_force = self.body_force(subdomains)
         equation = self.balance_equation(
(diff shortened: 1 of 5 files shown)


src/porepy/models/momentum_balance.py
```diff
@@ -151,7 +151,10 @@ def momentum_balance_equation(self, subdomains: list[pp.Grid]):
 
         """
         accumulation = self.inertia(subdomains)
-        stress = self.stress(subdomains)
+        # By the convention of positive tensile stress, the balance equation is
+        # accumulation - stress = body_force. The balance_equation method will *add* the
```
keileg, line 155:
```suggestion
        # acceleration - stress = body_force. The balance_equation method will *add* the
```

keileg (approved):
Looks good.

PR #872, opened 2023-04-24
Bug in front end function for mdg generation

keileg:
## Proposed changes
When the user-facing method `pp.create_mdg` was used to generate structured grids, the domain of the fracture network was never imposed on the fracture network. This had two negative consequences:
* If the fracture network was used later, for instance to add wells to the geometry, the domain was not available contrary to expectations.
* Fractures that were outside the domain could be projected to coincide with the boundary of the domain, which is not really meaningful in the way fractures are implemented in PorePy.

This PR introduces the following changes:
* The domain is now imposed also in structured meshing. Fractures outside the domain are deleted, and a warning is given (it is not clear the latter is the best option, let me know)
* Tests have been added to check that deletion is done correctly.

Also added:
* The class `pp.Domain` now has a `__contains__` method which allows for testing if a point is inside a domain.
* The same domain class has somewhat better test coverage. Please check that you agree with the placement of that test.

## Types of changes

- [ ] Other:

## Checklist

- [ ] NOT RELEVANT If new skipped tests have been introduced in this PR, `pytest` was run with the `--run-skipped` flag.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,6 @@ numpy
 requests
 scipy
 shapely
-shapely[vectorized]
 six
 sympy
 typing_extensions
--- a/src/porepy/fracs/fracture_network_2d.py
+++ b/src/porepy/fracs/fracture_network_2d.py
@@ -630,7 +630,6 @@ class FractureNetwork2d:
         )
 
     def _find_intersection_points(self, lines: np.ndarray) -> np.ndarray:
-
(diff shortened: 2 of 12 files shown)


OmarDuran (approved):
According to what @keileg  and I discussed, it has been accepted. In addition, the docstring documentation was properly formatted thanks to @solveigste  and @vlipovac.

keileg:
Thanks for helping out, @OmarDuran, @vlipovac and @solveigste!

PR #890, opened 2023-05-10
allow to create a grid from a domain described by a polytope (2d only)

alessiofumagalli:
## Proposed changes

When a pp.Domain is made of lines (polytope) the resulting grid is not created from its bounding box but by respecting the geometry. To allow for general shapes we assume that the lines are ordered. This functionality has been expanded only in 2d.

## Types of changes

- [ ] Other:

## Checklist

- [ ] Static typing is included in the update. NOT NEEDED
- [ ] The update is covered by the test suite (including tests added in the PR). I CAN ADD A TEST IF REQUESTED
- [ ] If new skipped tests have been introduced in this PR, `pytest` was run with the `--run-skipped` flag. NO NEW SKIPPED TESTS


IvarStefansson:
Please update the PR description briefly explaining the changes.

alessiofumagalli:
> Please update the PR description briefly explaining the changes.

The description has been updated

src/porepy/fracs/fracture_network_2d.py
```diff
@@ -829,6 +829,8 @@ def impose_external_boundary(
             domain: ``default=None``
 
                 Domain specification. If not provided, :attr:`domain` will be used.
+                When the domain is given as a set of lines create the grid respecting
```
jhabriel, line 832:
```suggestion
                If the domain is not a box, we assume the set of lines defining
                the polygon to be ordered either clockwise or counter-clockwise.
```

src/porepy/fracs/fracture_network_2d.py
```diff
@@ -829,6 +829,8 @@ def impose_external_boundary(
...
+                the lines.
```
jhabriel, line 833:
```suggestion
```

jhabriel (commented):
Hi @alessiofumagalli. Thanks a lot for the PR. I think a lot of people can benefit from this functionality.

However, before approving the PR, could you please add a unit test?

The test can be placed as a method here: https://github.com/pmgbergen/porepy/blob/4cc6dcc67877d9394f21158670af0c38d191aae5/tests/unit/test_fracture_network.py#L129

The test should check that the fracture network is correctly generated for a non-boxed domain (perhaps a pentagon or similar) with and without fractures.

Please, let me know if you have any questions.

alessiofumagalli:
Dear all, I've added few test cases to validate the proposed code. Please, let me know if I need to do something else. Thanks a lot, best regards

alessiofumagalli:
Dear all, I've added few test cases to validate the proposed code. Please, let me know if I need to do something else. Thanks a lot, best regards

tests/unit/test_fracture_network.py
```diff
@@ -314,6 +313,79 @@ def test_add_networks_preserve_tags(self):
...
+
+        sd_1d = mdg.subdomains(dim=1)[0]
+        self.assertTrue(np.isclose(sd_1d.num_cells, 2))
+        self.assertTrue(np.isclose(sd_1d.num_faces, 3))
+        self.assertTrue(np.isclose(sd_1d.num_nodes, 3))
+
+        sd_1d = mdg.subdomains(dim=1)[1]
+        self.assertTrue(np.isclose(sd_1d.num_cells, 2))
+        self.assertTrue(np.isclose(sd_1d.num_faces, 3))
+        self.assertTrue(np.isclose(sd_1d.num_nodes, 3))
```
keileg, line 387: Sorry, but this test will not work. The number of cells is not invariant if Gmsh changes its algorithms. Also, does such a acceptance criterion really pinpoint the part of the code you want to test?

tests/unit/test_fracture_network.py
```diff
@@ -314,6 +313,79 @@ def test_add_networks_preserve_tags(self):
...
+    def test_pentagon_domain_with_fracs(self):
+        line_1 = np.array([[0, 1], [0, 0]])
+        line_2 = np.array([[1, 1], [0, 1]])
+        line_3 = np.array([[1, 0.5], [1, 0.5]])
+        line_4 = np.array([[0.5, 0], [0.5, 1]])
+        line_5 = np.array([[0, 0], [1, 0]])
+        lines = [line_1, line_2, line_3, line_4, line_5]
+        domain = pp.Domain(polytope=lines)
+
+        pts = np.array([[0.2, 0.8], [0.6, 0.6]])
```
keileg, line 366: If I am not mistaken, this fracture will be partly outside the domain. Is this on purpose? If so, please document.

tests/unit/test_fracture_network.py
```diff
@@ -314,6 +313,79 @@ def test_add_networks_preserve_tags(self):
...
+        edges = np.array([[0], [1]])
+        fracs = pts_edges_to_linefractures(pts, edges)
```
keileg, line 368: Why not just create a `pp.LineFracture`?

tests/unit/test_fracture_network.py
```diff
@@ -314,6 +313,79 @@ def test_add_networks_preserve_tags(self):
         known_tags = np.array([tag1, tag1, tag2, tag2])
         self.assertTrue(np.all(together._edges[2] == known_tags))
 
+    def test_triangle_domain_no_fracs(self):
```
keileg, line 316: Please provide a few lines at the top of the test describing what is being tested and what a failure of the test will tell us.

tests/unit/test_fracture_network.py
```diff
@@ -314,6 +313,79 @@ def test_add_networks_preserve_tags(self):
...
+        line_3 = np.array([[0, 0], [1, 0]])
+        lines = [line_1, line_2, line_3]
+        domain = pp.Domain(polytope=lines)
+
+        mesh_size = 0.5
+        mesh_kwargs = {"mesh_size_frac": mesh_size}
+        mdg = pp.create_fracture_network(domain=domain).mesh(mesh_kwargs)
+        sd = mdg.subdomains(dim=2)[0]
+
+        self.assertTrue(np.isclose(sd.num_cells, 7))
```
keileg, line 328: See comments below regarding such acceptance criteria (I somehow placed them on the last test)

keileg (commented):
I think we need to work a bit more on the tests:
1. It is not clear to me what is being tested, and what a failure of this specific test will (most likely) signify. This is invaluable for debugging.
2. Counting the cell numbers is not a good criterion, it basically shows that i) Meshing did not crash, and ii) Gmsh has not changed since the test was written. The latter is bound to fail and the former may or may not test what you really want (cf. point 1.) Are there better ways of doing this?

jhabriel:
I agree with Eirik. But I think this has an easy fix with arguably simpler tests.

I think we are testing two things here: (1) That the fracture network is created with a non-boxed domain, and (2) That the fracture network can be meshed correctly, returning the expected `mdg`. I would suggest adding the tests for mesh generation as methods here: https://github.com/pmgbergen/porepy/blob/161bfe14245650ed5eda3434ae445d0f914ad4a5/tests/integration/test_fracture_mesh_generation.py#L27

Now, with regards to the tests themselves.

For (1), we need to check that the domain given as parameter for instantiating the `FractureNetwork2d` class is identical to the FractureNetwork2d.domain _after_ the `impose_external_boundary` has been applied. Note that `pp.Domain` has an `__eq__` method that will now be very useful. Simply put it, what we want to assert is `domain == fn.domain`, where `fn` is an instance of `FractureNetwork2d`.

For (2), we need to check that the mixed-dimensional grid is mesh correctly. Here, we should test that (a) Gmsh doesn't crash, (b) that the boundaries of the highest-dimensional grid match with the boundaries of the `domain`, and (c) the number of fractures. I don't think we need an `assert` for (a), meaning that the if the `mdg` is a consequence of Gmsh doesn't crashing.

Does this sound feasible?

alessiofumagalli:
Thanks a lot for your comments, I'm going to update the test accordingly

alessiofumagalli:
Dear all, I've been updated the tests according to your suggestions. Now they rely less on the actual number of elements given by gmsh but rather on what @jhabriel suggested. Let me know what do you think. Ciao ciao

tests/unit/test_fracture_network.py
alessiofumagalli: I added few lines for each test to briefly describe it

tests/unit/test_fracture_network.py
alessiofumagalli: It is the case, the documentation has been added

tests/unit/test_fracture_network.py
alessiofumagalli: Tests are now improved (I hope) according to @jhabriel suggestions

tests/unit/test_fracture_network.py
alessiofumagalli: thanks, now it has been used

tests/unit/test_fracture_network.py
alessiofumagalli: thanks, now I'm checking something that should be robust with respect to the resulting grid

keileg:
Thanks for the updated test, @alessiofumagalli, this is very useful. I'll make a few (mainly cosmetic) changes to the test before merging, but there is no need for further efforts from your side.

keileg (approved):

keileg:
Again, thanks for the contribution @alessiofumagalli!

keileg:
And sorry that it took a while to finalize this.

PR #893, opened 2023-05-12
FIX: Convection bc hint improved

Yuriyzabegaev:
## Proposed changes

This is a slight improvement of the hint of how to set the Dirichlet boundary condition for the convection term in the energy balance equation. Right now, it may cause errors because it doesn't warn that one must divide by viscosity. Dividing by viscosity is required since later these values are multiplied by the Darcy flux. It is lacking viscosity, so in the mass balance equation we [multiply](https://github.com/pmgbergen/porepy/blob/161bfe14245650ed5eda3434ae445d0f914ad4a5/src/porepy/models/fluid_mass_balance.py#L213) it by the mobility (inversed viscosity). Thus, we must do the same here. @keileg is notified about this change.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/energy_balance.py
+++ b/src/porepy/models/energy_balance.py
@@ -679,17 +679,15 @@ class BoundaryConditionsEnergyBalance:
         # Loop over subdomains to collect boundary values
         for sd in subdomains:
             vals = np.zeros(sd.num_faces)
-            # If you know the boundary temperature, do something like:
-            # boundary_faces = self.domain_boundary_sides(sd).all_bf
-            # vals[boundary_faces] = (
-            #     self.fluid.specific_heat_capacity()
-            #     * dirichlet_values_temperature
-            #     * self.fluid.density()
-            #     / self.fluid.viscosity()
-            # )
-            # Dividing by viscosity is an implementation detail since later these
-            # values are multiplied by `DarcysLaw.darcy_flux`. Those must be divided by
-            # viscosity from the outside to represent the Darcy velocity.
+            # If you know the boundary temperature, you might want to reconstruct
+            # the enthalpy flux at the boundary. Using Dirichlet BCs, the value assigned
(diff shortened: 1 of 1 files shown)


src/porepy/models/energy_balance.py
```diff
@@ -679,12 +679,18 @@ def bc_values_enthalpy_flux(self, subdomains: list[pp.Grid]) -> pp.ad.DenseArray
         # Loop over subdomains to collect boundary values
         for sd in subdomains:
             vals = np.zeros(sd.num_faces)
-            # If you know the boundary temperature, do something like: boundary_faces =
-            # self.domain_boundary_sides(sd).all_bf vals[boundary_faces] = (
-            # self.fluid.specific_heat_capacity()
-            # * dirichlet_values
-            # * self.fluid.density()
+            # If you know the boundary temperature, do something like:
```
IvarStefansson, line 682: For the reasons I explained in person, I think we should replace the comment by:

If you know the boundary temperature, you might want
to reconstruct the enthalpy flux at the boundary. Using Dirichlet
BCs, the value assigned herein will be multiplied with the Darcy
flux -K \nabla p. Thus, one would normally convert to fluid flux
multiplying by (density/viscosity). Then, if using EnthalpyFromTemperature,
one would compute enthalpy value by multiplying by specific heat
capacity and (boundary temperature - reference temperature).
Note that using non-default constitutive laws in the interior
might require changes to the above recipe.

keileg: Sounds good to me.

IvarStefansson: Will you implement please, @Yuriyzabegaev, so we can merge this PR?

IvarStefansson (approved):
Thanks, @Yuriyzabegaev! Please feel free to merge as soon as the tests have passed.

PR #907, opened 2023-05-26
Extend boundary grid functionality

### LINKED ISSUE #903: Store boundary data in the `mdg`.
There is no data storage for the boundary grids in the mdg now. We need to implement it in the same manner as for the `MortarGrid`s and regular grids.

Yuriyzabegaev:
## Proposed changes

This PR closes #903. It adds several methods to the mdg in order to work with boundary grids in the same manner as we work with subdomains and interfaces. Some things to notice:

It changes the place of creation of the boundary grids. Previously, they were created outside the mdg, in the function `subdomains_to_mdg`. Now they are created inside the class in order to maintain the agreement between the subdomains and their boundary grids. Thus, if the subdomain is registered in the mdg, the corresponding boundary grid is created automatically. The same applies to subdomain removal. The subdomain can be changed in the process of creation, for instance, when two intersecting fractures are divided. To avoid inconsistency, I added the method `compute_geometry` to the boundary grid, which fetches the data from the parent subdomain. It is called in the mdg only after the call to `compute_geometry` of the subdomains. At that point, all the changes must be made.

I covered the new functionality with new tests and modified some oldtests. In those tests that I modified, I updated the style from `unittest` to `pytest`, as we are leaning towards the latter.

Before the PR, the `BoundaryGrid` class was inherited from `Grid`, which was subject to change according to the note in the docstring. I discarded this inheritance and restored the lacking features. Luckily, there were not many of them: 
* Tracking the id. Now the ids of boundary grids are separate from the subdomain ids, the same as for interface ids.
* Ability to sort grids with `mdg.argsort_grids`. I changed the method signature to support this.

The latter change may suggest us to append `BoundaryGrid` to the `GridLike` type. I did not do it now, because `BoundaryGrid` lacks several attributes, such as `num_faces`. But from what I see now, it can be a good generalization when we start updating the AD in order to support boundary grids.

I removed several redeclarations of `GridLike`. Now they all reference one declaration.

I initialize the data storage of boundary grids at the same place as it is done for the other grid types - in the `EquationSystem.create_variable`. However, since boundary variables will not be the actual variables, I do not create ITERATEs and STATEs for them. Instead, I set an empty dict for each variable as a placeholder. Maybe when we start implementing boundary grid variables, it will be more clear how to store the values.

I changed the line `pp.grids.standard_grids.md_grids_2d.single_horizontal` to just `single_horizontal` and added the proper import to satisfy vscode type checker. Not sure why it can not resolve it how it was.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/fracs/meshing.py
+++ b/src/porepy/fracs/meshing.py
@@ -83,6 +83,10 @@ def subdomains_to_mdg(
     # will be added to the mixed-dimensional grid.
     create_interfaces(mdg, node_pairs)
 
+    # Set projections to the boundary grids (this must be done after having
+    # split the fracture faces, or else the projections will have the wrong dimension).
+    mdg.set_boundary_grid_projections()
+
     if time_tot is not None:
         logger.info(
             "Mesh construction completed. Total time " + str(time.time() - time_tot)
--- a/src/porepy/fracs/meshing.py
+++ b/src/porepy/fracs/meshing.py
@@ -82,10 +82,6 @@ def subdomains_to_mdg(
     # we are ready to create mortar grids on the interface between subdomains. These
     # will be added to the mixed-dimensional grid.
     create_interfaces(mdg, node_pairs)
-    # Add boundary grids to the MixedDimensionalGrid (this must be done after having
-    # split the fracture faces, or else the projection to the boundary grids will have
-    # the wrong dimension).
-    _add_boundary_grids(mdg)
 
     if time_tot is not None:
         logger.info(
@@ -534,22 +530,6 @@ def _assemble_mdg(
                 sd_pair_to_face_cell_map[(hsd, lsd)] = face_cell_map
 
     return mdg, sd_pair_to_face_cell_map
-
-
-def _add_boundary_grids(mdg: pp.MixedDimensionalGrid) -> None:
-    """Add boundary grids to a mixed-dimensional grid.
-
-    The grid is added to the dictionary
-    ``~porepy.grids.md_grid.MixedDimensionalGrid._subdomain_boundary_grids``
-
-    Parameters:
-        mdg: The mixed-dimensional grid where the boundary grids are added.
-
-    """
-    for sd in mdg.subdomains():
-        if sd.dim > 0:
-            # Generate a boundary grid for this grid
-            mdg._subdomain_boundary_grids[sd] = pp.BoundaryGrid(g=sd)
 
 
 def create_interfaces(
--- a/src/porepy/grids/boundary_grid.py
+++ b/src/porepy/grids/boundary_grid.py
@@ -63,7 +63,7 @@ class BoundaryGrid:
         grid. Initialized in :meth:`~compute_geometry`.
 
         """
-        self._projs: sps.csr_matrix
+        self._projections: sps.csr_matrix
         """Projection matrix from the parent grid to the boundary grid.
 
         Initialized in :meth:`~compute_geometry`.
@@ -92,7 +92,7 @@ class BoundaryGrid:
         self.cell_centers = self._parent.face_centers[:, parent_boundary]
 
         sz = self.num_cells
-        self._proj = sps.coo_matrix(
+        self._projections = sps.coo_matrix(
             (np.ones(sz), (np.arange(sz), np.where(parent_boundary)[0])),
             shape=(self.num_cells, self._parent.num_faces),
         ).tocsr()
@@ -106,7 +106,7 @@ class BoundaryGrid:
         which maps face-wise values on the parent grid to the boundary grid.
 
         """
-        return self._proj
+        return self._projections
 
     @property
     def id(self):
@@ -135,7 +135,8 @@ class BoundaryGrid:
                 f"Boundary grid of dimension {self.dim} "
                 f"containing {self.num_cells} cells.\n"
                 f"ID of parent grid: {self._parent.id}.\n"
-                f"Dimension of the projection from the parent grid: {self._proj.shape}."
+                "Dimension of the projection from the parent grid: "
+                f"{self._projections.shape}."
             )
 
         return (
--- a/src/porepy/grids/boundary_grid.py
+++ b/src/porepy/grids/boundary_grid.py
@@ -60,30 +60,30 @@ class BoundaryGrid:
         """Number of cells in the boundary grid.
 
         Will correspond to the number of faces on the domain boundary in the parent
-        grid. Initialized in :meth:`~compute_geometry`.
+        grid. Initialized in :meth:`~set_projections`.
 
         """
         self._projections: sps.csr_matrix
         """Projection matrix from the parent grid to the boundary grid.
 
-        Initialized in :meth:`~compute_geometry`.
+        Initialized in :meth:`~set_projections`.
 
         """
         self.cell_centers: np.ndarray
         """Cell centers of the boundary grid.
 
         Subset of the face centers of the parent grid. Initialized in
-        :meth:`~compute_geometry`.
+        :meth:`~set_projections`.
 
         """
 
-    def compute_geometry(self) -> None:
-        """Compute geometric quantities for the boundary grid.
+    def set_projections(self) -> None:
+        """Set projections from the parent grid and set the corresponding attributes.
 
-        Currently, there are no heavy computations here. However, the parent grid can
-        be modified during its construction, and the boundary grid must reflect these
-        changes. Thus, it is required to call this method after calling
-        :meth:`Grid.compute_geometry`. Usually, this is done by the MDG.
+        The parent grid can be modified during its construction, and the boundary grid
+        must reflect these changes. It is required to call this method after having
+        split the fracture faces in order to finish the initialization of
+        the boundary grid.
 
         """
         parent_boundary = self._parent.tags["domain_boundary_faces"]
@@ -107,6 +107,11 @@ class BoundaryGrid:
 
         """
         return self._projections
+
+    @property
+    def parent(self) -> pp.Grid:
+        """Subdomain associated with this boundary grid."""
+        return self._parent
 
     @property
     def id(self):
--- a/src/porepy/grids/boundary_grid.py
+++ b/src/porepy/grids/boundary_grid.py
@@ -3,6 +3,9 @@ boundary of a domain in the form of a grid. The intention is to use this class t
 boundary conditions in cases where constitutive laws are defined on the boundary.
 
 """
+from __future__ import annotations
+
+from itertools import count
 from typing import Optional
 
 import numpy as np
@@ -11,7 +14,7 @@ import scipy.sparse as sps
 import porepy as pp
 
 
-class BoundaryGrid(pp.Grid):
+class BoundaryGrid:
     """A grid representing the boundary of a domain.
 
     The BoundaryGrid is generated from a parent grid, and represents those faces in the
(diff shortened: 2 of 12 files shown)


src/porepy/grids/md_grid.py
```diff
@@ -396,6 +466,13 @@ def add_subdomains(self, new_subdomains: Union[pp.Grid, Iterable[pp.Grid]]) -> N
             # Add the grid to the dictionary of subdomains with an empty data dictionary.
             self._subdomain_data[sd] = {}
 
+        # Generate boundary grids for each subdomain.
```
IvarStefansson, line 469: @keileg, I believe you worked on this recently. Please have a look. Note that BoundaryGrid does not inherit from Grid anymore.

src/porepy/numerics/ad/equation_manager.py
```diff
@@ -10,13 +10,12 @@
 import scipy.sparse as sps
 
 import porepy as pp
+from porepy.utils.porepy_types import GridLike
```
IvarStefansson, line 13: Thanks!

src/porepy/numerics/ad/equation_system.py
```diff
@@ -399,57 +399,69 @@ def create_variables(
...
+        if subdomains is not None and interfaces is None:
+            grids = subdomains
+        elif subdomains is None and interfaces is not None:
+            grids = interfaces
+        elif subdomains is None and interfaces is None:
+            raise ValueError(
+                "Cannot create variable not defined on any subdomain or interface."
+            )
+        else:
+            # Yura: The current behavior is just to ignore the interfaces, which is
```
IvarStefansson, line 416: I agree this looks strange, but what do you mean by "ignore interfaces"?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -399,57 +399,69 @@ def create_variables(
...
 
-        if grids:
-            for grid in grids:
-                if subdomains:
-                    assert isinstance(grid, pp.Grid)  # mypy
-                    data = self.mdg.subdomain_data(grid)
-                else:
-                    assert isinstance(grid, pp.MortarGrid)  # mypy
-                    data = self.mdg.interface_data(grid)
+        def initialize_boundary_data(sd: pp.Grid):
```
IvarStefansson, line 425: I'm not convinced this is the right place to initialize boundary data. Let's discuss!

tests/unit/test_equation_system.py
```diff
@@ -116,9 +118,18 @@ def test_variable_creation():
         sys_man.num_dofs() == ndof_subdomains + ndof_single_subdomain + ndof_interface
     )
 
+    # Check that the variables registered in the boundary data storage.
```
IvarStefansson, line 121: Missing verb?
```suggestion
    # Check that the variables are registered in the boundary data storage.
```

tests/unit/test_md_grid.py
```diff
@@ -293,33 +322,91 @@ def test_subdomain_and_interface_iterators(self):
...
+        assert all(found.values())
+
+        # Finally, check the boundaries
+        parent_grids_found = {sd_1: False, sd_2: False, sd_3: False}
+        for bg in mdg.boundaries():
+            parent_sd = mdg.boundary_grid_to_subdomain(bg)
+            parent_grids_found[parent_sd] = True
+        assert all(parent_grids_found.values())
+
+    def test_iterators_dim_keyword(self):
```
IvarStefansson, line 344: Could you add some documentation to this test, please?

tests/unit/test_md_grid.py
```diff
@@ -293,33 +322,91 @@ def test_subdomain_and_interface_iterators(self):
...
+        sd_1 = MockGrid(1)
+        sd_2 = MockGrid(2)
+        sd_3 = MockGrid(3)
+        sd_4 = MockGrid(2)
+        for sd in [sd_1, sd_2, sd_3, sd_4]:
+            mdg.add_subdomains(sd)
+
+        intf_12 = mock_mortar_grid(sd_2, sd_1)
+        intf_32 = mock_mortar_grid(sd_3, sd_2)
+        all_interfaces = set([intf_12, intf_32])
```
IvarStefansson, line 355: Unused variable!

src/porepy/fracs/meshing.py
```diff
@@ -82,10 +82,6 @@ def subdomains_to_mdg(
     # we are ready to create mortar grids on the interface between subdomains. These
     # will be added to the mixed-dimensional grid.
     create_interfaces(mdg, node_pairs)
```
IvarStefansson, line 84: @keileg, note changes in this file.

IvarStefansson (commented):
Awesome work! Mainly minor comments. Let's discuss once @keileg has had a look.

src/porepy/grids/boundary_grid.py
```diff
@@ -56,24 +51,50 @@ def __init__(self, g: pp.Grid, name: Optional[str] = None) -> None:
...
+        self.dim = g.dim - 1
+        """Dimension of the boundary grid."""
+
+        self.num_cells: int
+        """Number of cells in the boundary grid.
+
+        Will correspond to the number of faces on the domain boundary in the parent
+        grid. Initialized in :meth:`~compute_geometry`.
+
+        """
```
IvarStefansson, line 65: I'm tempted to suggest self._projections

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: If the user passes both subdomains and interfaces, the line `subdomains if subdomains else interfaces` will silently discard the interfaces.

IvarStefansson: But I think passing both subdomains and interfaces is not allowed. If so, the old line `assert subdomains is not None or interfaces is not None` should probably have been an xor. In your changed implementation, I guess this corresponds to replacing the elif on 411 to else, right?

tests/unit/test_md_grid.py
Yuriyzabegaev: Fixed

tests/unit/test_md_grid.py
Yuriyzabegaev: Fixed

src/porepy/grids/boundary_grid.py
Yuriyzabegaev: Ok

tests/unit/test_equation_system.py
Yuriyzabegaev: Fixed

src/porepy/grids/md_grid.py
keileg: I actually moved the BoundaryGrid construction out of this method, because of the issue @Yuriyzabegaev hints to regarding splitting of faces when fracture intersections are identified. I agree that doing the update inside `MixedDimensionalGrid` is more natural (hence my original placement of the BC-grid generation was the same as suggested here), but I am not convinced about the solution of adding it to geometry computation:
1. The purist case, this is topology, not geometry.
2. More fundamentally, this is not an operation I would expect from a computation of geometry, it breaks the single responsibility principle.
Is it an option to have a separate method `mdg.update_boundary_topology()` or similar, and call this after having split the faces?

src/porepy/fracs/meshing.py
keileg: I think I covered this with my comment to `md_grid` , at least that was my intention. Let me know if I failed, or if we should discuss.

src/porepy/grids/boundary_grid.py
```diff
@@ -11,7 +14,7 @@
 import porepy as pp
 
 
-class BoundaryGrid(pp.Grid):
+class BoundaryGrid:
```
keileg, line 17: I agree this is a fair simplification for now. Then let's revisit once we start using this grids in earnest in the models.

src/porepy/grids/boundary_grid.py
```diff
@@ -56,24 +51,50 @@ def __init__(self, g: pp.Grid, name: Optional[str] = None) -> None:
...
+        Subset of the face centers of the parent grid. Initialized in
+        :meth:`~compute_geometry`.
 
         """
 
-        if g.dim == 0:
-            raise ValueError("Boundary grids are not supported for 0d grids.")
-        self.dim = g.dim - 1
-        """Dimension of the boundary grid."""
+    def compute_geometry(self) -> None:
```
keileg, line 80: As stated elsewhere, this is topology, not geometry. Consider renaming the method?

src/porepy/grids/md_grid.py
```diff
@@ -297,11 +336,29 @@ def subdomain_to_boundary_grid(self, sd: pp.Grid) -> Optional[pp.BoundaryGrid]:
...
-            domain boundary, None is returned.
+            The boundary grid associated with this subdomain. ``None`` is returned if
+            ``sd`` is not in this mixed-dimensional grid, or the dimension of ``sd`` is
+            0.
 
         """
-        return self._subdomain_boundary_grids.get(sd, None)
+        return self._subdomain_to_boundary_grid.get(sd, None)
+
+    def boundary_grid_to_subdomain(self, bg: pp.BoundaryGrid) -> Optional[pp.Grid]:
```
keileg, line 346: Depending on how we use the boundary grids in the model, we may or may not need this method. If it turns out we don't need it, let's try to remember to purge it again. We have fought hard to keep the model classes relatively clean.

It may be that it is obvious we will need this; if so, please bear with me.

keileg (commented):
Some more minor comments. At least one can be resolved after reading.

src/porepy/grids/md_grid.py
Yuriyzabegaev: I have used this method in tests several times, for instance [here](https://github.com/pmgbergen/porepy/blob/e85d18487f19a72e71ab43de4a65904ff96d9b90/tests/unit/test_equation_system.py#L125). I think it'd be natural to make the property `_parent` of `BoundaryGrid` public through getter since we anyway store the related subdomain there. Then we would get rid of this method in the mdg.

Yuriyzabegaev: We agreed on making `_parent` public.

src/porepy/grids/boundary_grid.py
Yuriyzabegaev: Renaming it

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: We decided to keep it here for now during the discussion.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: Added the error there.

src/porepy/grids/md_grid.py
Yuriyzabegaev: Added the separate method.

src/porepy/grids/md_grid.py
```diff
@@ -590,9 +655,19 @@ def compute_geometry(self) -> None:
         for intf in self.interfaces():
             intf.compute_geometry()
 
+    def set_boundary_grid_projections(self):
```
keileg, line 658:
```suggestion
    def set_boundary_grid_projections(self) -> None:
```

src/porepy/grids/md_grid.py
```diff
@@ -733,7 +819,7 @@ def dim_max(self) -> int:
         """
         # Access protected attribute instead of subdomains() to avoid sorting, which can
         # lead to a recursion error.
-        return np.max([sd.dim for sd in self._subdomain_data.keys()])
+        return max([sd.dim for sd in self._subdomain_data.keys()], default=0)
```
keileg, line 822: Any particular reason to have a default, and set this default value? I can imagine getting a 0 here will give trouble, e.g., for calculations of grid refinement levels.

keileg:
I added two minor comments. Sorry for the late introduction of these, I didn't spot them before.

src/porepy/grids/md_grid.py
keileg, line 822:
```suggestion
        return max([sd.dim for sd in self._subdomain_data.keys()])
```

keileg (approved):
I took care of the last few questions. Again, thanks for the thorough effort.

Issue #891, opened 2023-05-11
Lacking access to convert_units when creating grids 


jwboth:
At the moment, when creating grids, under the use of customized units (via pp.Units), it is not possible to use the convenient functionality `self.solid.convert_units(<val>, "m")`. See as example the tutorial single_phase_flow.ipynb. In order to fix this, the method `set_materials()` has to be called before `set_geometry()` within `SolutionStrategy.prepare_simulation()`. After the change, the tutorials and other places in the core should be also updated to suggest consistent use of convert_units.

PR #914, opened 2023-06-02
891 lacking access to convert units when creating grids

IngridKJ:
## Proposed changes

Solves issue  #891. Now the method `convert_units` can be used during grid creation. The core and tutorials (where it was possible) are updated to use, for intance, `self.solid.convert_units(val, "m")` instead of `val / self.units.m`.

In practice the change was only calling `self.set_materials` before `self.set_geometry` within `prepare_simulation` method in solution_strategy.py.

## Types of changes

- [ ] Other:

## Checklist


src/porepy/models/geometry.py
```diff
@@ -80,8 +80,7 @@ def set_domain(self) -> None:
         Override this method to define a geometry with a different domain.
 
         """
-        size = 1 / self.units.m
-        self._domain = nd_cube_domain(2, size)
+        self._domain = nd_cube_domain(2, 1.0 / self.units.m)
```
IvarStefansson, line 83:
```suggestion
        self._domain = nd_cube_domain(2, self.solid.convert_units(1.0, "m"))
```

IvarStefansson (changes requested):
Thanks for fixing the issue. Please fix the seaborn issue, then we can merge.

IvarStefansson (approved):
Nice work. Thanks for taking the trouble to fix the seaborn issues!

PR #915, opened 2023-06-02
STY: Add missing return type in method signature

jhabriel:
## Proposed changes

Small PR that adds a missing annotation for the return type of the method `interface_darcy_flux_equation` from the `DarcysLaw` mixing class and modifies the return type for `fluid_compressibility` from `pp.ad.Scalar` to the more general `pp.ad.Operator`.

## Types of changes

- [ ] Other:

## Checklist


src/porepy/models/constitutive_laws.py
```diff
@@ -431,7 +431,7 @@ class FluidDensityFromPressure:
 
     """
 
-    def fluid_compressibility(self, subdomains: list[pp.Grid]) -> pp.ad.Scalar:
```
IvarStefansson, line 434: Shouldn't we keep it as specific as possible?

jhabriel: I agree. But this means that we should change all the signatures of the methods belonging to `ConstantX`-type mixin classes. Since this method was the only one that returned a `pp.ad.Scalar`, I assumed there were other reasons for using `pp.ad.Operator` in the other methods.

I'm OK with either option as long as we are consistent. What do you think?

IvarStefansson: You're right. I think the idea of using Operator was to indicate that the methods may be extended/overridden to return an Operator. Opinions, @keileg? What's the most principled solution?

keileg: I recall the same rationale that you indicate: To allow for general inheritance, the return type should be an operator.

PR #918, opened 2023-06-07
Test numerical values returned by methods from the single phase flow model

jhabriel:
## Proposed changes

This PR introduces an integration test that compares known numerical values against returned values from all possible testable methods from the single-phase flow model. A testable method is one that has either `subdomains` or `interfaces` as input parameter and returns either a `pp.ad.Operator` or a `pp.ad.DenseArray`.

The current test includes checks for all 27 such methods associated with the single-phase flow model. In addition, a helper method that gets all testable methods for a given model was included in `setup_utils.py`. Note that it is required that all testable methods match _exactly_ with all tested methods.

Extension to other models should be tedious but straightforward. The recipe to follow is clear: 

1. Fix a minimal geometry.
2. Use "real" physical parameters to set up an instance of the model being tested.
3. Set non-trivial (but constant) states for the primary variables.
4. Prepare to simulate (but do not run the model).
5. Write the test for all testable methods with hard-coded expected values.

Full coverage of the models will require one integration test per model. Once this is obtained, we could think of removing the file `test_constitutive_laws.py`.

## Types of changes

- [ ] Other:

## Checklist


tests/integration/models/setup_utils.py
```diff
@@ -661,3 +658,46 @@ def compare_values(
         # Add a small absolute tolerance to avoid problems with zero values.
         rtol = 1e-5 * np.sum(np.abs(values_0))
         assert np.isclose(np.sum(values_0 - values_1), 0, atol=1e-10 + rtol)
+
+
+def get_testable_methods_names(model_setup) -> list[str]:
```
IvarStefansson, line 663: The name is a bit unfortunate. get_operator_producing_model_methods?

tests/integration/models/setup_utils.py
```diff
@@ -661,3 +658,46 @@ def compare_values(
...
+    Returns:
+        List of all possible testable method names for the given model.
+
+    """
+
+    # Get all public methods
+    all_methods = [method for method in dir(model_setup) if not method.startswith("_")]
+
+    # Get all testable methods
+    # A testable method is one that:
```
IvarStefansson, line 678: Perhaps move this definition to function docstring?

tests/integration/models/test_eval_methods_mass_balance.py
```diff
@@ -0,0 +1,420 @@
...
+
+    # Instantiate the model setup
+    setup = Model(params)
+
+    # Prepare to simulate
+    setup.prepare_simulation()
+    setup.set_well_network()
+
+    # Set constant but non-zero values for the primary variables
+    if hasattr(setup, "pressure_variable"):
```
IvarStefansson, line 80: Is this if needed?

tests/integration/models/test_eval_methods_mass_balance.py
```diff
@@ -0,0 +1,420 @@
...
+                    1.76405615e00,
+                    1.76405615e00,
+                    1.76405615e00,
+                    3.52811230e-02,
+                    3.52811230e-02,
+                    3.52811230e-02,
+                    3.52811230e-02,
+                    7.05622460e-04,
+                ]
+            ),
```
IvarStefansson, line 292: For derived quantities s.a. mass, do you think it'd be useful to add a short comment explaining how it was computed? E.g. rho * phi * V

tests/integration/models/test_eval_methods_mass_balance.py
```diff
@@ -0,0 +1,420 @@
...
+    domains = setup_utils.domains_from_method_name(
+        model_setup.mdg,
+        method,
+        dimension_restriction,
+    )
+
+    # Obtain operator in AD form.
+    operator = method(domains)
+
+    # Discretize (if necessary), evaluate, and retrieve numerical values.
```
IvarStefansson, line 411: Shouldn't this be taken care of by prepare_simulation?

IvarStefansson (commented):
Looks really good to me! Could we rename the new file to avoid "eval", please? Should we create issues for making same type of tests for the other models? Otherwise minor comments.

tests/integration/models/test_eval_methods_mass_balance.py
jhabriel: Not entirely. There are two reasons this is needed: (1) the `prepare_simulation()` does not consider the non-trivial primary variable states (since these are set afterwards), and (2) some methods have a domain restriction which must be respected, i.e., via `dimension_restriction`.

tests/integration/models/test_eval_methods_mass_balance.py
jhabriel: Not really. Now the primary variables should be constrained by each model. I will remove them.

tests/integration/models/setup_utils.py
jhabriel: Will do.

tests/integration/models/test_eval_methods_mass_balance.py
IvarStefansson: (2) should not be a problem, I think. (1) might be, I suppose.

tests/integration/models/setup_utils.py
IvarStefansson: Yes, I like that!

tests/integration/models/test_eval_methods_mass_balance.py
IvarStefansson: Great. But it should not be an absolute requirement (e.g. not fluxes)

jhabriel:
I think I have incorporated all the suggestions. Yes, we will benefit from some team effort here. I guess TH should be the next model we should tackle. Right?

IvarStefansson:
> I think I have incorporated all the suggestions. Yes, we will benefit from some team effort here. I guess TH should be the next model we should tackle. Right?

Or momentum balance

tests/integration/models/test_ad_methods_single_phase_flow.py
```diff
@@ -275,6 +277,7 @@ def test_tested_vs_testable_methods_single_phase_flow(
             ),
             None,
         ),
+        # fluid_mass = rho * phi * cell_volume
```
IvarStefansson, line 280: * specific volume?

IvarStefansson (approved):

tests/integration/models/test_ad_methods_single_phase_flow.py
jhabriel: Oops. My bad. Fixed.

jhabriel:
Yes. Probably momentum balance is a better option.

PR #927, opened 2023-06-15
Feat prune tree

Yuriyzabegaev:
## Proposed changes

As we discussed with @keileg, `operator.tree` slightly disturbs debugging the ad, so I moved its attributes to the `operator` itself.

The minor changes I made are an attempt to improve annotations:
* I reformatted the creation of the enum `Operator.Operations` so now static type checkers can see the enum entries. Values left the same.
* The other thing I changed is the type of `operator.children` (previously `operator.tree.children`) from `Sequence[Operator | AdArray]` to just `Sequence[Operator]`. All the methods inside operators.py treat children as if they are only operators and not the AdArrays. Everything works fine, but mypy complains about that. I removed AdArrays from the signature [here](https://github.com/pmgbergen/porepy/blob/03d2bfc848f3c2950e60de2caafbba301c298163/src/porepy/numerics/ad/operator_functions.py#L93), because it is never used and is the only place where `operator.children` are possibly initialized with AdArrays.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/ad/operators.py
+++ b/src/porepy/numerics/ad/operators.py
@@ -58,8 +58,10 @@ class Operator:
         interfaces (optional): List of interfaces in the mixed-dimensional grid on which
             the operator is defined. Will be empty for operators not associated with any
             interface. Defaults to None (converted to empty list).
-        tree (optional): The tree structure of child operators. Defaults to None
-            (converted to a tree with a single void operator). TODO
+        operation (optional): Arithmetic or other operation represented by this
+            operator. Defaults to void operation.
+        children (optional): List of children, other AD operators. Defaults to empty
+            list.
 
     """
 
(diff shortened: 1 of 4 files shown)


src/porepy/numerics/ad/operators.py
```diff
@@ -879,6 +894,7 @@ def _find_subtree_variables(self) -> Sequence[Variable]:
                     # Effectively, this node is one step from the leaf
                     var_list.append(var)
                 elif isinstance(var, list):
+                    assert False
```
keileg, line 897: Is this on purpose? If so, please provide a short documentation and consider deleting the code below.

keileg (changes requested):
Good work, one minor question only. Thanks for taking care of this.

src/porepy/numerics/ad/operators.py
Yuriyzabegaev: I forgot to remove it when was debugging. Deleting it.

keileg (approved):

PR #937, opened 2023-08-03
MAINT: Raising an error in `previous_iteration` and `previous_timestep`

### LINKED ISSUE #936: Incorrect behavior of `previous_iteration` and `previous_timestep`
I stumbled upon the problem of creating the ad variable both on the previous time step and the previous iteration. The code that does not work (imagine we are building a model where the accumulation term is evaluated on the previous time step):

```
import porepy as pp
from porepy.models.fluid_mass_balance import SinglePhaseFlow 
class SinglePhaseModelBreakdown(SinglePhaseFlow):

    def fluid_mass(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
        pressure_prev_it = self.pressure(subdomains).previous_iteration()
        exp = pp.ad.Function(pp.ad.exp, "density_exponential")
        rho_0 = pp.ad.Scalar(self.fluid.density())
        density_prev_it = rho_0 * exp(self.fluid_compressibility(subdomains) * pressure_prev_it)

        mass_density = density_prev_it * self.porosity(subdomains)
        mass = self.volume_integral(mass_density, subdomains, dim=1)
        mass.set_name("fluid_mass")
        return mass
    

model = SinglePhaseModelBreakdown()
pp.run_stationary_model(model, {})
```

The result is:
```
KeyError: Variable pressure with id 5 on grid 0
Degrees of freedom: cells (1), faces (0), nodes (0)
Evaluated at the previous iteration.
```

I understand that these functions are not meant to be used together, and the previous Newton iteration for the previous time step is not defined. However, I faced this problem when I was making a custom stabilization term for preconditioning, and I had to use `previous_iteration` in the accumulation term. I only needed the Jacobian of that operator, so I assumed that the whole term related to the previous time step will cut down. Instead, it raised this error and took some time to figure out what causes it in the ad implementation. Thus, I propose to raise an error [here](https://github.com/pmgbergen/porepy/blob/c9bbe84f8ae6caa0676da56828ead33c759f4aed/src/porepy/numerics/ad/operators.py#L1803), to show more clearly, that this is not the intended use. If you agree, I can make this change.

Yuriyzabegaev:
## Proposed changes

This PR attempts to close #936. By this, we raise `ValueError` early if the user attempts to create an ad variable both on the previous time step and the previous iteration. I also raise `NotImplementedError` if the user attempts to create a variable on more than one previous iteration or time step behind, because it is currently not supported: [here](https://github.com/pmgbergen/porepy/blob/c9bbe84f8ae6caa0676da56828ead33c759f4aed/src/porepy/numerics/ad/operators.py#L826) we take only one `original_variable`, which leads again to an unclear error message.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/ad/operators.py
+++ b/src/porepy/numerics/ad/operators.py
@@ -1623,8 +1623,8 @@ class Variable(Operator):
 
         Raises:
             ValueError:
-                If the variable is a representation of the previous time
-                iteration, previously set by :meth:`~previous_iteration`.
+                If the variable is a representation of the previous iteration,
+                previously set by :meth:`~previous_iteration`.
 
             NotImplementedError:
                 If the variable is already a representation of the previous time step.
@@ -1830,8 +1830,8 @@ class MixedDimensionalVariable(Variable):
 
         Raises:
             ValueError:
-                If the variable is a representation of the previous time
-                iteration, previously set by :meth:`~previous_iteration`.
+                If the variable is a representation of the previous iteration,
+                previously set by :meth:`~previous_iteration`.
 
             NotImplementedError:
                 If the variable is already a representation of the previous time step.
--- a/src/porepy/numerics/ad/operators.py
+++ b/src/porepy/numerics/ad/operators.py
@@ -1621,13 +1621,14 @@ class Variable(Operator):
     def previous_timestep(self) -> Variable:
         """Return a representation of this variable on the previous time step.
 
-        If the variable is already a representation of the previous time step, the
-        method returns itself.
-
         Raises:
             ValueError:
                 If the variable is a representation of the previous time
                 iteration, previously set by :meth:`~previous_iteration`.
+
+            NotImplementedError:
+                If the variable is already a representation of the previous time step.
+                Currently, we support creating only one previous time step.
 
         Returns:
             A representation of this variable at the previous time step,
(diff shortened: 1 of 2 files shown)


src/porepy/numerics/ad/operators.py
```diff
@@ -1621,16 +1621,31 @@ def set_name(self, name: str) -> None:
     def previous_timestep(self) -> Variable:
         """Return a representation of this variable on the previous time step.
 
-        If the variable is already a representation of the previous time step, the
-        method returns itself.
+        Raises:
+            ValueError:
+                If the variable is a representation of the previous time
```
keileg, line 1626: Should this be previous iteration (not time?)

keileg:
Thanks. Two minor questions only.

src/porepy/numerics/ad/operators.py
Yuriyzabegaev: Indeed!

src/porepy/numerics/ad/operators.py
```diff
@@ -1791,15 +1828,20 @@ def previous_timestep(self) -> MixedDimensionalVariable:
         """Return a representation of this mixed-dimensional variable on the previous
         time step.
 
-        If the md-variable is already defined on the previous time step, return itself.
+        Raises:
+            ValueError:
+                If the variable is a representation of the previous time
```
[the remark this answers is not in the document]
Yuriyzabegaev: I agree

keileg (approved):
Nice work. Thanks.

PR #942, opened 2023-08-22
Switch to pip-based docker images

keileg:
## Proposed changes
This PR replaces the old conda-based docker image with a less complex pip-based installation. 

The main update is to the file defining the base image, but the files defining the development and stable images have been updated accordingly. Note that the image for the stable branch on DockerHub will not be updated until the next release.

## Types of changes

- [x] Other: Docker

## Checklist

No change to the code base.

### DIFF
--- a/dockerfiles/dev/Dockerfile
+++ b/dockerfiles/dev/Dockerfile
@@ -29,7 +29,7 @@ RUN git checkout develop
 # Do the install
 RUN pip install -e .
 
-# Run a pip freeze it is always good to know whas been installed
+# Run a pip freeze. It is always good to know what been installed
 RUN pip freeze
 
 # Add PorePy home to the pythonpath. This may or may not be necessary.
--- a/dockerfiles/dev/Dockerfile
+++ b/dockerfiles/dev/Dockerfile
@@ -23,21 +23,14 @@ FROM porepy/base
 ## Download PorePy from github
 # Make sure we are in the home directory
 WORKDIR ${POREPY_HOME}
-USER porepy
-# Clone from Github
-RUN git clone https://github.com/pmgbergen/porepy.git ${POREPY_SRC}
-
-
-# Enter source directory
-WORKDIR ${POREPY_SRC}
 # Make sure we're on the develop branch
 RUN git checkout develop
 
 # Do the install
 RUN pip install -e .
 
-# Run a conda list, it is always good to know whas been installed
-RUN conda list
+# Run a pip freeze it is always good to know whas been installed
+RUN pip freeze
 
 # Add PorePy home to the pythonpath. This may or may not be necessary.
 ENV PYTHONPATH $POREPY_HOME:$PYTHONPATH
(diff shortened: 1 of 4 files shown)


dockerfiles/dev/Dockerfile
```diff
@@ -23,21 +23,14 @@ FROM porepy/base
...
-RUN git clone https://github.com/pmgbergen/porepy.git ${POREPY_SRC}
-
-
-# Enter source directory
-WORKDIR ${POREPY_SRC}
 # Make sure we're on the develop branch
 RUN git checkout develop
 
 # Do the install
 RUN pip install -e .
```
IvarStefansson, line 30: Why not install in base file?

dockerfiles/dev/Dockerfile
```diff
@@ -23,21 +23,14 @@ FROM porepy/base
...
 
-# Run a conda list, it is always good to know whas been installed
-RUN conda list
+# Run a pip freeze it is always good to know whas been installed
```
IvarStefansson, line 32:
```suggestion
# Run a pip freeze. It is always good to know what been installed
```

dockerfiles/stable/Dockerfile
```diff
@@ -22,16 +22,14 @@ FROM porepy/base
...
-WORKDIR ${POREPY_SRC}
-# Make sure we're on the main branch
+# Make sure we're on the develop branch
 RUN git checkout main
 
 # Do the install
-RUN pip install --user -e .
+RUN pip install -e .
+
+# Run a pip freeze it is always good to know whas been installed
```
IvarStefansson, line 31:
```suggestion
# Run a pip freeze. It is always good to know what been installed.
```

IvarStefansson (commented):
Have you tested the two images?

dockerfiles/dev/Dockerfile
keileg: To avoid a double install when building the stable version. It is a marginal difference (at least for something as simple as PorePy), and one can think of both options as conceptually simpler. Basically, I ended up using the same workflow as fenics applies, on the assumption they know what they are doing.

keileg:
> Have you tested the two images?

Yes, I tested the build (both base and develop, the latter runs pytest in the container as part of the build) locally. Both ran as hoped.

dockerfiles/dev/Dockerfile
IvarStefansson: Do I get this right: if you install, then check out a different branch, there's some sort of re-install going on?

keileg: Not really. Install of PorePy proper is done in the relevant front image (dev or stable), while base prepares the ground. If you check out, PorePy would be updated since we install with the -e option (which could be taken as a sign we should not use that option in stable, but then, playing with git in the stable docker image is far from intended usage).

There are probably several ways one can do this, but this is something that just should work.

PR #950, opened 2023-09-07
Boundary grids ad

Yuriyzabegaev:
## Proposed changes

## Types of changes

- [ ] Other:

## Checklist


src/porepy/models/fluid_mass_balance.py
```diff
@@ -534,9 +503,15 @@ def create_variables(self) -> None:
             tags={"si_units": f"m^{self.nd} * Pa"},
         )
 
-    def pressure(self, subdomains: list[pp.Grid]) -> pp.ad.MixedDimensionalVariable:
-        p = self.equation_system.md_variable(self.pressure_variable, subdomains)
-        return p
+    def pressure(
+        self, grids: list[pp.Grid] | list[pp.BoundaryGrid]
+    ) -> pp.ad.MixedDimensionalVariable:
```
IvarStefansson, line 508: Fix return type

src/porepy/models/fluid_mass_balance.py
```diff
@@ -534,9 +503,15 @@ def create_variables(self) -> None:
...
+        if len(grids) > 0 and isinstance(grids[0], pp.BoundaryGrid):
+            return pp.ad.TimeDependentDenseArray(
+                name=self.bc_data_pressure_key, domains=grids
+            )
+
+        return self.equation_system.md_variable(self.pressure_variable, grids)
```
IvarStefansson, line 514: I think we want (better method name needed!):
    
    return self.get_dynamic_quantity(self.pressure_name, grids)

    def get_dynamic_quantity(name, grids) -> Operator:
        if len(grids) > 0 and isinstance(grids[0], pp.BoundaryGrid):
            return pp.ad.TimeDependentDenseArray(
                name=name, domains=grids
            )

        return self.equation_system.md_variable(name, grids)

This can be reused for temperature and displacement

src/porepy/models/fluid_mass_balance.py
```diff
@@ -743,6 +722,26 @@ def set_discretization_parameters(self) -> None:
                 },
             )
 
+    def set_boundary_data(self) -> None:
```
IvarStefansson, line 725: Should we call this from update_time_dependent_arrays?

src/porepy/models/fluid_mass_balance.py
```diff
@@ -743,6 +722,26 @@ def set_discretization_parameters(self) -> None:
...
+        """Set values for the boundary conditions.
+
+        Both Neuman and Dirichlet data should be initialized here.
+
+        """
+        for bc_var_name in [self.bc_data_pressure_key, self.bc_data_darcy_flux_key]:
+            for bg, bg_data in self.mdg.boundaries(return_data=True):
+                bc_values = np.zeros(bg.num_cells)
```
IvarStefansson, line 733: This needs to be a separate method. See how this is handled in poromechanics for time dependent mechanical conditions.

IvarStefansson: The setter method could be generalized using a e.g. a list self._boundary_names: List[str]. The assignment using a "separate method" would then be

for name in self._boundary_names:
    for bg, bg_data in self.mdg.boundaries(return_data=True):
        bc_values = getattr(name+"_boundary_values")(bg)

This could cover all BCs for all physics.
Ask if clarification is needed!

src/porepy/numerics/ad/equation_system.py
```diff
@@ -423,14 +427,6 @@ def create_variables(
             if var.name == name and var.domain in grids:
                 raise KeyError(f"Variable {name} already defined on {var.domain}.")
 
-        def initialize_data_with_key(data: dict, key: str) -> None:
```
IvarStefansson, line 426: Why this change?

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -186,7 +187,7 @@ def _error_message(self):
         return "Argument should be a subdomain grid or a list of subdomain grids"
 
 
-class MortarProjections(Operator):
+class MortarProjections:
```
IvarStefansson, line 190: Please motivate change.

src/porepy/numerics/ad/operators.py
```diff
@@ -1295,7 +1295,7 @@ def parse(self, mdg: pp.MixedDimensionalGrid) -> np.ndarray:
         return self._values
 
 
-class TimeDependentDenseArray(DenseArray):
+class TimeDependentDenseArray(Operator):
```
IvarStefansson, line 1298: Please explain!

src/porepy/utils/porepy_types.py
```diff
@@ -1,14 +1,26 @@
...
+    "fracture_network",
+]
 
-GridLike = Union["pp.Grid", "pp.MortarGrid"]
+GridLike = Union["pp.Grid", "pp.MortarGrid", "pp.BoundaryGrid"]
 """Type for grids and mortar grids."""
+SubdomainsOrBoundaries = Sequence["pp.Grid"] | Sequence["pp.BoundaryGrid"]
+"""Type for sequence of subdomains or sequence of boundary grids."""
+
+GridsLike = SubdomainsOrBoundaries | Sequence["pp.MortarGrid"]
```
IvarStefansson, line 22: Please explain change. But I like the changes to this class!

src/porepy/numerics/ad/operators.py
```diff
@@ -1375,22 +1358,20 @@ def __init__(
 
         """
 
-        self._initialize_children()
+        super().__init__(name=name)
+
+    def grid_type(self) -> Literal["subdomains", "interfaces", "boundary grids"]:
```
IvarStefansson, line 1363: Is GridLikeSequence better, perhaps?

IvarStefansson (commented):
Good draft! Please see my comments and feel free to come and discuss if needed.

IvarStefansson (commented):
Good draft! Please see my comments and feel free to come and discuss if needed.

src/porepy/grids/md_grid.py
```diff
@@ -343,6 +344,13 @@ def subdomain_to_boundary_grid(self, sd: pp.Grid) -> Optional[pp.BoundaryGrid]:
         """
         return self._subdomain_to_boundary_grid.get(sd, None)
 
+    def subdomains_to_boundary_grids(
```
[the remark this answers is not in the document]
Yuriyzabegaev: Sounds right!

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: It was me who originally made this inner function. I made it at the first stage of implementing boundary grids. At that moment, I thought that it would be useful to initialize BG data arrays with the variable names: `data[key][name] = {}`. Now we use separate keys for boundary data, not the variable names, e.g. "bc_data_darcy_flux". Thus, fields with variable names on boundary grids are never used. I didn't see a reason to keep the unified initialization in a separate inner function anymore, so I got rid of it and came back to how it was implemented before me. Do you prefer having the inner function?

src/porepy/models/constitutive_laws.py
```diff
@@ -934,6 +936,7 @@ def pressure_trace(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
         """
         interfaces: list[pp.MortarGrid] = self.subdomains_to_interfaces(subdomains, [1])
         projection = pp.ad.MortarProjections(self.mdg, subdomains, interfaces, dim=1)
+        projection_bound = pp.ad.BoundaryProjection(self.mdg, subdomains=subdomains)
```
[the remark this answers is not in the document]
Yuriyzabegaev: Agreed

src/porepy/numerics/ad/equation_system.py
IvarStefansson: I think it may make sense to use the (primary or secondary) variable names, i.e, "pressure" and "darcy_flux". Do you agree? In that case, I think keeping the inner function makes sense.

Yuriyzabegaev: This happens inside `EquationSystem.create_variable`, so the field for pressure is created when we're passing name="pressure". "darcy_flux" won't come inside, since we're not creating a primary variable with this name. Or how do you see this?

src/porepy/models/fluid_mass_balance.py
```diff
@@ -362,6 +368,10 @@ class BoundaryConditionsSinglePhaseFlow:
 
     """
 
+    mdg: pp.MixedDimensionalGrid
```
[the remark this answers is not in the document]
Yuriyzabegaev: Agreed

src/porepy/models/fluid_mass_balance.py
```diff
@@ -195,6 +195,11 @@ def fluid_mass(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
         mass.set_name("fluid_mass")
         return mass
 
+    def mobility_rho(
+        self, grids: Sequence[pp.Grid] | Sequence[pp.BoundaryGrid]
+    ) -> pp.ad.Operator:
```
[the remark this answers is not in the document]
Yuriyzabegaev: Agreed

src/porepy/numerics/ad/equation_system.py
IvarStefansson: Good point. I still think that we may want to use the same names on the boundary, though.

src/porepy/models/fluid_mass_balance.py
Yuriyzabegaev: Thanks

src/porepy/models/fluid_mass_balance.py
Yuriyzabegaev: Looks good to me. Where would you locate this method? We don't have an abstract BoundaryConditionMixin, it would fit there perfectly. Maybe VariableMixin?

src/porepy/models/fluid_mass_balance.py
Yuriyzabegaev: There's something weird with this method, as for my knowledge. A search through the project files shows that this method `update_time_dependent_ad_arrays` is called only in the poromechanics model. Should I add it to the common solution strategy?

src/porepy/models/fluid_mass_balance.py
Yuriyzabegaev: One thing worries me. How will this list work in a multiphysics case? Say, we have single phase flow mixin with its list, and a mechanics mixin with its list. How and where would these lists merge into one? Maybe it's better to make a common method, which is responsible just for setting the value to iterate and moving the previous value to state, and then call it by hand from each mixin for its variables?

src/porepy/numerics/ad/operators.py
Yuriyzabegaev: I modified some things inside TimeDependentDenseArray, and noticed that it repeats several things from the operator initialization, so I decided to clean it up a bit. 

As I understand it, the point why we inherit from DenseArray is that we might use a check `isinstance(..., DenseArray)`, and want to catch TimeDependentDenseArrays as well. However, I don't see when this can be actually used, because DenseArray and TimeDependentDenseArrays have completely different attributes, so if one wants to do something specific only if the operator is a DenseArray, they will anyway have to add one more check `isinstance(..., TimeDependentDenseArray)` to treat them differently. Tell me if I didn't get it right, please.

src/porepy/numerics/ad/grid_operators.py
Yuriyzabegaev: I also removed the same parent from `BoundaryProjection`, and did so for `MortarProjections` just for the unity. These projection classes are never (as for my knowledge) used as operators directly - they don't have the `parse` method, so it won't work anyway. Instead, they are used as a collection of several projection operators, each of which is a proper ad operator, so there is no reason to make it an operator itself. Please, let me know if I'm wrong.

src/porepy/utils/porepy_types.py
Yuriyzabegaev: This brings us to the earlier discussion. Now, all the methods that take a list of subdomains can also take a list of boundary grids. So, the alias `SubdomainsOrBoundaries` (the name is discussible) will be useful in models. I also defined a convenient alias `GridsLike` (I don't like the name) aiming to use it instead of `list[GridLike]`, because we agreed that operators on several types of grids are banned (am I right?).

I also try to use `Sequence[...]` instead of `list[...]` everywhere in annotations, because it's a protocol for immutable collections, which makes mypy happy, by saying that we won't modify the passed list inside the function. Mypy considers it a good practice, see [this](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#tuple-types). Is it okay, or should I stick to lists in annotations?

src/porepy/numerics/ad/operators.py
Yuriyzabegaev: Is this about the name GridsLike -> GridLikeSequence? If yes, I like your option more than mine.

src/porepy/models/fluid_mass_balance.py
IvarStefansson: Or rather just make that abstract class for BCs?

src/porepy/models/fluid_mass_balance.py
IvarStefansson: Do you mean the implementation? Let's discuss in person.

src/porepy/models/fluid_mass_balance.py
IvarStefansson: You'd merge them in the multi-physics combination.

src/porepy/numerics/ad/grid_operators.py
IvarStefansson: Your argument makes sense to me. Do you agree, @keileg?

src/porepy/numerics/ad/operators.py
IvarStefansson: I agree. @keileg?

src/porepy/utils/porepy_types.py
IvarStefansson: Sorry about this. My codespace messed up the location of my comments.

src/porepy/numerics/ad/operators.py
IvarStefansson: Sorry about this. My codespace messed up the location of my comments. Yes, that's what I meant.

Yuriyzabegaev: Then I rename it.

src/porepy/numerics/ad/grid_operators.py
keileg: Agree, this is a change in the right direction.

src/porepy/numerics/ad/operators.py
keileg: Makes sense.

src/porepy/models/fluid_mass_balance.py
```diff
@@ -392,63 +402,20 @@ def bc_type_mobrho(self, sd: pp.Grid) -> pp.BoundaryCondition:
...
-        Returns:
-            Ad array representing the boundary condition values for the Darcy flux.
-
-        """
-        num_faces = sum([sd.num_faces for sd in subdomains])
-        # Ignore typing error below, the parameter in _ad_wrapper forces it to be an
-        # Array.
-        return pp.wrap_as_ad_array(0, num_faces, "bc_values_darcy")
-
-    def bc_values_mobrho(self, subdomains: list[pp.Grid]) -> pp.ad.DenseArray:
```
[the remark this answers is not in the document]
keileg: In my understanding, Neumann-type conditions for conservation equations can either be specified as the total influx of the conserved quantity (so and so much mass enters through this face), or by specifying the influx of fluid together with the concentration of the conserved quantitiy (this would be a Neumann condition for the Darcy flux, but a Dirichlet-type condition for the density or similar). I think we need to be able to do both. What that means in terms of methods and return types, I am not sure of.

src/porepy/grids/md_grid.py
```diff
@@ -12,6 +12,7 @@
 import porepy as pp
 from porepy.grids import mortar_grid
 from porepy.utils.porepy_types import GridLike
+from typing import Sequence
```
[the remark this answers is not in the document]
Yuriyzabegaev: Thanks

src/porepy/models/geometry.py
```diff
@@ -203,6 +203,13 @@ def interfaces_to_subdomains(
                     subdomains.append(sd)
         return self.mdg.sort_subdomains(subdomains)
 
+    def subdomains_to_boundary_grids(
+        self, subdomains: Sequence[pp.Grid]
+    ) -> Sequence[pp.BoundaryGrid]:
+        """TODO"""
+        boundary_grids = [self.subdomain_to_boundary_grid(sd) for sd in subdomains]
```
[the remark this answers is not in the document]
Yuriyzabegaev: I'm slightly confused. We have `mdg.subdomain_to_interfaces` because one subdomain may have several associated interfaces, and also we have a function `subdomains_to_interfaces`, which is a shortcut that does the same, but for several subdomains. In the same logic, we have `mdg.subdomain_to_boundary_grid`, singular because only one BG is associated with the subdomain, and the shortcut function `subdomains_to_boundary_grids`. Do you propose to keep only the plural ones for both interfaces and boundary grids?

src/porepy/models/abstract_equations.py
```diff
@@ -183,3 +184,57 @@ def perturbation_from_reference(self, variable_name: str, grids: list[pp.Grid]):
...
         d_var.set_name(variable_name + "_perturbation")
         return d_var
+
+
+class BoundaryConditionsMixin:
+    """TODO"""
+
+    mdg: pp.MixedDimensionalGrid
+
+    def update_boundary_conditions(self, initial: bool) -> None:
```
IvarStefansson, line 194: I think "set_boundary_condition" is even more precise. This is already reflected in the docstring.

Should this be an abstract method?

src/porepy/models/abstract_equations.py
```diff
@@ -183,3 +184,57 @@ def perturbation_from_reference(self, variable_name: str, grids: list[pp.Grid]):
         d_var = var(grids) - var_ref(grids)
         d_var.set_name(variable_name + "_perturbation")
         return d_var
+
+
+class BoundaryConditionsMixin:
```
IvarStefansson, line 189: I think I'd prefer singular, i.e. BoundaryCondition. And does this perhaps belong in a separate file?

src/porepy/models/abstract_equations.py
```diff
@@ -183,3 +184,57 @@ def perturbation_from_reference(self, variable_name: str, grids: list[pp.Grid]):
...
+
+        for bg, data in self.mdg.boundaries(return_data=True):
+            # Set the known time step value.
+            if initial:
+                vals = function([bg])
+                pp.set_solution_values(
+                    name=name, values=vals, data=data, time_step_index=0
+                )
+            else:
+                vals = data[pp.ITERATE_SOLUTIONS][name][0]
```
IvarStefansson, line 224: @IngridKJ, should we make a get_solution_values method?

src/porepy/models/abstract_equations.py
```diff
@@ -183,3 +184,57 @@ def perturbation_from_reference(self, variable_name: str, grids: list[pp.Grid]):
...
+                )
+            # Set the iterate value.
+            vals = function([bg])
+            pp.set_solution_values(name=name, values=vals, data=data, iterate_index=0)
+
+    def make_boundary_operator(
+        self, name: str, domains: Sequence[pp.BoundaryGrid]
+    ) -> pp.ad.Operator:
+        """TODO
+        Yury: this function wraps two line of code, but I decided to make it, because we
```
IvarStefansson, line 236: Agreed.

src/porepy/models/fluid_mass_balance.py
```diff
@@ -368,9 +369,19 @@ class BoundaryConditionsSinglePhaseFlow:
 
     """
 
-    mdg: pp.MixedDimensionalGrid
-    pressure: Callable[[list[pp.Grid]], pp.ad.Operator]
-    darcy_flux: Callable[[Sequence[pp.Grid]], pp.ad.Operator]
+    bc_data_darcy_flux_key: str = "darcy_flux"
+    """TODO"""
+    bc_data_pressure_key: str = "pressure"
```
IvarStefansson, line 374: I think we should rather reuse the variable name.

src/porepy/models/fluid_mass_balance.py
```diff
@@ -412,11 +423,30 @@ def bc_values_darcy(self, subdomains: Sequence[pp.Grid]) -> pp.ad.Operator:
...
+        boundary_grids = self.subdomains_to_boundary_grids(subdomains)
         pressure_dirichlet = self.pressure(boundary_grids)
         flux_neumann = self.darcy_flux(boundary_grids)
         return pressure_dirichlet + flux_neumann
 
+    def update_boundary_conditions(self, initial: bool) -> None:
+        """Set values for the pressure and the darcy flux on boundaries."""
+        super().update_boundary_conditions(initial=initial)
+
+        def pressure(boundary_grids: Sequence[pp.BoundaryGrid]) -> np.ndarray:
```
IvarStefansson, line 435: If we define these functions outside as e.g. _boundary_pressure(...), we can reuse the outer function and tailor the inner ones in a run script.

src/porepy/models/solution_strategy.py
```diff
@@ -622,4 +635,4 @@ def update_time_dependent_ad_arrays(self, initial: bool) -> None:
                 updated by copying the iterate.
 
         """
-        pass
+        self.update_boundary_conditions(initial=initial)
```
IvarStefansson, line 638: Again, the "initial" strongly suggests that "set" is better than "update"

IvarStefansson (commented):
Good improvements!

src/porepy/numerics/ad/operators.py
vlipovac, line 1363: @Yuriyzabegaev @keileg @IvarStefansson 
I propose using the property decorator for such cases, since it is not really a callable and the private return value is static once set.

I marked all of you, since this is a general comment. Might be relevant at several other spots in porepy.

src/porepy/models/poromechanics.py
```diff
@@ -218,33 +201,35 @@ def update_time_dependent_ad_arrays(self, initial: bool) -> None:
...
-                )
-
-            vals = self.time_dependent_bc_values_mechanics([sd])
-            pp.set_solution_values(
-                name=self.bc_values_mechanics_key,
-                values=vals,
-                data=data,
-                iterate_index=0,
-            )
+        assert False, "Assert we can remove this function"
```
vlipovac, line 204: Is there surely no other use for this then updating BC values?

src/porepy/numerics/ad/operators.py
```diff
@@ -1332,11 +1332,11 @@ class TimeDependentDenseArray(Operator):
     def __init__(
         self,
         name: str,
-        domains: GridsLike,
+        domains: GridLikeSequence,
```
vlipovac, line 1335: Should we generalize this signature to the base class Operator?
With the property _grid_type. 
All children of operator can then have a simplified check of grid_type in the implementations of their methods.
But also other objects operating on instances of Operator.

src/porepy/numerics/ad/operators.py
IvarStefansson: I agree

src/porepy/models/poromechanics.py
IvarStefansson: Good question. Am I right that the reason you propose to delete this one is because it should be handled in a general method (not specific to poromechanics), @Yuriyzabegaev?

src/porepy/numerics/ad/operators.py
IvarStefansson: I don't understand exactly what you mean by generalizing the signature. Could you elaborate, please?

src/porepy/models/abstract_equations.py
IngridKJ: Sounds reasonable. I assume it should/can be analogous to the pp.set_solution_values()? If so, I already have one in my private repo. Should be super quick to add to porepy

src/porepy/models/poromechanics.py
Yuriyzabegaev: @IvarStefansson @vlipovac yes, at least I don't see if it does something more than that. I just tried to be careful with changing the logic of the poromechanics setup.

src/porepy/numerics/ad/operators.py
Yuriyzabegaev: I agree, changing it to the property

Yuriyzabegaev: I just noticed that I currently don't use it anywhere outside the class. Do we really need it, or maybe I should remove it completely for brevity?

src/porepy/models/solution_strategy.py
Yuriyzabegaev: Agreed, changing it.

src/porepy/models/fluid_mass_balance.py
Yuriyzabegaev: Agreed, changing it

src/porepy/models/abstract_equations.py
IvarStefansson: That would be nice. If you do, please check for occurrences  of pp.ITERATE_SOLUTIONS and pp.TIME_STEP_SOLUTIONS throughout the code and replace by new function where appropriate.

src/porepy/numerics/ad/operators.py
IvarStefansson: I'd say leave it, could be useful. Perhaps rename to domain_type? @keileg?

src/porepy/models/fluid_mass_balance.py
Yuriyzabegaev: The variable name is not available at the initialization time of this class, because it's initialized in the init method of `SolutionStrategySinglePhaseFlow`. I can also add the init method for the boundary condition class, but it restricts the rule of thumb, that mixins must not have their own init methods. 

I agree with reusing it, so I move the definition of the name as the private global variable inside the same file and refer to it in both places. Please, let me know if you see a more elegant solution.

src/porepy/models/abstract_equations.py
Yuriyzabegaev: BoundaryCondition or BoundaryConditionMixin? I agree on moving to a separate file.

src/porepy/models/fluid_mass_balance.py
IvarStefansson: No, I mean to remove the attribute `bc_data_pressure_key` and use `pressure_variable` instead. Optionally renaming to `pressure_(variable_)key`

src/porepy/models/abstract_equations.py
IvarStefansson: I would prefer BoundaryCondition, but that's already taken in bc.py (as is AbstractBoundaryCondition). I can't think of a better name than BoundaryConditionMixin. Feel free to suggest!

src/porepy/models/abstract_equations.py
Yuriyzabegaev: I agree

src/porepy/models/abstract_equations.py
Yuriyzabegaev: I like BoundaryConditionMixin

src/porepy/numerics/ad/operators.py
Yuriyzabegaev: @vlipovac do you propose to use `GridLikeSequence` for the base class Operator, and move `_grid_type` to the base class? If so, I agree

src/porepy/models/fluid_mass_balance.py
Yuriyzabegaev: Ahhh, got it

src/porepy/models/abstract_equations.py
IngridKJ: Sure. I take it the method is to be part of this PR? And are there any suggestions about where to put it?

src/porepy/numerics/ad/operators.py
vlipovac: @IvarStefansson @Yuriyzabegaev exactly like Yuriy summarized it.
class Operator currently has interfaces and subdomains as arguments in the constructor.
It may be more general (and consistent with derived classes) to use the same approach as here.

src/porepy/models/abstract_equations.py
IvarStefansson: No, separate PR. Should go same place at the set method, shouldn't it?

src/porepy/numerics/ad/operators.py
IvarStefansson: I like the idea. I think we need @keileg's opinion as well, though.

src/porepy/models/abstract_equations.py
IngridKJ: I agree they should be together, question is rather if we want them both within equation_system.py or not. equation_system.py is where the set function is now. But I will start working on an issue/PR for this, and we can discuss placement of the functions further elsewhere (if there is anything to discuss)

src/porepy/numerics/ad/operators.py
keileg: Agree, let's keep it. And domain_type is a better name.

src/porepy/numerics/ad/operators.py
keileg: Sounds good.

src/porepy/models/fluid_mass_balance.py
Yuriyzabegaev: Closing after discussion in person.

src/porepy/models/fluid_mass_balance.py
Yuriyzabegaev: Closing after discussion in person: added an abstract class for BCs.

src/porepy/models/fluid_mass_balance.py
Yuriyzabegaev: Closing after discussion in person.

src/porepy/numerics/ad/operators.py
Yuriyzabegaev: Renamed to domain_type.

src/porepy/models/boundary_condition.py
```diff
@@ -0,0 +1,64 @@
...
+            else:
+                vals = data[pp.ITERATE_SOLUTIONS][name][0]
+                pp.set_solution_values(
+                    name=name, values=vals, data=data, time_step_index=0
+                )
+            # Set the iterate value.
+            vals = function([bg])
+            pp.set_solution_values(name=name, values=vals, data=data, iterate_index=0)
+
+    def make_boundary_operator(
```
IvarStefansson, line 55: I think I prefer `create_boundar_operator` for consistency with other PorePy names.

Yuriyzabegaev: I agree

src/porepy/models/boundary_condition.py
```diff
@@ -0,0 +1,64 @@
...
+        boundary conditions. The specific boundary condition values should be updated in
+        concrete overrides. By default, it does nothing.
+
+        Note:
+            One can use the convenience method `_update_boundary_condition` for each
+            boundary condition value.
+
+        """
+
+    def _update_boundary_condition(
```
IvarStefansson, line 27: Still not convinced by name. Could we discuss in person, please?

Yuriyzabegaev: Following what we discussed in person, I got rid of the parameter `initial`, and renamed the method back to "update".

IvarStefansson: To me, it seems the "initial" is still a parameter. Might I suggest that the docstring explains that this method "updates" in the same sense as for a dictionary? And that we make this method public, as it is something the user will be exposed to?

Yuriyzabegaev: I agree, I'll come back to adding the docstring later, so keeping this conversation open for now.

src/porepy/models/boundary_condition.py
```diff
@@ -0,0 +1,80 @@
...
+                vals = function(bg)
+            pp.set_solution_values(name=name, values=vals, data=data, time_step_index=0)
+
+            # Set the unknown time step values.
+            vals = function(bg)
+            pp.set_solution_values(name=name, values=vals, data=data, iterate_index=0)
+
+    def create_boundary_operator(
+        self, name: str, domains: Sequence[pp.BoundaryGrid]
+    ) -> pp.ad.Operator:
```
vlipovac, line 73: Can we narrow the return type down to TimeDependentDenseArray?
Also, the typing of the declaration of this mixed-in method elsewhere is off
@Yuriyzabegaev @IvarStefansson

src/porepy/models/boundary_condition.py
```diff
@@ -0,0 +1,80 @@
...
+    time_manager: pp.TimeManager
+    """Time manager. Normally set by an instance of a subclass of
+    :class:`porepy.models.solution_strategy.SolutionStrategy`.
+
+    """
+
+    units: "pp.Units"
+    """Units object, containing the scaling of base magnitudes."""
+
+    @abstractmethod
```
vlipovac, line 34: @Yuriyzabegaev Can this be safely deleted?
It's re-defined below

src/porepy/models/boundary_condition.py
```diff
@@ -0,0 +1,80 @@
...
+    def update_boundary_conditions(self) -> None:
```
vlipovac, line 35: @Yuriyzabegaev  The naming of this is very dangerous, since only on `s` distinguishes it from the part where there is an actual implementation to be used in child classes.

Yuriyzabegaev: @vlipovac I agree. What about `update_all_boundary_conditions`?

vlipovac: I included that refactoring in the changes I am about to make with the energy balance

src/porepy/models/energy_balance.py
```diff
@@ -657,8 +717,14 @@ def bc_values_fourier(self, subdomains: list[pp.Grid]) -> pp.ad.DenseArray:
...
 
         """
-        num_faces = sum([sd.num_faces for sd in subdomains])
-        return pp.wrap_as_ad_array(0, num_faces, "bc_values_fourier")
+        boundary_projection = pp.ad.BoundaryProjection(self.mdg, subdomains=subdomains)
+        boundary_grids = self.subdomains_to_boundary_grids(subdomains)
+        temperature_dirichlet = self.temperature(boundary_grids)
+        flux_neumann = self.fourier_flux(boundary_grids)
+        result = temperature_dirichlet + flux_neumann  # TODO heat conductivity ?
+        result = boundary_projection.boundary_to_subdomain @ result
```
IvarStefansson, line 724: This violates the pp guideline about variable redefinition. Please improve variable naming.

src/porepy/models/energy_balance.py
```diff
@@ -657,8 +717,14 @@ def bc_values_fourier(self, subdomains: list[pp.Grid]) -> pp.ad.DenseArray:
             Ad array representing the boundary condition values for the Fourier flux.
 
         """
-        num_faces = sum([sd.num_faces for sd in subdomains])
-        return pp.wrap_as_ad_array(0, num_faces, "bc_values_fourier")
+        boundary_projection = pp.ad.BoundaryProjection(self.mdg, subdomains=subdomains)
+        boundary_grids = self.subdomains_to_boundary_grids(subdomains)
+        temperature_dirichlet = self.temperature(boundary_grids)
+        flux_neumann = self.fourier_flux(boundary_grids)
+        result = temperature_dirichlet + flux_neumann  # TODO heat conductivity ?
```
IvarStefansson, line 723: No need for heat conductivity, it should be included in the bound_discr object that will hit these bc values.

src/porepy/models/energy_balance.py
```diff
@@ -673,30 +739,47 @@ def bc_values_enthalpy_flux(self, subdomains: list[pp.Grid]) -> pp.ad.DenseArray
...
+        boundary_projection = pp.ad.BoundaryProjection(self.mdg, subdomains=subdomains)
+        boundary_grids = self.subdomains_to_boundary_grids(subdomains)
+
+        enthalpy_dirichlet = self.fluid_enthalpy(boundary_grids) * self.fluid_density(
+            boundary_grids
+        )
+        enthalpy_neumann = self.enthalpy_flux(boundary_grids)
+
+        result = enthalpy_dirichlet + enthalpy_neumann
+        result = boundary_projection.boundary_to_subdomain @ result
```
IvarStefansson, line 751: Please improve variable name.

src/porepy/models/fluid_mass_balance.py
```diff
@@ -208,9 +217,15 @@ def fluid_flux(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
 
         """
         discr = self.mobility_discretization(subdomains)
-        mob_rho = self.fluid_density(subdomains) * self.mobility(subdomains)
-
-        bc_values = self.bc_values_mobrho(subdomains)
+        mob_rho = self.mobility_rho(subdomains)
+
+        boundary_projection = pp.ad.BoundaryProjection(self.mdg, subdomains=subdomains)
+        # NOTE Here (in the background), some Dirichlet-type data for p and T have
```
IvarStefansson, line 223: This does not seem to me the right plase for this comment.

src/porepy/models/boundary_condition.py
```diff
@@ -70,10 +70,19 @@ def update_boundary_condition(
...
+        Raises:
+            AssertionError: If the passed sequence of domains does not consist entirely
+                of instances of boundary grid.
+
+        Returns:
+            An operator of given name representing time-dependent value on given
+            sequence of boundary grids.
 
         """
         assert all(isinstance(x, pp.BoundaryGrid) for x in domains)
```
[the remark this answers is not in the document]
Yuriyzabegaev: Changed it.

src/porepy/models/momentum_balance.py
```diff
@@ -841,21 +845,41 @@ def bc_type_mechanics(self, sd: pp.Grid) -> pp.BoundaryConditionVectorial:
...
+            `update_all_boundary_conditions`.
 
         """
-        num_faces = sum([sd.num_faces for sd in subdomains])
-        return pp.wrap_as_ad_array(0, num_faces * self.nd, "bc_vals_mechanics")
+        boundary_grids = self.subdomains_to_boundary_grids(subdomains)
+        boundary_projection = pp.ad.BoundaryProjection(
+            self.mdg, subdomains=subdomains, dim=self.nd
+        )
+        # TODO: Neumann?
```
IvarStefansson, line 865: I think that should be included by adding self.stress(boundary_grids) and extending said method.

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -524,17 +520,27 @@ def __init__(
...
         """Projection from subdomain faces to boundary grids cells."""
+        if len(mat) > 0:
+            self._projection = sps.bmat([[m] for m in mat], format="csr")
+        else:
+            self._projection = sps.csr_matrix((0, 0))
 
-    def subdomain_to_boundary(self) -> sps.spmatrix:
-        return self._projection
+    @property
+    def subdomain_to_boundary(self) -> Operator:
```
IvarStefansson, line 532: Missing docstrings

src/porepy/numerics/ad/operators.py
```diff
@@ -1767,6 +1748,9 @@ def __init__(self, variables: list[Variable]) -> None:
         # assigned to a problem where the grid happened not to have any fractures.
         self._no_variables = len(variables) == 0
 
+        # It should be defined in the parent class, but we do not call super().__init__
```
IvarStefansson, line 1751: What do you mean it "should be defined"?

tests/functional/setups/manu_flow_comp_2d_frac.py
```diff
@@ -605,42 +604,39 @@ def matrix_boundary_density(self, sd_matrix: pp.Grid, time: float) -> np.ndarray
             on the exterior boundary faces for the given ``time``.
 
         """
-        # Symbolic variables
-        x, y, t = sym.symbols("x y t")
+        assert False, "We should not need it anymore."
```
IvarStefansson, line 607: Please purge if possible

tests/functional/setups/manu_flow_comp_2d_frac.py
```diff
@@ -660,19 +656,20 @@ def bc_type_mobrho(self, sd: pp.Grid) -> pp.BoundaryCondition:
             boundary_faces = self.domain_boundary_sides(sd).all_bf
             return pp.BoundaryCondition(sd, boundary_faces, "neu")
 
-    def bc_values_darcy(self, subdomains: list[pp.Grid]) -> pp.ad.DenseArray:
-        """Set boundary condition values for the elliptic discretization."""
-        bc_values = pp.ad.TimeDependentDenseArray(
-            name="darcy_bc_values", subdomains=subdomains, previous_timestep=True
-        )
-        return bc_values
+    def boundary_pressure(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
```
IvarStefansson, line 659: Missing docstring

tests/visualization/times.json
```diff
@@ -0,0 +1 @@
+{"time": [0, 1], "dt": [1.0, 1.0]}
```
IvarStefansson, line 1: Please purge file

IvarStefansson (commented):
Lots of great stuff, guys! Some work remaining, though, mainly regarding documentation (please search for TODO!).

src/porepy/models/boundary_condition.py
```diff
@@ -0,0 +1,89 @@
...
+
+        """
+        pass
+
+    def update_boundary_condition(
+        self,
+        name: str,
+        function: Callable[[pp.BoundaryGrid], np.ndarray],
+    ) -> None:
+        """TODO
```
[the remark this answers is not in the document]
Yuriyzabegaev: done

src/porepy/models/boundary_condition.py
```diff
@@ -0,0 +1,89 @@
...
+
+        TODO: Somewhere we must have an assertion that there is only 1 time step behind
+        stored, and 1 iterate.
+        """
+        for bg, data in self.mdg.boundaries(return_data=True):
+            # Set the known time step values.
+            if name in data[pp.ITERATE_SOLUTIONS]:
+                # Use the values at the unknown time step from the previous time step.
+                vals = data[pp.ITERATE_SOLUTIONS][name][0]
```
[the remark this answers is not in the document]
Yuriyzabegaev: done!

src/porepy/models/fluid_mass_balance.py
```diff
@@ -534,9 +540,13 @@ def create_variables(self) -> None:
             tags={"si_units": f"m^{self.nd} * Pa"},
         )
 
-    def pressure(self, subdomains: list[pp.Grid]) -> pp.ad.MixedDimensionalVariable:
-        p = self.equation_system.md_variable(self.pressure_variable, subdomains)
-        return p
+    def pressure(self, grids: pp.SubdomainsOrBoundaries) -> pp.ad.Operator:
```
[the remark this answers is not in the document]
Yuriyzabegaev: added

src/porepy/models/momentum_balance.py
Yuriyzabegaev: I agree

src/porepy/numerics/ad/grid_operators.py
Yuriyzabegaev: added

src/porepy/models/fluid_mass_balance.py
vlipovac: Should I put it into a Note-sections in the docstring?

IvarStefansson: Yes, I think that'd be better

vlipovac: Alright.
Did the same for the enthalpy flux in `energy_balance.EnergyBalanceEquations`
Can you think of another another such case somewhere in pp?

IvarStefansson: No, I don't think there are more advective fluxes at the moment.

tests/visualization/times.json
vlipovac: Done,
We might wanna think about adding `tests/*.json` to `.gitignore`.

src/porepy/models/fluid_mass_balance.py
```diff
@@ -341,27 +373,25 @@ class ConstitutiveLawsSinglePhaseFlow(
...
 
+    bc_data_fluid_flux_key: str = "fluid_flux"
+    bc_data_darcy_flux_key: str = "darcy_flux"
+    """Name of the boundary data for the Neuman boundary condition."""
+    pressure_variable: str
+    """Name of the pressure variable."""
+    pressure: Callable[[pp.SubdomainsOrBoundaries], pp.ad.Operator]
+    darcy_flux: Callable[[pp.SubdomainsOrBoundaries], pp.ad.Operator]
+
     def bc_type_darcy(self, sd: pp.Grid) -> pp.BoundaryCondition:
```
IvarStefansson, line 395: I think we should rename to bc_type_darcy_flux

IvarStefansson: And would "bc_type_fluid_flux" be an improvement on "bc_type_mobrho"?

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -524,17 +520,31 @@ def __init__(
...
+    @property
+    def boundary_to_subdomain(self) -> Operator:
+        return SparseArray(
+            self._projection.transpose().tocsc(),
+            name="boundaries to subdomains projection",
+        )
+    
+    # @property
+    # def dirichlet_to_subdomain(self) -> Operator:
+        # dirichlet =
```
keileg, line 546:
```suggestion
```suggestion
    
    # @property
    # def dirichlet_to_subdomain(self) -> Operator:
        # dirichlet = 


```
```

File: src/porepy/numerics/ad/grid_operators.py [hunk 47]
```diff
@@ -524,17 +520,31 @@ def __init__(
...
```
keileg, line 545:
```suggestion

```

keileg, line 545:
```suggestion

```
Can we delete this?

tests/integration/models/setup_utils.py
```diff
@@ -117,17 +117,14 @@ def set_well_network(self) -> None:
         )
 
 
-class BoundaryConditionsMassAndEnergyDirNorthSouth(
-    pp.mass_and_energy_balance.BoundaryConditionsFluidMassAndEnergy
+class BoundaryConditionsMassDirNorthSouth(
+    # pp.fluid_mass_balance.BoundaryConditionsSinglePhaseFlow
```
keileg, line 121: Can we delete?

tests/integration/models/setup_utils.py
```diff
@@ -371,7 +314,7 @@ def time_dependent_bc_values_mechanics(
         return values.ravel("F")
 
 
-class NoPhysics(pp.ModelGeometry, pp.SolutionStrategy, pp.DataSavingMixin):
+class NoPhysics(pp.ModelGeometry, pp.SolutionStrategy, pp.DataSavingMixin, pp.BoundaryConditionMixin):
```
keileg, line 317: Please run black

tests/integration/models/test_constitutive_laws.py
```diff
@@ -55,8 +56,6 @@
         ("momentum_balance", "friction_coefficient", 1),
         ("momentum_balance", "friction_bound", 1),
         # Energy balance
-        ("energy_balance", "bc_values_enthalpy_flux", None),
```
keileg, line 58: Were these removed on purpose?

tests/integration/models/test_rediscretization.py
```diff
@@ -61,31 +61,29 @@ def assemble_linear_system(self) -> None:
...
-        for sd in subdomains:
-            domain_sides = self.domain_boundary_sides(sd)
-            vals_loc = np.zeros(sd.num_faces)
-            np.random.seed(0)
-            vals_loc[domain_sides.north] = np.random.rand(domain_sides.north.sum())
-            vals.append(vals_loc)
-        return pp.wrap_as_ad_array(np.hstack(vals), name="bc_values_darcy")
+        domain_sides = self.domain_boundary_sides(boundary_grid)
+        vals_loc = np.zeros(boundary_grid.num_cells)
+        np.random.seed(0)
```
keileg, line 84:
```suggestion
        # Fix the random seed to make it possible to debug in the future.
        np.random.seed(0)
```

tests/unit/ad/test_ad_operator_tree.py
```diff
@@ -257,13 +257,30 @@ def test_time_dependent_array():
...
-    sd_array = pp.ad.TimeDependentDenseArray("foo", subdomains=mdg.subdomains())
-    intf_array = pp.ad.TimeDependentDenseArray("bar", interfaces=mdg.interfaces())
+    sd_array = pp.ad.TimeDependentDenseArray("foo", domains=mdg.subdomains())
+    intf_array = pp.ad.TimeDependentDenseArray("bar", domains=mdg.interfaces())
+    bg_array = pp.ad.TimeDependentDenseArray("foobar", domains=mdg.boundaries())
+
+    # Check correct domain types
+    assert sd_array.domain_type == sd_array_top.domain_type == "subdomains"
+    assert intf_array.domain_type == "interfaces"
+    assert bg_array.domain_type == "boundary grids"
```
keileg, line 283: Should we not also test evaluation of bg_array?

tests/unit/ad/test_grid_operators.py
```diff
@@ -248,30 +247,31 @@ def test_boundary_grid_projection(mdg: pp.MixedDimensionalGrid, scalar: bool):
...
+    assert np.sum(subdomain_to_boundary[:, ind0:ind1]) == 2 * proj_dim
     # Check that the projection matrix for the second fracture is non-zero.
     ind0 = ind1
     ind1 += g_2.num_faces * proj_dim
-    assert np.sum(projection.subdomain_to_boundary()[:, ind0:ind1]) == 2 * proj_dim
+    assert np.sum(subdomain_to_boundary[:, ind0:ind1]) == 2 * proj_dim
     # The projection matrix for the intersection should be zero.
     ind0 = ind1
-    assert np.sum(projection.subdomain_to_boundary()[:, ind0:]) == 0
+    assert np.sum(subdomain_to_boundary[:, ind0:]) == 0
```
keileg, line 274: We don't have any tests of boundary_to_subdomain (I know, it is my fault). I think this should be the transpose, so perhaps we can add a test 

assert np.allclose((subdomain_to_boundary - boundary_to_subdomain.T).data, 0)

Or roughly so.

keileg (commented):
Excellent work, it is really nice to see how much natural treatment of boundary conditions have become. I have left some comments and suggestions, all of them really minor. Perhaps the two most important ones are.
1) In constitutive laws there are helper methods create_boundary_operator and make_boundary_operators. If you have decided the close names are not a problem, it is okay with me, but if not, please evaluate.
2) We may want to expand the documentation of the setter-methods for bcs (the ones that return numpy arrays) so that the user understands this is where numbers should be inserted. But if you think the current explatation is sufficient, we please leave it as it is.

tests/functional/setups/manu_flow_comp_3d_frac.py
```diff
@@ -509,25 +505,27 @@ def matrix_boundary_density(
             on the exterior boundary faces for the given ``time``.
 
         """
-        # Symbolic variables
-        x, y, z, t = sym.symbols("x y z t")
+        assert False, "We should not need it anymore."
```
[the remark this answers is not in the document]
Yuriyzabegaev: Purged

tests/unit/ad/test_grid_operators.py
Yuriyzabegaev: Added

tests/unit/ad/test_ad_operator_tree.py
Yuriyzabegaev: Included.

tests/integration/models/setup_utils.py
Yuriyzabegaev: Running black on tests/ modify 16 files. Should I commit them?

tests/integration/models/setup_utils.py
Yuriyzabegaev: Deleting it

src/porepy/models/fluid_mass_balance.py
```diff
@@ -392,63 +422,49 @@ def bc_type_mobrho(self, sd: pp.Grid) -> pp.BoundaryCondition:
...
-
-        # Concatenate to single array and wrap as ad.DenseArray
-        # We have forced the type of bc_values_array to be an ad.DenseArray, but mypy does
-        # not recognize this. We therefore ignore the typing error.
-        bc_values_array: pp.ad.DenseArray = pp.wrap_as_ad_array(  # type: ignore
-            np.hstack(bc_values), name="bc_values_mobility"
+        return np.zeros(boundary_grid.num_cells)
+
+    def bc_values_fluid_flux(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
+        """TODO"""
```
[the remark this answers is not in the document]
Yuriyzabegaev: Removed it, added docstring.

src/porepy/models/fluid_mass_balance.py
```diff
@@ -392,63 +422,49 @@ def bc_type_mobrho(self, sd: pp.Grid) -> pp.BoundaryCondition:
         # Define boundary condition on all boundary faces.
         return pp.BoundaryCondition(sd, boundary_faces, "dir")
 
-    def bc_values_darcy(self, subdomains: list[pp.Grid]) -> pp.ad.DenseArray:
-        """Boundary condition values for the Darcy flux.
+    def bc_values_pressure(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
```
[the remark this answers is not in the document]
Yuriyzabegaev: I agree

src/porepy/models/energy_balance.py
```diff
@@ -281,24 +288,54 @@ def interface_energy_flux(self, interfaces: list[pp.MortarGrid]) -> pp.ad.Operat
...
 
+        Note:
+            Note:
+            The advected entity in the enthalpy flux is a product of density,
+            enthalpy and mobility of the fluid.
+            When using upwinding, Dirichlet-type data for pressure and temperature
+            must also be provided on the Neumann-boundary when there is
+            an in-flux into the domain.
+            The advected entity must provide values on the boundary in this case, since
+            the upstream value of it is on the boundary.
```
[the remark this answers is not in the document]
Yuriyzabegaev: Now it can. I think this note is outdated.

src/porepy/models/constitutive_laws.py
```diff
@@ -963,17 +985,37 @@ def darcy_flux(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
...
+                name=self.bc_data_darcy_flux_key, domains=subdomains
+            )
+
         interfaces: list[pp.MortarGrid] = self.subdomains_to_interfaces(subdomains, [1])
-        projection = pp.ad.MortarProjections(self.mdg, subdomains, interfaces, dim=1)
+        intf_projection = pp.ad.MortarProjections(
+            self.mdg, subdomains, interfaces, dim=1
+        )
+
+        boundary_operator = self._make_boundary_operator(
```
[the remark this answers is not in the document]
Yuriyzabegaev: I agree that we should find better names.

src/porepy/models/boundary_condition.py
```diff
@@ -0,0 +1,201 @@
...
+        result.set_name(name)
+
+        # self.update_all_boundary_conditions()  # TODO: Remove this line
+
+        return result
+
+    @cached_property
+    def __bc_type_storage(self) -> dict[str, Callable[[pp.Grid], pp.BoundaryCondition]]:
+        """TODO"""
+        return {}
```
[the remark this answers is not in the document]
Yuriyzabegaev: Fixed.

src/porepy/models/energy_balance.py
```diff
@@ -673,30 +739,47 @@ def bc_values_enthalpy_flux(self, subdomains: list[pp.Grid]) -> pp.ad.DenseArray
...
+        result.set_name("bc_values_enthalpy")
+        return result
+
+    def boundary_temperature(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
+        """TODO"""
+        return self.fluid.temperature() * np.ones(boundary_grid.num_cells)
+
+    def boundary_fourier_flux(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
+        """TODO"""
+        return np.zeros(boundary_grid.num_cells)
```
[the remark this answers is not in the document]
Yuriyzabegaev: With boundary grids, this is not a problem anymore. Previously, it was a problem, because you had to set values for Dirichlet enthalpy flux. Now this is done by the ad machinery.

src/porepy/models/poromechanics.py
Yuriyzabegaev: Fixed.

src/porepy/models/fluid_mass_balance.py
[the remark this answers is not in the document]
Yuriyzabegaev: I guess, now it works as it should. There are 3 cases for the single phase fluid flow (the logic is the same for the thermal model):
1. Neumann `bc_type_mobrho` - we're setting the total influx of the conserved quantity.
2. Dirichlet `bc_type_mobrho`, Neumann `bc_type_darcy` - we're setting the fluid volumetric influx and pressure. It is multiplied by `mobility_rho` which takes pressure values.
3. Dirichlet `bc_type_mobrho` and `bc_type_darcy` - we're setting pressure values.

tests/integration/models/setup_utils.py
IvarStefansson: Yes, please

src/porepy/models/fluid_mass_balance.py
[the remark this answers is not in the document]
IvarStefansson: I'll create an issue to update the explanation in the tutorials.

tests/integration/models/setup_utils.py
Yuriyzabegaev: Done

src/porepy/models/constitutive_laws.py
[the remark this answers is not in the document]
Yuriyzabegaev: As @IvarStefansson proposed, I renamed `_make_boundary_operator` to `_combine_boundary_operators`.

src/porepy/models/energy_balance.py
```diff
@@ -633,70 +745,115 @@ def bc_type_fourier(self, sd: pp.Grid) -> pp.BoundaryCondition:
...
+        """Boundary conditions on all external boundaries for the advective flux in the
+        energy equation.
+
+        Warning:
+            This must not be different from mixed in definition of
+            :meth:`~porepy.models.fluid_mass_balance.BoundaryConditionsSinglePhaseFlow.
+            bc_type_darcy`.
+            Advective enthalpy flux is due to energy associated with mass entering the
+            system, which is the same as for the mass balance equation.
+            Unphysical systems arise otherwise.
```
vlipovac, line 757: I think this is important to address. And I also feel like there is some "unnatural" flexibility in how the user can set BC conditions and values, which cause confusion and unphysical systems if not done correctly.

src/porepy/models/fluid_mass_balance.py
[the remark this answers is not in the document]
keileg: I think we can resolve this for now, and then hope that the bigger question of consistency between boundary conditions will be addressed in the near future. @IvarStefansson Do you agree?

tests/integration/models/test_constitutive_laws.py
IvarStefansson: Yes, bc value methods are no longer of the type DomainList -> Operator.

src/porepy/models/boundary_condition.py
IvarStefansson: Can we resolve, @Yuriyzabegaev?

Yuriyzabegaev: The docstrings are updated, so resolving it

tests/functional/setups/manu_flow_comp_2d_frac.py
Yuriyzabegaev: I added missing docstrings, so resolving this.

src/porepy/models/solution_strategy.py
```diff
@@ -199,6 +204,7 @@ def prepare_simulation(self) -> None:
         self.initial_condition()
         self.reset_state_from_file()
         self.set_equations()
+        self.update_time_dependent_ad_arrays()  # This must be done after set_equations.
```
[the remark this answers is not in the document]
Yuriyzabegaev: This line is deleted now. Currently, `update_all_boundary_conditions` is called in:
*  `initial_condition`
*  `reset_state_from_file` only if we are actually resetting from a file
*  `before_nonlinear_loop`

@IvarStefansson, am I right?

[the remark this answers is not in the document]
IvarStefansson: Yes

keileg (approved):

PR #952, opened 2023-09-14
MAINT: improved input validation on cartesian- and tensor-grids

OmarDuran:
## Proposed changes

The input validation for cell_sizes has been improved. Code now raises a warning when the domain size is smaller than the informed cell size.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/grids/mdg_generation.py
+++ b/src/porepy/grids/mdg_generation.py
@@ -533,22 +533,44 @@ def _preprocess_cartesian_args(
 
     # If provided compute all the information from cell_size
     nx_cells: list[int] = [-1 for i in range(domain.dim)]
+    n_x: int = -1
+    n_y: int = -1
+    n_z: int = -1
 
     # if provided overwrite information in x-direction
     if cell_size_x is not None:
-        n_x: int = round((xmax - xmin) / cell_size_x)
+        n_x = round((xmax - xmin) / cell_size_x)
+        if np.abs((xmax - xmin)) - cell_size_x < 0.0:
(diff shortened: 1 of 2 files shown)


src/porepy/grids/mdg_generation.py
```diff
@@ -536,19 +536,34 @@ def _preprocess_cartesian_args(
 
     # if provided overwrite information in x-direction
     if cell_size_x is not None:
-        n_x: int = round((xmax - xmin) / cell_size_x)
+        if np.floor((xmax - xmin) / cell_size_x) < 1.0:
+            warnings.warn(
+                "In the x-direction, cell_size_x is greater than the domain. The domain "
+                "size is used instead."
+            )
+        n_x: int = int(np.max([np.floor((xmax - xmin) / cell_size_x), 1]))
```
IvarStefansson, line 544: I think we should keep round. Otherwise, passing a slightly too large cell size will yield unexpected results. 
MVE:

```
import numpy as np
ds = 1
cs = .066667 # approx 1/15
frac = ds / cs
fl = int(np.max([np.floor(frac), 1]))
ro = int(np.max([np.round(frac), 1]))
print(frac, fl, ro)
```

OmarDuran: I agree with your statement. Now the code is calculating the number of cells or points using round instead of floor.

IvarStefansson (approved):
Thanks!

IvarStefansson (approved):

Issue #953, opened 2023-09-15
Get solution values function


IngridKJ:
Getting variable values is easily done by using the method pp.get_variable_values(). If one wants to gather values that are not tied to a variable, the data dictionary needs to be approached "directly". I.e., something along the lines of:

data["parameter name"][pp.TIME_STEP_SOLUTION/pp.ITERATE_SOLUTION][index value] 

needs to be written. 

This issue is for making a get_solution_value function, analogous to the set_solution_value function, such that fetching values for non-variable quantities is easier. All instances of directly accessing the data dictionary for fetching values should be updated accordingly.

keileg:
Where do you plan to implement such a method? I think nobody will miss writing the direct call into the data dictionary.

IngridKJ:
As of now it is put at the bottom of equation_system.py together with the set_solution_values function. but I am not sure if that is the best location for it. I discussed only briefly with @IvarStefansson that such a method could be created, not too deeply exactly where to put it. Suggestions?

PR #956, opened 2023-09-21
953 get solution values function

IngridKJ:
## Proposed changes

Solves issue #953. Adds a function for fetching values from a data dictionary using the parameter name. Especially useful for obtaining values that are not associated with a variable. Also updates all known/found use of directly accessing the data dictionary in the brute force way. 

In addition to this, the set_solution_values function is moved to be within _ad_utils.py. This is also where get_solution_values function is put.

## Types of changes

- [ ] Other:

## Checklist


src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -175,6 +184,127 @@ def discretize_from_list(
...
+
+    Raises:
+        ValueError: If neither of `time_step_index` or `iterate_index` have been
+            assigned a non-None value.
+
+    """
+    if time_step_index is None and iterate_index is None:
+        raise ValueError(
+            "At least one of time_step_index and iterate_index needs to be different"
+            " from None"
```
IvarStefansson, line 220:
```suggestion
            " from None."
```

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -175,6 +184,127 @@ def discretize_from_list(
...
+def get_solution_values(
+    name: str,
+    data: dict,
+    time_step_index: Optional[int] = None,
+    iterate_index: Optional[int] = None,
+) -> np.ndarray:
+    """Function for fetching values stored in the data dictionary.
+
+    This function should be used for obtaining solution values that are not related to a
+    variable. This is to avoid the time consuming alternative of writing e.g.:
```
IvarStefansson, line 254:
```suggestion
    variable. This is to avoid the cumbersome alternative of writing e.g.:
```

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -175,6 +184,127 @@ def discretize_from_list(
...
+        KeyError: If there are no data values assigned to the time step/iterate index.
+
+    Returns:
+        An array containing the solution values.
+
+    """
+    if (time_step_index is None and iterate_index is None) or (
+        time_step_index is not None and iterate_index is not None
+    ):
+        raise ValueError(
```
IvarStefansson, line 282: I suggest splitting this in two separate checks.

tests/unit/ad/test__ad_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    pp.set_solution_values(
+        name=parameter_name, values=values_2, data=data, time_step_index=1
+    )
+    pp.set_solution_values(
+        name=parameter_name, values=values_3, data=data, iterate_index=1
+    )
+
+    # Values gathered from data dictionary directly
+    values_0_gathered_directly = data[pp.TIME_STEP_SOLUTIONS][parameter_name][0]
```
IvarStefansson, line 39: Can't you just use values_0 etc. for comparison below? Is there any reason to retrieve them from the dict?

tests/unit/ad/test__ad_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    assert (values_2_gathered_directly == values_2).all()
+    assert (values_3_gathered_directly == values_3).all()
+
+    assert (values_0_test_get == values_0_gathered_directly).all()
+    assert (values_1_test_get == values_1_gathered_directly).all()
+    assert (values_2_test_get == values_2_gathered_directly).all()
+    assert (values_3_test_get == values_3_gathered_directly).all()
+
+    # Check that the expected errors are raised
+    with pytest.raises(ValueError):
```
IvarStefansson, line 73: Very good that you 1) check that the errors are raised and 2) use pytest functionality!!!

IvarStefansson (commented):
Excellent work! A few very minor suggestions.

tests/unit/ad/test__ad_utils.py
IngridKJ: My line of thought was to test the set and get methods somewhat separately. Retrieving values directly from the dictionary shows that the set-method works without relying on the get method to work. But I see now that the get-method test somewhat relies on that the set method works.. Maybe I should just skip the intermediate step and get rid of the directly gathered values.

IvarStefansson: Or you keep them and test that values_0_gathered_directly == values_0

IvarStefansson: Right! Sorry, I missed it. In that case, let's keep it as is.

IvarStefansson (approved):

Issue #908, opened 2023-05-26
Unit test fails locally


mariusnevland:
The test dual_hybrid_vem_2d_ani_simplex fails locally on my computer. The problem seems to be a slightly too strict tolerance (10^-15). If I increase the tolerance to 10^-14, the test runs successfully.

mariusnevland:
I am using Windows 10 with an Intel processor. All of the required packages have been installed with conda. I am using:

Numba 0.57.0
Numpy 1.24.3
Scipy 1.10.1

IvarStefansson:
What's the status on this, @mariusnevland?

mariusnevland:
I haven't looked more into this yet @IvarStefansson.

PR #987, opened 2023-10-02
Convergence criterion

mariusnevland:
## Proposed changes

This PR adds a residual based convergence criterion to the non-linear solver, in addition to the currently used criterion based on the update increment. Some notes:

- A new parameter "nl_convergence_tol_res" has been introduced for the residual error tolerance. The tolerance for the increment error is set by the parameter "nl_convergence_tol" as before. I considered renaming it to "nl_convergence_tol_inc", but decided against it in order to not break already existing code.
- The default value of the residual error tolerance has been set to infinity. Hence, the default convergence criterion is the same as before.
- Progress bars for now only include increments. Adding residuals is not a big deal but would add more output per line. Suggestion actually is to add another standardized progress output which is more lightweight and prints all increment and residual errors for each iteration. The code is ready for that but left for a future PR.
- No new tests have been introduced, since there are already no tests for nonlinear solvers, as far as I know. See also one comment on tests in the discussion below.

## Types of changes

- [ ] Other:

## Checklist

I will also mention that I am still having issue #908 about one test failing locally on my computer.


src/porepy/models/solution_strategy.py
```diff
@@ -437,13 +438,14 @@ def after_nonlinear_convergence(
         self.save_data_time_step()
 
     def after_nonlinear_failure(
-        self, solution: np.ndarray, errors: float, iteration_counter: int
+        self, solution: np.ndarray, errors: dict, iteration_counter: int
```
IvarStefansson, line 441: I don't like changing the signature based on the needs identified in this particular PR. I appreciate the effort to make it more flexible by using a dict instead of, say, having the parameters errors_incr and errors_res. However, I think passing information between methods in this way is tricky when we can't predict what information may be required by future/tailor made versions of the methods. What about having a single object (dictionary, tailored class) for the current state of solver-relevant information? Perhaps simply as part of the (Non)LinearSolver object?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -55,9 +56,10 @@ def solve(self, model) -> tuple[float, bool, int]:
         Returns:
             A 3-tuple containing:
 
-            float:
+            dict:
 
-                The error estimate.
+                Dictionary containing the residual errors and increment errors for each
```
IvarStefansson, line 61: Isn't the return value the current errors only?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -165,15 +169,26 @@ def newton_step() -> None:
...
+            np.ndarray: Residual vector of nonlinear system.
+            np.ndarray: Solution to linearized system, i.e. the Newton update
+                increment.
+
         """
         model.assemble_linear_system()
+        _, res = model.linear_system  # Extract residual
         sol = model.solve_linear_system()
-        return sol
+        return res, sol
```
IvarStefansson, line 194: It's incorrect to use this residual, collected before solving, for the error computation. We need to re-evaluate after solving and updating the solution.

IvarStefansson (commented):
Thanks, Marius, this will turn out useful. Since we're entering somewhat uncharted territory by extending the solution strategy functionality, we need to consider design choices carefully. I'm afraid this requires some discussion and planning before we proceed.

src/porepy/models/solution_strategy.py
mariusnevland: If I understood correctly: The "errors" dictionary should possibly be an attribute of the NewtonSolver class? That does seem like a better solution. In that case though, wouldn't you still have to change the signature of the after_nonlinear_convergence/failure methods? At least if you would like to have both the residual and increment errors as input.

I can also mention that the previous "errors" object is a subset of the new "errors" object. The previous errors is a list, and the new errors is a dictionary of lists, so no information is lost. Related to this, is the current signature of after_nonlinear_convergence/failure wrong, because errors is a list and not a float?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
mariusnevland: Previously the method returned only the current error, but I changed it so it instead returns the entire dictionary containing the errors from every iteration. Is it important which option you choose here?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
mariusnevland: Yes, I realized this was not entirely the right way to do it. I am guessing we do not want to assemble the same system twice, though? So we need to assemble the system for extracting the residual, and then find a way to reuse this system for the next Newton iteration?

IvarStefansson:
I converted to draft, since this seems to be WIP. Please convert back and request a re-review from me once you've finished.

src/porepy/models/solution_strategy.py
jwboth: This discussion should now be resolved due the to introduction of the new `pp.SolverStatistics` object which is meant to be a medium of "whatever" is required. With internal tracking of these quantities, the signatures now changed again, as moving around data is not required anymore.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
jwboth: Again, this comment should now be irrelevant due to existence of `pp.SolverStatistics`.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
jwboth: This comment has been integrated in the recent changes. The residual is considered after update.

jwboth:
The original comments by Ivar from longtime ago should be taken care of. Thus, I do not see any issues related to those to be still open. Instead I would like to draw attention to recent updates and in particular some specific issues which involve a code design choice which may be discussed. These are highlighted below to ease the review.

src/porepy/models/solution_strategy.py
```diff
@@ -508,26 +516,78 @@ def check_convergence(
...
             # is returned. We check for this.
-            diverged = bool(np.any(np.isnan(solution)))
+            diverged = bool(np.any(np.isnan(solution_increment)))
             converged: bool = not diverged
-            error: float = np.nan if diverged else 0.0
-            return error, converged, diverged
+            residual_error: float = np.nan if diverged else 0.0
+            increment_error: float = np.nan if diverged else 0.0
+            # TODO decide whether to omit logging of errors for linear problems
+            # return residual_error, increment_error, converged, diverged
```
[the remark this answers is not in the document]
jwboth: Note also that a unified way allows a simple access to residuals and increments (only true for all subproblems) for debugging in case a solution strategy of a linear problem does not generate the correct solution.

src/porepy/models/solution_strategy.py
```diff
@@ -508,26 +516,78 @@ def check_convergence(
...
+            logger.info(f"Normalized increment error: {increment_error:.2e}")
+            logger.info(f"Normalized residual error: {residual_error:.2e}")
+            # Check convergence
+            converged_inc = increment_error < nl_params["nl_convergence_tol"]
+            # Allow for nan values in the residual error for effectively disabled
+            # convergence check.
+            converged_res = (
+                residual_error < nl_params["nl_convergence_tol_res"]
+                or nl_params["nl_convergence_tol_res"] is np.inf
+            )
```
jwboth, line 544: To explain: I managed to define characteristic functions which produce nan values (due to division by zero). Even when these in the end produce return 0 values, the residual still is nan. If this is confusing, I suggest removing these extra treatments from this PR. I just had them included as they have been part of active code development.

src/porepy/models/solution_strategy.py
```diff
@@ -508,26 +516,78 @@ def check_convergence(
...
+            converged = converged_inc and converged_res
             diverged = False
-            return error, converged, diverged
+
+        # Log the errors, increments and residuals
+        self.nonlinear_solver_statistics.log_error(increment_error, residual_error)
+        self.nonlinear_solver_statistics.log_increment(solution_increment)
+        self.nonlinear_solver_statistics.log_residual(residual, init_residual)
```
jwboth, line 551: The separation at the moment allows for usage of custom metrics for the global problem which are defined in the solution strategy and can be changed through mixins, while the subproblem increments and residuals are always measure in L2-like norms.

src/porepy/models/solution_strategy.py
```diff
@@ -560,6 +620,10 @@ def assemble_linear_system(self) -> None:
         self.linear_system = self.equation_system.assemble()
         logger.debug(f"Assembled linear system in {time.time() - t_0:.2e} seconds.")
 
+        # Definition of nonlinear solver strategy not finalized - redefine FIXME
+        if not self.nonlinear_solver_statistics.init_complete:
+            self.set_nonlinear_solver_statistics()
```
jwboth, line 625: The solver statistics requires info about equation names and variable names. For this it requires an active assembly. This is not elegant but kept for now to start the discussion of the overall PR.

src/porepy/models/solution_strategy.py
```diff
@@ -508,26 +516,78 @@ def check_convergence(
             # At least for the default direct solver, scipy.sparse.linalg.spsolve, no
             # error (but a warning) is raised for singular matrices, but a nan solution
             # is returned. We check for this.
-            diverged = bool(np.any(np.isnan(solution)))
+            diverged = bool(np.any(np.isnan(solution_increment)))
             converged: bool = not diverged
-            error: float = np.nan if diverged else 0.0
-            return error, converged, diverged
+            residual_error: float = np.nan if diverged else 0.0
+            increment_error: float = np.nan if diverged else 0.0
```
jwboth, line 522: It may be worth reflecting on the choice of naming here: Do we want to call the norm of the residual and increment `residual_error` and `increment_error` as right now? Or do we e.g. rather choose `norm_residual` and `norm_increment`. I actually rather prefer the second choice. Currently this PR rather consistently uses the former.

src/porepy/viz/solver_statistics.py
```diff
@@ -0,0 +1,323 @@
...
+        self.increment_errors: list[float] = []
+        """List of increments for each non-linear iteration."""
+        self.residual_errors: list[float] = []
+        """List of residual for each non-linear iteration."""
+        self.sub_increment_errors: dict["str", list[float]] = {}
+        """Dictionary of subincrements for each non-linear iteration."""
+        self.sub_residual_errors: dict["str", list[float]] = {}
+        """Dictionary of subresiduals for each non-linear iteration."""
+        self.init_sub_residual_errors: dict["str", float] = {}
+        """Dictionary of subinitial residuals for each time step."""
```
jwboth, line 150: A consistent choice of names is aimed at here. Yet, in my opinion, it starts becoming more cumbersome. The simplest would be to use `increments`, `residuals`, `increment_components`, `residual_components`. This would be my preferred suggestion. Consistently, in the other files.

jwboth (commented):
Hope these flashlights help to get a grasp of the code.

src/porepy/models/solution_strategy.py
```diff
@@ -508,26 +516,78 @@ def check_convergence(
...
+
+        Returns:
+            float: Update error.
+        """
+        # Simple but fairly robust convergence criterions. More advanced options are
+        # e.g. considering errors for each variable and/or each grid separately,
+        # possibly using _l2_norm_cell
+        # We normalize by the size of the solution vector.
+        # Enforce float to make mypy happy
+        increment_error = np.linalg.norm(solution_increment) / np.sqrt(
```
mariusnevland, line 587: Could these comments be removed? Or possibly moved to the check_convergence() method instead?

src/porepy/models/solution_strategy.py
OmarDuran: I vote for second choice `norm_residual` and `norm_increment` or the inflections `residual_norm` and `increment_norm`.

src/porepy/models/solution_strategy.py
```diff
@@ -476,26 +480,30 @@ def after_simulation(self) -> None:
...
                 implemented.
 
         Returns:
             The method returns the following tuple:
 
             float:
-                Error, computed to the norm in question.
+                Residual error, computed to the norm in question.
+            float:
+                Increment error, computed to the norm in quesiton.
```
jwboth, line 506:
```suggestion
                Increment error, computed to the norm in question.
```

IvarStefansson (commented):
Partial review. Main comment: Discussion with @jwboth resulted in a decision to try to split the SolverStatistics into a Model level mixin with functionality to compute metrics and a data storage class to be added as a model component/attribute.

keileg:
> Partial review. Main comment: Discussion with @jwboth resulted in a decision to try to split the SolverStatistics into a Model level mixin with functionality to compute metrics and a data storage class to be added as a model component/attribute.

Makes sense to me. Regarding storage, this sounds like a dataclass to me.

src/porepy/models/solution_strategy.py
```diff
@@ -187,6 +184,9 @@ def __init__(self, params: Optional[dict] = None):
 
         """
 
+        self.nonlinear_solver_statistics: pp.SolverStatistics
```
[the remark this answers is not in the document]
jwboth: I agree that it would be sound to make it part of the NewtonSolver. Yet, as `self.nonlinear_solver_statistics` is meant to be the interface to any stats as number of nonlinear iterations, and e.g. the `DataSavingModelMixin` is using it, it has been made part of the solution strategy.

src/porepy/models/solution_strategy.py
jwboth: @mariusnevland will set up an overview which will summarize all chosen namings. After that an overall discussion will be initiated.

jwboth:
> Partial review. Main comment: Discussion with @jwboth resulted in a decision to try to split the SolverStatistics into a Model level mixin with functionality to compute metrics and a data storage class to be added as a model component/attribute.



> > Partial review. Main comment: Discussion with @jwboth resulted in a decision to try to split the SolverStatistics into a Model level mixin with functionality to compute metrics and a data storage class to be added as a model component/attribute.
> 
> Makes sense to me. Regarding storage, this sounds like a dataclass to me.

SolverStatistics is now a dataclass where the functionality was stripped down to the bare minimum.

After giving it some thought and trying to squeeze all in, to me somehow it feels very strange to use a mixin which always(!) would need to be defined alongside SolutionStrategy, just to make it customizable. I therefore suggest removing all more advanced components from SolverStatistics from this PR. My sense instead is, the "dream" of having a dedicated object/class/mixin whatever being responsible for all metrics in a simulation (also for increments and residuals) will transfer any computed metrics to the solver statistics object. With this customization of "which" metrics to be computed, and with which mathematical "norm" essentially components that a user would want to steer are controllable and propagated to a sole accesible object (`self.nonlinear_solver_statistics`). 

@mariusnevland would this result in you being unhappy? I am a bit unsure whether you still check residuals of single equations on single grids, or whether you just plot full residuals. With the current stripped-down version, you would again need to hardcode the access to these quantities. I will most likely just do that in my stuff and enhance my local extension of PP.

@IvarStefansson @keileg I assume you do not care about the lack of that particular functionality at the moment?

mariusnevland:
> > Partial review. Main comment: Discussion with @jwboth resulted in a decision to try to split the SolverStatistics into a Model level mixin with functionality to compute metrics and a data storage class to be added as a model component/attribute.
> 
> > > Partial review. Main comment: Discussion with @jwboth resulted in a decision to try to split the SolverStatistics into a Model level mixin with functionality to compute metrics and a data storage class to be added as a model component/attribute.
> > 
> > 
> > Makes sense to me. Regarding storage, this sounds like a dataclass to me.
> 
> SolverStatistics is now a dataclass where the functionality was stripped down to the bare minimum.
> 
> After giving it some thought and trying to squeeze all in, to me somehow it feels very strange to use a mixin which always(!) would need to be defined alongside SolutionStrategy, just to make it customizable. I therefore suggest removing all more advanced components from SolverStatistics from this PR. My sense instead is, the "dream" of having a dedicated object/class/mixin whatever being responsible for all metrics in a simulation (also for increments and residuals) will transfer any computed metrics to the solver statistics object. With this customization of "which" metrics to be computed, and with which mathematical "norm" essentially components that a user would want to steer are controllable and propagated to a sole accesible object (`self.nonlinear_solver_statistics`). But this remains then a future project.
> 
> @mariusnevland would this result in you being unhappy? I am a bit unsure whether you still check residuals of single equations on single grids, or whether you just plot full residuals. With the current stripped-down version, you would again need to hardcode the access to these quantities. I will most likely just do that in my stuff and enhance my local extension of PP.
> 
> @IvarStefansson @keileg I assume you do not care about the lack of that particular functionality at the moment?

I am not checking residuals of single equations/grids at the moment, but will probably do so in the future, so it would be nice to have this functionality included in SolverStatistics. However, I am also fine with it being removed, since I also have my own scripts for doing this.

IvarStefansson:
> > Partial review. Main comment: Discussion with @jwboth resulted in a decision to try to split the SolverStatistics into a Model level mixin with functionality to compute metrics and a data storage class to be added as a model component/attribute.
> 
> > > Partial review. Main comment: Discussion with @jwboth resulted in a decision to try to split the SolverStatistics into a Model level mixin with functionality to compute metrics and a data storage class to be added as a model component/attribute.
> > 
> > 
> > Makes sense to me. Regarding storage, this sounds like a dataclass to me.
> 
> SolverStatistics is now a dataclass where the functionality was stripped down to the bare minimum.
> 
> After giving it some thought and trying to squeeze all in, to me somehow it feels very strange to use a mixin which always(!) would need to be defined alongside SolutionStrategy, just to make it customizable. I therefore suggest removing all more advanced components from SolverStatistics from this PR. My sense instead is, the "dream" of having a dedicated object/class/mixin whatever being responsible for all metrics in a simulation (also for increments and residuals) will transfer any computed metrics to the solver statistics object. With this customization of "which" metrics to be computed, and with which mathematical "norm" essentially components that a user would want to steer are controllable and propagated to a sole accesible object (`self.nonlinear_solver_statistics`). But this remains then a future project.
> 
> @mariusnevland would this result in you being unhappy? I am a bit unsure whether you still check residuals of single equations on single grids, or whether you just plot full residuals. With the current stripped-down version, you would again need to hardcode the access to these quantities. I will most likely just do that in my stuff and enhance my local extension of PP.
> 
> @IvarStefansson @keileg I assume you do not care about the lack of that particular functionality at the moment?

I do not need the functionality immediately. But I'm sure it will be useful to include at least part of it (and further additions) in a (hopefully not too distant) future PR!

src/porepy/models/solution_strategy.py
jwboth: Here is an overview of used/introduced terminology concerned with measuring norms for different files (as it is not consistent atm):

**File: src/porepy/models/solution_strategy.py**
pp.SolverStatistics: Object monitoring the performance of the nonlinear solver in terms of iterations, norms of increments and residuals as computed throughout the simulation
solution_increment (np.ndarray): solution vector to Newton's method
residual (np.ndarray): full residual vector 
init_residual (np.ndarray): full residual vector at first iteration
nonlinear_residual_error (method): Method to determine the residual_error (See below)
nonlinear_increment_error (method): Method to determine the increment_error (See below)
residual_error (float): norm of the residual
increment_error (float): norm of the increment
In the DOC it is also written 'update' or 'update increment' in some places. 

**File: src/porepy/numerics/linear_solvers.py**
error_res (float): norm of the residual
error_inc (float): norm of the increment

**File: src/porepy/numerics/nonlinear/nonlinear_solvers.py**
error_res (float): norm of the residual
error_inc (float): norm of the increment
sol (np.ndarray): increment vector
res (np.ndarray): residual vector
DOC: Use of 'Newton update increment'

**File: src/porepy/viz/solver_statistics.py**
increment_error (float): norm of increment vector
residual_error (Float): norm or residual vector
increment_errors (list): list of norms of increments vectors
residual_errors (list): list of norms of residual vectors

--------

To turn around the overview, here is a list of relevant quantities that ask for consistent naming:

residual vector
increment vector
norm of residual vector
norm of increment vector
method to determine norm of residual vector
method to determine norm of increment vector

I do not have strong opinions, but here are some suggestions, respectively:

residual
increment
residual_norm
increment_norm
determine_residual_norm()
determine_increment_norm()

Changes in solverstatistics would follow accordingly.

src/porepy/models/solution_strategy.py
jwboth: These comments have been inherited from the original code. I do not agree that check_convergence is a better place as check_convergence uses nonlinear_increment_error() only blindly, and it could possibly be replaced through a mixin.

The comment on mypy should be obsolete, though.

jwboth:
> > > Partial review. Main comment: Discussion with @jwboth resulted in a decision to try to split the SolverStatistics into a Model level mixin with functionality to compute metrics and a data storage class to be added as a model component/attribute.
> > 
> > 
> > > > Partial review. Main comment: Discussion with @jwboth resulted in a decision to try to split the SolverStatistics into a Model level mixin with functionality to compute metrics and a data storage class to be added as a model component/attribute.
> > > 
> > > 
> > > Makes sense to me. Regarding storage, this sounds like a dataclass to me.
> > 
> > 
> > SolverStatistics is now a dataclass where the functionality was stripped down to the bare minimum.
> > After giving it some thought and trying to squeeze all in, to me somehow it feels very strange to use a mixin which always(!) would need to be defined alongside SolutionStrategy, just to make it customizable. I therefore suggest removing all more advanced components from SolverStatistics from this PR. My sense instead is, the "dream" of having a dedicated object/class/mixin whatever being responsible for all metrics in a simulation (also for increments and residuals) will transfer any computed metrics to the solver statistics object. With this customization of "which" metrics to be computed, and with which mathematical "norm" essentially components that a user would want to steer are controllable and propagated to a sole accesible object (`self.nonlinear_solver_statistics`). But this remains then a future project.
> > @mariusnevland would this result in you being unhappy? I am a bit unsure whether you still check residuals of single equations on single grids, or whether you just plot full residuals. With the current stripped-down version, you would again need to hardcode the access to these quantities. I will most likely just do that in my stuff and enhance my local extension of PP.
> > @IvarStefansson @keileg I assume you do not care about the lack of that particular functionality at the moment?
> 
> I do not need the functionality immediately. But I'm sure it will be useful to include at least part of it (and further additions) in a (hopefully not too distant) future PR!

Great. This means the reduction to the minimum functionality is finalized.

src/porepy/models/solution_strategy.py
```diff
@@ -450,21 +460,17 @@ def after_nonlinear_convergence(
             values=solution, time_step_index=0, additive=False
         )
         self.convergence_status = True
-
+        self.nonlinear_solver_statistics.save()
```
IvarStefansson, line 463: Perhaps this should be done as part of save_data_time_step()?

src/porepy/models/solution_strategy.py
IvarStefansson: I like your suggestion. Two questions:
1. Does it make sense to specify that this is the "nonlinear" increment?
2. Is "compute" instead of "determine" more consistent with rest of code?

src/porepy/models/solution_strategy.py
```diff
@@ -476,26 +482,30 @@ def after_simulation(self) -> None:
...
 
         Parameters:
-            solution: Newly obtained solution vector prev_solution: Solution obtained in
-            the previous non-linear iteration. init_solution: Solution obtained from the
-            previous time-step. nl_params: Dictionary of parameters used for the
-            convergence check.
+            solution_increment: Newly obtained solution increment vector
+            residual: Residual vector of non-linear system, evaluated at the newly
+            obtained solution vector.
+            init_residual: Reference residual vector of non-linear system, evaluated
```
IvarStefansson, line 496: Isn't "reference_residual" a better name?

src/porepy/models/solution_strategy.py
```diff
@@ -508,26 +518,68 @@ def check_convergence(
...
-            error = float(np.linalg.norm(solution)) / np.sqrt(solution.size)
-            logger.info(f"Normalized residual norm: {error:.2e}")
-            converged = error < nl_params["nl_convergence_tol"]
+                return np.nan, np.nan, False, True
+
+            # Increment based error
+            increment_error = self.nonlinear_increment_error(solution_increment)
+            # Residual based error
+            residual_error = self.nonlinear_residual_error(residual, init_residual)
+            logger.info(f"Normalized increment error: {increment_error:.2e}")
```
IvarStefansson, line 535: Please merge logger messages to one line of output. Also, as we now have the option of printing to file, please consider to change logger level to debug.

src/porepy/models/solution_strategy.py
```diff
@@ -508,26 +518,68 @@ def check_convergence(
...
+
+    def nonlinear_residual_error(
+        self, residual: np.ndarray, init_residual: np.ndarray
+    ) -> float:
+        """Compute the residual error for a nonlinear iteration.
+
+        Parameters:
+            residual: Residual of current iteration.
+            init_residual: Reference residual value (initial residual expected), allowing
+                for definiting relative criteria.
```
IvarStefansson, line 556:
```suggestion
                for defining relative criteria.
```

src/porepy/models/solution_strategy.py
```diff
@@ -508,26 +518,68 @@ def check_convergence(
...
+
+        """
+        residual_error = np.linalg.norm(residual) / np.sqrt(residual.size)
+        return residual_error
+
+    def nonlinear_increment_error(self, solution_increment: np.ndarray) -> float:
+        """Compute the error based on the update increment for a nonlinear iteration.
+
+        Parameters:
+            solution_increment: Solution to the Newton linearization.
```
IvarStefansson, line 569:
```suggestion
            solution_increment: Solution to the linearization.
```

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, params=None) -> None:
         default_options = {
             "max_iterations": 10,
             "nl_convergence_tol": 1e-10,
+            "nl_convergence_tol_res": np.inf,
```
IvarStefansson, line 39: Why this tolerance?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -147,34 +147,39 @@ def newton_step() -> None:
                     )
                     newton_step()
                     solver_progressbar.update(n=1)
-                    solver_progressbar.set_postfix_str(f"Error {error_norm}")
+                    solver_progressbar.set_postfix_str(f"Inc. {error_inc}")
```
IvarStefansson, line 150: Isn't "Inc." too ambiguous?

src/porepy/viz/solver_statistics.py
```diff
@@ -0,0 +1,91 @@
...
+class SolverStatistics:
+    """Statistics object for non-linear solver loop.
+
+    This object keeps track of the number of non-linear iterations performed for the
+    current time step, as well as increments and residuals for each iteration.
+
+    Example:
+
+        After storing solver statistics to file, we can load the file and plot
+        analogous data, here for the first time step.
```
IvarStefansson, line 24: In what sense "analogous"?

src/porepy/viz/solver_statistics.py
```diff
@@ -0,0 +1,91 @@
...
+
+        >>> import matplotlib.pyplot as plt
+        >>> import json
+        >>> with open("solver_statistics.json", "r") as f:
+        >>>     history = json.load(f)
+        >>> ts = str(1)
```
IvarStefansson, line 30:
```suggestion
        >>> time_step = str(1)
```

src/porepy/viz/solver_statistics.py
```diff
@@ -0,0 +1,91 @@
...
+        >>> err = history[ts]["residual_errors"]
```
IvarStefansson, line 31:
```suggestion
        >>> err = history[time_step]["residual_errors"]
```

src/porepy/viz/solver_statistics.py
```diff
@@ -0,0 +1,91 @@
...
+        >>> plt.ylabel("Residual")
+        >>> plt.title("Residual error")
+        >>> plt.show()
+
+    """
+
+    num_iteration: int = 0
+    """Number of non-linear iterations performed for current time step."""
+    increment_errors: list[float] = field(default_factory=list)
+    """List of increments for each non-linear iteration."""
```
IvarStefansson, line 43:
```suggestion
    """List of increment magnitudes for each non-linear iteration."""
```

src/porepy/viz/solver_statistics.py
```diff
@@ -0,0 +1,91 @@
...
+    residual_errors: list[float] = field(default_factory=list)
+    """List of residual for each non-linear iteration."""
+    path: Optional[Path] = None
+    """Path to save the statistics object to."""
+
+    def log_error(self, increment_error: float, residual_error: float) -> None:
```
IvarStefansson, line 49: If I want to keep track of other errors, I'd have to change the signature of this method, right? Is it worth considering to replace the explicit arguments by kwargs?

src/porepy/viz/solver_statistics.py
```diff
@@ -0,0 +1,91 @@
...
+        """Reset the statistics object."""
+        self.num_iteration = 0
+        self.increment_errors.clear()
+        self.residual_errors.clear()
+
+    def save(self) -> None:
+        """Save the statistics object to a JSON file.
+
+        Parameters:
+            path (Path): Path to the file to save the statistics object to.
```
IvarStefansson, line 70: Wrong documentation. The method takes no parameters.

IvarStefansson (commented):
Thanks! I think we're almost there. Some relatively minor suggestions.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
jwboth: The default case was intended to not break with the previous state of the code. This tolerance deactivates checking residuals to accept convergence. With this, the solver behavior of existing simulations should not change with this PR.

src/porepy/models/solution_strategy.py
jwboth: Done.

src/porepy/models/solution_strategy.py
jwboth: Done.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
jwboth: Changed to "Increment" now. Better? As this is going to be printed to the screen, I would be cautious with too long names.

src/porepy/viz/solver_statistics.py
jwboth: Changed to "stored".

src/porepy/viz/solver_statistics.py
keileg: Or a combination. Always storing the main errors seem sound to me.

src/porepy/models/solution_strategy.py
keileg: > I like your suggestion. Two questions:
> 
> 1. Does it make sense to specify that this is the "nonlinear" increment?

To me, yes. There will also be a time increment.

> 2. Is "compute" instead of "determine" more consistent with rest of code?

'Compute' fits better with how I picture the process (determine sounds more cumbersome and possibly less accurate), but I do not have very strong feelings here.

src/porepy/models/solution_strategy.py
jwboth: Would be OK for me. However, I would then aim for a consistent change. This would include:
- Move the handling of the path where to store the performance data also to DataSavingMixin

In addition, I would advocate for saving the data also after failed convergence, as now, such that in the end, the routine would not entirely vanish.

src/porepy/viz/solver_statistics.py
jwboth: If you would want to customize what errors are kept track of/stored to file, more changes would be needed at the moment. So my answer atm would be "no", but in the near future, I may change my view.

jwboth: Sorry, was working on outdated window and did not see comments by @keileg. The main idea here was just to store the main errors for now. The functionality for storing other errors has been removed for this PR.

keileg: Fair enough

src/porepy/models/solution_strategy.py
jwboth: Done as commented here now.

src/porepy/models/solution_strategy.py
IvarStefansson: I agree. We could, though, consider calling `save_data_time_step` after failure as well.

src/porepy/viz/solver_statistics.py
IvarStefansson: Yes, but we should avoid changing signatures in two months.

jwboth: @keileg @IvarStefansson I have no strong opinion here. I would keep the signature, as I think the standard errors will remain no mater how new functionality will enter. And optional errors will then be added through kwargs. Of course one could also consider using kwargs for all of them. If you want to decide, give me a sign and I will update the signature here accordingly. This is the last open comment.

src/porepy/models/solution_strategy.py
```diff
@@ -414,34 +414,29 @@ def before_nonlinear_iteration(self) -> None:
...
         self.equation_system.set_variable_values(
-            values=solution_vector, additive=True, iterate_index=0
+            values=nonlinear_increment, additive=True, iterate_index=0
         )
+        self.nonlinear_solver_statistics.num_iteration += 1
 
-    def after_nonlinear_convergence(
-        self, solution: np.ndarray, errors: float, iteration_counter: int
-    ) -> None:
+    def after_nonlinear_convergence(self, solution: np.ndarray) -> None:
```
jwboth, line 433: Is there a historical or important reason why `solution` has been offered as part of the signature here? Otherwise, I would remove it.

src/porepy/models/solution_strategy.py
```diff
@@ -450,21 +445,16 @@ def after_nonlinear_convergence(
             values=solution, time_step_index=0, additive=False
         )
         self.convergence_status = True
-
         self.save_data_time_step()
 
-    def after_nonlinear_failure(
-        self, solution: np.ndarray, errors: float, iteration_counter: int
-    ) -> None:
+    def after_nonlinear_failure(self, solution: np.ndarray) -> None:
```
jwboth, line 450: Same comment regarding `solution` here.

src/porepy/models/solution_strategy.py
jwboth: @keileg @IvarStefansson two more questions, for which you possibly have an answer.

IvarStefansson: I guess the answer is yes, there is some historical reason. But I see no reason to keep it, as evident from the fact that it's fetched from storage in line 442.

keileg: Agreed.

src/porepy/viz/solver_statistics.py
keileg: Then let us add `kwargs` as a way to include future extensions, so that the new signature is 

```python
    def log_error(self, increment_error: float, residual_error: float, **kwargs) -> None:
```

jwboth: Done.

src/porepy/models/solution_strategy.py
jwboth: Removed.

IvarStefansson (approved):
Looks good to me. Thanks for the great work!

PR #1004, opened 2023-10-06
Issue 969

vlipovac:
Refactored tests in relevant files

Note: 'test_mdg_generation.py` only moved, not modified since already based on pytest


tests/grids/test_grid_extrusion.py
```diff
@@ -0,0 +1,686 @@
...
+
+    z = np.array([0, 1, 2])
+    mdg_new, _ = pp.grid_extrusion.extrude_grid_bucket(mdg, z)
+
+    for dim in range(mdg.dim_max()):
+        assert len(mdg.subdomains(dim=dim)) == len(mdg_new.subdomains(dim=dim + 1))
+
+    # Get hold of interface grids. We know there is a single interface in each
+    # mixed-dimensional grid, so make a loop and abort it.
+    for intf in mdg.interfaces():
```
vlipovac, line 594: It looks like this test was in its original form incomplete, or more was intended at least

tests/grids/test_grid_refinement.py
```diff
@@ -17,104 +15,104 @@
...
-        self.assertTrue(np.sum(parent == 0) == 4)
-        self.assertTrue(np.sum(parent == 1) == 4)
-        self.assertTrue(np.allclose(np.bincount(parent, h.cell_volumes), 0.5))
+        assert np.sum(parent == 0) == 4
+        assert np.sum(parent == 1) == 4
+        assert np.allclose(np.bincount(parent, h.cell_volumes), 0.5)
 
 
-class TestRefinementMortarGrid(unittest.TestCase):
+class TestRefinementMortarGrid:
```
keileg, line 113: The structure of this set of tests seem very similar: Generate an md-grid, do some replacement operation (often of the mortar grid), verify that the new mortar projections have the right form. Have you considered whether it is possible to do parametrization of sorts here? I realize the answer may be the tests are sufficiently different for this not to be meaningful, but I thought it best to be sure.

tests/grids/test_grid_refinement.py
```diff
@@ -17,104 +15,104 @@
...
     def test_mortar_grid_1d(self):
```
keileg, line 114: Despite the name of this class, I think these tests should rather go in test_mortar_grid, since they mainly check that the mortar mappings are correctly updated. Does this make sense?

tests/grids/test_grid_extrusion.py
```diff
@@ -0,0 +1,686 @@
...
+    assert np.allclose(gn.nodes, coord_known)
+    assert check_cell_map(cell_map, g, z.size - 1)
+    assert check_face_map(face_map, g, z.size - 1)
+
+
+class TestGridExtrusion2D:
+    """Testing grid extrusion in 2D.
+
+    Note:
+        (VL) Since all test are quite individual, a parametrization using pytest would
```
keileg, line 118: I think this a good judgement.

tests/grids/test_grid_refinement.py
```diff
@@ -5,10 +5,8 @@
 """
```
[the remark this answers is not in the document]
vlipovac: Done

tests/grids/test_grid_refinement.py
vlipovac: So this method `test_mortar_grid_1d` should not be in this class, but a test as an independent function?

tests/grids/test_grid_refinement.py
vlipovac: I thought so as well, but I did not find a satisfying solution for a unification of the tests, since all of them perform different operations on the mortar grid in terms of how it is refined
the first test refines all side grids with a fixed level
the second test makes a refinement based on some side grid related value
the third test makes yet another uniform refinement, plus a substitution of one grid, with another

vlipovac: Let me come back to this

tests/grids/test_grid_refinement.py
vlipovac: I put this test into `tests/grids/test_mortar_grid.py`

tests/grids/test_grid_refinement.py
vlipovac: I was able to make two groupings of the tests
1 parametrized test where all side grids are refined
1 parametrized test where only one side grid is refined, which has another assertion included

The parametrization is not pretty though...

PR #1088, opened 2023-12-29
bug match grids

enricoballini:
## Proposed changes

There is a mistake in the function match_grids_along_1d_mortar which occurs when a new 2D grid, different from the the old 2D, is replaced. The tests fail to spotlight the error because theirs meshes are made of too few elements

## Types of changes

- [ ] Other:

## Checklist


keileg:
Thank you, @enricoballini. I may make some additions before merging, but it may take a few days before I get around to do so.

keileg:
Note to self: The bugfix is correct, one of the two matrices in the expression must be transposed, as they are formed by the same function, which creates mappings between grids of different dimensions.

This bug also indicates that the current suite of tests for non-matching grids is insufficient. I will file a separate issue.

keileg (approved):
This makes sense. Thanks for the bugfix!

PR #1090, opened 2024-01-03
BUG: Allow .value calls for ad functions

IvarStefansson:
## Proposed changes

After replacing .evaluate with .value and .value_and_jacobian, a call to the former for a function wrapped as ad.Function requires handling of arguments of non-AdArray type (e.g. numpy array). This was not implemented for all members of the ad.functions module. PR also contains improved typing of said module.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/ad/functions.py
+++ b/src/porepy/numerics/ad/functions.py
@@ -20,7 +20,7 @@ Examples:
 """
 from __future__ import annotations
 
-from typing import Callable, TypeVar, Union, overload
+from typing import Callable, TypeVar
 
 import numpy as np
 import scipy.sparse as sps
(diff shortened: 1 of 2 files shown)


src/porepy/numerics/ad/functions.py
```diff
@@ -401,10 +399,22 @@ def maximum(var_0: pp.ad.AdArray, var_1: pp.ad.AdArray | np.ndarray) -> pp.ad.Ad
...
-def characteristic_function(tol: float, var: pp.ad.AdArray):
+
+@overload
+def characteristic_function(tol: float, var: AdArray) -> AdArray:
+    ...
+
+
+def characteristic_function(
+    tol: float, var: Union[AdArray, np.ndarray, float]
+) -> np.ndarray | AdArray:
```
keileg, line 417: This is the place where you highlighted the choice between overload and TypeVar? If so, I much prefer TypeVar if that is fine with you.

keileg (commented):
One remark only. Thanks for taking care of this.

src/porepy/numerics/ad/functions.py
IvarStefansson: Meaning that we change the return type to float, rather than bool? If yes, do you want me to do it?

IvarStefansson:
> One remark only. Thanks for taking care of this.

Thanks for the quick review! I think avoiding waiting for reviews makes us significantly more efficient overall.

src/porepy/numerics/ad/functions.py
keileg: Yes and yes, please.

PR #1092, opened 2024-01-10
Vectorized block diagonal mat inversion

OmarDuran:
## Proposed changes

This PR refactor and introduce a vectorized solution for the function `invert_diagonal_blocks_python` in file `matrix_operations.py`. The function is 2x faster than numba function for a large block diagonal matrix with irregular block sizes.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/linalg/matrix_operations.py
+++ b/src/porepy/numerics/linalg/matrix_operations.py
@@ -555,9 +555,6 @@ def invert_diagonal_blocks(
         """
         Invert block diagonal matrix using pure python code.
 
-        The implementation is slow for large matrices, consider to use the
-        numba-accelerated method invert_invert_diagagonal_blocks_numba instead
-
         Parameters
         ----------
         a: sps.crs-matrix, to be inverted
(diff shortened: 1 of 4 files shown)


src/porepy/numerics/linalg/matrix_operations.py
```diff
@@ -567,19 +567,24 @@ def invert_diagonal_blocks_python(a: sps.spmatrix, sz: np.ndarray) -> np.ndarray
...
+        row, col = a.nonzero()
+        idx_blocks = np.cumsum([0] + list(sz))
+        idx_nnz = np.searchsorted(row, idx_blocks)
+
+        def sub_block(ib):
+            lr = row[idx_nnz[ib] : idx_nnz[ib + 1]] - idx_blocks[ib]
+            lc = col[idx_nnz[ib] : idx_nnz[ib + 1]] - idx_blocks[ib]
+            dense_block = np.zeros((sz[ib], sz[ib]))
+            dense_block[lr, lc] = a.data[idx_nnz[ib] : idx_nnz[ib + 1]]
+            return dense_block
```
keileg, line 583: Please document this part.

src/porepy/numerics/linalg/matrix_operations.py
```diff
@@ -694,15 +699,15 @@ def inv_python(indptr, ind, data, sz):
 
     # Remove blocks of size 0
     s = s[s > 0]
-    # Variable to check if we have tried and failed with numba
-    if method == "numba" or method is None:
+    # Select python vectorized function
+    if method == "python" or method is None:
```
keileg, line 703: I agree with making this the default, but to do so properly, we should also change the default value of the inverter in mpfa, mpsa (and possibly Biot).

src/porepy/numerics/linalg/matrix_operations.py
```diff
@@ -694,15 +699,15 @@ def inv_python(indptr, ind, data, sz):
...
+        inv_vals = invert_diagonal_blocks_python(mat, s)
```
keileg, line 704: (emtpy comment, just to start a formal review). Disregard.

keileg (commented):
Control question: You have verified that the inverted matrices are correct?

Given the performance results, I think we should make this into the default inverter. This will require changing parameter values in mpfa, mpsa and biot. A repository-wide search for 'inverter' may be needed.

src/porepy/numerics/linalg/matrix_operations.py
OmarDuran: I have documented the function.

src/porepy/numerics/linalg/matrix_operations.py
OmarDuran: I have updated the default on those files.

PR #1096, opened 2024-01-14
Mixed-dimensional grids for case 3, flow benchmark 3d

jhabriel:
## Proposed changes

This PR introduces the possibility to create `mdg`s for the geometry corresponding to case 3 of the 3d flow benchmark. Four level of refinements are available (levels 0 and 1 are the ones used by UiB in the benchmark). A weak _in-situ_ test is incorporated in the function.

Please, feel free to modify and extend the functionality as you see fit.

Screenshot 2024-01-14 at 14 34 27

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/applications/md_grids/mdg_library.py
+++ b/src/porepy/applications/md_grids/mdg_library.py
@@ -14,6 +14,8 @@ import porepy as pp
 from porepy.fracs.fracture_network_2d import FractureNetwork2d
 from porepy.fracs.fracture_network_3d import FractureNetwork3d
 
+from pathlib import Path
+
 from . import domains, fracture_sets
 
 
@@ -172,3 +174,53 @@ def benchmark_regular_2d(
     if is_coarse:
         pp.coarsening.coarsen(mdg, "by_volume")
     return mdg, fracture_network
+
+
+def benchmark_3d_case_3(
+        refinement_level: Literal[0, 1, 2, 3] = 0,
+        ) -> pp.MixedDimensionalGrid:
+    """
+    Create a mixed-dimensional grid for the geometry of case 3 from [1].
+
+    Note:
+        The mixed-dimensional grid is created by reading a `geo` file, so there is no
+        direct way of prescribing meshing arguments.
+
+    Reference:
+        Berre, I., Boon, W. M., Flemisch, B., Fumagalli, A., Gläser, D., Keilegavlen,
+        E., ... & Zulian, P. (2021). Verification benchmarks for single-phase flow in
+        three-dimensional fractured porous media. Advances in Water Resources, 147,
+        103759.
+
+    Parameters:
+        refinement_level: An integer denoting the level of refinement. Use `0` to
+            generate a mixed-dimensional grid with approximately 30K 2D cells,
+            `1` for 140K 2D cells, `2` for 350K cells, and `3` for 500K cells.
+            Default is `0`.
+
+    Returns:
+        Mixed-dimensional grid.
+
+    """
+    # Get directory pointing to the `geo` file
+    abs_path = Path(__file__)
+    benchmark_path = abs_path.parent / Path("geo_library") / Path("benchmark_3d_case_3")
+    if refinement_level == 0:
+        full_path = benchmark_path / Path("mesh30k.geo")
+    elif refinement_level == 1:
+        full_path = benchmark_path / Path("mesh140k.geo")
+    elif refinement_level == 2:
+        full_path = benchmark_path / Path("mesh350k.geo")
+    elif refinement_level == 3:
+        full_path = benchmark_path / Path("mesh500k.geo")
+    else:
+        raise NotImplementedError("Refinement level not available.")
+
+    # Create mixed-dimensional grid
+    mdg = pp.fracture_importer.dfm_from_gmsh(str(full_path), dim=3)
+
+    # In-situ weak test
+    assert (mdg.dim_max() == 3) and (mdg.dim_min() == 1)
+    assert (len(mdg.subdomains()) == 16) and (len(mdg.interfaces()) == 22)
+
+    return mdg
--- a/src/porepy/applications/md_grids/mdg_library.py
+++ b/src/porepy/applications/md_grids/mdg_library.py
@@ -6,6 +6,7 @@ Mainly for use in tests. Other usage should be covered by the model_geometries.
 
 from __future__ import annotations
 
+from pathlib import Path
 from typing import Literal, Optional, cast
 
 import numpy as np
@@ -13,8 +14,6 @@ import numpy as np
 import porepy as pp
 from porepy.fracs.fracture_network_2d import FractureNetwork2d
 from porepy.fracs.fracture_network_3d import FractureNetwork3d
-
-from pathlib import Path
 
 from . import domains, fracture_sets
 
@@ -177,8 +176,8 @@ def benchmark_regular_2d(
 
 
 def benchmark_3d_case_3(
-        refinement_level: Literal[0, 1, 2, 3] = 0,
-        ) -> pp.MixedDimensionalGrid:
+    refinement_level: Literal[0, 1, 2, 3] = 0,
+) -> pp.MixedDimensionalGrid:
     """
     Create a mixed-dimensional grid for the geometry of case 3 from [1].
 
--- a/src/porepy/applications/md_grids/geo_library/benchmark_3d_case_3/mesh140k.geo
+++ b/src/porepy/applications/md_grids/geo_library/benchmark_3d_case_3/mesh140k.geo
@@ -1,0 +1,470 @@
+Geometry.Tolerance = 3.33333333333e-05;
+
+// Mesh size parameters. Roughly grouped into classes with apparently similar geometric complexity.
+// Feel free to refine this.
+
+// Domain corners
+h_domain = 0.06;
+
+// Fracture 1, left (low x) and right ends
+h_1_left = 0.05;
+h_1_right = 0.05;
+
+// Fracture 2
+h_2 = 0.05;
+
+// Intersection of 1 and 2
+h_1_2 = 0.05;
+
+// Intersection of fractures 1 and 3
+h_1_3 = 0.02;
+// Other points on fracture 3
+h_3 = 0.05;
+
+// Points on fracture 4, close to f1 and far away
+h_4_close = 0.02;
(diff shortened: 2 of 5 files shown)


keileg:
Thanks @jhabriel. We will have a look.

src/porepy/applications/md_grids/mdg_library.py
```diff
@@ -172,3 +173,53 @@ def benchmark_regular_2d(
...
+            Default is `0`.
+
+    Returns:
+        Mixed-dimensional grid.
+
+    """
+    # Get directory pointing to the `geo` file
+    abs_path = Path(__file__)
+    benchmark_path = abs_path.parent / Path("geo_library") / Path("benchmark_3d_case_3")
+    if refinement_level == 0:
```
IvarStefansson, line 207: Could we do something like
```
num_cells = [30, 140, 350, 500][refinement_level]
full_path = benchmark_path/Path(f"mesh{num_cells}k.geo")
```

src/porepy/applications/md_grids/mdg_library.py
```diff
@@ -172,3 +173,53 @@ def benchmark_regular_2d(
...
+
+    Reference:
+        Berre, I., Boon, W. M., Flemisch, B., Fumagalli, A., Gläser, D., Keilegavlen,
+        E., ... & Zulian, P. (2021). Verification benchmarks for single-phase flow in
+        three-dimensional fractured porous media. Advances in Water Resources, 147,
+        103759.
+
+    Parameters:
+        refinement_level: An integer denoting the level of refinement. Use `0` to
+            generate a mixed-dimensional grid with approximately 30K 2D cells,
```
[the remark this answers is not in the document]
jhabriel: Yes.

keileg:
I will add some testing of the new file.

keileg:
@IvarStefansson: I added some testing and renamed the folder as we discussed. Could you please have a look at the changes that I introduced?

I have no further comments to the PR.

src/porepy/applications/md_grids/gmsh_file_library/benchmark_3d_case_3/mesh140k.geo
IvarStefansson, line 1: Are the geo files the same, except from mesh sizes? If so, is there a way of reusing definition of lines, points etc. between the four refinement levels?

src/porepy/applications/md_grids/mdg_library.py
```diff
@@ -172,3 +173,51 @@ def benchmark_regular_2d(
...
+        103759.
+
+    Parameters:
+        refinement_level: An integer denoting the level of refinement. Use `0` to
+            generate a mixed-dimensional grid with approximately 30K 3D cells,
+            `1` for 140K 3D cells, `2` for 350K 3D cells, and `3` for 500K 3D cells.
+            Default is `0`.
+
+    Returns:
+        Mixed-dimensional grid for the given refinement level, and the fracture network.
```
IvarStefansson, line 201: Strictly speaking, this does not follow the doc style for multiple returns. I'm fine with not fixing now, but wanted to draw your attention to it.

IvarStefansson (commented):
Only one substantial comment

jhabriel:
@keileg: I think you forgot to add the `CSV` file to the subdirectory `benchmark_3d_case_3`. If you are copying and pasting from `porepy_mesh_factory`, please note that the first line (i.e., the domain specification) is not correctly defined. Instead of `0, 0, 0, 1, 1, 1` it has to be `0, 0, 0, 1, 2.5, 1`.

Another point to consider is that we might want to retrieve only the fracture network. For these cases, creating the `mdg` will be quite time-consuming. Would you consider adding an `only_network: bool = False` input parameter to cover such cases.

Finally, this might be a rare case, but: How do we know that there is a 1-to-1 match between the fracture network and the mdg? Both, in terms of coordinates and indexing.

src/porepy/applications/md_grids/gmsh_file_library/benchmark_3d_case_3/mesh140k.geo
keileg: I believe they are the same, although I have not checked in details, and it has been years since I made those files.

I am quite sure it is possible to include one `.geo` file into another, most likely it is even quite simple, but I would have to look up how to do that. Most likely we will never touch those files again, so may I suggest we revisit the question if we need to modify them somehow?

keileg:
In my understanding, pytest fails for the following reason: Before running GH actions, PorePy is installed as a standard package (not in editable mode as we usually do) and placed in a directory `.../python3.11/site-packages/porepy/`. The `.geo` file is not part of the pip installation, thus it is only available via the repository. In `mdg_library.benchmark_3d_case_3()`, the data file is specified as a relative path, which apparently is interpreted based on the location of the installed files (`site-packages` etc.). This does not work.

Note that this, AFAIK, is the only test that uses pre-generated data files, thus this is a problem we have not encountered previously. Hence, the solution we end up with may set precedence for other cases, though I doubt there will be many of them.

Options:
1. Make GH actions install in editable mode. This would cure the symptom, but would require any user who wants to generate this mesh to also install in editable mode, which is hardly ideal.
2. Somehow make the file available in the test, perhaps by copying files around on the GH server. This will have the same problem for users.
3. Remove this test. Same problem.
4. Check if the file is present, and if not, download it from GH and place it somewhere accessible. This is hardly beautiful, but it will at least work.
5. Decide that, after all, we do not want to support this geometry. The same will by extension apply to other geometries that cannot readily be specified by the standard PorePy approach, and for any other future functionality that may rely on access to data files (as part of the repository).

I don't like any of these options, but no 4 is the only one that will not crash or set strict constraints on future development. 

@jhabriel @IvarStefansson @OmarDuran: 
1. Does my reasoning make sense, or do you see other explanation for this  [error messsage](https://github.com/pmgbergen/porepy/actions/runs/7538888665/job/20520278738?pr=1096 )?
2. Do you have other suggestions

src/porepy/applications/md_grids/gmsh_file_library/benchmark_3d_case_3/mesh140k.geo
IvarStefansson: Sure

jhabriel:
I'm not so sure this has to do with porepy being installed in editable mode or not. I believe the problem is that Actions cannot find the `geo` files since it interprets that it has to be provided by the porepy installation. Since Actions installs the last develop version, it simply does not notice the newly added `geo` files.

A potential fix here could be to make the `geo` files available in a separate PR so that the `dev` branch is updated. Again, this might or might not work.

IvarStefansson:
> In my understanding, pytest fails for the following reason: Before running GH actions, PorePy is installed as a standard package (not in editable mode as we usually do) and placed in a directory `.../python3.11/site-packages/porepy/`. The `.geo` file is not part of the pip installation, thus it is only available via the repository. In `mdg_library.benchmark_3d_case_3()`, the data file is specified as a relative path, which apparently is interpreted based on the location of the installed files (`site-packages` etc.). This does not work.
> 
> Note that this, AFAIK, is the only test that uses pre-generated data files, thus this is a problem we have not encountered previously. Hence, the solution we end up with may set precedence for other cases, though I doubt there will be many of them.
> 
> Options:
> 
>     1. Make GH actions install in editable mode. This would cure the symptom, but would require any user who wants to generate this mesh to also install in editable mode, which is hardly ideal.
> 
>     2. Somehow make the file available in the test, perhaps by copying files around on the GH server. This will have the same problem for users.
> 
>     3. Remove this test. Same problem.
> 
>     4. Check if the file is present, and if not, download it from GH and place it somewhere accessible. This is hardly beautiful, but it will at least work.
> 
>     5. Decide that, after all, we do not want to support this geometry. The same will by extension apply to other geometries that cannot readily be specified by the standard PorePy approach, and for any other future functionality that may rely on access to data files (as part of the repository).
> 
> 
> I don't like any of these options, but no 4 is the only one that will not crash or set strict constraints on future development.
> 
> @jhabriel @IvarStefansson @OmarDuran:
> 
>     1. Does my reasoning make sense, or do you see other explanation for this  [error messsage](https://github.com/pmgbergen/porepy/actions/runs/7538888665/job/20520278738?pr=1096)?
> 
>     2. Do you have other suggestions

As you know, this is not exactly my home turf. But could it be that this can be fixed in setup.py? See this: https://stackoverflow.com/questions/11848030/how-include-static-files-to-setuptools-python-package

keileg:
> I'm not so sure this has to do with porepy being installed in editable mode or not. I believe the problem is that Actions cannot >find the geo files since it interprets that it has to be provided by the porepy installation. Since Actions installs the last develop version, it simply does not notice the newly added geo files.

If you look at the path where python expects the data file to be, it is at a location which
i) is set relative to the location of the file trying to load the data file
ii) is outside the repository.

Since the source file has been moved, presumably due to the installation process, but the data file has not, this is bound to fail. We should not try to give the correct relative path, that would be bad practice for so many reasons. The `.geo` files are already available (or else the updated test would not have been available), so splitting this into two PRs will not make the problem go away.

IvarStefansson:
It seems my suggestion might work. See [this test run](https://github.com/pmgbergen/porepy/actions/runs/7543247449/job/20533828168) on [this branch](https://github.com/pmgbergen/porepy/tree/include-static-files-in-install), which is based off of the branch of this PR. Note that I tried and failed at doing something a bit more targeted (specific folders only). My attempts are documented in the branch's git history. I am sure we can find a way to make that work given a bit more stubbornness and technical expertise.

keileg:
> It seems my suggestion might work. See [this test run](https://github.com/pmgbergen/porepy/actions/runs/7543247449/job/20533828168) on [this branch](https://github.com/pmgbergen/porepy/tree/include-static-files-in-install), which is based off of the branch of this PR. Note that I tried and failed at doing something a bit more targeted (specific folders only). My attempts are documented in the branch's git history. I am sure we can find a way to make that work given a bit more stubbornness and technical expertise.

Thanks, @IvarStefansson, this looks promising. I will go ahead with your suggestion, but as you suggest, also read up a bit more before finalizing.

keileg:
@IvarStefansson I took your suggestion, which seems like an accepted solution on the internet, but narrowed the scope down to only files living within the directory of gmsh_files. We can of course expand later, but for now, caution seems preferrable.

keileg (approved):
With the latest changes, everything should be fine now.

Issue #939, opened 2023-08-16
Inconsistent number of time steps


IngridKJ:
For certain parameter values in TimeManager, an inconsistent number of time steps is run. When the simulation reaches its final time step, it varies (due to floating points) whether the model.time_manager.time has reached set final time. So, the condition in run_models.py, line 141, is fulfilled though it shouldn't.

E.g., for 10 assigned time steps (see code snippet below), the final time is set to be 1.0. However, when the final time step has been run, the model.time_manager.time only holds a value of .9999999999999999. Therefore it runs an additional time step.

For e.g. 20 time steps, the expected number of time steps is run. So there is an inconsistency here.

Code snippet is in this file:
[inconsistent_time_steps.txt](https://github.com/pmgbergen/porepy/files/12357249/inconsistent_time_steps.txt)

IvarStefansson:
Should be a simple matter of using np.isclose. Beware the tolerance pitfalls.

jhabriel:
Indeed, this is simply a matter of introducing `np.isclose` for all time comparisons. We should consider passing a `tol` input at instantiation with a default value and make sure `np.isclose` is throughout the the class.

keileg:
Resolved in #1100

PR #1100, opened 2024-01-18
Fix of issue #0939

jwboth:
## Proposed changes

Fix for issue #0939. By introducing a new method for `TimeManager` to check when a final time is reached, the time manager now behaves as expected. 

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/time_step_control.py
+++ b/src/porepy/numerics/time_step_control.py
@@ -404,6 +404,15 @@ class TimeManager:
 
         return s
 
+    def final_time_reached(self) -> bool:
+        """Check whether the time manager has reached the end of the schedule.
+
+        Returns:
+            Whether the final time has reached or been overstepped.
+
+        """
+        return self.time > self.time_final or np.isclose(self.time, self.time_final)
+
     def compute_time_step(
         self, iterations: Optional[int] = None, recompute_solution: bool = False
     ) -> Union[float, None]:
@@ -431,7 +440,7 @@ class TimeManager:
         self._iters = iterations
 
         # First, check if we reach final simulation time
-        if self.time >= self.time_final:
+        if self.final_time_reached():
             return None
 
         # If the time step is constant, always return that value
--- a/src/porepy/numerics/time_step_control.py
+++ b/src/porepy/numerics/time_step_control.py
@@ -159,6 +159,8 @@ class TimeManager:
         allowable
             number of consecutive recomputation attempts.
         print_info. Whether to print on-the-fly time-stepping information or not.
+        rtol: Relative tolerance parameter for float point equality.
+        atol: Absolute tolerance parameter for float point equality.
 
     Example:
         # The following is an example on how to initialize a time-stepping object
@@ -206,6 +208,8 @@ class TimeManager:
         recomp_factor: float = 0.5,
         recomp_max: int = 10,
         print_info: bool = False,
+        rtol: float = 1e-10,
+        atol: float = 1e-16,
     ) -> None:
         schedule = np.array(schedule)
         # Sanity checks for schedule
@@ -223,6 +227,9 @@ class TimeManager:
             raise ValueError(
                 "Initial time step cannot be larger than final simulation time."
             )
+
+        self.rtol = rtol
+        self.atol = atol
 
         # If dt_min_max is not given, set dt_min=0.001*final_time and
         # dt_max=0.1*final_time
@@ -319,13 +326,11 @@ class TimeManager:
             # If the time step is constant, check that the scheduled times and the time
             # step are compatible. See documentation of ``schedule``.
             sim_times = np.arange(schedule[0], schedule[-1] + dt_init, dt_init)
-            compat_rtol = 1e-05
-            compat_atol = 1e-08
             is_compatible = self.is_schedule_in_simulated_times(
                 schedule,
                 sim_times,
-                rtol=compat_rtol,
-                atol=compat_atol,
+                rtol=self.rtol,
+                atol=self.atol,
             )
             if not is_compatible:
                 msg = (
@@ -411,7 +416,9 @@ class TimeManager:
             Whether the final time has reached or been overstepped.
 
         """
-        return self.time > self.time_final or np.isclose(self.time, self.time_final)
+        return self.time > self.time_final or np.isclose(
+            self.time, self.time_final, rtol=self.rtol, atol=self.atol
+        )
 
     def compute_time_step(
         self, iterations: Optional[int] = None, recompute_solution: bool = False
@@ -648,8 +655,8 @@ class TimeManager:
     def is_schedule_in_simulated_times(
         schedule: np.ndarray,
         sim_times: np.ndarray,
-        rtol: float = 1e-05,
-        atol: float = 1e-08,
+        rtol: float = 1e-10,
+        atol: float = 1e-16,
     ) -> bool:
         """Checks if ``schedule`` is a proper subset of ``sim_times`` for given
            tolerances
--- a/tests/numerics/test_time_step_control.py
+++ b/tests/numerics/test_time_step_control.py
@@ -16,6 +16,7 @@ from pathlib import Path
 
 import numpy as np
 import pytest
+
 import porepy as pp
 from porepy.models.fluid_mass_balance import SinglePhaseFlow
 
--- a/tests/numerics/test_time_step_control.py
+++ b/tests/numerics/test_time_step_control.py
@@ -356,8 +356,6 @@ class TestParameterInputs:
             (3, 1.0e-14),
             (1, 1.0),
             (3, 1.0),
-            (1, 1.0e14),
-            (3, 1.0e14),
         ],
     )
     def test_number_time_steps(self, num_time_steps, final_time):
--- a/tests/numerics/test_time_step_control.py
+++ b/tests/numerics/test_time_step_control.py
@@ -348,6 +348,29 @@ class TestParameterInputs:
             )
         assert msg in str(excinfo.value)
 
+    @pytest.mark.parametrize("num_time_steps", [10, 20])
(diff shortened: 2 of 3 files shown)


jwboth:
The above workflow shall first locate the issue (by adding a breaking test) and then a suggestion for a solution (added functionality for fixing the test). 

Two issues remain which I happily hear your feedack:
- Is the test well placed (it includes running a model)?
- Do you know what possibly makes the run time of the test so long (ca. 2 x 3 seconds)?

tests/numerics/test_time_step_control.py
```diff
@@ -348,6 +348,30 @@ def test_number_of_recomp_attempts_greater_than_zero(
             )
         assert msg in str(excinfo.value)
 
+    @pytest.mark.parametrize("num_time_steps", [10, 20])
```
keileg, line 351: @jwboth Regarding your question about the time it takes to run the test: Would it work to say, e.g. `[3, 5]` here? Or `[10, 11]`?

I'm not very worried about the cost of this test, but if we can reduce it further, that would be great.

tests/numerics/test_time_step_control.py
```diff
@@ -348,6 +348,30 @@ def test_number_of_recomp_attempts_greater_than_zero(
...
+            dt_init=1.0 / num_time_steps,
+            constant_dt=True,
+        )
+
+        params = {
+            "time_manager": time_manager,
+            "suppress_export": True,
+        }
+
+        from porepy.models.momentum_balance import MomentumBalance
```
keileg, line 367: Any reason to use momentum balance here, and not a simpler mass_balance?

src/porepy/numerics/time_step_control.py
```diff
@@ -404,6 +404,15 @@ def __repr__(self) -> str:
...
         return s
 
+    def final_time_reached(self) -> bool:
+        """Check whether the time manager has reached the end of the schedule.
+
+        Returns:
+            Whether the final time has reached or been overstepped.
+
+        """
+        return self.time > self.time_final or np.isclose(self.time, self.time_final)
```
keileg, line 414: This will become a problem if the time step is tiny compared to the final time. Will it work if we insist on using a relative tolerance (so, set `np.isclose(..., atol=self.time_final * 1e-30)` or something), and then scale the relative tolerance with the time step, `np.isclose(..., rtol=self.dt * 1e-8)`. This will also fail if the time step must be severely cut just before the end of simulation (thus `dt` is dramatically decreased), but it should be more robust.

Let me know what you think about this, but don't accept my logic by default, I don't trust it myself.

keileg (commented):
I have a few questions, see details.

tests/numerics/test_time_step_control.py
OmarDuran: Now the test is parametrized with:
-  number of time steps [1, 3]
-  final time [1.0e-14,1.0e0,1.0e+14]

tests/numerics/test_time_step_control.py
OmarDuran: SinglePhaseFlow was included instead of momentum balance.

src/porepy/numerics/time_step_control.py
OmarDuran: The scale you mention is already performed in the implementation of [np.isclose](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html). Tolerances for float point equality were exposed to the user with default:

- rtol: float = 1e-10
- atol: float = 1e-16

The parametrization of the test was updated to cover time steps and final times values of the order or (1e-14,1e0,1e+14). See comment above.

tests/numerics/test_time_step_control.py
jwboth: Not sure if this is relevant: @OmarDuran, did you test whether one of the options would have still broken the test before? Not every value for `num_time_steps` has led to the unexpected behavior of `TimeManager`. 

I like the idea of testing small final times to address the tolerance question.

src/porepy/numerics/time_step_control.py
jwboth: If `atol=self.time_final * 1e-30` should make sense, why not setting to zero and entirely deactivating the absolute criterion?

I think I like @keileg 's suggestion more regarding relative tolerances. Just using a fixed value for `rtol` without relaing to the time step size will not necessarily fix the issue when the final time is large while the time step size is small (although in a practical situation one most likely would be absolutely fine if the final simulation time is not 10 years, but 10 years minus 0.03 seconds).

tests/numerics/test_time_step_control.py
OmarDuran: @jwboth with the implementation you did in `def final_time_reached(self)` the code works correctly for any `num_time_steps`. However the question that remains was to test values very close to the final time.

Let set `a = self.time` and `b = self.time_final`. To mimic the situation `a` very close to `b`, it is enough to test very small values of `a` and `b` in the sense of `absolute(a - b) <= (atol + rtol * absolute(b))`. This cover all the cases for regular time steps discretizations. I believe possible situations associated with irregular time steps belong to another issue and a future pull request.

src/porepy/numerics/time_step_control.py
OmarDuran: See comment above.

keileg: As @jwboth writes, any fixed `rtol` will be insufficient if the problem is sufficiently difficult, or the user has prescribed sufficiently small time steps. Whether it makes any practical difference if the simulation stops `epsilon` time before what was intended is a different matter. 

Come to think of it, precisely the same problem will arise when [adapting the time step to the schedule](https://github.com/pmgbergen/porepy/blob/83581f3a5fdd0bf3358a0c3c2d3ad306edd5b245/src/porepy/numerics/time_step_control.py#L663). Should we fix this at the same time?

tests/numerics/test_time_step_control.py
keileg: @OmarDuran: What do you mean by regular time step? To me, this is a highly problematic concept, for the following reason:

In addition to running simulation on the unit interval, I regularly consider cases where the simulation first runs for a couple of hundred years to reach a steady state, followed by a period where injection rates is changed on a schedule of, say, minutes. On the other hand, if we were to do simulations inspired by the FluidFlower CO2 experiments, the time step size needed in OPM are commonly down to fractions of milliseconds, and PorePy would likely need even smaller ones. Unless we make the code robust in such wide ranges, I don't think we can consider the issue closed.

tests/numerics/test_time_step_control.py
```diff
@@ -348,6 +349,36 @@ def test_number_of_recomp_attempts_greater_than_zero(
             )
         assert msg in str(excinfo.value)
 
+    @pytest.mark.parametrize(
+        "num_time_steps, final_time",
+        [
+            (1, 1.0e-14),
+            (3, 1.0e-14),
+            (1, 1.0),
+            (3, 1.0),
```
keileg, line 358: As I understand it, the problem we are trying to deal with is related to rounding errors. Perhaps contradicting my previous remark regarding these parameters, I think the number of time steps should be set to maximize the chance for such errors. In that regard, 1 is a bad number, 3 is good, and so would, say, 7 be. In the same spirit, would it make sense to change the final time, say, to `pi`?

keileg (commented):
Also see my response to comments.

I think we are converging on a solution. However, as noted in one of my comment, I think we should deal with the method that adapts the time step to the schedule at the same time (the same reasoning should be valid)

tests/numerics/test_time_step_control.py
OmarDuran: @keileg: I believe this  PR do fix the issue #939. In this case, it is not about running simulations on a unit interval nor dealing with different time step sizes in the same simulation, but rather to ensure that the number of time steps performed is the one requested. This works perfectly for regular time step sizes (constant time step size) and large simulation times. You have elaborated on an issue of different nature when it comes to irregular time step sizes (no constant time steps). Considering the case you describe above, I believe we should create another issue specific to this case.

jwboth: Hereby, after a quick local test, I can assure that the values chosen by @OmarDuran would not break the initial test given the old implementation of checking the final state of the simulation. Also values [3,5] for num_time_steps pass, even for final time math.pi.  Finally, the combination of 7 time steps and final time 1e-14 breaks the test. My suggestions would be to aim at four tests with 2 breaking the previous implementation - one per final time. Thus, althoug I see the critcism around final_time=1, I would suggest the following selection:
   `@pytest.mark.parametrize(
        "num_time_steps, final_time",
        [
            (3, 1.0e-14),
            (7, 1.0e-14),
            (3, pi),
            (10, 1.),
        ],
    )`
Here, the original implementation would break for the second and fourth test. Repalcing pi with 1 in the case (3, pi) would be sensible to be as well - no strong opinion here.

tests/numerics/test_time_step_control.py
jwboth: See comment above.

src/porepy/numerics/time_step_control.py
keileg: Update: Adaptation to time step will not be a problem, the implementation is slightly different from what I recalled.

tests/numerics/test_time_step_control.py
keileg: Let's do that.

tests/numerics/test_time_step_control.py
keileg: @jwboth: Agree. Ping me when you have implemented, then I will approve.

tests/numerics/test_time_step_control.py
jwboth: @keileg Done. I flipped the coin and chose not to include pi via an extra import.

keileg (approved):

PR #1102, opened 2024-01-19
Setup for case 3, flow benchmark 3d

jhabriel:
## Proposed changes

This PR introduces the setup for case 3, flow benchmark 3d, as a porepy example. Tests for the sanity of benchmark parameters specification (e.g., effective permeabilities and boundary conditions) are included. 

Idea for a proper test: Ideally, we should try to sample the reference value of the plot over line from the benchmark and compare them with the approximate pressure solutions. Might be a nice little side project for the user group.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/examples/flow_benchmark_3d_case_3.py
+++ b/examples/flow_benchmark_3d_case_3.py
@@ -2,7 +2,7 @@
 This module contains the implementation of Case 3 from the 3D flow benchmark [1].
 
 Note:
-    The class `Benchmark3dCase3Model` admits the keyword `refinement_level`,
+    The class `FlowBenchmark3dCase3Model` admits the keyword `refinement_level`,
     which can take values 0, 1, 2, 3, to control the mesh refinement level. Level `0`
     contains approximately 30K three-dimensional cells, level `1` contains 140K
     three-dimensional cells, level `2` contains 350K three-dimensional cells,
@@ -122,16 +122,4 @@ class FlowBenchmark3dCase3Model(  # type:ignore[misc]
     FlowBenchmark3dCase3BoundaryConditions,
     pp.fluid_mass_balance.SinglePhaseFlow,
 ):
-    """Mixer class for case 3 from 3D flow benchmark."""
-
-
-# %% Runner
-model = FlowBenchmark3dCase3Model({"material_constants": {"solid": solid_constants}})
-pp.run_time_dependent_model(model, {})
-pressure_norm = np.linalg.norm(
-    model.pressure(model.mdg.subdomains()).value(model.equation_system)
-)
-
-# %%
-print("Pressure norm", pressure_norm)
-np.testing.assert_allclose(pressure_norm, 13.337947636587648)
+    """Mixer class for case 3 from the 3d flow benchmark."""
--- a/examples/flow_benchmark_3d_case_3.py
+++ b/examples/flow_benchmark_3d_case_3.py
@@ -1,0 +1,137 @@
+"""
+This module contains the implementation of Case 3 from the 3D flow benchmark [1].
+
+Note:
+    The class `Benchmark3dCase3Model` admits the keyword `refinement_level`,
+    which can take values 0, 1, 2, 3, to control the mesh refinement level. Level `0`
+    contains approximately 30K three-dimensional cells, level `1` contains 140K
+    three-dimensional cells, level `2` contains 350K three-dimensional cells,
+    and level `3` contains 500K three-dimensional cells.
+
+References:
+    [1] Berre, I., Boon, W. M., Flemisch, B., Fumagalli, A., Gläser, D., Keilegavlen,
+        E., ... & Zulian, P. (2021). Verification benchmarks for single-phase flow in
+        three-dimensional fractured porous media. Advances in Water Resources, 147,
+        103759.
+
+"""
+from typing import Callable
+
+import numpy as np
+
+import porepy as pp
+from porepy.applications.md_grids.mdg_library import benchmark_3d_case_3
+from porepy.fracs.fracture_network_3d import FractureNetwork3d
+from porepy.models.constitutive_laws import DimensionDependentPermeability
+
+solid_constants = pp.SolidConstants(
+    {
+        "residual_aperture": 1e-2,
+        "normal_permeability": 1e4,
+    }
+)
+
+
+class FlowBenchmark3dCase3Geometry(pp.ModelGeometry):
+    """Define Geometry as"""
+
+    params: dict
+    """User-defined model parameters."""
+
+    def set_geometry(self) -> None:
+        """Read geometry from the `geo` files associated to the benchmark."""
+
+        # Create mixed-dimensional grid and fracture network
+        self.mdg, self.fracture_network = benchmark_3d_case_3(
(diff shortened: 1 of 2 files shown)


tests/functional/test_benchmark_3d_case_3.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    Test if the permeability values are consistent with the benchmark specification.
+
+    Parameters:
+        model: ModelWithEffectivePermeability
+            Solved model. Returned by the `model()` fixture.
+
+    """
+    for sd, d in model.mdg.subdomains(return_data=True):
+        val = np.mean(
```
IvarStefansson, line 94: Why do you use mean?

tests/functional/test_benchmark_3d_case_3.py
```diff
@@ -0,0 +1,145 @@
...
+            d[pp.PARAMETERS][model.darcy_keyword]["second_order_tensor"].values[0][0]
+        )
+        if sd.dim == 3:
+            np.testing.assert_allclose(val, 1)
```
IvarStefansson, line 98: Please either explain the origin of these hard-coded values or (even better) define a method effective_tangential_permeability which computes them (using Eq. 6a, right?) from the values provided in table 5.

tests/functional/test_benchmark_3d_case_3.py
```diff
@@ -0,0 +1,145 @@
...
+        the domain. At the outlet boundary, we check that the pressure value is zero.
+
+    """
+    bg, data_bg = model.mdg.boundaries(return_data=True, dim=2)[0]
+
+    # Inlet boundary
+    south_side = model.domain_boundary_sides(bg).south
+    inlet_flux = np.sum(data_bg["iterate_solutions"]["darcy_flux"][0][south_side])
+    np.testing.assert_allclose(inlet_flux, desired=-1/3, atol=1e-5)
+
```
IvarStefansson, line 141: Could you please comment (on GH) on the use of .testing? And do we really need allclose? Shouldn't isclose suffice?

tests/functional/test_benchmark_3d_case_3.py
IvarStefansson, line 1: Allthough I agree with you that a stricter test would be even better, I like the implementation a lot.

IvarStefansson (changes requested):
Thanks! Nice work. I was pleasantly surprised how few LOC this ammounts to. A few relatively minor comments.

IvarStefansson (commented):
Thanks! Nice work. I was pleasantly surprised how few LOC this ammounts to. A few relatively minor comments.

IvarStefansson (commented):
Thanks! Nice work. I was pleasantly surprised how few LOC this ammounts to. A few relatively minor comments.

examples/flow_benchmark_3d_case_3.py
```diff
@@ -0,0 +1,125 @@
...
+
+import porepy as pp
+from porepy.applications.md_grids.mdg_library import benchmark_3d_case_3
+from porepy.fracs.fracture_network_3d import FractureNetwork3d
+from porepy.models.constitutive_laws import DimensionDependentPermeability
+
+solid_constants = pp.SolidConstants(
+    {
+        "residual_aperture": 1e-2,
+        "normal_permeability": 1e4,
```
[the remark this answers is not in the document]
jhabriel: This is true. @keileg: You might benefit from checking these calculations:

https://www.dropbox.com/scl/fi/m7m7jxw9q3d2yboxb6z8y/permeability_calculations.pdf?rlkey=riq9vont0u3cznye0cgjsfqoi&dl=0

tests/functional/test_benchmark_3d_case_3.py
jhabriel: Good point, `np.mean` is actually not needed here.

tests/functional/test_benchmark_3d_case_3.py
jhabriel: I did not do that because the second order tensor stored in the data dictionary is actually the effective tangential permeability (the specific volume multiplies the tangential permeability [here](https://github.com/pmgbergen/porepy/blob/e4d484f915cde39cc17639e24c1ffe5ab37c1f99/src/porepy/models/fluid_mass_balance.py#L837)).

But this is not a problem, I will add a method `effective_tangential_permeability` to the `ModelWithEffectivePermeability` class with a more in-depth explanation.

tests/functional/test_benchmark_3d_case_3.py
jhabriel: Yes, this is a scalar, `np.isclose()` suffices.

jhabriel:
Thanks Ivar. I will work on incorporating the suggestions now.

tests/functional/test_benchmark_3d_case_3.py
jhabriel: Thanks. I wish we could have some tests on the produced results, but without comparing them to the reference solution, I don't really see how we can achieve this. I'm open to suggestions though.

tests/functional/test_benchmark_3d_case_3.py
jhabriel: I have added the method `effective_tangential_permeability` as explained above. Please, let me know if this is more or less what you envisioned.

jhabriel:
@IvarStefansson: I believe I've incorporated all your suggestions now.

tests/functional/test_benchmark_3d_case_3.py
IvarStefansson: I'm content as is. @keileg, please make a decision.

tests/functional/test_benchmark_3d_case_3.py
```diff
@@ -81,59 +117,61 @@ def model() -> ModelWithEffectivePermeability:
 def test_effective_tangential_permeability_values(model) -> None:
     """Test if the permeability values are consistent with the benchmark specification.
 
+    The values are specified in Table 5 from [1]. We expect a value of normal
+    permeability = 2E6 for 2d interfaces, and 2e4 for 1d interfaces.
```
IvarStefansson, line 121: This comment seems to be misplaced.

IvarStefansson (commented):
Looks good!

tests/functional/test_benchmark_3d_case_3.py
jhabriel: Fixed.

IvarStefansson (commented):
I'm happy. Over to you for final review, @keileg.

examples/flow_benchmark_3d_case_3.py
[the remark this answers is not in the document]
keileg: I agree, the values make sense.

tests/functional/test_benchmark_3d_case_3.py
keileg: I think this is a fair compromise.

keileg (commented):
Thanks a lot for the contribution, @jhabriel. The files are fine with me, but let us come back to the placement (as you know, where to put such functionality has been a long standing issue). Just to be sure, I don't formally approve the PR yet, but there should be no need for any action from your side.

IvarStefansson (approved):

PR #1111, opened 2024-01-24
Bug in Differentiable Tpfa related to fracture tips and internal boundaries

### LINKED ISSUE #1104: Bug in AdTpfaFlux
There seems to be a bug in the AdTpfaFlux mixin class related to Neumann conditions on internal domains.

A likely partial cause is that, in `AdTpfaFlux.diffusive_flux`, filtering on Neumann conditions is only done on external boundaries (contrast this with `AdTpfaFlux.pressure_trace`, where internal boundaries is also filtered). It could also be that both methods need filtering on fracture tips, it seems likely this is currently not done.

To get this right, and properly tested, we need to solve a flow problem with both intersections and fracture tips.

@IvarStefansson: Feel free to add to the problem description.

### LINKED ISSUE #1107: Name mistake in test function
https://github.com/pmgbergen/porepy/blob/e4d484f915cde39cc17639e24c1ffe5ab37c1f99/tests/models/test_thermoporomechanics.py#L341

Function should be called test_THERMOporomechanics

### LINKED ISSUE #1108: Name mistake in test function
https://github.com/pmgbergen/porepy/blob/e4d484f915cde39cc17639e24c1ffe5ab37c1f99/tests/models/test_thermoporomechanics.py#L341

Function should be called something like test_thermoporomechanics_well

keileg:
## Proposed changes
The previous implementation of differentiable Tpfa contained a bug which gave non-zero transmissibilities on fracture tips and internal boundaries (where the transmissibilities should be zero, but were not). This PR introduces a fix which by filtering sets the transmissibilites to zero at the relevant faces.

A similar fix was also introduced for the potential_trace function, but this is only relevant for the fracture tips.

A new test that covers the fixed code has been added.

Resolves #1104
Resolves #1107 
Resolves #1108

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/constitutive_laws.py
+++ b/src/porepy/models/constitutive_laws.py
@@ -1700,9 +1700,6 @@ class AdTpfaFlux:
         )
         # Also a separate filter for internal boundaries, which are always Neumann.
         internal_boundary_filter = diff_discr.internal_boundary_filter(subdomains)
-        # And a filter for tip faces, which are also Neumann, but where the flux is
-        # known to be zero (by assumptions embedded in the modeling framework).
-        tip_filter = diff_discr.tip_filter(subdomains)
 
         # Face contribution to boundary potential is 1 on Dirichlet faces, -1/t_f_full
         # on Neumann faces (both external and internal - see Tpfa.discretize). Named
@@ -1711,8 +1708,6 @@ class AdTpfaFlux:
         bound_pressure_face_discr = dir_filter - (
             neu_filter + internal_boundary_filter
         ) * (one / t_f_full)
-        # On tip faces, the face contribution to the boundary potential is zero.
-        bound_pressure_face_discr = bound_pressure_face_discr * (one - tip_filter)
 
         # Project the interface flux to the primary grid, preparing for discretization
         # on internal boundaries.
--- a/src/porepy/models/constitutive_laws.py
+++ b/src/porepy/models/constitutive_laws.py
@@ -1509,22 +1509,33 @@ class AdTpfaFlux:
 
         # Treatment of boundary conditions.
         one = pp.ad.Scalar(1)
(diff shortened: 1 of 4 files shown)


src/porepy/models/constitutive_laws.py
```diff
@@ -1697,6 +1711,8 @@ def potential_trace(
         bound_pressure_face_discr = dir_filter - (
             neu_filter + internal_boundary_filter
         ) * (one / t_f_full)
+        # On tip faces, the face contribution to the boundary potential is zero.
+        bound_pressure_face_discr = bound_pressure_face_discr * (one - tip_filter)
```
IvarStefansson, line 1715: Is this neccessary? If the filters are disjoint, I think this line should not impact the result

tests/models/test_thermoporomechanics.py
```diff
@@ -338,8 +338,8 @@ def meshing_arguments(self) -> dict:
         return mesh_sizes
 
 
-def test_poromechanics():
-    """Test that the poromechanics model runs without errors."""
+def test_thermoporomechanics_well():
```
IvarStefansson, line 341: Thanks!

IvarStefansson (commented):
Thanks! One bugfix (pushed) and one substantial question.

src/porepy/models/constitutive_laws.py
keileg: The point is that bound_pressure_face_discr should be zero on tips, which this line enforces. Note that this is different from the other kinds of boundaries, where there is a contribution.

I can move 1715 a few lines up to multiply t_f_full instead if you think that is more transparent, but the filter must be applied somewhere.

IvarStefansson: I'm missing something. Let's just talk in person.

IvarStefansson (approved):

PR #1112, opened 2024-01-24
Setups for cases 1 and 3, flow benchmark 2d

jhabriel:
## Proposed changes

This PR introduces the setup for case 3 (both top-to-bottom and left-to-right flows) from the two-dimensional flow benchmark. The benchmark is heavily based on the tutorial. Tests for the sanity of the parameters (permeability and boundary condition specifications) are included.

A new mixin class `FluxDiscretization` was added to the `applications` package. This class introduces the possibility to define either `tpfa` or `mpfa` as the flux discretisation scheme via the model parameter dictionary.

## Some important information
1. Currently, the tests are not passing, since I removed the class `_SetFluxDiscretizations` from the tests (see, e.g., commit 876f9928181cc5cde6de78df909185b419844100)
2. I'm currently marking the tests as slow, although they really don't take too long. Please, feel free to remove the pytest mark if you decide to run the test as part of the standard workflow.
3. The benchmark tutorial, at some point, must be updated. Possibly, this could wait for the Geiger 2d benchmark to be part of the examples (if that's going to be the case).

## Types of changes

- [ ] Other:

## Checklist


src/porepy/applications/test_utils/benchmarks.py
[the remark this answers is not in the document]
jhabriel: I was trying to make it a bit more general in case we need to host other classes/methods for the benchmarks, but I have no problem renaming the module to `effective_permeability.py`.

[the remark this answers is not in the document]
IvarStefansson: I was thinking "general" in a different way: Perhaps computing effective permeability may be relevant in other tests. But come to think of it, you're probably right, since the definition of the effective permeability is somewhat specific to the benchmark papers (i.e., the definition might be different in other contexts). Let's keep the module name for now.

IvarStefansson:
@jhabriel, we're waiting for #1115 before merging this. Otherwise, I think I'm mostly done. If you should feel like it, please have a look and let me know what you think.

jhabriel:
I'll have a quick look now @IvarStefansson

examples/flow_benchmark_2d_case_1.py
```diff
@@ -0,0 +1,114 @@
...
+      media. Advances in Water Resources, 111, 239-258.
+
+"""
+import numpy as np
+
+import porepy as pp
+from porepy.applications.discretizations.flux_discretization import FluxDiscretization
+from porepy.models.constitutive_laws import DimensionDependentPermeability
+
+solid_constants = pp.SolidConstants({"residual_aperture": 1e-4})
```
jhabriel, line 22: I believe we're missing the `normal_permeability` declaration in the solid constants.

examples/flow_benchmark_2d_case_1.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class FlowBenchmark2dCase1Model(  # type:ignore[misc]
+    FluxDiscretization,
+    Geometry,
+    Permeability,
+    BoundaryConditions,
+    pp.fluid_mass_balance.SinglePhaseFlow,
+):
+    """Complete model class for case 1 from the 2d flow benchmark."""
```
jhabriel: I would also add the following in the model docstring to guide the user on how to set up the two variants of the benchmark:

The benchmark specifies two variants; one with conductive fractures and one with blocking fractures. To set up an instance of the first variant you have to pass the keyword `'fracture_permeability'` with a value `1e4` as a model parameter. To set up an instance of the second variant, set the value to `1e-4` instead.

examples/flow_benchmark_2d_case_1.py
```diff
@@ -0,0 +1,114 @@
...
+
+class BoundaryConditions:
+    """Boundary conditions for Case 1 of the 2D flow benchmark.
+
+    Inflow on west (left) and prescribed pressure on east (right).
+
+    """
+
+    def bc_values_pressure(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
+        """Pressure value of 1 Pa on east side."""
```
jhabriel: West side?

examples/flow_benchmark_2d_case_1.py
```diff
@@ -0,0 +1,114 @@
...
+    def bc_type_darcy_flux(self, sd: pp.Grid) -> pp.BoundaryCondition:
+        """Assign Dirichlet to the east boundary. The rest are Neumann by default."""
+        bounds = self.domain_boundary_sides(sd)
+        bc = pp.BoundaryCondition(sd, bounds.east, "dir")
+        return bc
+
+    def bc_values_darcy_flux(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
+        """Inflow on the west boundary.
+
+        Per PorePy convention, the sign is negative for inflow and the value is
```
jhabriel: Excellent clarification b.t.w.

examples/flow_benchmark_3d_case_3.py
```diff
@@ -98,7 +99,7 @@ def intersection_permeability(self, subdomains: list[pp.Grid]) -> pp.ad.Operator
         return self.isotropic_second_order_tensor(subdomains, permeability)
 
 
-class FlowBenchmark3dCase3BoundaryConditions:
+class BoundaryConditions:
```
jhabriel: Now that I saw you included the scaling of the influx in the Geiger 2d, I remember I haven't done that for this example. Maybe we want to do it for consistency? The change should only affect L133 of this module.

jhabriel:
I'm done @IvarStefansson. I only left a few minor comments.

examples/flow_benchmark_2d_case_1.py
IvarStefansson: You're right. And we have the related question of how to define tangential fracture permeability. This prompted me to think a bit on the risk involved in the most tempting solution, which would be passing them in self.params. To avoid the user misspelling the parameter name, forgetting to specify it or otherwise get confused, I introduced tailored SolidConstants. Please let me know if you approve of the approach.

jhabriel: I support 100% this approach. Especially, when it comes to benchmarks, where the number 1 priority is to reproduce the setup.

Great job!

IvarStefansson (commented):
Comments much appreciated, @jhabriel!

examples/flow_benchmark_2d_case_1.py
```diff
@@ -0,0 +1,114 @@
+"""
+This module contains the implementation of Case 1 from the 2D flow benchmark [1].
+
+Note:
+    We provide the two variants of this benchmark, i.e., Case 1a (conductiv fractures)
```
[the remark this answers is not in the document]
jhabriel: Well, there weren't any variants of the model before. But now I think this is correct.

examples/flow_benchmark_2d_case_1.py
jhabriel: Yes, the proposed solution is way more elegant. I would not include this comment now.

examples/flow_benchmark_2d_case_1.py
jhabriel: Oh! Luckily you spotted that one.

examples/flow_benchmark_3d_case_3.py
jhabriel: I was thinking more about consistency in physical scales (`m` and `s`), in case someone would like to change the units that is.

examples/flow_benchmark_2d_case_1.py
jhabriel: Do you think unit tests are prudent?

examples/flow_benchmark_2d_case_1.py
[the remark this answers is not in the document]
IvarStefansson: No, you're right. I have fixed the note.

tests/functional/test_benchmark_2d_case_3.py
```diff
@@ -0,0 +1,189 @@
+"""
+Module containing tests for cases 3a and 3b from the 2d benchmark study [1].
+
+The tests check whether effective permeabilities and boundary conditions are correctly
+assigned. Since we solved the actual model, we also implicitly check that the model does
+not crash. However, we are currently not checking that the solution is the "correct"
+one.
```
keileg, line 7: Is it correct to say that these tests (and those for 3d case 3) verifies that the benchmark setup and the tests are consistently set up, but not that the values are actually the right ones (for that, the flow problem must be solved, errors computed etc. etc.)? If so, these tests will fail:
* If someone changes the setup, but not the test;
* If we change the way a single phase fluid model is set up, for instance by changing method names for specifying boundary conditions, but do not change the test;
* if we introduce an error in the permeability scaling with specific volume - this is the only non-trivial quantitative test that I can see here.
I also realize that running the tutorials check will verify that the 2d models also run, though not the 3d case.

I see four alternatives (or two pairs of alternatives) with regards to testing:
1. Leave the tests as they are, but introduce some verification that the 3d case actually runs. If the 3d case is slow, this can be a skipped test that only runs once a week.
2. Same as 1., but also include some weak check that detects changes in the solution (in the spirit of functional tests), say, in the form of total flux through the in- or outlet boundaries.
3. Introduce tests similar to those in this file, also for 2d case 1 and 3d case 3.
4. Drop the tests of the models, except verification that they run, with the motivation that the tests add more entropy and possibly maintenance than they enhance security.
Of course, both 3. and 4. can in a sense be combined with 1. or 2.

A verification that 3d case 3 does not crash should be made before this PR is merged. My preference would be to implement 2., but I think we can postpone that somewhat, and possibly reassign the task. With regards to the present file, I lean towards 4., but I am open to other opinions - let me know.

keileg (commented):
One major consideration regarding testing - let me know if you want to discuss.

Comments on the notebooks will be posted soon.

IvarStefansson:
<p>Thanks!</p>

---

 <a href='https://app.reviewnb.com/pmgbergen/porepy/pull/1112/discussion/#comment-1923515737'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tutorials/mandels_problem.ipynb
```diff
@@ -39,22 +39,10 @@
    "metadata": {
```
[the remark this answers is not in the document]
IvarStefansson: <p>Sorry, my bad. Sloppy commit. Will fix.</p>

tests/functional/test_benchmark_2d_case_3.py
IvarStefansson: Alternative 2 could be accommodated in the tutorial (with an assert on the solution, say).

jhabriel: If I may give my two cents on this...

I believe a decent compromise between maintenance and having some level of confidence in the benchmarks is testing the boundary conditions and permeability specifications. These are, essentially, the two aspects that are crucial to get it right when it comes to the flow benchmarks (in addition to the geometry that is). At the end of the day, the incompressible flow model I'd say is quite well tested via funcional and integration tests.

The next level of rigour, which is checking the solution, is indeed non-trivial, since this will probably be quite sensitive to the mesh that `gmsh` produces (this is why we suggest to compare order of convergence in the rigorous paper for system tests). Self-convergence could be an option, but it will require some effort in terms of implementation.

keileg: These are valid points, @jhabriel. Let's leave the tests as they are.

keileg (approved):

keileg:
Feel free to merge when the final issues are resolved.

Issue #740, opened 2022-10-13
Add possibility to export scheduled times


jhabriel:
Once the `TimeManager` is fully integrated into the models, we should provide the following exporting options:

1. no time
2. every time
3. scheduled times

This could be implemented, for example, inside `after_netwon_convergence()` as a method, say:
`export_solution(times_to_export: Literal[none, all, scheduled] = none)`,
where `none` refers to no time, `all` refers to all times, and `scheduled` to scheduled times.

The default behaviour is to export no time.

IvarStefansson:
I think the definition of when to export should be in the parameter dictionary rather than the signature of the method

jhabriel:
Yes, that would be better.

Something like:

times_to_export = self.params.get("times_to_export", "none")
export_times(times_to_export)

PR #1114, opened 2024-01-25
ENH: Allow to export only specified times

IngridKJ:
## Proposed changes

Addresses issue #740.  

To be fixed which is not addressed yet:
* Export initial condition or not
* Where to put the export_times method. Currently in the bottom of solution_strategy.py

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/solution_strategy.py
+++ b/src/porepy/models/solution_strategy.py
@@ -11,7 +11,7 @@ import logging
 import time
 import warnings
 from pathlib import Path
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Optional, Union
 
 import numpy as np
 import scipy.sparse as sps
@@ -449,7 +449,9 @@ class SolutionStrategy(abc.ABC):
             values=solution, time_step_index=0, additive=False
         )
         self.convergence_status = True
-        self.save_data_time_step()
+
+        times_to_export = self.params.get("times_to_export", "none")
+        self.export_solution(times_to_export=times_to_export)
 
     def after_nonlinear_failure(
         self, solution: np.ndarray, errors: float, iteration_counter: int
@@ -635,3 +637,22 @@ class SolutionStrategy(abc.ABC):
 
         """
         self.update_all_boundary_conditions()
+
+    def export_solution(self, times_to_export: Union[str, np.ndarray]) -> None:
+        """Method for exporting only specified solutions.
+
+        Parameters:
+            times_to_export: Contains information about which, if any, solution time
+                steps should be exported. It may hold the value "all", "none" or an
+                array of specified time step indices.
+
+        """
+        if type(times_to_export) is np.ndarray:
+            if self.time_manager.time_index in times_to_export:
+                self.save_data_time_step()
+        elif times_to_export == "all":
+            self.save_data_time_step()
+        elif times_to_export == "none":
+            pass
+        else:
+            raise ValueError("Invalid specification of times_to_export.")


IvarStefansson:
As the devil's advocate: What's the explicit motivation for this? As in, where exactly is it needed?

keileg:
> As the devil's advocate: What's the explicit motivation for this? As in, where exactly is it needed?

It makes sense to me to facilitate this. For instance, for highly non-linear problems (advection-dominated multiphase transport comes to mind), I would expect that the time step can be severely limited by the convergence radius of the Newton solver. In this case, it will be useful to easily skip exporting all time steps.

jwboth:
> > As the devil's advocate: What's the explicit motivation for this? As in, where exactly is it needed?
> 
> It makes sense to me to facilitate this. For instance, for highly non-linear problems (advection-dominated multiphase transport comes to mind), I would expect that the time step can be severely limited by the convergence radius of the Newton solver. In this case, it will be useful to easily skip exporting all time steps.

I can only confim that I wished to have a functionality like this for recent simulations which required small time steps and large grids even for simple transport simulations. In this case, both sparse access to specific time steps is of only interest in terms of inspection of the data and this way the memory output can be limited. I would thus also appreciate core fuctionality instead of hacking an own solution.

IvarStefansson:
Just to be clear: It's not that I can't see cases where this is useful. What is less obvious is whether this should be part of porepy or left to the user. A reason why the second option might be sensible is that the exact needs may depend on the application and be difficult to predict.

jwboth:
Of course, a very valid point! I have not a too good overview of `TimeManager`. But at least connecting exporting to the schedule without leaving this to the user would sound to me like a very standard use case. I will remain silent from now and just wanted to express my personal interest.

IvarStefansson:
> Of course, a very valid point! I have not a too good overview of `TimeManager`. But at least connecting exporting to the schedule without leaving this to the user would sound to me like a very standard use case. I will remain silent from now and just wanted to express my personal interest.

Yes, thanks a lot for the feedback. That was exactly why I posed the question!

keileg:
> Of course, a very valid point! I have not a too good overview of `TimeManager`. But at least connecting exporting to the schedule without leaving this to the user would sound to me like a very standard use case. I will remain silent from now and just wanted to express my personal interest.

This sounds like a fair compromise also to me.

IngridKJ:
If I understand the schedule correctly, it prescribes the exact time values to be run. If the exporting of files then is to be related to the schedule, won't the exporting happen for every time step that is run? Or have I perhaps misunderstood you guys' comments and/or the schedule functionality

IvarStefansson:
You have misunderstood the schedule. The scheduled times are a (possibly small) subset of all time steps.

IngridKJ:
That makes total sense, thank you for the clarification @IvarStefansson. I will take these comments into account when I work further on the implementation.

src/porepy/models/solution_strategy.py
```diff
@@ -449,7 +449,9 @@ def after_nonlinear_convergence(
             values=solution, time_step_index=0, additive=False
         )
         self.convergence_status = True
-        self.save_data_time_step()
+
+        times_to_export = self.params.get("times_to_export", None)
+        self.export_solution(times_to_export=times_to_export)  # type: ignore
```
IvarStefansson, line 454: I think we should keep using save_data_time_step and keep the method in the visualization part of the model.

src/porepy/models/solution_strategy.py
```diff
@@ -635,3 +637,22 @@ def update_time_dependent_ad_arrays(self) -> None:
 
         """
         self.update_all_boundary_conditions()
+
+    def export_solution(self, times_to_export: Union[list, np.ndarray, None]) -> None:
+        """Method for exporting only specified solutions.
```
IvarStefansson, line 642:
```suggestion
        """Method for exporting specified solutions.
```

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -45,16 +45,37 @@ class DataSavingMixin:
 
     def save_data_time_step(self) -> None:
         """Export the model state at a given time step, and log time."""
```
IvarStefansson, line 47: Please provide a short explanation of the main options:
None - export all
list - export if time in list. If list is empty, no times are exported.

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -45,16 +45,37 @@ class DataSavingMixin:
...
+        else:
+            # If times are specified, export should only occur if the current time is in
+            # the list of times to export.
+            do_export = bool(
+                np.any(np.isclose(self.time_manager.time, times_to_export))
+            )
+
+        # The suppress_export property takes presedence over all other criteria, hence
+        # check that last.
+        do_export = do_export and not self.suppress_export
```
IvarStefansson, line 63: Didn't we agree to remove this option, since it's covered by times_to_export=[]?

tests/viz/test_data_saving_model_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+
+from porepy.models.momentum_balance import MomentumBalance
+
+import pytest
+
+
+class DataSavingModelMixinSetup(MomentumBalance):
+    """Model setup for testing."""
+
+    def nd_rect_domain(self, x, y) -> pp.Domain:
```
IvarStefansson, line 19: Please use a Geometry class from library in applications

tests/viz/test_data_saving_model_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+    def set_domain(self) -> None:
+        x = 1.0 / self.units.m
+        y = 1.0 / self.units.m
+        self._domain = self.nd_rect_domain(x, y)
+
+    def meshing_arguments(self) -> dict:
+        mesh_args: dict[str, float] = {"cell_size": 0.1 / self.units.m}
+        return mesh_args
+
+    def bc_type_mechanics(self, sd: pp.Grid) -> pp.BoundaryConditionVectorial:
```
IvarStefansson, line 35: Why is this needed??

tests/viz/test_data_saving_model_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+        bounds = self.domain_boundary_sides(sd)
+        bc = pp.BoundaryConditionVectorial(
+            sd,
+            bounds.all_bf,
+            "dir",
+        )
+        return bc
+
+    def bc_values_displacement(self, bg: pp.BoundaryGrid) -> np.ndarray:
```
IvarStefansson, line 44: Same as above

tests/viz/test_data_saving_model_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    time_steps = 10
+    tf = 1.0
+    dt = tf / time_steps
+
+    time_manager = pp.TimeManager(
+        schedule=[0.0, tf],
+        dt_init=dt,
+        constant_dt=True,
+        iter_max=10,
```
IvarStefansson, line 91: Hardly needed for constant_dt. Same goes for print_info, no?

tests/viz/test_data_saving_model_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+
+    model = DataSavingModelMixinSetup(params)
+    model.exported_times = []
+    pp.run_time_dependent_model(model, params)
+
+    # The actual test of exported times based on the log stored in model.exported_times:
+    if times_to_export is None:
+        scheduled_times = np.linspace(0.0, tf, time_steps + 1)
+        assert len(model.exported_times) == len(scheduled_times)
+    elif isinstance(times_to_export, list):
```
IvarStefansson, line 108: Can this be converted into else?

tests/viz/test_data_saving_model_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+        assert np.all(np.isclose(model.exported_times, np.sort(times_to_export)))
```
IvarStefansson, line 109: You can also use np.allclose

tests/viz/test_data_saving_model_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    model = DataSavingModelMixinSetup(params)
+    model.exported_times = []
+    pp.run_time_dependent_model(model, params)
+
+    # The actual test of exported times based on the log stored in model.exported_times:
+    if times_to_export is None:
+        scheduled_times = np.linspace(0.0, tf, time_steps + 1)
+        assert len(model.exported_times) == len(scheduled_times)
```
IvarStefansson, line 107: Why not same assert as below? Then you could simply do: 
if times_to_export is None:
   times_to_export = np.linspace(0.0, tf, time_steps + 1)
assert np.allclose(model.exported_times, times_to_export)

IvarStefansson (commented):
Nice work, some minor comments.

tests/viz/test_data_saving_model_mixin.py
IngridKJ: So, just remove constant_dt and print_info?

src/porepy/viz/data_saving_model_mixin.py
IngridKJ: Ahh yes we did... nut I forgot, I will get to it.

tests/viz/test_data_saving_model_mixin.py
IvarStefansson: No, iter_max and print_info

IvarStefansson:
I'm happy. Since I've been involved in the process, could either of you please have a look, @OmarDuran or @keileg?

OmarDuran (commented):
Nice work @IngridKJ. I left a single comment.

tests/viz/test_data_saving_model_mixin.py
```diff
@@ -0,0 +1,70 @@
...
+        constant_dt=True,
+    )
+
+    params = {
+        "time_manager": time_manager,
+        "times_to_export": times_to_export,
+    }
+
+    model = DataSavingModelMixinSetup(params)
+    model.exported_times = []
```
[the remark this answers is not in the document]
IngridKJ: It is only for testing purposes!:)

PR #1121, opened 2024-01-29
BLD: Speed up import by delaying import of dependencies

keileg:
## Proposed changes
Reduce the import time of PorePy by delaying import of dependencies that may not be needed. 

The dependencies can roughly be divided into three groups:
* Those that are almost always used, e.g. `numpy`
* Those that are only used in special cases, e.g., `seaborn`
* Those that, in an average session, may or may not be used.

This PR tries to postpone the import of the second class until the packages are actually needed. Moreover, some imports of the third type of dependencies are also postponed, but the choice of when to do this is necessarily pragmatic rather than principled. In particular, while I would have liked to postpone import of `matplotlib`, this would have required too intrusive changes to the code.

Anecdotally, import time was cut by 20-30% by these changes.

The only change not directly related to imports was the deletion of one method in `FractureNetwork2d` which was used hardly if at all.

## Types of changes

- [ ] Other:

## Checklist

- [] If new skipped tests have been introduced in this PR, `pytest` was run with the `--run-skipped` flag.

### DIFF
--- a/src/porepy/fracs/fracture_network_2d.py
+++ b/src/porepy/fracs/fracture_network_2d.py
@@ -9,7 +9,6 @@ import warnings
 from typing import Optional, Union
 
 import meshio
-import networkx
 import numpy as np
 
 import porepy as pp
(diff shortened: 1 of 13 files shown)


src/porepy/fracs/fracture_network_2d.py
```diff
@@ -1269,53 +1268,6 @@ def copy_with_split_intersections(
...
-            for key, value in split_network.tags.items():
-                tags[key] = value[ei]
-            G.add_edge(edges[0, ei], edges[1, ei], labels=tags)
-
-        if split_intersections:
-            return G, split_network
-        else:
-            return G
-
     # --------- Utility functions below here
```
OmarDuran, line 1271: Just a matter of taste, but could you remove recurrences of `---------`?

OmarDuran (approved):
Looks good to me. I left a minor comment.

src/porepy/fracs/fracture_network_2d.py
keileg: With pleasure!

PR #1132, opened 2024-02-15
Documentation and minor change in the Peaceman well model

keileg:
## Proposed changes
This PR introduce contains a slight modification of the Peaceman well model, and more importantly better documentation of that model. Moreover, the units of some constitutive laws were added.

Consider in particular:
1. For the specific volume in Peaceman, should we take the volume integral over the interface or the subdomain? It seems the former has the better chance of being correct for 1d-3d couplings, but I am not sure.
2. Please pay attention to the formatting in the documentation of units. I prefer this style, but if we want changes, it is best to do that now, and define whatever comes out of it as a template. Also, should we always document in SI units, or are derived units (Pa, N etc.) also okay?

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/constitutive_laws.py
+++ b/src/porepy/models/constitutive_laws.py
@@ -466,7 +466,7 @@ class FluidDensityFromPressure:
             subdomains: List of subdomain grids.
 
         Returns:
-            Fluid density as a function of pressure [kg*m^-3].
+            Fluid density as a function of pressure [kg * m^-3].
 
         """
         # The reference density is taken from the fluid constants..
@@ -634,7 +634,7 @@ class FluidMobility:
             subdomains: List of subdomains.
 
         Returns:
-            Operator representing the mobility.
+            Operator representing the mobility [m * s * kg^-1].
 
         """
         return pp.ad.Scalar(1) / self.fluid_viscosity(subdomains)
@@ -677,15 +677,15 @@ class ConstantViscosity:
     """
 
     def fluid_viscosity(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
-        """Fluid viscosity [Pa s].
+        """Fluid viscosity .
 
         Parameters:
             subdomains: List of subdomain grids. Not used in this implementation, but
                 included for compatibility with other implementations.
 
         Returns:
-            Operator for fluid viscosity, represented as an Ad operator. The value is
-            picked from the fluid constants.
+            Operator for fluid viscosity [Pa * s], represented as an Ad operator. The
+            value is picked from the fluid constants.
 
         """
         return Scalar(self.fluid.viscosity(), "viscosity")
@@ -2166,6 +2166,8 @@ class PeacemanWellFlux:
     def well_flux_equation(self, interfaces: list[pp.MortarGrid]) -> pp.ad.Operator:
         """Equation for well fluxes.
 
+        The equation has units [kg * m^2 * s^-2].
+
         Parameters:
             interfaces: List of interfaces where the well fluxes are defined.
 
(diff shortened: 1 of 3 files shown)


src/porepy/models/constitutive_laws.py
```diff
@@ -2181,17 +2183,35 @@ def well_flux_equation(self, interfaces: list[pp.MortarGrid]) -> pp.ad.Operator:
...
+        # well within one cell. For a 0d-2d coupling, this will be the aperture of the
+        # 2d fracture cell; in practice the number can be obtained by multiplying with
+        # the specific volume, or by taking a volume integral over the mortar cell
+        # (which will incorporate the specific volume of the higher-dimensional
+        # neighbor, that is, the fracture). For a 1d-3d coupling, we will need the
+        # length of the well within the 3d cell (see the MRST book, p.128, for comments
+        # regarding deviated wells). Again, this could be obtained by a volume integral
+        # over the mortar cell; however, as 1d-3d couplings have not yet been
+        # implemented, we will raise an error in this case.
+        if any([sd.dim == 3 for sd in subdomains]):
```
IvarStefansson, line 2200: Should we check that no subdomains are intersections as well? I think something else would break before reaching this code, but better to be explicit about assumptions/cases not yet considered.

src/porepy/models/constitutive_laws.py
```diff
@@ -2181,17 +2183,35 @@ def well_flux_equation(self, interfaces: list[pp.MortarGrid]) -> pp.ad.Operator:
...
+        e_i = self.e_i(subdomains, i=0, dim=9).T  # type: ignore[call-arg]
+
+        # To get a transmissivity, we multiply the permeability with the length of the
+        # well within one cell. For a 0d-2d coupling, this will be the aperture of the
+        # 2d fracture cell; in practice the number can be obtained by multiplying with
+        # the specific volume, or by taking a volume integral over the mortar cell
+        # (which will incorporate the specific volume of the higher-dimensional
+        # neighbor, that is, the fracture). For a 1d-3d coupling, we will need the
+        # length of the well within the 3d cell (see the MRST book, p.128, for comments
+        # regarding deviated wells). Again, this could be obtained by a volume integral
```
IvarStefansson, line 2197: Since both cases could be covered by volume integral, I don't see why you change from that to multiplication by specific volume.

src/porepy/models/constitutive_laws.py
```diff
@@ -2166,6 +2166,8 @@ class PeacemanWellFlux:
     def well_flux_equation(self, interfaces: list[pp.MortarGrid]) -> pp.ad.Operator:
         """Equation for well fluxes.
 
+        The equation has units [kg * m^2 * s^-2].
```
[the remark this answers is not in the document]
IvarStefansson: And should we add the equation somewhere?

IvarStefansson (commented):
I'm fine with the suggested formatting. Some comments on the well equation.

src/porepy/models/constitutive_laws.py
keileg: That would require either calling this function with the wrong set of interfaces, or a very strange misuse of the geometry definition? Unless I have understood your point, I don't think we need to cover those; we generally do not do broad sanity checks of inputs (with good reason).

src/porepy/models/constitutive_laws.py
keileg: I'll change to the specific volume of the interface. To me this is more explicit (though of course equivalent) to the volume integral.

keileg: Now I got it. It will be a volume integral.

src/porepy/models/constitutive_laws.py
IvarStefansson: I think at least one of us is misunderstanding/overlooking something. Let's discuss.

IvarStefansson (approved):

PR #1133, opened 2024-03-04
Update test of fracture propagation

keileg:
## Proposed changes
Use pytest.autouse on a few setup methods. Change the name of one class to make it visible to pytest.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/tests/numerics/fracture_deformation/test_fracture_propagation.py
+++ b/tests/numerics/fracture_deformation/test_fracture_propagation.py
@@ -555,7 +555,7 @@ class MockPropagationModel(pp.ConformingFracturePropagation):
         return vals
 
 
-class PropagationCriteria:
+class TestPropagationCriteria:
     """Test of functionality to compute sifs and evaluate propagation onset and angle.
 
     Test logic:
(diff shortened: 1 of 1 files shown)


tests/numerics/fracture_deformation/test_fracture_propagation.py
```diff
@@ -582,7 +582,7 @@ class PropagationCriteria:
         values for u and parameters.
 
     """
-
+    @pytest.fixture(autouse=True)
```
IvarStefansson, line 585: Should we be explicit about the scope of the fixture?

keileg: I don't use it that way, but I don't know if I follow best practice.

IvarStefansson: Let me rephrase: Have you considered what the scope of this fixture is? 
I think that it is `function`, i.e., it is invoked once per `test_...` method of this class. Do you agree? And is this the intended behaviour?

keileg: No, I have not considered this.

This is a fix of functionality which in practice is not used, so honestly, I put a low priority on maintaining the tests beyond making sure they run, and do take too much time (this class runs pretty fast).

PR #1135, opened 2024-03-07
Tensorial coupling in (thermo)poromechanical systems

### LINKED ISSUE #395: Implement tensor BiotAlpha in DivU term.
In the current implementation of MPSA for Biot, we allow for a matrix biot modulus in the gradP term, but only a scalar in the divU. 

The implementation should not be difficult; a natural option seems to be to modify Biot._subcell_gradient_to_cell_scalar().

Once this is in place, we should also allow for input of non-scalar BiotAlpha.

keileg:
### Overview
This PR introduces tensorial coupling terms ('tensor Biot coefficients') for thermoporomechanical systems. The main changes can be summarized as:
* The Biot discretization class has been updated to discretize tensorial coupling coefficients.
* The Ad wrapper for discretizations has been changed to accommodate the new functionality. Some technical details are outlined below.
* The models for poromechanics and thermoporomechanics, and the corresponding constitutive laws, have been adapted accordingly. By default, scalar coupling terms are used, but this can be changed by owerwriting the relevant constitutive laws.
* Functional tests for thermoporomechanics have been introduced. 

### Modifications to the governing equations
* In addition to the `biot_coefficient` we now have a `biot_tensor` (set in constitutive law `BiotCoefficient.biot_tensor()`). The tensor variant enters what we known as the `grad_p` and `div_u` terms (following Coussy, Eq. (4.13)), while the coefficient is still used for the pressure-dependent modification of porosity (following Coussy, eq (4.13b) and (4.35))
* Correspondingly, the `solid_thermal_expansion` has been split into a `solid_thermal_expansion_coefficient` and a `solid_thermal_expansion_tensor`. The coefficient is used in the porosity-temperature relation (Coussy, 4.13b and 4.59-61), while the tensor is used is used in the momentum balance (Coussy 4.13a).

The default behavior is as before: Both Biot and thermal expansions are scalars set as material constants, and expanded to tensors where needed.

### Accessing Ad discretization matrices, changes to the Biot discretization
Since the Biot discretization now handles an arbitrary number of coupling terms, we need a system to distinguish between, say, different `grad_p`-terms. I have experimented with several variants, and ended up with the following solution:
1. The discretization matrices (technically `MergedOperators`) are now methods of the Ad discretizations rather than attributes. Thus, instead of `mpfa.flux`, write `mpfa.flux()`.
2. The different coupling terms are distinguished by giving an argument to the method, so, write `biot.grad_p('foo')`, 'biot.grad_p('bar')' etc.
3. To make points 1. and 2. work, I had to make some adjustments in the wrapping of discretizations into Ad discretizations: A discretization which contains coupling terms must specify the name of the wrapped terms (so, `BiotAd` must declear `grad_p` to be a coupling term, so that the method `grad_p()` will accept keywords). 
4. The Biot discretization accepts a dictionary of coupling coefficients/tensors, and will discretize each of these separately. For an example of how these tensors are set on the model side, see https://github.com/pmgbergen/porepy/blob/9579320ad1ee119744729a67bd48d362475e2818/src/porepy/models/poromechanics.py#L174C1-L177C43

### Testing
The design of the discretization classes makes unit-like testing challenging for this extension. I have run extensive convergence tests based on the new THM functional test (the contributed new tests can be viewed as a summary of this testing), and I am quite confident the implementation is correct.

### Clarifications needed
A few clarifications are needed:
* The name of the coupling terms are still `grad_p`, `div_u` and `stabilization`. For the latter, `consistency` would be a more descriptive name, while`div_u` is may be fair enough (for a scalar coupling coefficient, the term still represents the divergence of the displacement tensor). `grad_p` is clearly a misnomer. Suggestions for what to do here?
* The inclusion of tensor quantities into the models should be checked:
    * Are the names and placements of new constitutive laws reasonable?
    * Is my interpretation of Coussy, and the inclusion of the constitutive laws, correct?

### Changes not implemented
During implementation, it became clear to me that we can merge the `mpsa.py` and `biot.py` to have a single discretization of vector problems. This may simplify the code structure significantly. However, this will be a major undertaking by itself, and should be considered at a later stage.

This PR resolves #395 

## Types of changes

- [ ] Other:

## Checklist


keileg:
> What about `displacement_divergence` and `scalar_gradient` as coupling term names? 

Agree, those are good names.

> Could we be more specific than `consistency`?

Regarding consistency, the term is inherently linked to mpsa, so we could say `mpsa_consistency`?

src/porepy/models/constitutive_laws.py
```diff
@@ -3645,6 +3665,10 @@ class ThermoPressureStress(PressureStress):
     instance of
     :class:`~porepy.models.momentum_balance.SolutionStrategyMomentumBalance`.
 
+    """
+    temperature_variable: str
+    """Name of the pressure variable. Normally set by a mixin instance of
```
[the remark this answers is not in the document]
IvarStefansson: Any idea what happens if we wish to use enthalpy as primary variable?

src/porepy/models/constitutive_laws.py
```diff
@@ -4477,10 +4516,16 @@ def displacement_divergence(
...
+        self, subdomains: list[pp.Grid], variable_name: str
+    ) -> pp.ad.Operator:
+        """Consistency term for Biot-type discretizations.
 
-        TODO: Determine if this is the correct place to include stabilization.
+        This function returns a diffusion-type term that is needed to ensure an
+        MPSA-type discretization of poromechanics (or
+        thermomechanics/thermoporomechanics) is stable in the limit of vanishing time
+        steps and permeability. The term arises naturally from the MPSA discretization,
+        see Nordbotten 2016 (SIAM) for details.
```
IvarStefansson, line 4528: Could we have a DOI, please?

src/porepy/models/constitutive_laws.py
```diff
@@ -4493,12 +4538,14 @@ def biot_stabilization(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
         if not all(sd.dim == self.nd for sd in subdomains):
             raise ValueError("Biot stabilization only defined in nd.")
 
-        discr = pp.ad.BiotStabilizationAd(self.darcy_keyword, subdomains)
+        discr = pp.ad.BiotAd(self.stress_keyword, subdomains)
+
+        # discr = pp.ad.BiotStabilizationAd(self.darcy_keyword, subdomains)
```
IvarStefansson, line 4543: Delete?

src/porepy/models/constitutive_laws.py
```diff
@@ -4477,10 +4516,16 @@ def displacement_divergence(
         div_u.set_name("div_u")
         return div_u
 
-    def biot_stabilization(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
-        """Biot stabilization term.
+    def biot_stabilization(
+        self, subdomains: list[pp.Grid], variable_name: str
```
IvarStefansson, line 4520: The exposed method mostly have only subdomains in the signature. We might want to discuss.

src/porepy/models/poromechanics.py
```diff
@@ -137,24 +137,55 @@ class SolutionStrategyPoromechanics(
...
+        Parameters:
+            params: Dictionary of parameters.
+
+        """
+        # Initialize the solution strategy for the fluid mass balance subproblem.
+        mass.SolutionStrategySinglePhaseFlow.__init__(self, params=params)
+        # Initialize the solution strategy for the momentum balance subproblem.
+        momentum.SolutionStrategyMomentumBalance.__init__(self, params=params)
+
+        self.biot_coefficient_key = "biot_alpha"
```
IvarStefansson, line 160: Can't this be a class attribute? Possibly not, but if so, I'm not sure I like this solution of having to explicitly call the other inits.

src/porepy/models/thermoporomechanics.py
```diff
@@ -174,14 +186,32 @@ def set_discretization_parameters(self) -> None:
...
-            pp.initialize_data(
-                sd,
-                data,
-                self.stress_keyword,
-                {
-                    "biot_alpha": self.solid.biot_coefficient(),  # TODO: Rename in Biot
-                },
-            )
+            if self.stress_keyword in data[pp.PARAMETERS]:
+                # TODO: I think we can always assume that the stress is defined in the
```
IvarStefansson, line 190: Otherwise, we need an else, right?

IvarStefansson (commented):
Partial review

src/porepy/models/constitutive_laws.py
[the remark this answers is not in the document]
keileg: Good question, I have thought about it, but not deeply. Let's discuss in person.

src/porepy/models/poromechanics.py
keileg: I think it can be deleted, old code. Thanks for spotting.

src/porepy/models/constitutive_laws.py
keileg: agree

src/porepy/numerics/fv/biot.py
```diff
@@ -340,7 +297,17 @@ def discretize(self, sd: pp.Grid, sd_data: dict) -> None:
...
         )
 
-        alpha: float = parameter_dictionary["biot_alpha"]
+        scalar_vector_mappings: dict = parameter_dictionary["scalar_vector_mappings"]
+        coupling_keywords: list[str] = list(scalar_vector_mappings.keys())
+
+        alpha: dict[str, pp.SecondOrderTensor] = {}
+
+        for key, alpha_input in scalar_vector_mappings.items():
+            # TODO: Revisit 'biot_coupling_coefficient
```
IvarStefansson, line 306: Please search for TODOs before merging. Perhaps leave this as unresolved as a reminder?

src/porepy/numerics/fv/biot.py
```diff
@@ -394,13 +366,27 @@ def discretize(self, sd: pp.Grid, sd_data: dict) -> None:
...
         active_bound_displacement_face = sps.csr_matrix((nf * nd, nf * nd))
-        active_bound_displacement_pressure = sps.csr_matrix((nf * nd, nc))
+
+        active_scalar_gradient, active_displacement_divergence, active_bound_displacement_divergence, active_stabilization = (
+            {},
+            {},
+            {},
+            {},
+        )
+        active_bound_displacement_pressure = {}
```
IvarStefansson, line 379: Please choose between defining all active_... with one statement and on individual lines.

src/porepy/numerics/fv/biot.py
```diff
@@ -532,13 +534,10 @@ def discretize(self, sd: pp.Grid, sd_data: dict) -> None:
...
-        # IMPLEMENTATION NOTE: With the current implementation, this scaling is only
-        # needed for the vector quantities, that is, not for the stabilization and div_u
-        # terms. This is because the latter are computed cell-wise rather than
-        # face-wise, and it so happens that this is sufficient to avoid double counting.
-        # If we ever change the implementation (e.g., when introducing tensor Biot
-        # coefficients), it is likely the scaling will be needed also for the scalar
-        # terms.
+        # IMPLEMENTATION NOTE: This scaling is only needed for the vector quantities,
+        # that is, not for the stabilization and displacement_divergence terms. The latter are computed
+        # cell-wise rather than face-wise, and will only discretized once, even for
```
IvarStefansson, line 539:
```suggestion
        # cell-wise rather than face-wise, and will only be discretized once, even for
```

src/porepy/numerics/fv/biot.py
```diff
@@ -744,270 +759,223 @@ def _local_discretization(
...
 
         # We obtain the reconstruction of displacements. This is equivalent as for
         # mpsa, but we get a contribution from the pressures.
         dist_grad, cell_centers = self._reconstruct_displacement(
             sd, subcell_topology, eta
         )
+        # The boundary discretization of the displacement_divergence term is represented
+        # directly on the cells, instead of going via the faces.
+        scalar_gradient, bound_displacement_divergence, displacement_divergence, stabilization, disp_pressure = {}, {}, {}, {}, {}
+        for key in alpha:
```
IvarStefansson, line 777: Could you expand the documentation a little bit, please?

src/porepy/numerics/fv/biot.py
```diff
@@ -744,270 +759,223 @@ def _local_discretization(
...
-        """
+        # Background/motivation: In the extension of MPSA from pure elasticity to
+        # poroelasticity (or thermoelasticity etc.), the consistent discretization of
+        # the pressure force, represented as ``alpha p`` where ``alpha`` is a 2-tensor
+        # consists of two terms (to see this, see sections 3 and 4 in Nordbotten 2016 -
+        # but be aware a careful read will be needed): First, the pressure force is
+        # computed on a face as ``n\dot (alpha p)`` where n is the normal vector; this
+        # is the variable scalar_gradient_face below. Second, imbalances in the pressure force
+        # between neighboring cells give rise to local mechanical deformation, which
+        # again induce stresses; this is variable rhs_jumps below.  The discretization
```
IvarStefansson, line 856:
```suggestion
        # again induce stresses; this is the variable rhs_jumps below. The discretization
```

src/porepy/numerics/fv/biot.py
```diff
@@ -340,7 +297,17 @@ def discretize(self, sd: pp.Grid, sd_data: dict) -> None:
             "inverter", "numba"
         )
 
-        alpha: float = parameter_dictionary["biot_alpha"]
+        scalar_vector_mappings: dict = parameter_dictionary["scalar_vector_mappings"]
+        coupling_keywords: list[str] = list(scalar_vector_mappings.keys())
+
+        alpha: dict[str, pp.SecondOrderTensor] = {}
```
IvarStefansson, line 303: This would make the distinction between a single tensor and the collection of tensors clearer:
```suggestion
        alphas: dict[str, pp.SecondOrderTensor] = {}
```

src/porepy/numerics/fv/biot.py
```diff
@@ -1065,336 +1072,3 @@ def _subcell_gradient_to_cell_scalar(
...
-        stress_discr = matrix_dictionary["stress"]
-        bound_stress = matrix_dictionary["bound_stress"]
-        bound_val = sd_data[pp.PARAMETERS][self.mechanics_keyword]["bc_values"]
-        d = u[: sd.dim * sd.num_cells]
-
-        stress = stress_discr * d + (bound_stress * bound_val)
-        return stress
-
-
-class GradP(Discretization):
```
IvarStefansson, line 1118: Hooray!

src/porepy/models/thermoporomechanics.py
```diff
@@ -166,6 +166,18 @@ class SolutionStrategyThermoporomechanics(
     """Discretization of the Fourier flux. Normally provided by a mixin instance of
     :class:`~porepy.models.constitutive_laws.FouriersLaw`.
 
+    """
+    temperature_variable: str
+    """Name of the pressure variable. Normally set by a mixin instance of
```
IvarStefansson, line 171:
```suggestion
    """Name of the temperature variable. Normally set by a mixin instance of
```

src/porepy/models/constitutive_laws.py
```diff
@@ -3665,23 +3689,12 @@ def thermal_stress(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
...
-            beta
-            * k
-            / alpha
-            * (
-                discr.grad_p @ self.temperature(subdomains)
-                - discr.grad_p @ self.reference_temperature(subdomains)
-            )
+        stress: pp.ad.Operator = discr.scalar_gradient(
+            self.temperature_variable
+        ) @ self.temperature(subdomains) - discr.scalar_gradient(
```
IvarStefansson, line 3694: This would equal  discr.scalar_gradient(self.temperature_variable) @ self.perturbation_from_reference(subdomains, "temperature"), wouldn't it?

src/porepy/models/constitutive_laws.py
```diff
@@ -3665,23 +3689,12 @@ def thermal_stress(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
...
-        stress: pp.ad.Operator = (
-            beta
-            * k
-            / alpha
-            * (
-                discr.grad_p @ self.temperature(subdomains)
-                - discr.grad_p @ self.reference_temperature(subdomains)
-            )
+        stress: pp.ad.Operator = discr.scalar_gradient(
+            self.temperature_variable
```
IvarStefansson, line 3693: Although in a way I like passing self.temperature_variable here, I think a more consistent thing would be to link it to the discretization keyword ("mechanics"). That's the identifier for discretizations and their parameters and matrices.

IvarStefansson (commented):
I have not looked at the tests yet.

src/porepy/numerics/fv/fvutils.py
```diff
@@ -1257,16 +1251,7 @@ def partial_update_discretization(
...
-    # according to the left/right, cell/face and scalar/vector specifications.
-    # Also eliminate contributions to rows that will also be updated (but not
-    # columns, a non-updated row should keep its information about a column
-    # to be updated).
-    mat_dict = data[pp.DISCRETIZATION_MATRICES][keyword]
-    mat_dict_copy = {}
-    for key, val in mat_dict.items():
-        mat = val
-
+    def update_matrix(key, mat):
```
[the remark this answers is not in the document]
keileg: When the function is nested like this, I don't think it matters much, and it is not the first place I will strive to be consistent. I usually do not include a leading underscore, probably out of habit.

If the function is not nested, I think it should be marked with a leading underscore.

src/porepy/numerics/fv/fvutils.py
```diff
@@ -1750,3 +1731,46 @@ def partial_discretization(
...
+    loc_c.values = loc_c.values[::, ::, loc_cells]
+    # Also restrict the lambda and mu fields; we will copy the stiffness tensors
+    # later.
+    loc_c.lmbda = loc_c.lmbda[loc_cells]
+    loc_c.mu = loc_c.mu[loc_cells]
+    return loc_c
+
+
+def restrict_second_order_tensor_to_subgrid(
+    K: pp.SecondOrderTensor, loc_cells: np.ndarray
```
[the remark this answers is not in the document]
keileg: I agree with unification. Tensor is fine, I will implement.

IvarStefansson:
> > What about `displacement_divergence` and `scalar_gradient` as coupling term names?
> 
> Agree, those are good names.
> 
> > Could we be more specific than `consistency`?
> 
> Regarding consistency, the term is inherently linked to mpsa, so we could say `mpsa_consistency`?

Yes, if it's linked to mpsa and not the mpsa-mpfa combination.

src/porepy/models/thermoporomechanics.py
keileg: Turns out we can do without the if-else

src/porepy/models/constitutive_laws.py
keileg: The discretization keyword is common for all coupling terms, since they are all triggered by the `mechanics` equation. So the identifier of the specific coupling term must either be connected to the variable with which it is associated (e.g. temperature), or with the equation with which the coupling term is associated (`energy_equation`  or similar). Of these options, the variable is to me the natural one.

IvarStefansson: No, I meant the "other" discretization keyword, e.g. "flow".

keileg:
> > > What about `displacement_divergence` and `scalar_gradient` as coupling term names?
> > 
> > 
> > Agree, those are good names.
> > > Could we be more specific than `consistency`?
> > 
> > 
> > Regarding consistency, the term is inherently linked to mpsa, so we could say `mpsa_consistency`?
> 
> Yes, if it's linked to mpsa and not the mpsa-mpfa combination.

Closure: The term will be `mpsa_stability` in the general setting (ex: `constitutive_laws.mpsa_consistency()`), but simply `consistency` when there should be no room for confusion, such as `BiotAd().consistency()`

src/porepy/models/constitutive_laws.py
keileg: I agree, using 'physics keywords' would be preferable. For the pressure coupling term, the natural key is accessed as `self.darcy_keyword`. For the thermal coupling, there is no single keyword associated with energy balance, but we do have both `self.enthalpy_keyword` and `self.fourier_keyword`. I am inclined to use `enthalpy_keyword` for now, and then possibly revisit the whole keyword concept at a later stage. Does this make sense?

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+        # but simpler, implementation in manu_poromech_nofrac_2d. EK tried this but
+        # found out that, while the number of lines of code could be reduced somewhat,
+        # the price was a significant increase in code complexity. It was therefore
+        # decided to keep the current structure.
+
+        mdg: pp.MixedDimensionalGrid = self.mdg
+        sd: pp.Grid = mdg.subdomains()[0]
+        t: number = self.time_manager.time
+
+        # Collect data
```
IvarStefansson, line 170: As a matter of taste, I would have done this in a loop to save code lines.

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+            time=t,
+        )
+
+        return collected_data
+
+
+class ManuThermoPoroMechExactSolution2d:
+    """Class containing the exact manufactured solution for the verification setup.
+
+    The exact solutions for the primary variables pressure, displacement, temperature,
```
IvarStefansson, line 272:
```suggestion
    The exact solutions for the primary variables pressure, displacement and temperature,
```

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+    are defined below, as well as the exact solutions for the secondary variables Darcy
+    flux and poroelastic force.
```
IvarStefansson, line 274: Do you mean?
```suggestion
    flux and thermoporoelastic force.
```

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+    take different values in the region x > 0.5 and y > 0.5. The primary variables
+    pressure and displacement are scaled with the heterogeneity, but in a reciprocal
+    way, so that the derived fluxes and stresses are continuous across the interface
+    between the regions with different material parameters. Moreover, the primary
+    variables are constructed to be zero at the interface, so that they are continuous
+    across the interface. This is necessary to ensure that the exact solution can be
+    used to verify convergence of the numerical solution.
+
+    The temperature variable was not scaled with the heterogeneity, since this would be
+    difficult to make work with the fluid conductivity (thus effective heterogeneity)
```
IvarStefansson, line 286: I don't get the parenthesis.
```suggestion
    difficult to make work with the fluid heat conductivity (thus effective heterogeneity)
```

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+        ]
+
+        # Face-centered Darcy fluxes
+        q_fc: np.ndarray = (
+            q_fun[0](fc[0], fc[1], time) * fn[0] + q_fun[1](fc[0], fc[1], time) * fn[1]
+        )
+
+        return q_fc
+
+    def poroelastic_force(self, sd: pp.Grid, time: float) -> np.ndarray:
```
IvarStefansson, line 755: Thermoporoelastic?

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+        source_mech: list[np.ndarray] = [
+            source_mech_fun[0](cc[0], cc[1], time) * vol,
+            source_mech_fun[1](cc[0], cc[1], time) * vol,
+        ]
+
+        # Flatten array
+        source_mech_flat: np.ndarray = np.asarray(source_mech).ravel("F")
+
+        # Invert sign according to sign convention.
+        return -source_mech_flat
```
IvarStefansson, line 845: Why is a minus required here and not for flow?

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+            name="source_flow",
+            domains=self.mdg.subdomains(),
+            previous_timestep=True,
+        )
+
+        # Add up contribution of internal and external sources of fluid
+        fluid_sources = internal_sources + external_sources
+        fluid_sources.set_name("Fluid source")
+
+        return fluid_sources  # - prod
```
IvarStefansson, line 998:
```suggestion
        return fluid_sources
```

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+            domains=self.mdg.subdomains(),
+            previous_timestep=True,
+        )
+
+        return external_sources
+
+    def energy_source(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
+        """Thermal source."""
+
+        # Internal sources are inherit from parent class
```
IvarStefansson, line 1014:
```suggestion
        # Internal sources are inherited from parent class.
```

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+        # Zero internal sources for unfractured domains, but we keep it to maintain
+        # good code practice
+        internal_sources: pp.ad.Operator = super().energy_source(subdomains)
+
+        # External sources are retrieved from pp.TIME_STEP_SOLUTIONS and wrapped as an
+        # AdArray.
+        external_sources = pp.ad.TimeDependentDenseArray(
+            name="source_energy",
+            domains=self.mdg.subdomains(),
+            previous_timestep=True,
```
IvarStefansson, line 1024: Why previous timestep?

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+        )
+
+        # Add up contribution of internal and external sources of fluid
```
IvarStefansson, line 1027:
```suggestion
        # Add up contribution of internal and external sources of energy.
```

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+    """List of SaveData objects."""
+
+    def __init__(self, params: dict):
+        """Constructor for the class."""
+        super().__init__(params)
+
+        self.exact_sol: ManuThermoPoroMechExactSolution2d
+        """Exact solution object."""
+
+        self.results: list[ManuThermoPoroMechSaveData] = []
```
IvarStefansson, line 1056: There are duplicate versions of self.results: Class and instance attribute.

IvarStefansson: Same goes for exact_sol

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+        """Results object that stores exact and approximated solutions and errors."""
+
+        self.flux_variable: str = "darcy_flux"
```
IvarStefansson, line 1059: I don't see why this and the stress variable are not class variables. That's the common practice in the models.

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+            name="source_energy", values=energy_source, data=data, time_step_index=0
+        )
+
+    def _is_nonlinear_problem(self) -> bool:
+        """The problem is non-linear due to the coupling between fluid flux and
+        advective energy transport.
+        """
+        return True
+
+    def bulk_modulus(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
```
IvarStefansson, line 1103: Does this belong to the SolutionStrategy, or should it be a constitutive law? Motivation: Being precise on classification/modularity may facilitate reuse in the future.

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
...
+            if isinstance(bulk, (float, int)):
+                bulk = bulk * np.ones(sd.num_cells)
+
+            return pp.wrap_as_dense_ad_array(bulk, name="bulk_modulus")
+
+    def set_discretization_parameters(self) -> None:
+        """Set parameters for the subproblems and the combined problem.
+
+        In addition to the standard fields (call to super), the permeability, stiffness
+        parameters, and the Biot and thermal stress tensors are set.
```
IvarStefansson, line 1129: Wouldn't it be more precise to say that the super call's values are overwritten?

tests/functional/setups/manu_thermoporomech_nofrac_3d.py
```diff
@@ -0,0 +1,1053 @@
...
+        if self.params["grid_type"] == "simplex":
+            # Mark the fractures added as constraints (not to be represented as
+            # lower-dimensional objects).
+            return {"constraints": [0, 1, 2]}
+        else:
+            return {}
+
+
+# -----> Solution strategy
+class ManuThermoPoroMechSolutionStrategy3d(
```
IvarStefansson, line 873: I think this class allows for partial reuse of the 2d equivalent.

tests/functional/test_manu_poromech_nofrac.py
```diff
@@ -131,43 +131,43 @@ def desired_l2_errors() -> list[list[dict[str, float]]]:
     """
     desired_errors_2d = [
         {  # t = 0.2 [s]
-            "error_pressure": 0.25151427228540035,
-            "error_flux": 0.18937166702134517,
-            "error_displacement": 0.3956281463918882,
-            "error_force": 0.1637416678810283,
+            "error_pressure": 0.2545484162739883,
```
IvarStefansson, line 134: Do you know what caused these changes?

tests/viz/test_exporter.py
```diff
@@ -614,7 +614,7 @@ def run_simulation_save_results(units: pp.Units, file_name: str):
             "specific_heat_capacity": 7.2,  # [J * kg^-1 * K^-1]
             "specific_storage": 4e-1,  # [Pa^-1]
             "thermal_conductivity": 3.1,  # [W * m^-1 * K^-1]
-            "thermal_expansion": 9.6e-1,  # [K^-1]
+            "thermal_expansion": 9.6e-2,  # [K^-1]2
```
IvarStefansson, line 617: What happened here?

tests/numerics/fracture_deformation/test_propagate_fracture.py
```diff
@@ -151,9 +151,18 @@ def set_param(sd, data):
...
-                if diff.data.size > 0:
-                    assert np.max(np.abs(diff.data)) < 1e-10
+                if isinstance(mat, dict):
+                    # Special treatment of Biot coupling terms
+                    for sub_key, sub_mat in mat.items():
+                        mat2 = new_d[pp.DISCRETIZATION_MATRICES][eq_type][key][sub_key]
+                        assert sub_mat.shape == mat2.shape
+                        diff = sub_mat - mat2
+                        if diff.data.size > 0:
+                            assert np.max(np.abs(diff.data)) < 1e-10
```
IvarStefansson, line 161:
```suggestion
                            assert np.allclose(diff.data, 0, atol=1e-10))
```

tests/numerics/fracture_deformation/test_propagate_fracture.py
```diff
@@ -151,9 +151,18 @@ def set_param(sd, data):
...
+                else:
+                    mat2 = new_d[pp.DISCRETIZATION_MATRICES][eq_type][key]
+                    assert mat.shape == mat2.shape
+                    diff = mat - mat2
+
+                    if diff.data.size > 0:
+                        assert np.max(np.abs(diff.data)) < 1e-10
```
IvarStefansson, line 168:
```suggestion
                        assert np.allclose(diff.data, 0, atol=1e-10))
```

IvarStefansson (approved):
Impressive work with the tests!! Have you considered whether the new ones make some of the old tests redundant?

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -0,0 +1,1219 @@
+r"""
+This module contains the setup for a 2d verification test for the coupled
+thermo-poromechanical problem. The problem is defined on a unit square domain, and
+consists of a fluid flow equation, a mechanical equation, and an energy equation.
+
+The problem definition contains a heterogeneity in the permeability and Lamé parameters.
```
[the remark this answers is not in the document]
keileg: No.

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
keileg: It is rather specialized, so I think this placement will be fine.

PR #1141, opened 2024-04-03
Make code compatible with scipy v1.13

keileg:
## Proposed changes
Updates:
1. Disabled two tests related to the coarsening module. The scipy update permuted the output of what is essentially a sparse.find, leading the test to fail a comparison with a known truth. As the coarsening moduleis deprecated, thus fixing the tests was not prioritized. 
2. Enforce integer format of grid topology information. 

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/grids/simplex.py
+++ b/src/porepy/grids/simplex.py
@@ -103,7 +103,7 @@ class TriangleGrid(Grid):
         # Cell face relation
         num_faces_per_cell = 3
         cell_faces = cell_faces.reshape(num_faces_per_cell, num_cells).ravel("F")
-        cf_data = cf_data.reshape(num_faces_per_cell, num_cells).ravel("F")
+        cf_data = cf_data.reshape(num_faces_per_cell, num_cells).ravel("F").astype(int)
 
         indptr = np.hstack(
             (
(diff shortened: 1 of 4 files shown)


src/porepy/grids/simplex.py
```diff
@@ -112,9 +112,7 @@ def __init__(
             )
         )
         cell_faces = sps.csc_matrix(
-            (cf_data, cell_faces, indptr),
-            shape=(num_faces, num_cells),
-            dtype=float,
```
IvarStefansson, line 117: Interesting!

IvarStefansson (commented):
Seems good to me. I'll defer to @OmarDuran, though. He has a better understanding of this than me.

OmarDuran (changes requested):
@keileg thank you for enforcing types in topological information. The only suggestion I have is to implement a test that checks that the topological information is indeed an integer for all the grids.

OmarDuran:
> > @keileg thank you for enforcing types in topological information. The only suggestion I have is to implement a test that checks that the topological information is indeed an integer for all the grids.
> 
> I'm not sure if I agree: The attributes will be `int`s now and remain so unless we remove that line, which is highly unlikely. Moreover, it is not a disaster if the attributes become floats again - the previous code worked perfectly fine for several years. I realize that sparse find etc. may be less stable for floats, but again, this will not happen unless we revert the code.
> 
> So, I would say the benefits of testing this feature do not warrant the cost of dealing with extra tests. Do you agree, or have I missed something?

In addition to being a suggestion, keep in mind that tracking types correctly correlates with code stability. Nevertheless, I agree with you that as long as users do not mess with types associated with topology, these changes are sufficient.

PR #1166, opened 2024-05-03
Maintenance in porepy.ad

vlipovac:
## Proposed changes

1. Introduction of intermediate Operator base classes handling arbitrary depth of previous iteration and time step.
2. Re-work in forward evaluation: Removing double recursion by introducing `parse` for variables and replacing projection of global forward AdArray by slicing (less expensive than matrix multiplication)
3. EquationSystem works internally now solely on Variable ID: Changes in some private data maps and fixing of global DOF order.
4. Maintenance on Operator functions: Proper type setting and abstraction compatible with multiple inheritance including intermediate base classes from Change No 1.
5. Adding tests to cover changes in prev iter and time step, as well as some basic tests of the operator function class.

Since the changes are a bit extensive, I am giving a walk-through in the form of a self-review.

## EDIT:

Following the 1st round of reviews, the index convention for previous time steps is now unified with the convention for previous iterates:

- 0 represents now the current value
- 1 represents the first previous value

Since current time (unknown value) is given by `time_step_index = 0` and any `iterate_index >= 0`, it would be ambiguous to use the time step index. (and also unnecessary to store same data twice).
Therefore `time_step_index = 0` is now not supported anymore.

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue).
- [ ] Other:

## Checklist


src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -294,6 +294,42 @@ def discretize_from_list(
                     pass
 
 
+def _validate_indices(
```
vlipovac, line 297: Adding general validation of time step and iteration indices.
Using ValueError to alert invalid usage (index increasing with depth)
Using KeyError to alert user that evaluation is attempted at index for which no data is set.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -168,22 +168,31 @@ def __init__(self, mdg: pp.MixedDimensionalGrid) -> None:
 
         """
 
-        self._variables: list[Variable] = list()
```
vlipovac, line 171: Internal maps are now all based on the variable ID for quick access.
_variable is a general collection containing the variables as created chronologically.
Other maps are slightly optimized (after each variable creation) to reflect a grid-wise clustering,

src/porepy/numerics/ad/equation_system.py
```diff
@@ -572,56 +584,28 @@ def get_variable_values(
...
         Raises:
-            ValueError: If neither of ``time_step_index`` or ``iterate_index`` have been
-            assigned a non-None value.
-            ValueError: If both ``time_step_index`` and ``iterate_index`` have been
-            assigned a value.
-            KeyError: If no values are stored for the VariableType input.
             ValueError: If unknown VariableType arguments are passed.
 
         """
-        if time_step_index is None and iterate_index is None:
```
vlipovac, line 583: get and set variable values uses now the respective ad utility functions (no code duplicate anymore)
Error handling of indices is implemented there.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -851,56 +818,49 @@ def _append_dofs(self, variable: pp.ad.Variable) -> None:
...
-        subdomains and interfaces.
 
         """
         # Data stracture for the new order of dofs.
         new_variable_counter: int = 0
-        new_variable_numbers: dict[Variable, int] = dict()
+        new_variable_numbers: dict[int, int] = dict()
         new_block_dofs: list[int] = list()
 
-        for variable in self._variables:
```
vlipovac, line 891: Grid-wise clustering of variable DOFs was broken in some previous PR. Fixed again

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1035,31 +997,40 @@ def dofs_of(self, variables: VariableList) -> np.ndarray:
...
 
         """
         variables = self._parse_variable_type(variables)
+        var_ids = [var.id for var in variables]
         global_variable_dofs = np.hstack((0, np.cumsum(self._variable_num_dofs)))
 
         # Storage of indices per requested variable.
+        # NOTE Loop over current order in variable numbers to preserve order of dofs
         indices = list()
-        for variable in variables:
```
vlipovac, line 1049: dof_of returned the indices depending on the order of the VariableList argument.
Fixed s.t. dofs are always ascending (reflecting order imposed by equation system).

src/porepy/numerics/ad/forward_mode.py
```diff
@@ -100,6 +100,56 @@ def __repr__(self) -> str:
         s += " elements"
         return s
 
+    def __getitem__(self, key: slice | np._ArrayLikeInt) -> AdArray:
```
vlipovac, line 103: AdArrays can now be sliced row-wise.
The getter is clear I think.
The setter allows to set the rows in val and jac using an AdArray of respective size.
It also allows to set val only, if the setter receives a numpy array only, and not an AdArray.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -9,97 +16,90 @@
...
 
 import porepy as pp
 from porepy.numerics.ad.forward_mode import AdArray
 
+from .functions import FloatType
 from .operators import Operator
 
 __all__ = [
-    "Function",
-    "ConstantFunction",
```
vlipovac, line 24: ConstantFunction was removed, since this is a duplicate functionality of Wrapped array or scalar.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -9,97 +16,90 @@
...
-        ### PUBLIC
-
-        self.func: Callable = func
-        """Callable passed at instantiation"""
-
-        self.array_compatible: bool = array_compatible
-        """Indicator whether the callable can process arrays."""
-
-        super().__init__(name=name, operation=Operator.Operations.evaluate)
+        name: Optional[str] = None,
```
vlipovac, line 61: The AbstractFunction has a general implemenation of the `func` member now, with get_values and get_jacobian remaining abstract.
Constructor changed to be compatible with multiple inheritance including other base clases such as the new IterativeOperator and TimeDependentOperator.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -109,59 +109,127 @@ def __repr__(self) -> str:
         s = f"AD Operator function '{self._name}'"
         return s
 
-    def __mul__(self, other):
-        raise RuntimeError(
-            "AD Operator functions are meant to be called, not multiplied."
+    def __neg__(self) -> Operator:
```
vlipovac, line 112: Updating the arithmetic overloads, which were introduced for the class Operator.
Introducing the TypeError to avoid users operating on uncalled functions.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -170,188 +238,117 @@ def get_values(self, *args: AdArray) -> np.ndarray:
...
+        jacs = [
+            arg.jac * m
+            for arg, m in zip(args, self._multipliers)
+            if isinstance(arg, AdArray)
+        ]
+        return sum(jacs).tocsr()
 
 
 class Function(AbstractFunction):
-    """Ad representation of an analytically given function,
```
vlipovac, line 235: The Function class receives the callable as before, but it internally uses it in the new, inherited `func` member

src/porepy/numerics/fracture_deformation/propagation_model.py
```diff
@@ -170,7 +170,7 @@ def _map_variables(self, x: np.ndarray) -> np.ndarray:
             # Only cell-based dofs have been considered so far.
             # It should not be difficult to handle other types of variables,
             # but the need has not been there.
-            dofs = self.equation_system._variable_dof_type[var]
```
vlipovac, line 173: This usage reflects that the framework as it was and still is does not fully support changing numbers of dofs

tests/numerics/ad/test_equation_system.py
```diff
@@ -60,6 +60,8 @@ def test_evaluate_variables():
             name=var_name, values=vals_sol, data=d, time_step_index=0
         )
         pp.set_solution_values(name=var_name, values=vals_it, data=d, iterate_index=0)
```
vlipovac, line 62: I don't know how this test passed earlier, since no values were set for this iterate index?

tests/numerics/ad/test_forward_mode.py
```diff
@@ -198,3 +200,57 @@ def test_exp_scalar_times_ad_var():
...
+
+
+@pytest.mark.parametrize(
+    'index,index_c', [  # indices and their complement for tested array
+        (1, [0, 2, 3, 4, 5, 6, 7, 8, 9]),
+        (slice(0, 10, 2), slice(1, 10, 2)),
+        (np.array([0, 2, 4, 6, 8], dtype=int), np.array([1, 3, 5, 7, 9], dtype=int)),
+    ]
+)
+def test_get_set_slice_ad_var(index, index_c):
```
vlipovac, line 212: Testing the new slicing functionality of the AdArray

tests/numerics/ad/test_operator_functions.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+from __future__ import annotations
+
+import pytest
+
+import numpy as np
+import porepy as pp
+
+
+def test_ad_function():
```
vlipovac, line 14: Basic test for operator functions.
Since the values of various ad.functions is tested separately, this test covers the general usage of `Function`

tests/numerics/ad/test_operators.py
```diff
@@ -412,34 +418,6 @@ def test_ad_variable_creation():
     assert mvar_1_copy.id == mvar_1.id
     assert mvar_1_deepcopy.id == mvar_1.id
 
-    # Get versions of the variables at previous iteration and time step.
```
vlipovac, line 415: ID tests shifted to testing of prev iter and prev time variables.

tests/numerics/ad/test_operators.py
```diff
@@ -651,6 +629,170 @@ def _compare_ad_objects(a, b):
     assert np.allclose(true_state[ind1], v1_prev.value(eq_system, true_iterate))
 
 
+def test_ad_variable_prev_iter():
```
vlipovac, line 632: Testing of arbitrary prev iter depth.
@keileg  Note that I changed the Variable.id usage: Prev iter and time have for now the same ID since the framework does not fully support changes in dofs.
This requires more thinking and work on the internal data structures.

tests/numerics/ad/test_operators.py
```diff
@@ -651,6 +629,170 @@ def _compare_ad_objects(a, b):
...
+    # Testing index values. Should be increasing with depth, while time step index stays
+    # None (current time)
+    for i in range(1, depth - 1):
+        assert vars_exp[i - 1].iterate_index == i
+        assert vars_exp[i - 1].time_step_index is None
+        assert vars_rec[i - 1].iterate_index == i
+        assert vars_rec[i - 1].time_step_index is None
+
+
+def test_ad_variable_prev_time():
```
vlipovac, line 712: Testing of arbitrary prev time depth.
Indexation logic is different (prev time start with index 0, where as prev iter starts with index 1.
Otherwise the code is quite similar to prev iter test.

src/porepy/numerics/ad/operators.py
```diff
@@ -38,6 +40,62 @@ def _get_shape(mat):
         return mat.shape
 
 
+def _get_previous_time_or_iterate(
```
vlipovac, line 43: An old form of this recursion was hidden inside the scope of some method in `Operator`
Now it is outside, and generalized to handle the abstraction of `IterativeOperator` and `TimeDependentOperator`

src/porepy/numerics/ad/operators.py
```diff
@@ -108,6 +166,28 @@ def __init__(
                 " interfaces, subdomains or boundary grids."
             )
 
+        self.func: Callable[..., float | np.ndarray | AdArray]
```
vlipovac, line 169: I introduced the declaration of this attribute to handle the abstraction in `AbstractFunction`.
This will also come in handy in the future, in case we optimize the evaluation of the operator tree towards a more efficient, functional approach, (E.g., any operation in the AD operator sense can be represented as a function, which is a more general format)

src/porepy/numerics/ad/operators.py
```diff
@@ -181,6 +261,23 @@ def is_leaf(self) -> bool:
         """
         return len(self.children) == 0
 
+    @property
+    def is_original_operator(self) -> bool:
```
vlipovac, line 265: Base class has now a check to see if the operator represents the current time and iter.

src/porepy/numerics/ad/operators.py
```diff
@@ -190,72 +287,25 @@ def set_name(self, name: str) -> None:
         """
         self._name = name
 
-    def previous_timestep(self) -> pp.ad.Operator:
```
vlipovac, line 193: Base class has now a prev iter and time step method which invokes the recursion.

src/porepy/numerics/ad/operators.py
```diff
@@ -275,63 +325,91 @@ def parse(self, mdg: pp.MixedDimensionalGrid) -> Any:
...
+        #    proceed to the non-void operation.
+
+        # Case 1), Some numeric data, or already evaluated operator.
+        if isinstance(op, AdArray | np.ndarray | pp.number):
             return op
 
-        elif op.is_leaf():
-            # Case 2
-            return op.parse(mdg)  # type:ignore
+        # to continue, we must assert it is an actual, unparsed operator
```
vlipovac, line 367: Operator parsing changed. Usage of Variable.parse is introduced to handle arbitrary iter and time step depth.
Also, slicing of the global AdArray is introduced.
Overall more efficient and clearer code.

src/porepy/numerics/ad/operators.py
```diff
@@ -447,24 +525,20 @@ def _parse_operator(self, op: Operator, mdg: pp.MixedDimensionalGrid):
...
-            assert len(results) > 1
-            func_op = results[0]
+            # Operator functions should have at least 1 child (themselves)
+            assert len(results) >= 1, "Operator functions must have at least 1 child."
+            assert hasattr(op, "func"), (
+                f"Operators with operation {operation} must have a functional"
+                + f" representation `func` implemented as a callable member."
+            )
 
-            # if the callable can be fed with AdArrays, do it
```
vlipovac, line 454: The mess of this uneccessary flags is removed.
The `Operation.evaluate` now checks that we are dealing with operator functions and that `func` is available.
Notice that Operator functions do not require themselves as children anymore, which shortens the recursion tree a bit.

src/porepy/numerics/ad/operators.py
```diff
@@ -742,207 +816,31 @@ def _evaluate(
             depends on the operator.
 
         """
-        # Get the mixed-dimensional grid used for the dof-manager.
```
vlipovac, line 745: These internal data structurs are not required anymore with `Variable.parse` being implemented

src/porepy/numerics/ad/operators.py
```diff
@@ -742,207 +816,31 @@ def _evaluate(
...
+            ad_base = state
+
+        # 2. Parse operators. This is left to a separate function to facilitate the
         # necessary recursion for complex operators.
-        eq = self._parse_operator(self, mdg)
+        eq = self._parse_operator(self, system_manager, ad_base)
 
         return eq
 
-    def _identify_variables(
```
vlipovac, line 859: This recursion is eliminated, since not required anymore.

src/porepy/numerics/ad/operators.py
```diff
@@ -1234,15 +1132,232 @@ def _parse_other(self, other):
...
+                "Operator functions must be called before applying any operation."
+            )
+        elif isinstance(other, Operator):
+            # Put Operator at end, because Seconary and Abstract are also operators
+            return [self, other]
         else:
             raise ValueError(f"Cannot parse {other} as an AD operator")
 
 
+class TimeDependentOperator(Operator):
```
vlipovac, line 1155: New intermediate base class for operators with time-dependent behavior.
Implements the pushing back of the time step index in a general manner.

src/porepy/numerics/ad/operators.py
```diff
@@ -1234,15 +1132,232 @@ def _parse_other(self, other):
...
+        else:
+            op.original_operator = self.original_operator
+
+        return op
+
+
+_TimeDependentOperator = TypeVar("_TimeDependentOperator", bound=TimeDependentOperator)
+
+
+class IterativeOperator(Operator):
```
vlipovac, line 1261: Analogous to the abstraction in `TimeDependentOperator`, but for iterative indices.

src/porepy/numerics/ad/operators.py
```diff
@@ -1371,7 +1486,7 @@ def parse(self, mdg: pp.MixedDimensionalGrid) -> np.ndarray:
         return self._values
 
 
-class TimeDependentDenseArray(Operator):
+class TimeDependentDenseArray(TimeDependentOperator):
```
vlipovac, line 1489: Usage of new intermediate base class

src/porepy/numerics/ad/operators.py
```diff
@@ -1546,26 +1647,38 @@ def set_value(self, value: float) -> None:
         self._value = value
 
 
-class Variable(Operator):
+class Variable(TimeDependentOperator, IterativeOperator):
```
vlipovac, line 1650: Usage of both, new, intermediate base classes

src/porepy/numerics/ad/operators.py
```diff
@@ -1750,88 +1802,111 @@ def __repr__(self) -> str:
...
 
     def __init__(self, variables: list[Variable]) -> None:
-        ### PUBLIC
-
-        self.sub_vars = variables
-        """List of sub-variables passed at instantiation, each defined on a separate
-        domain.
 
-        """
+        time_indices = []
```
vlipovac, line 1835: The MD variable has now new assumptions: Atomic variables must not only have the same name, but be at the same instance of time and iterate.

src/porepy/numerics/ad/operators.py
```diff
@@ -1861,99 +1936,35 @@ def copy_common_sub_tags(self) -> None:
...
-    def copy(self) -> "MixedDimensionalVariable":
-        """Copy the mixed-dimensional variable.
-
-        Returns:
-            A shallow copy should be sufficient here; the attributes are not expected to
-            change.
-
-        """
-        return copy.deepcopy(self)
+    def parse(self, mdg: pp.MixedDimensionalGrid) -> Any:
```
vlipovac, line 1947: MD variables are not parsed, but their sub vars

src/porepy/numerics/ad/operators.py
```diff
@@ -1861,99 +1936,35 @@ def copy_common_sub_tags(self) -> None:
...
+            + " Use ``value(equation_system)`` instead."
+        )
 
-    def __repr__(self) -> str:
-        if self._no_variables:
-            return (
-                "Mixed-dimensional variable defined on an empty list of "
-                "subdomains or interfaces."
-            )
+    def previous_timestep(self, steps: int = 1) -> MixedDimensionalVariable:
```
vlipovac, line 1954: Custom prev time and iter behavior for MD variables, invoking the shift for all sub vars after super call

src/porepy/numerics/ad/operators.py
```diff
@@ -38,6 +40,62 @@ def _get_shape(mat):
...
+
+    """
+
+    # The recursion reached an atomic operator, which has some time- or
+    # iterate-dependent behaviour
+    if isinstance(op, TimeDependentOperator) and prev_time:
+        return op.previous_timestep(steps=steps)
+    elif isinstance(op, IterativeOperator) and not prev_time:
+        return op.previous_iteration(steps=steps)
+    # NOTE The previous_iteration of a time-dependent operator will return the operator
```
IvarStefansson, line 75: Why is this? And does it hold even for steps > 1?

src/porepy/numerics/ad/operators.py
```diff
@@ -1750,88 +1798,111 @@ def __repr__(self) -> str:
...
+            time_indices = [-1]
+            iterate_indices = [0]
+            names = ["empty_md_variable"]
 
-        self.id = next(Variable._ids)
-        """ID counter. Used to identify variables during operator parsing."""
+        ### PRIVATE
+        self._id = next(Variable._ids)
+        # NOTE private time step index is -1 if public time step index of atomic
+        # variables is None (current time)
```
IvarStefansson, line 1865: Why not use same (None or +1) in both cases?

tests/numerics/ad/test_operator_functions.py
```diff
@@ -0,0 +1,78 @@
...
+    func = lambda x: x  # identity
+
+    F = pp.ad.Function(func, 'identity')
+
+    g = pp.CartGrid(np.array([3, 2]))
+    mdg = pp.meshing.subdomains_to_mdg([g])
+    mdg.compute_geometry()
+
+    sds = mdg.subdomains()
+    eqs = pp.ad.EquationSystem(mdg)
```
IvarStefansson, line 35: Can we reserve "eqs" for "equations" and use e.g. "eqsys" for an EquationSystem?

IvarStefansson (commented):
This looks extremely useful, thanks! I did a curiosity motivated review, deferring to @keileg or @OmarDuran for the proper review. Please bear with my stupid questions and somewhat random suggestions.

src/porepy/numerics/ad/operators.py
OmarDuran: Great!

src/porepy/numerics/ad/equation_system.py
vlipovac: It seems like there is no test covering it as of now.

src/porepy/numerics/ad/equation_system.py
vlipovac: The output was an array with indices, simply concatenating indices of variables found in the input argument.
So the output was sensitive to the order of the input arguments.
Now, the method loops over the current dof order and computes indices for each variable in the input argument.
The output is always ordered ascending and reflects the DOF order.

src/porepy/numerics/ad/operators.py
vlipovac: Sure.
This name is based on previous code, where the TimeDependentArray class had an attribute `original_operator`
How about `is_current_time` or `is_current_iter`. The latter one probably is more precise, because current iter is always current time.

src/porepy/numerics/ad/operators.py
vlipovac: I should comment at this point, that the `TimeDependentArray` parses itself with a grid-wise order.
So we have the order of equations/dofs implemented in multiple places in porepy. @keileg @IvarStefansson @OmarDuran 
One solution would be, to give the `EquationSystem` as argument to parse, and implement a method in `EquationSystem` returning the row indices per grid.
Then an MD variable could also parse itself, and the row indices would be very useful for e.g. optimizing the sparsity pattern

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -109,59 +109,127 @@ def __repr__(self) -> str:
...
-            The instance itself.
+            If ``*args`` contains only numpy arrays, it returns the result of
+            :meth:`get_values`.
+            If it contains an Ad array, it combines the results of :meth:`get_values`
+            and :meth:`get_jacobian` in an Ad array and returns it.
 
         """
-        return self
+
+        values = self.get_values(*args)
```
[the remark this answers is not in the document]
vlipovac: The split is done for the most general case, where values and derivatives are obtained independently.
For the case where the actual AD is implemented for example in `Function`, there is no split because that class assumes that the callables can handle AdArrays.
The more general cases include the interpolation table, or the secondary operator which will follow after this merger.

src/porepy/numerics/ad/operators.py
```diff
@@ -190,72 +287,25 @@ def set_name(self, name: str) -> None:
...
 
-    def previous_timestep(self) -> pp.ad.Operator:
-        """Return an operator that represents the value of this operator at the previous
-        timestep.
-
-        The operator tree at the previous time step is created as a shallow copy, and will
-        thus be identical to the original operator, except that all time dependent operators
-        are evaluated at the previous time step.
+    def previous_timestep(self, steps: int = 1) -> pp.ad.Operator:
+        """Base method to trigger a recursion over the operator tree and create a
```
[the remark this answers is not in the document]
vlipovac: Because Operator classes inheriting from `TimeDependentOperator` or `IterativeOperator` are overriding this method implementing the shift.
They are leaves in these recursions.

src/porepy/numerics/ad/operators.py
```diff
@@ -1234,15 +1132,228 @@ def _parse_other(self, other):
...
+        but represent a leaf in the recursion tree.
+
+        Note:
+            You cannot create operators at the previous time step from operators which
+            are at some previous iterate. Use the :attr:`original_operator` instead.
+
+        Parameters:
+            steps: ``default=1``
+
+                Number of steps backwards in time.
```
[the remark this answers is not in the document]
vlipovac: I agree it's confusing, but this is how it was implemented so far.
Maybe Operators at current time should have a default time step index 0 (which is never used, because the iterate index is used), and previous times should start with 1

src/porepy/numerics/ad/operators.py
vlipovac: It is steps-independent, yes.
And at this point two things can happen:
The time-dependent operator is original (current time), and it will return itself.
If it is at a previous time, the override in TimeDependentOperator will raise an error, that you cannot take the previous iterate. (But only if the operator inherits from both, time and iterative operator, like the variables do).

src/porepy/numerics/ad/operators.py
vlipovac: I sticked with the convention that the first previous time step is represented by the index 0. Should I change it?

IvarStefansson: But don't you use None for variables and -1 for md variables?

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
IvarStefansson: That could be an option, yes. Opinions, @keileg?

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
IvarStefansson: I see. As a follow-up question: Say I have
```
a: TimeDependentOperator = ...
b: TimeDependentOperator = ...
c: Operator = ... # (not TimeDependentOperator)
d = a * b
e = a + c
```
Are any of d and e `TimeDependentOperator`?

src/porepy/numerics/ad/operators.py
vlipovac: This issue is related to your other comment, https://github.com/pmgbergen/porepy/pull/1166#discussion_r1590249513

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
vlipovac: See also discussion in https://github.com/pmgbergen/porepy/pull/1166#discussion_r1590274116

src/porepy/numerics/ad/operators.py
IvarStefansson: Sounds reasonable (but I'd personally prefer "iteration" or "iterate"). On that, though. What exactly is the benefit of having `TimeDependentOperator` separate from `IterativeOperator`?

vlipovac: Looking for example at the current TImeDependentARRAY.
It has obviously a behavior in time.
But it can also be part of multiple operators at different iterates, always returning the value corresponding to its time index.

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
vlipovac: d and e are created in the arithmetic overloads of `Operator`, therefore they are basic operators with a defined `Operation` and have no time or iterate index per se. One of their children has them though, and this has an effect when the operator is parsed.
d and e do not represent leaves in the operator tree, they are operations in the sense of the word operation.
Where as `a` and `b` are leaves and do not represent an operation, but an term.

src/porepy/numerics/ad/operators.py
IvarStefansson: I see (and thanks for the capitalization of Array! I'm sure I would've been confused without it).

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
IvarStefansson: Thanks, that's what I thought. I just wanted to be sure I hadn't missed something.

tests/numerics/ad/test_operators.py
```diff
@@ -651,6 +629,170 @@ def _compare_ad_objects(a, b):
...
+    # Test only 1 variable, the rest should be covered by other tests
+    mdg, _ = pp.mdg_library.square_with_orthogonal_fractures(
+        "cartesian",
+        {"cell_size": 0.2},
+        fracture_indices=[1],
+    )
+    eqs = pp.ad.EquationSystem(mdg)
+
+    # Integer to test the depth of prev _*, could be a test parameter, but no need
+    depth = 5
```
keileg, line 642: General comment: Do we gain anything from having depth = 5 and not, say, 2 or 3? If not, keeping it lower will somewhat limit the cost of the test. In the same spirit, can we set cell_size to 0.5 and have (on my count) 6 dofs in total instead of 30?

tests/numerics/ad/test_operators.py
```diff
@@ -651,6 +629,170 @@ def _compare_ad_objects(a, b):
...
+    eqs.create_variables(
+        var_name, dof_info={"cells": 1}, subdomains=mdg.subdomains()
+    )
+    var = eqs.md_variable(var_name)
+
+    # Starting point is time step index is None, iterate index is 0
+    # (current time and iter)
+    assert var.time_step_index is None
+    assert var.iterate_index == 0 
+
```
keileg, line 655: Unification of this would be a good thing.

tests/numerics/ad/test_operators.py
keileg: Does this mean we can unify the test if we unify the index scheme?

keileg (commented):
Nice work! Some remarks:
1. Thanks for taking care of the parsing of variables in the EquationSystem. The previous code will not be missed!
2. It seems there is a consensus to let the indexes of time steps and iterates both start at 0. 
3. Overall, I think the cleanup in EquationSystem is to the good, although you are surely right this needs further clarifications, and quite likely refinement of the logic you have introduced.
4. My only major concern is the changes proposed to dofs_of, which (I believe) are contrary to the way I would expect such a method to behave. This is particularly relevant when working with solvers and similar. I suggest we hear from the solver guys before making any decisions there.

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
vlipovac: Alright. Should I implement a raise error to prohibit users storing data at `time_step_index=0`?
I don't think there is a reason to ever use that key for storage, given the usage of `iterate_index` for the current time.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -410,39 +406,47 @@ def __init__(
...
-    def get_values(self, *args: AdArray) -> np.ndarray:
+    def get_values(self, *args: float | np.ndarray | AdArray) -> np.ndarray:
         # stacking argument values vertically for interpolation
-        X = np.vstack([x.val for x in args])
+        args_: list[float | np.ndarray] = []
+        for a in args:
+            if isinstance(a, AdArray):
+                args_.append(a.val)
+            else:
+                args_.append(a)
```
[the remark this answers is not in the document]
vlipovac: I don't use it in the case of flash calculations, and Omar implemented his VTK structure for the OBL approach.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -410,39 +406,47 @@ def __init__(
...
-
-            # add blocks to complete Jacobian
-            jac += partial_jac
+            if isinstance(arg, AdArray):
+                # The trivial Jacobian of one argument gives us the correct position for
+                # the entries as ones
+                partial_jac = arg.jac
+                # replace the ones with actual values
+                # Since csr, we can simply replace the data array with the values of the
+                # derivative
```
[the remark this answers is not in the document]
vlipovac: Yes I think this is indirectly assumed here. I will add respective documentation.

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -484,8 +486,9 @@ def __init__(
         self,
         func: Optional[Callable] = None,
         ad_function_type: Type[AbstractFunction] = Function,
-        operator_kwargs: dict = {},
+        operator_kwargs: dict = {"name": "unnamed_function"},
```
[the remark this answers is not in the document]
vlipovac: Fixed it.

src/porepy/numerics/ad/operators.py
vlipovac: Changed to `is_current_iterate`

src/porepy/numerics/ad/operators.py
```diff
@@ -275,63 +325,91 @@ def parse(self, mdg: pp.MixedDimensionalGrid) -> Any:
         """
         raise NotImplementedError("This type of operator cannot be parsed right away")
 
-    def _parse_operator(self, op: Operator, mdg: pp.MixedDimensionalGrid):
-        """TODO: Currently, there is no prioritization between the operations; for
+    def _parse_operator(
+        self,
+        op: Operator,
+        eqs: pp.ad.EquationSystem,
+        ad_base: AdArray | np.ndarray,
```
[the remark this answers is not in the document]
vlipovac: It is what used to be `self._ad`, which contained either projections of `state` or the global AdArray.
Now it's either `state` or the global ad array with `state` as `ad_base.val`, in global form.

IvarStefansson:
@keileg, regarding your comment 
> 2. It seems there is a consensus to let the indexes of time steps and iterates both start at 0.
I think it's wise to have a second look at all indexes and tags for iterates and time steps, including None and where we start counting (current time/iterate vs first previous time/iterate).

src/porepy/numerics/ad/operators.py
```diff
@@ -742,207 +816,31 @@ def _evaluate(
...
-            # Gather all indices for this variable
-            if len(ind_var) > 0:
-                inds.append(np.hstack([i for i in ind_var]))
-            else:
-                inds.append(np.array([], dtype=int))
-
-        return inds, variable_ids, prev_time, prev_iter
-
-    def _find_subtree_variables(self) -> Sequence[Variable]:
+    def find_variables_in_tree(self) -> Sequence[Variable | MixedDimensionalVariable]:
```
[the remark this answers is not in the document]
vlipovac: My motivation is that any operator should have a way to find it's independent variables, and the *set* of grids on which it is defined (the latter I did not implement for this PR, but the recursion is analogous).
I think this will be required for any kind of optimization in the operator tree.
Also, since porepy represents equations as operators, this is an easy way to find the independent variables (and domains of definition) which are part of the equation.

src/porepy/numerics/ad/operators.py
```diff
@@ -1234,15 +1132,228 @@ def _parse_other(self, other):
...
+        """
+        if isinstance(self, IterativeOperator):
+            if self.prev_iter:
+                raise ValueError(
+                    "Cannot create an operator representing a previous time step,"
+                    + " if it already represents a previous iterate."
+                )
+
+        assert steps > 0, "Number of steps backwards must be strictly positive."
+        # TODO copy or deepcopy? Is this enough for every operator class?
```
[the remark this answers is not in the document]
vlipovac: Since this would require changes in several files (storing discr. matrices with time or iterate indices), I would suggest to put this in a `Note` in the module documentation. Also to notify users.

src/porepy/numerics/ad/operators.py
```diff
@@ -1234,15 +1132,228 @@ def _parse_other(self, other):
...
+    ) -> None:
+        super().__init__(
+            name=name, domains=domains, operation=operation, children=children
+        )
+
+        self.original_operator: Operator
+        """Reference to the operator representing this operator at the current time amd
+        iterate.
+
+        This attribute is only available in operators representing previous time steps.
```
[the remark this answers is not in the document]
vlipovac: We could make this a property returning `self` if there is no private reference to the original operator stored.

src/porepy/numerics/ad/operators.py
```diff
@@ -1234,15 +1132,228 @@ def _parse_other(self, other):
...
+
+        """
+
+        self._iterate_index: int = 0
+        """iterate index, starting with 0 (current iterate at current time) and
+        increasing for previous iterates."""
+
+    @property
+    def prev_iter(self) -> bool:
```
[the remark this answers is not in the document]
vlipovac: Renamed to `is_previous_time` and `is_previous_iterate`

src/porepy/numerics/ad/operators.py
```diff
@@ -1628,9 +1716,24 @@ def __init__(
...
     def domain(self) -> GridLike:
-        """The grid or mortar grid on which this variable is defined."""
+        """The grid or mortar grid on which this variable is defined.
+
+        Note:
+            Not to be confused with :meth:`domains`, which has the grid in a sequence
+            of length 1.
+
+            This is for inheritance reasons, since :class:`Variable` inherits from
+            :class:`Operator`.
```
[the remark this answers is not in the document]
vlipovac: Should I make the clean up for this PR? What do you have in mind?

tests/numerics/ad/test_equation_system.py
vlipovac: Yes it would crash, because it can't find values for `var.previous_iteration()`
But it did not crash before that... My guess there was a logic missing when accessing data with `prev_iter=True` but I could not find that part.

tests/numerics/ad/test_operators.py
vlipovac: You are right, changed it.

tests/numerics/ad/test_operators.py
vlipovac: yes, we can unify the test and make 1 parametrized test, after unifying the index logic.

vlipovac:
I pushed the easy/trivial suggestions and resolved respective conversations (including my self-review).
Open discussions are what's left.

src/porepy/numerics/ad/equation_system.py
jwboth: I have been so far at most going via the private methods of the `DiagnosticsMixin,` like `_variable_data` - very useful and possible worth extracting somewhere or making it somewhat publicly available TBC. See below for an example code snippet:
image

Yuriyzabegaev: I do not rely on the order returned from `dofs_of`. Instead, I store and manipulate the indices of variables outside the model. The `dofs_of` is used only to retreat the indices associated with an individual MD-variable.

vlipovac:
@keileg @IvarStefansson 
Regarding the discussion on unifying the time step indexation with the iterate indexation (time index 0  is current time):
This requires to introduce a new convention.
Since time index 0 is current time and we do not want to store same values twice, I propose making `time_step_index=0` and `iterate_index=0` equivalent throughout the framework.
(By "current time" we always refer to the current iterate, since otherwise it would ambiguous).
This convention must be implemented in `pp.get_solution_values` and `pp.set_solutuon_values`.
Any comments regarding that?

Also, this will introduce minor changes in several files, where `time_step_index=0` was used for previous time,

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
keileg: No, let's just add the reminder.

keileg:
> 2\. I think it's wise to have a second look at all indexes and tags for iterates and time steps, including None and where we start counting (current time/iterate vs first previous time/iterate).

I'll see if I can have a look at some code later today, and will get back to this (soon, I hope).

src/porepy/numerics/ad/equation_system.py
keileg: I suggest the following resolutions:
* The variable id is considered a private field (conversion to _id can be postponed, since work on other PRs should have higher priority). The user is expected to identify variables in terms of its name and domain.
* Internally to the EquationSystem, we are free to organize variable lookup and ordering of dofs as we best see fit. I believe the changes introduced in this PR is this regard are sound, so I'm mostly fine with the changes herein.
* The method dofs_of is the way the user can fetch the global DOFs for a list of variables. The user does not know the order the EqSys has chosen for the variables, thus this method should give the dofs in the order of the input list of variables. That is, *the changes to this specific function should be reverted*.

Does this make sense, @vlipovac?

I will document more clearly the intention for the Variable.id and push a small commit to this branch.

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
keileg: Having thought more and discussed with @IvarStefansson, I suggest the following principles (written in terms of time_step_index, but it should be equally valid, and implemented, for iterate_index or whatever that is called):

1. We define the attribute time_step_index to tell where in the pp.STATE arrays to look for the values of this variable. This is one of two possible, reasonably consistent schemes (the other one is that the solution under construction is assigned index 0 for both time and index, but since that is an unknown state, this is somehow less useful).
2. The time step (iteration) being computed (call it the current one, or the one under construction) does not yet have a solution that can be looked up. To mark this, I actually prefer setting the index to None, since the alternative -1 actually can be used for array indexing, and that is not a nice bug to hunt down.
3. It should be possible to do some safeguarding in the get/set methods to make sure the above principles are followed.

Does this make sense? 

If so, follow-up question/comment to @vlipovac: While it would be great to get this super-clean, we also need to get this PR done, so that you can continue with your more important work on the compositional branch. It is therefore fine with me to be somewhat pragmatic in how polished these concepts are in the context of this PR, although the code should be working and not be more confusing. So, feel free to apply some cost-benefit thinking in this direction, and feel free to ask if you need guidance. Again, does that make sense?

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
keileg: Let's not overcomplicate things, but leave it as it is for now.

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
keileg: Fair enough.

tests/numerics/ad/test_operators.py
keileg: Noted.

tests/numerics/ad/test_operators.py
keileg: My comment regarding pragmatism when polishing code (posted somewhere in operators.py) applies here: Feel free to postpone this unification in the interest of efficiency.

src/porepy/numerics/ad/equation_system.py
keileg: I made a placeholder [issue](https://github.com/pmgbergen/porepy/issues/1167).

src/porepy/numerics/ad/operators.py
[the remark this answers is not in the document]
vlipovac: @keileg  I think I can spend 1-2 hours in getting the time_step_index right, but I was not able to understand your conclusion from various comments.

In your last comment, you say unification is the goal, here you say you prefer `time_step_index` to stay None for the solution under construction?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1029,37 +997,46 @@ def projection_to(self, variables: Optional[VariableList] = None) -> sps.csr_mat
...
-            var_number = self._variable_numbers[variable]
-            var_indices = np.arange(
-                global_variable_dofs[var_number],
-                global_variable_dofs[var_number + 1],
-                dtype=int,
-            )
-            indices.append(var_indices)
+        indices: list[np.ndarray] = []
+
+        for var in variables:
```
vlipovac, line 1018: @keileg I reverted the changes for `dofs_of`. Order corresponds now again to input.
Validation was added that variable is known to equation system.

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -294,6 +294,60 @@ def discretize_from_list(
...
+    if time_step_index is not None:
+        # Some previous time.
+        if time_step_index > 0:
+            out.append((pp.TIME_STEP_SOLUTIONS, time_step_index))
+        # Current time. NOTE this is ambigous since the current time is an unknown and
+        # has multiple iterate values.
+        # Alternatively, we could associate time_step_index = 0 with iterate_index = 0
+        # the below elif branch introduces the convention that
+        # time step = iterate step = 0 are equivalent.
+        elif time_step_index == 0:
```
vlipovac, line 334: @keileg I unified the time step indexation with the iterate indexation.
0 is now the current time and every time-dependent operator starts with that value.
Though current time (unknown value) is ambiguous when referring to it using time step indices.
I added respective value error to alert the user to use iterate_index instead.

src/porepy/numerics/ad/equation_system.py
vlipovac: I turned the ID into a property, with a private `_id`.
Users have access to the public `id`, without being able to modify it.

src/porepy/numerics/ad/_ad_utils.py
keileg: I agree, this is a fair state to leave it for now.

keileg (approved):
Thanks for the effort, @vlipovac

PR #1169, opened 2024-05-13
Limit-cases-of-ad-functions

IvarStefansson:
## Proposed changes

This PR introduces two minor modifications of the ad.functions:
1. Uses first argument in the case of equality of the arguments of the max function. This impact the tangential contact mechanics and the aperture calculation.
2. Assigns unitary jacobian values for l2_norm of an argument with val=0. Thus, derivative information is nonzero and we avoid special cases where a block of the linear system vanishes.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/ad/functions.py
+++ b/src/porepy/numerics/ad/functions.py
@@ -91,9 +91,10 @@ def l2_norm(dim: int, var: pp.ad.AdArray) -> pp.ad.AdArray:
     """L2 norm of a vector variable.
 
     For the example of dim=3 components and n vectors, the ordering is assumed
-    to be ``[u0, v0, w0, u1, v1, w1, ..., un, vn, wn]``
+    to be ``[u0, v0, w0, u1, v1, w1, ..., un, vn, wn]``.
 
-    Vectors satisfying ui=vi=wi=0 are assigned zero entries in the jacobi matrix
+    Vectors satisfying ui=vi=wi=0 are assigned positive entries in the Jacobi
+    matrix.
 
     Note:
         See module level documentation on how to wrap functions like this in ad.Function.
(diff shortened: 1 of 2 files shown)


keileg:
@mariusnevland will have a look at the impact of the suggested changes on some of his setups.

keileg:
@mariusnevland: Will you have time to have a look at this in the nearest future, or should we proceed without your input?

mariusnevland:
> @mariusnevland: Will you have time to have a look at this in the nearest future, or should we proceed without your input?

I have taken a quick look at it already. I ran the same simulations that I presented on the previous MaPSI meeting (where the dilation angle and the angle between two intersecting fractures are varied), and the performance was generally much better with the changes in this PR; the Newton solver converged for all combinations. I have not tested for any other setups though.

I will be quite busy the next couple of weeks, so I might not be able to look further into it for now.

One comment: I am assuming that the change in the max-function is primarily felt by the `b_p` variable in the tangential equation? I believe that, from a theoretical standpoint, the max-function in `b_p` is redundant, because the distinction between open and closed cells is already being handled by the characteristic function. In other words, changing the definition to `b_p=self.friction_bound(subdomains)` should give the same solution. However, I can confirm through some experimentation that this change will affect the behavior of the solver (not sure yet if in a good or bad way).

IvarStefansson:
@jwboth and I have revisited the usage and confirmed that the changes make sense from a "heuristic theoretical" point of view. We will keep probing the implementation a bit (approx. a week?) before merging. Marking as draft in the meantime.

src/porepy/numerics/ad/functions.py
```diff
@@ -375,7 +381,7 @@ def maximum(var_0: FloatType, var_1: FloatType) -> FloatType:
     # the case.
     assert isinstance(vals[0], np.ndarray) and isinstance(vals[1], np.ndarray)
     # Maximum of the two arrays
-    inds = (vals[1] >= vals[0]).nonzero()[0]
+    inds = (vals[1] > vals[0]).nonzero()[0]
```
IvarStefansson, line 384: Note to self: Run skipped tests before merging.

IvarStefansson:
@jwboth, have you had a chance to test this?

jwboth:
> @jwboth, have you had a chance to test this?

I integrated the branch in my recent numerical studies. But I am not really able to conclude. I did not really do a proper study, comparing both side by side for many cases. I only quickly compared the variants on this branch and develop for one single test case. And I could not see a real difference in the performance. 

It will be a bit difficult to do a proper study this week. If you do not want to wait, I still feel like the proposed changes are meaningful, and would be OK with integrating them. Otherwise, I make a proper study in the beginning of next week.

IvarStefansson:
> Quote reply



> > @jwboth, have you had a chance to test this?
> 
> I integrated the branch in my recent numerical studies. But I am not really able to conclude. I did not really do a proper study, comparing both side by side for many cases. I only quickly compared the variants on this branch and develop for one single test case. And I could not see a real difference in the performance.
> 
> It will be a bit difficult to do a proper study this week. If you do not want to wait, I still feel like the proposed changes are meaningful, and would be OK with integrating them. Otherwise, I make a proper study in the beginning of next week.

Any updates?

keileg (approved):

PR #1170, opened 2024-05-14
Introducing Operator class wrapping externally computed values and derivatives (CF Merger # 1)

vlipovac:
## Proposed changes

This is merger # 1 in a series of mergers regarding the multiphase, multicomponent extension of PorePy.

It introduces a new Operator representing data which are externally calculated and provided by the user.
This secondary operator is produced by a factory class, which defines the dependencies of the expression and provides management functionalities to update values and derivative values in both time and iterate sense, on all grid types.
It also supports expressions with arbitrary DOFs (cells, nodes, faces), potentially laying the ground work to express any kind of operator or discretization as a secondary quantity, whose numerical representation is obtained (more efficiently) elsewhere.

## Types of changes

- [X] Minor change (e.g., dependency bumps, broken links).
- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/ad/_ad_utils.py
+++ b/src/porepy/numerics/ad/_ad_utils.py
@@ -456,8 +456,6 @@ def shift_solution_values(
 
     The shift is implemented s.t. values at index ``i`` are copied to index ``i + 1``,
     based on the number of stored values.
-    I.e., shifts do not occur if nothing is stored at an index ``i``, or only values
-    for 1 index are stored.
 
     Note:
         The data stored must have support for ``.copy()`` in order to avoid faulty
@@ -465,9 +463,9 @@ def shift_solution_values(
 
     Parameters:
         name: Key in ``data`` for which quantity the shift should be performed.
-        value: The new value to be stored at index 0 at given ``location``.
         data: A grid data dictionary.
-        location: Either ``pp.TIME_STEP_SOLUTIONS`` or ``pp.ITERATE_SOLUTIONS``
+        location: Either :data:`~porepy.utils.common_constants.TIME_STEP_SOLUTIONS`
+            or :data:`~porepy.utils.common_constants.ITERATE_SOLUTIONS`.
 
     Raises:
         ValueError: If unsupported ``location`` is passed.
@@ -476,19 +474,21 @@ def shift_solution_values(
     if location not in data:
         data[location] = {}
     if name not in data[location]:
+        # TODO should we return here instead? Nothing to be shifted, and
+        # set_solution_values is the method which prepares the data dict
         data[location][name] = {}
 
     num_stored = len(data[location][name])
 
     if location == pp.ITERATE_SOLUTIONS:
-        first_index = 0
+        range_ = range(num_stored, 0, -1)
     # previous time step values start with index 1.
     elif location == pp.TIME_STEP_SOLUTIONS:
-        first_index = 1
+        range_ = range(num_stored + 1, 1, -1)
     else:
         raise ValueError(f"Shifting values not implemented for location {location}")
 
-    for i in range(num_stored, first_index, -1):
+    for i in range_:
         data[location][name][i] = data[location][name][i - 1].copy()
 
 
--- a/src/porepy/numerics/ad/_ad_utils.py
+++ b/src/porepy/numerics/ad/_ad_utils.py
@@ -312,20 +312,38 @@ def _validate_indices(
 
     out = []
 
-    if time_step_index is not None:
-        if time_step_index < 0:
-            raise ValueError(
-                "Use increasing, non-negative integers for previous time step values."
-            )
-        else:
-            out.append((pp.TIME_STEP_SOLUTIONS, time_step_index))
     if iterate_index is not None:
-        if iterate_index < 0:
+        # Some iterate values of the current time
+        if iterate_index >= 0:
+            out.append((pp.ITERATE_SOLUTIONS, iterate_index))
+        # Negative iterate indices are not supported
+        else:
             raise ValueError(
                 "Use increasing, non-negative integers for (previous) iterate values."
             )
+
+    if time_step_index is not None:
+        # Some previous time.
+        if time_step_index > 0:
+            out.append((pp.TIME_STEP_SOLUTIONS, time_step_index))
+        # Current time. NOTE this is ambigous since the current time is an unknown and
+        # has multiple iterate values.
+        # Alternatively, we could associate time_step_index = 0 with iterate_index = 0
+        # the below elif branch introduces the convention that
+        # time step = iterate step = 0 are equivalent.
+        elif time_step_index == 0:
+            # if (pp.ITERATE_SOLUTIONS, 0) not in out:
+            #     out.append((pp.ITERATE_SOLUTIONS, 0))
+            raise ValueError(
+                "Using time_step_index = 0 (current time) is ambiguous."
+                + " Specify iterate_index instead."
+                + " First previous time step value is time_step_index = 1."
+            )
+        # Negative time step indices are not supported
         else:
-            out.append((pp.ITERATE_SOLUTIONS, iterate_index))
+            raise ValueError(
+                "Use increasing, non-negative integers for previous time step values."
+            )
 
     return out
 
--- a/src/porepy/numerics/ad/_ad_utils.py
+++ b/src/porepy/numerics/ad/_ad_utils.py
@@ -471,12 +471,12 @@ def shift_solution_values(
         ValueError: If unsupported ``location`` is passed.
 
     """
+    # NOTE return because nothing to be shifted. Avoid confusion by introducing data
+    # dictionaries for values which were never set using pp.set_solution_values
     if location not in data:
-        data[location] = {}
+        return
     if name not in data[location]:
-        # TODO should we return here instead? Nothing to be shifted, and
-        # set_solution_values is the method which prepares the data dict
-        data[location][name] = {}
+        return
 
     num_stored = len(data[location][name])
 
--- a/src/porepy/numerics/ad/_ad_utils.py
+++ b/src/porepy/numerics/ad/_ad_utils.py
@@ -447,6 +447,51 @@ def get_solution_values(
     return value
 
 
+def shift_solution_values(
+    name: str,
+    data: dict,
+    location: Any,
+) -> None:
+    """Function to shift numerical values stored in the data dictionary.
+
+    The shift is implemented s.t. values at index ``i`` are copied to index ``i + 1``,
+    based on the number of stored values.
+    I.e., shifts do not occur if nothing is stored at an index ``i``, or only values
+    for 1 index are stored.
+
+    Note:
+        The data stored must have support for ``.copy()`` in order to avoid faulty
+        referencing (e.g., numpy arrays or sparse matrices).
+
+    Parameters:
+        name: Key in ``data`` for which quantity the shift should be performed.
+        value: The new value to be stored at index 0 at given ``location``.
+        data: A grid data dictionary.
+        location: Either ``pp.TIME_STEP_SOLUTIONS`` or ``pp.ITERATE_SOLUTIONS``
+
+    Raises:
+        ValueError: If unsupported ``location`` is passed.
+
+    """
+    if location not in data:
+        data[location] = {}
+    if name not in data[location]:
+        data[location][name] = {}
+
+    num_stored = len(data[location][name])
+
+    if location == pp.ITERATE_SOLUTIONS:
+        first_index = 0
+    # previous time step values start with index 1.
+    elif location == pp.TIME_STEP_SOLUTIONS:
+        first_index = 1
+    else:
+        raise ValueError(f"Shifting values not implemented for location {location}")
+
+    for i in range(num_stored, first_index, -1):
+        data[location][name][i] = data[location][name][i - 1].copy()
+
+
 class MergedOperator(operators.Operator):
     """Representation of specific discretization fields for an Ad discretization.
 
--- a/src/porepy/numerics/ad/_ad_utils.py
+++ b/src/porepy/numerics/ad/_ad_utils.py
@@ -313,36 +313,23 @@ def _validate_indices(
     out = []
 
     if iterate_index is not None:
-        # Some iterate values of the current time
+        # Some valid iterate value.
         if iterate_index >= 0:
             out.append((pp.ITERATE_SOLUTIONS, iterate_index))
         # Negative iterate indices are not supported
         else:
             raise ValueError(
-                "Use increasing, non-negative integers for (previous) iterate values."
+                "Use increasing, non-negative integers for iterate indices."
             )
 
     if time_step_index is not None:
         # Some previous time.
-        if time_step_index > 0:
+        if time_step_index >= 0:
             out.append((pp.TIME_STEP_SOLUTIONS, time_step_index))
-        # Current time. NOTE this is ambigous since the current time is an unknown and
-        # has multiple iterate values.
-        # Alternatively, we could associate time_step_index = 0 with iterate_index = 0
-        # the below elif branch introduces the convention that
-        # time step = iterate step = 0 are equivalent.
-        elif time_step_index == 0:
-            # if (pp.ITERATE_SOLUTIONS, 0) not in out:
-            #     out.append((pp.ITERATE_SOLUTIONS, 0))
-            raise ValueError(
-                "Using time_step_index = 0 (current time) is ambiguous."
-                + " Specify iterate_index instead."
-                + " First previous time step value is time_step_index = 1."
-            )
         # Negative time step indices are not supported
         else:
             raise ValueError(
-                "Use increasing, non-negative integers for previous time step values."
+                "Use increasing, non-negative integers for time step indices."
             )
 
     return out
@@ -363,16 +350,20 @@ def set_solution_values(
         name: Name of the quantity that is to be assigned values.
         values: The values that are set in the data dictionary.
         data: Data dictionary corresponding to the subdomain or interface in question.
-        time_step_index (optional): Determines the key of where `values` are to be
-            stored in `data[pp.TIME_STEP_SOLUTIONS][name]`. 0 means it is the most
-            recent set of values, 1 means the one time step back in time, 2 is two time
-            steps back, and so on.
-        iterate_index (optional): Determines the key of where `values` are to be
-            stored in `data[pp.ITERATE_SOLUTIONS][name]`. 0 means it is the most
-            recent set of values, 1 means the values one iteration back in time, 2 is
-            two iterations back, and so on.
-        additive: Flag to decide whether the values already stored in the data
-            dictionary should be added to or overwritten.
+        time_step_index: ``default=None``
+
+            Determines the key of where ``values`` are to be stored in
+            ``data[pp.TIME_STEP_SOLUTIONS][name]``.
+            0 is the most recent time step, 1 the one before that and so on.
+        iterate_index: ``default=None``
+
+            Determines the key of where ``values`` are to be stored in
+            ``data[pp.ITERATE_SOLUTIONS][name]``.
+            0 is the current iterate, 1 the previous iterate, and so on.
+        additive: ``default=False``
+
+            Flag to decide whether the values already stored in the data dictionary
+            should be added to or overwritten.
 
     Raises:
         ValueError: In the case of inconsistent usage of indices
@@ -409,21 +400,29 @@ def get_solution_values(
     """Function for fetching values stored in the data dictionary, for some
     time-dependent or iterative term.
 
+    Note:
+        Compared to :func:`set_solution_values` the getter works only for 1 defined
+        index, whereas the setter can take both a time and iterate index.
+
     Parameters:
         name: Name of the parameter whose values we are interested in.
         data: The data dictionary.
-        time_step_index: Which time step we want to get values for. 0 is current, 1 is
-            one time step back in time. This is only limited by how many time steps are
-            stored from before.
-        iterate_index: Which iterate we want to get values for. 0 is current, 1 is one
-            iterate back in time. This is only limited by how many iterates are stored
-            from before.
+        time_step_index: ``default=None``
+
+            Determines the key of where ``values`` are to be stored in
+            ``data[pp.TIME_STEP_SOLUTIONS][name]``.
+            0 is the most recent time step, 1 the one before that and so on.
+        iterate_index: ``default=None``
+
+            Determines the key of where ``values`` are to be stored in
+            ``data[pp.ITERATE_SOLUTIONS][name]``.
+            0 is the current iterate, 1 the previous iterate, and so on.
 
     Raises:
         ValueError: In the case of inconsistent usage of indices
             (both None or negative values).
-        AssertionError: If the user attempts to get iterate and time step values
-            simultanously. Only 1 index is permitted in getter
+        ValueError: If the user attempts to get multiple iterate and time step values
+            simultanously. Only 1 index is permitted in the getter.
         KeyError: If no values are stored for the passed index.
 
     Returns:
@@ -431,9 +430,10 @@ def get_solution_values(
 
     """
     loc_index = _validate_indices(time_step_index, iterate_index)
-    assert (
-        len(loc_index) == 1
-    ), "Cannot get value from both iterate and time step at once. Call separately."
+    if len(loc_index) != 1:
+        raise ValueError(
+            "Cannot get value from both iterate and time step at once. Call separately."
+        )
 
     loc, index = loc_index[0]
 
@@ -451,26 +451,41 @@ def shift_solution_values(
     name: str,
     data: dict,
     location: Any,
+    max_index: Optional[int] = None,
 ) -> None:
     """Function to shift numerical values stored in the data dictionary.
 
-    The shift is implemented s.t. values at index ``i`` are copied to index ``i + 1``,
-    based on the number of stored values.
+    The shift is implemented s.t. values at index ``i`` are copied to index ``i + 1``.
 
     Note:
         The data stored must have support for ``.copy()`` in order to avoid faulty
         referencing (e.g., numpy arrays or sparse matrices).
+
+    Note:
+        After this operation, values at index 0 and 1 will be the same.
+        Use :meth:`set_solution_values` to update the latest values at index 0.
 
     Parameters:
         name: Key in ``data`` for which quantity the shift should be performed.
         data: A grid data dictionary.
         location: Either :data:`~porepy.utils.common_constants.TIME_STEP_SOLUTIONS`
             or :data:`~porepy.utils.common_constants.ITERATE_SOLUTIONS`.
+        max_index: ``default=None``
+
+            A non-negative integer, capping the range of the shift operation to
+            ``i -> max_index``.
+            If called repeatedly with ``None``, the depth in ``location`` keeps
+            increasing. To be used in schemes with a defined maximal depth of stored
+            iterate or time step values.
 
     Raises:
         ValueError: If unsupported ``location`` is passed.
+        ValueError: if ``max_index`` is negative.
 
     """
+    if location not in [pp.ITERATE_SOLUTIONS, pp.TIME_STEP_SOLUTIONS]:
+        raise ValueError(f"Shifting values not implemented for location {location}")
+
     # NOTE return because nothing to be shifted. Avoid confusion by introducing data
     # dictionaries for values which were never set using pp.set_solution_values
     if location not in data:
@@ -480,15 +495,13 @@ def shift_solution_values(
 
     num_stored = len(data[location][name])
 
-    if location == pp.ITERATE_SOLUTIONS:
-        range_ = range(num_stored, 0, -1)
-    # previous time step values start with index 1.
-    elif location == pp.TIME_STEP_SOLUTIONS:
-        range_ = range(num_stored + 1, 1, -1)
-    else:
-        raise ValueError(f"Shifting values not implemented for location {location}")
+    if max_index is not None:
+        if max_index < 0:
+            raise ValueError("Maximal index must be non-negative.")
 
-    for i in range_:
+        num_stored = int(np.min([num_stored, max_index]))
+
+    for i in range(num_stored, 0, -1):
         data[location][name][i] = data[location][name][i - 1].copy()
 
 
--- a/src/porepy/numerics/ad/_ad_utils.py
+++ b/src/porepy/numerics/ad/_ad_utils.py
@@ -28,7 +28,7 @@ Classes:
 from __future__ import annotations
 
 from abc import ABCMeta
-from typing import Optional
+from typing import Any, Optional
 
 import numpy as np
 import scipy.sparse as sps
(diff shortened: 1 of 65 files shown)


src/porepy/numerics/ad/equation_system.py
```diff
@@ -722,55 +723,10 @@ def shift_iterate_values(
                 be shifted.
 
         """
-        self._shift_variable_values(location=pp.ITERATE_SOLUTIONS, variables=variables)
```
vlipovac, line 725: This functionality is generalized and moved to _ad_utils

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -447,6 +447,51 @@ def get_solution_values(
...
+
+    if location == pp.ITERATE_SOLUTIONS:
+        first_index = 0
+    # previous time step values start with index 1.
+    elif location == pp.TIME_STEP_SOLUTIONS:
+        first_index = 1
+    else:
+        raise ValueError(f"Shifting values not implemented for location {location}")
+
+    for i in range(num_stored, first_index, -1):
```
vlipovac, line 491: @keileg  in the original function in EquationSystem, the range for the case `pp.ITERATE_SOLUTIONS` was `range(num_stored - 1, 0, -1)`
I don't know if I am introducing a breaking change or if this is a fix (no test coverage), but the original implementation had the (I think unwanted) consequence, that iterate values where not shifted, if no previous iterate was set (index > 0) (so values at index 0 were never shifted to index 1)

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -447,6 +447,51 @@ def get_solution_values(
...
+        data[location] = {}
+    if name not in data[location]:
+        # TODO should we return here instead? Nothing to be shifted, and
+        # set_solution_values is the method which prepares the data dict
+        data[location][name] = {}
+
+    num_stored = len(data[location][name])
+
+    if location == pp.ITERATE_SOLUTIONS:
+        range_ = range(num_stored, 0, -1)
```
vlipovac, line 484: @keileg in the original function in EquationSystem, the range for the case `pp.ITERATE_SOLUTIONS` was `range(num_stored - 1, 0, -1)`.
I don't know if I am introducing a breaking change or if this is a fix (no test coverage), but the original implementation had the (I think unwanted) consequence, that iterate values where not shifted, if no previous iterate was set (index > 0) (so values at index 0 were never shifted to index 1)

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+                # At this point we assume that the derivative has only an identity block
+                # Shape checks for derivatives are done in the factory class
+                jacs.append(
+                    sps.csr_matrix((d_i, arg.jac.nonzero()), shape=arg.jac.shape)
+                )
+
+        return sum(jacs).tocsr()
+
+
+class SecondaryExpression:
```
keileg, line 288: Is it correct that this is assumed used as a mixin for a model class, where (due to the `__call__` method) will function similar to any standard method defining a constitutive expression? At least the part about the mixin should be explicitly documented (or have I missed it?), as this seems to be an implicit assumption on usage.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+    ) -> np.ndarray:
+        """Function fetching the data stored for this secondary expression, represented
+        in Ad form by ``op``.
+
+        ``op`` has a time step or iterate index, which specifies which data should be
+        fetched.
+
+        This function is assigned to :attr:`SecondaryOperator.fetch_data`
+        (a work-around such that the Ad operator has no reference to the md-grid like
+        the other operators).
```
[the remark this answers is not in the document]
keileg: Having read more code, I can guess what the answer will be, but I still think it will be useful to discuss this briefly, and perhaps document the answer as an implementation note for future reference.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+        else:
+            raise TypeError(f"Unsupported type of grid {type(grid)}.")
+
+    @property
+    def name(self) -> str:
+        """Name of this expression assigned at instantiation."""
+        return self._name
+
+    @property
+    def _name_diffs(self) -> str:
```
keileg, line 569: `_name_derivatives`? I found `_name_diffs` very confusing for some reason.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+            (iterate index = 0). The shape is consistent with ``val``.
+
+        """
+        vals = []
+        for _, data in self.mdg.boundaries(return_data=True):
+            vals.append(pp.get_solution_values(self.name, data, iterate_index=0))
+        if len(vals) > 0:
+            return np.hstack(vals)
+        else:
+            return np.zeros(0, dtype=float)
```
keileg, line 612: Will this give an array of dimension 0 or 1, or a numpy scalar? I very much prefer 1d-arrays, and would really like to avoid numpy scalars, as these bring all kind of corner cases to the parsing (don't ask)

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+            The current iterate derivative values of same shape as ``val``.
+
+        """
+        vals = []
+        for _, data in self.mdg.subdomains(return_data=True):
+            vals.append(pp.get_solution_values(self._name_diffs, data, iterate_index=0))
+        if len(vals) > 0:
+            return np.hstack(vals)
+        else:
+            return np.zeros(self.num_dependencies, dtype=float)
```
keileg, line 691: What happens here? What would the else correspond to?

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -447,6 +447,51 @@ def get_solution_values(
...
+            or :data:`~porepy.utils.common_constants.ITERATE_SOLUTIONS`.
+
+    Raises:
+        ValueError: If unsupported ``location`` is passed.
+
+    """
+    if location not in data:
+        data[location] = {}
+    if name not in data[location]:
+        # TODO should we return here instead? Nothing to be shifted, and
```
keileg, line 477: I think I agree with not doing anything here. Such creation of dictionary items by stealth can be very confusing under debugging.

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -447,6 +447,51 @@ def get_solution_values(
...
+    location: Any,
+) -> None:
+    """Function to shift numerical values stored in the data dictionary.
+
+    The shift is implemented s.t. values at index ``i`` are copied to index ``i + 1``,
+    based on the number of stored values.
+
+    Note:
+        The data stored must have support for ``.copy()`` in order to avoid faulty
+        referencing (e.g., numpy arrays or sparse matrices).
```
keileg, line 462: This is probably the safer option, and certainly consistent with existing code, so fine with me. I'm not sure we have the right data model, using copies all over the place, but that is a different matter.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+
+        The update process of boundary values is different from the process on
+        subdomains and interfaces.
+
+        1. Boundary values have only a single iterate value (current time step)
+        2. Boundary values can have multiple previous time steps, if expression is
+           time-dependent.
+
+        Therefore, values for the current time can always be set, but values will be
+        shifted backwards in time only if the quantity is time-dependent.
```
keileg, line 809: How far back will it be shifted? From my reading of `shift_solution_values`, it will keep the length of the list (or dictionary). Does this mean that the user is expected to initialize the list of boundary values with the correct number of items? If so, this should be documented, should it not? The same applies (I believe) to subdomain and interfaces, both iterates and time steps.

Let me know if you want to discuss this.

keileg (commented):
@vlipovac: I have added quite a few comments and questions. Let me know if you want to discuss anything.

I need to look more carefully at the test, but I don't have the energy now, and may not have time tomorrow.

tests/numerics/ad/test_secondary_operator.py
```diff
@@ -0,0 +1,300 @@
...
+    assert sop_pt.iterate_index == 0
+    assert sop_pt.time_step_index == 1
+    assert all(v.iterate_index == 0 for v in sop_pt.children)
+    assert all(v.time_step_index == 1 for v in sop_pt.children)
+
+    # check that arithmetic overloads work for secondary operator, even though it
+    # inherits from AbstractFunction (correct MOR)
+    a = sop + vars[0]
+    try:
+        for a in ["+", "-", "*", "@", "**"]:
```
keileg, line 155: If the `SecondaryOperator` is a genuinely new type of `Operator` (and we need to discuss this), I would prefer that correct parsing of arithmetic operations are included in the [general test for such methods](https://github.com/pmgbergen/porepy/blob/0e94301536d6640cea9ae91816d368d8efa90d75/tests/numerics/ad/test_operators.py#L1896). This would require a careful definition of which combinations of arithmetic operations and operands can be permitted for a `SecondaryOperator`, but that will be required for any new `Operator.`

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+        the other operators).
+
+        Note:
+            This method uses PorePy's utility functions to get data stored in a grid's
+            data dictionary. Respective errors will be raised if no data stored.
+
+        Parameters:
+            op: An AD representation of this expression at a specific time and iterate.
+            grid: A grid in the md-grid for which the data should be fetched.
+            get_derivatives: If True, it fetches the derivative values, otherwise it
```
keileg, line 481: Should we explicitly raise an error if the derivative is requested for an operator which is not at the current time step? I see from the test that this will result in a failed lookup in the data dictionary, hence a `KeyError` will be raised, but that will not be very helpful for a user.

tests/numerics/ad/test_secondary_operator.py
```diff
@@ -0,0 +1,300 @@
...
+        assert np.all(sop_pt.value(eqsys) == expr.subdomain_values)
+    # No derivatives progressed in time
+    for g in domains:
+        with pytest.raises(KeyError):
+            expr.fetch_data(sop_pt, g, True)
+
+    # previous iterate and time step have no derivative values, but the parsing should
+    # still work because the dependencies at previous iter and time are parsed as
+    # numpy arrays: The function of the SOP (AbstractFunction) should not call
+    # get_jacobian of the SOP
```
keileg, line 242: This implies that `SecondaryExpression.fetch_data(..., True)` has behavior different from `SecondaryExpression.value_and_jacobian`. I understand why this is done, but is it wise to have two public methods for evaluation? An alternative would be to make `fetch_data` private (this would break with one standard for private methods, in that it will be called by other classes, but it would comply with the notion that private methods are part of the backend, not the API), another possibility is to overload `value` and `value_and_jacobian`. This brings us back to the question of where parsing should be located.

src/porepy/numerics/ad/secondary_operator.py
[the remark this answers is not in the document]
keileg: Having thought even more: Is this really a `_parse` method, like the other operators have?

keileg (commented):
@vlipovac Having thought a bit more, I think there is one rather fundamental questions we should discuss before diving further into the details: The `SecondaryOperator` is suggested as a new type of `Operator`, but it is also closely related to the existing hierarchy of `pp.ad.Function`. What is the relation, why is a new class needed etc.? Also note that, if we introduce a new fundamental type, we should clearly define which types of arithmetic operations it can be involved and with which other operands it can be used (and this must be tested).

tests/numerics/ad/test_secondary_operator.py
```diff
@@ -0,0 +1,300 @@
...
+    assert sop_pi.time_step_index == 0
+    assert all(v.iterate_index == 1 for v in sop_pi.children)
+    assert all(v.time_step_index == 0 for v in sop_pi.children)
+    assert sop_pt.iterate_index == 0
+    assert sop_pt.time_step_index == 1
+    assert all(v.iterate_index == 0 for v in sop_pt.children)
+    assert all(v.time_step_index == 1 for v in sop_pt.children)
+
+    # check that arithmetic overloads work for secondary operator, even though it
+    # inherits from AbstractFunction (correct MOR)
```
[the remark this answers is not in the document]
vlipovac: No it doesn't, this is an artifact. I will remove it

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -447,6 +447,51 @@ def get_solution_values(
     return value
 
 
+def shift_solution_values(
+    name: str,
+    data: dict,
+    location: Any,
```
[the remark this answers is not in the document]
vlipovac: I tried, but I got a warning from some of the static checks that Literal accepts only strings explicitly, and not implicitly through a variable `pp.TIME_STEP_SOLUTIONS` for example.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+
+    def __call__(self, *args: Operator) -> Operator:
+        """By calling this operator with new children, the user can change the
+        structure of the Jacobian.
+
+        The dependencies/children are replaced and hence the non-trivial entries in
+        the Jacobian in the md-setting.
+
+        Note:
+            Call this operator only if you know what you are doing.
```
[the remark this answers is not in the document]
vlipovac: I have 2 functionalities in mind:

The user can restrict the represented expression to a subset of grids, without going through the machinery of the factory class, and without messing with time and iterate index.
The user can dynamically change the dependencies (and hence the structure of the Jacobian), but this is probably insane and I can't think of a use case.

Since it also is a kind of artifact from the original intention to inherit from abstract function, I think the call-overload can be removed.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+            This class cannot inherit from
+            :class:`~porepy.numerics.ad.operator_functions.AbstractFunction` because
+            that parent would block arithmetic overloads.
+            But it uses the operation ``evaluate`` and the ``AbstractFunction.func``
+            with ``self`` as an explicit argument.
+
+        """
+        # mypy complains that self is not an instance of AbstractFunction because of the
+        # hack
+        return pp.ad.AbstractFunction.func(self, *args)  # type:ignore[arg-type]
```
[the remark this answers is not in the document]
vlipovac: Do you mean the return type of `def func(self, *args: FloatType) -> float | np.ndarray | AdArray:`?

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+        Note:
+            This class cannot inherit from
+            :class:`~porepy.numerics.ad.operator_functions.AbstractFunction` because
+            that parent would block arithmetic overloads.
+            But it uses the operation ``evaluate`` and the ``AbstractFunction.func``
+            with ``self`` as an explicit argument.
+
+        """
+        # mypy complains that self is not an instance of AbstractFunction because of the
+        # hack
```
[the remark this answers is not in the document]
vlipovac: I would require here the same code as implemented in `AbstractFunction` to call `get_values` and `get_jacobian`, this "hack" avoids code duplicates.
It is a "hack" because it does not inherit from `AbstractFunction` because of the blocking of arithmetic overloads.

I could resolve this by refactoring the `AbstractFuncion` to be a factory as intended, and not an operator.

src/porepy/numerics/ad/_ad_utils.py
vlipovac: Changed to return if no location or name in data dictionary found.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+        op = SecondaryOperator(
+            name=self.name,
+            domains=self.domains,  # type:ignore[arg-type]
+            children=args,  # type:ignore[arg-type]
+        )
+        op.fetch_data = self.fetch_data
+        return op
+
+    def previous_timestep(self, steps: int = 1) -> SecondaryOperator:
+        """Secondary operators have children which also need to be obtained at
```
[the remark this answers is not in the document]
vlipovac: Documentation updated.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+            that parent would block arithmetic overloads.
+            But it uses the operation ``evaluate`` and the ``AbstractFunction.func``
+            with ``self`` as an explicit argument.
+
+        """
+        # mypy complains that self is not an instance of AbstractFunction because of the
+        # hack
+        return pp.ad.AbstractFunction.func(self, *args)  # type:ignore[arg-type]
+
+    def get_values(self, *args: float | np.ndarray | AdArray) -> np.ndarray:
```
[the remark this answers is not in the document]
vlipovac: Discussed in person, documentation added.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+        """Fetches the values stored for this secondary operator at its time or iterate
+        index."""
```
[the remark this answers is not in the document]
vlipovac: Documentation (docstring and in-code) added

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+        return np.hstack([self.fetch_data(self, g, False) for g in self.domains])
+
+    def get_jacobian(self, *args: float | np.ndarray | AdArray) -> sps.spmatrix:
+        """Fetches the derivative values stored for this secondary operator at its time
+        or iterate index.
+
+        Uses the structure of the Jacobians of arguments to insert the values,
+        assuming the Jacobians contain only identity blocks.
```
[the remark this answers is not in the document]
vlipovac: Documentation in docstring and in-code added.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+
+        """
+
+        derivatives = np.hstack([self.fetch_data(self, g, True) for g in self.domains])
+
+        # list of jacs per dependency, assuming porepy.ad makes consistent shapes
```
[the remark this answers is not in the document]
vlipovac: Added a shape consistency check and respective in-code documentation

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+        jacs: list[sps.csr_matrix] = []
+
+        for arg, d_i in zip(args, derivatives):
```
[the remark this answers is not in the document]
vlipovac: Added documentation in-code

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+            if isinstance(arg, AdArray):
+                # At this point we assume that the derivative has only an identity block
```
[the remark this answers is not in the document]
vlipovac: Updated documentation

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+            :meth:`~porepy.numerics.ad.equation_system.EquationSystem.create_variables`.
+
+            The number of DOFs of this expression is used to validate the shape of
+            values and derivative values set.
+            Defaults to cell-wise, scalar expression.
+
+    Raises:
+        ValueError: If there are no ``*dependencies``. The user should use other
+            solutions in ``porepy.ad`` for this case.
+
```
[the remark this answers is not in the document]
vlipovac: Added reference to module-level documentation at top of docstring for both classes.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+        # On subdomains or interfaces, create the secondary operators
+        elif all(isinstance(g, pp.Grid) for g in domains) or all(
+            isinstance(g, pp.MortarGrid) for g in domains
+        ):
+            # for mypy
+            domains_ = cast(Sequence[pp.Grid] | Sequence[pp.MortarGrid], domains)
+            children = [child(domains_) for child in self._dependencies]
+
+            # Check if first-order dependency
+            assert all(isinstance(child, pp.ad.Variable) for child in children), (
```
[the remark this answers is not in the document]
vlipovac: No particular reason.
My thinking is, that the callables `self._dependencies` are given `__init__`, there I would use a `TypeError`, if I would actually try to call them.
Here I *assert* that everything is as required.

src/porepy/numerics/ad/secondary_operator.py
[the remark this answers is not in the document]
vlipovac: Discussed in person, documentation updated in the *Notes* sections in the module-level docstring.

src/porepy/numerics/ad/secondary_operator.py
vlipovac: I left this, just in case that for some the user stores the derivative values at previous time or iterates.
Maybe an error would be better, because Variables evaluated at prev time or iter give an AD array with jac=0.
It would theoretically trigger the `get_jacobian` in `SecondaryOperator`, and there the `KeyError` is raised.
But conceptually, even without `KeyError`, we cannot use the previous derivatives because the variables have `jac=0`.

Will that ever change?

src/porepy/numerics/ad/secondary_operator.py
vlipovac: Changed to _name_derivatives

src/porepy/numerics/ad/secondary_operator.py
vlipovac: It's an instance of `np.ndarray`, with `ndim=1`, `shape=(0,)`, definitely not a scalar.

src/porepy/numerics/ad/secondary_operator.py
```diff
@@ -0,0 +1,897 @@
...
+        The getter fetches the values in the order imposed by the md-grid.
+        The setter shifts iterate values backwards and sets the given value as the
+        current iterate.
+
+        Important:
+            The order of derivatives should reflect the order of ``dependencies``
+            passed at instantiation.
+
+        Parameters:
+            val: ``shape=(num_dependencies, N)``
```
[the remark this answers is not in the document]
vlipovac: I assume you are looking for a way to use sparse matrices?

The reason is, that this contains the actual values, which are then inserted into the (much bigger) sparse matrix upon evaluation.

So this `np.ndarray` is associated with the `.data` field of the resulting sparse matrix, stored here in a specific order.

src/porepy/numerics/ad/secondary_operator.py
vlipovac: `shift_solution_values` will extend the length by 1, because `i -> i + 1`.
Then the entry at `i= 0` and `i = 1` are identical, until the user provides a new value (both time and iter sense).

This is related to the discussion with @IvarStefansson  on the indexation. 
We probably should update that first in `develop`, before adapting this class

tests/numerics/ad/test_secondary_operator.py
```diff
@@ -0,0 +1,300 @@
...
+
+    """
+
+    bounding_box = {"xmin": 0, "xmax": 1, "ymin": 0, "ymax": 1}
+    domain = pp.Domain(bounding_box=bounding_box)
+
+    frac1 = pp.LineFracture(np.array([[0.5, 0.5], [0.2, 0.8]]))
+    frac2 = pp.LineFracture(np.array([[0.2, 0.8], [0.5, 0.5]]))
+
+    network_2d = pp.create_fracture_network([frac1, frac2], domain)
```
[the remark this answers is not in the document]
vlipovac: I was not aware of that method, switched to it!

tests/numerics/ad/test_secondary_operator.py
vlipovac: Discussed in person, removing this part.

tests/numerics/ad/test_secondary_operator.py
vlipovac: `SecondaryExpression` has only `fetch_data` and it does only that, fetching data for specified time or iterate.
In `SecondaryExpression`, I would leave it public to give the user the opportunity to inspect the stored data.
Maybe make it private in the `SecondaryOPERATOR`? Because that class has `value_and_jacobian`, and only that one should be used.

src/porepy/numerics/ad/secondary_operator.py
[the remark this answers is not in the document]
keileg: If you think we can remove this, I suggest we do so.

FYI: Restricting operators to a subset of grids may be part of a future enhancement of the Ad framework.

src/porepy/numerics/ad/secondary_operator.py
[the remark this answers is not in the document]
keileg: Let's not complicate this PR by changing other parts of the code.

Please be a bit more specific on what is meant by 'the hack', to avoid someone being confused at a later stage.

src/porepy/numerics/ad/secondary_operator.py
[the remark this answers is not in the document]
keileg: Rather, why is a `pp.ad.AbstractFunction.func` returned?

src/porepy/numerics/ad/secondary_operator.py
keileg: I don't think we can rule out that, in the future, evaluation at a previous time step or iterate may return a numpy arrays instead of an Ad array.

src/porepy/numerics/ad/secondary_operator.py
keileg: Yes, explicit raising of errors are usually preferable.

src/porepy/numerics/ad/secondary_operator.py
keileg: Agreed, let's do the cleanup before polishing this PR.

tests/numerics/ad/test_secondary_operator.py
keileg: Agreed, let's make it private in SecondaryOperator (with an explanation why this is done).

src/porepy/numerics/ad/secondary_operator.py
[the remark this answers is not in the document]
vlipovac: Call-overload removed.

src/porepy/numerics/ad/secondary_operator.py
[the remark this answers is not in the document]
vlipovac: Added better documentation.

src/porepy/numerics/ad/secondary_operator.py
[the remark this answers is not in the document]
vlipovac: Related to https://github.com/pmgbergen/porepy/pull/1170#discussion_r1602123034

src/porepy/numerics/ad/secondary_operator.py
vlipovac: Added `ValueError`, if `op` is at previous time or derivative and the user attempts to get derivative values for it.

tests/numerics/ad/test_secondary_operator.py
vlipovac: `SecondaryOperator._fetch_data` is now private. Updated documentation.

src/porepy/numerics/ad/secondary_operator.py
vlipovac: A generic check for type and shape of values to be set is implemented.
Note: currently we allow arrays which are "too large" in the global setters. The grid-wise slicing allows. Should we add errors requiring the new values to be of exact shape, and not "at least of shape" for the global setters?

vlipovac:
**Re-naming the new operator**

Since this has been brought to attention several times, let us discuss the name of the class.

1. **Secondary**Operator: Motivated by the fact that this is used to represent a dependent (secondary) term like ``density(pressure, temperature)``
2. **Surrogate**Operator: Motivated by the fact that this instance replaces the actual ``Operation.evaluate`` with user-given data.

Other candidates could be **External**, **Dummy** or **DataOperator**, where the motivation is clear I think.
Please feel free to suggest other names.

src/porepy/numerics/ad/secondary_operator.py
keileg: >  Should we add errors requiring the new values to be of exact shape, and not "at least of shape" for the global setters?

I think we can do without this. Feel free to resolve the comment if you consider yourself done here.

tests/numerics/ad/test_secondary_operator.py
```diff
@@ -0,0 +1,300 @@
...
+
+    mesh_args: dict[str, float] = {"cell_size": 0.1, "cell_size_fracture": 0.1}
+    mdg = pp.create_mdg("simplex", mesh_args, network_2d)
```
keileg, line 38:
```suggestion
    mdg = pp.create_mdg("cartesian", mesh_args, network_2d)
```

tests/numerics/ad/test_secondary_operator.py
```diff
@@ -0,0 +1,300 @@
...
+
+    bounding_box = {"xmin": 0, "xmax": 1, "ymin": 0, "ymax": 1}
+    domain = pp.Domain(bounding_box=bounding_box)
+
+    frac1 = pp.LineFracture(np.array([[0.5, 0.5], [0.2, 0.8]]))
+    frac2 = pp.LineFracture(np.array([[0.2, 0.8], [0.5, 0.5]]))
+
+    network_2d = pp.create_fracture_network([frac1, frac2], domain)
+
+    mesh_args: dict[str, float] = {"cell_size": 0.1, "cell_size_fracture": 0.1}
```
keileg, line 37:
```suggestion
    mesh_args: dict[str, float] = {"cell_size": 0.5, "cell_size_fracture": 0.5}
```
I think this should work, and the test will run faster.

keileg:
@vlipovac: I walked through the comments again, and have resolved all but two. One relates to #1172 and is put on hold. 

The other issue is that I don't fully understand the intended relation between this function and the `AbstractFunction` hierarchy. As  we discussed, the fault likely lies with the poorly formulated concept of the `Function`s, and we will not go there now. I suggest we add a TODO in `SecondaryOperator`, likely around [here](https://github.com/pmgbergen/porepy/blob/54c0a6a95ead85755b7ae67acacc5ef0351ef8d1/src/porepy/numerics/ad/secondary_operator.py#L249) that this construction should be revisited if the Function hierarchy is refactored. Does that make sense.

I also added two minor suggestions in the test, that should make it run faster without changing the behavior (I believe, if I am wrong, please ignore my suggestions).

keileg:
Then the final comment: We need a better name for `SecondaryOperator`, though I am blank. @OmarDuran, what do you think?

When we have settled the new name, I suggest the `SecondaryExpression` is renamed to `NewNameFactory`.

vlipovac:
> Then the final comment: We need a better name for `SecondaryOperator`, though I am blank. @OmarDuran, what do you think?
> 
> When we have settled the new name, I suggest the `SecondaryExpression` is renamed to `'NewName'Factory`.

See my previous comment
https://github.com/pmgbergen/porepy/pull/1170#issuecomment-2122381747

keileg:
> **Re-naming the new operator**
> 
> Since this has been brought to attention several times, let us discuss the name of the class.
> 
> 1. **Secondary**Operator: Motivated by the fact that this is used to represent a dependent (secondary) term like `density(pressure, temperature)`
> 2. **Surrogate**Operator: Motivated by the fact that this instance replaces the actual `Operation.evaluate` with user-given data.
> 
> Other candidates could be **External**, **Dummy** or **DataOperator**, where the motivation is clear I think. Please feel free to suggest other names.

Of these, I prefer `DataOperator`, though I'm not 100% happy with that either.

src/porepy/numerics/ad/secondary_operator.py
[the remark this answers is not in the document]
vlipovac: @keileg  See line 250 regarding your request for leaving a TODO note.
Please resolve this conversation, if you are satisfied.

OmarDuran:
> > **Re-naming the new operator**
> > Since this has been brought to attention several times, let us discuss the name of the class.
> > 
> > 1. **Secondary**Operator: Motivated by the fact that this is used to represent a dependent (secondary) term like `density(pressure, temperature)`
> > 2. **Surrogate**Operator: Motivated by the fact that this instance replaces the actual `Operation.evaluate` with user-given data.
> > 
> > Other candidates could be **External**, **Dummy** or **DataOperator**, where the motivation is clear I think. Please feel free to suggest other names.
> 
> Of these, I prefer `DataOperator`, though I'm not 100% happy with that either.

@vlipovac  and @keileg I do prefer the name **Surrogate**Operator which resembles a substitute for "true" derivatives.

keileg:
> > Of these, I prefer `DataOperator`, though I'm not 100% happy with that either.
> 
> @vlipovac and @keileg I do prefer the name **Surrogate**Operator which resembles a substitute for "true" derivatives.

I can live with `SurrogateOperator`, so let's go for that.

keileg (approved):

PR #1171, opened 2024-05-16
Deactivate redundant residual computation before check_convergence.

jwboth:
## Proposed changes

Prior to checking the convergence in `check_convergence`, the residual is always computed to ask for the recent residual. However, if the tolerance is `np.inf`, it will actually not be required. The PR only updates the residual if the tolerance is not `np.inf`.

## Types of changes

- [ ] Other:

## Checklist

No updates to tests are added, as the change only works on improving the logic.


jwboth:
> The following thought occurred to me when looking at the change: Does the residual update call perhaps belong inside `check_convergence`?

I had the identical intuitive reaction when making the changes. If we both agree on that conceptual thought, I can suggest a change for discussion.

PR #1172, opened 2024-05-21
Time step and iteration indexation in pp.AD

vlipovac:
## Proposed changes

DRAFT: @IvarStefansson @keileg  for the indexation issues we discussed after the last user meeting.

Time step indexation reverted to start with 0 for first previous time step.
Time-dependent AD operators have `None` as their time step index, as long as they are at current time.
Iterative operators have always an integer value as iteration index, starting with 0 for the first previous iterate.
The first call to `.previous_iteration()` does not push the iterate index up (stays 0), but flags the instance as as previous iterate, removing it's capability to get a proper AdArray during parsing (`.values_and_jacobian()` will give a trivial `.jac`)

The "flag" for is previous iterate (also for time step) is computed using a PRIVATE index, which starts at -1, and is consistently increased with `+= 1`.

Note also, I propose here to return `iterate_index` as `None` for operators which are both `TimeDependentOperators` and `IterativeOperators`, and at a previous time step.

1. The aim is to make it more clear to users, that there are no iteration indices for previous time steps.
2. Also, code in various parts of the framework will be slightly shorter, because the `get_values` and `set_values` methods do not need an if-branch regarding whether the variable/operator is at previous time. Variables will always have a pair of time step index and iterate index, which are consistent with the intended usage of various get and set values methods in the Ad framework.

## Relevant changes

The essential changes are in the following files:

src/numerics/ad/_ad_utils.py
src/numerics/ad/equation_system.py
src/numerics/ad/operators.py

changes in the other file revert essentially the changes introduced in PR #1166

### DIFF
--- a/src/porepy/__init__.py
+++ b/src/porepy/__init__.py
@@ -160,6 +160,7 @@ from porepy.numerics.ad.operators import wrap_as_dense_ad_array, wrap_as_sparse_
 from porepy.numerics.ad.equation_system import EquationSystem
 from porepy.numerics.ad._ad_utils import set_solution_values
 from porepy.numerics.ad._ad_utils import get_solution_values
+from porepy.numerics.ad._ad_utils import shift_solution_values
 
 # Time stepping control
 from porepy.numerics.time_step_control import TimeManager
--- a/src/porepy/models/boundary_condition.py
+++ b/src/porepy/models/boundary_condition.py
@@ -92,7 +92,7 @@ class BoundaryConditionMixin:
                 # No previous time step exists. The method was called during
                 # the initialization.
                 vals = function(bg)
-            pp.set_solution_values(name=name, values=vals, data=data, time_step_index=1)
+            pp.set_solution_values(name=name, values=vals, data=data, time_step_index=0)
 
             # Set the unknown time step values.
             vals = function(bg)
--- a/src/porepy/models/solution_strategy.py
+++ b/src/porepy/models/solution_strategy.py
@@ -264,7 +264,7 @@ class SolutionStrategy(abc.ABC):
             An array of the indices of which time step solutions will be stored.
 
         """
-        return np.array([1])
+        return np.array([0])
 
     @property
     def iterate_indices(self) -> np.ndarray:
@@ -302,9 +302,9 @@ class SolutionStrategy(abc.ABC):
                     time_index,
                     times_file,
                 )
-            vals = self.equation_system.get_variable_values(time_step_index=1)
+            vals = self.equation_system.get_variable_values(time_step_index=0)
             self.equation_system.set_variable_values(
-                vals, iterate_index=0, time_step_index=1
+                vals, iterate_index=0, time_step_index=0
             )
             # Update the boundary conditions to both the time step and iterate solution.
             self.update_time_dependent_ad_arrays()
@@ -442,7 +442,7 @@ class SolutionStrategy(abc.ABC):
         solution = self.equation_system.get_variable_values(iterate_index=0)
         self.equation_system.shift_time_step_values()
         self.equation_system.set_variable_values(
-            values=solution, time_step_index=1, additive=False
+            values=solution, time_step_index=0, additive=False
         )
         self.convergence_status = True
         self.save_data_time_step()
--- a/src/porepy/numerics/ad/_ad_utils.py
+++ b/src/porepy/numerics/ad/_ad_utils.py
@@ -313,36 +313,23 @@ def _validate_indices(
     out = []
 
     if iterate_index is not None:
-        # Some iterate values of the current time
+        # Some valid iterate value.
         if iterate_index >= 0:
             out.append((pp.ITERATE_SOLUTIONS, iterate_index))
         # Negative iterate indices are not supported
         else:
             raise ValueError(
-                "Use increasing, non-negative integers for (previous) iterate values."
+                "Use increasing, non-negative integers for iterate indices."
             )
 
     if time_step_index is not None:
         # Some previous time.
-        if time_step_index > 0:
+        if time_step_index >= 0:
             out.append((pp.TIME_STEP_SOLUTIONS, time_step_index))
-        # Current time. NOTE this is ambigous since the current time is an unknown and
-        # has multiple iterate values.
-        # Alternatively, we could associate time_step_index = 0 with iterate_index = 0
-        # the below elif branch introduces the convention that
-        # time step = iterate step = 0 are equivalent.
-        elif time_step_index == 0:
-            # if (pp.ITERATE_SOLUTIONS, 0) not in out:
-            #     out.append((pp.ITERATE_SOLUTIONS, 0))
-            raise ValueError(
-                "Using time_step_index = 0 (current time) is ambiguous."
-                + " Specify iterate_index instead."
-                + " First previous time step value is time_step_index = 1."
-            )
         # Negative time step indices are not supported
         else:
             raise ValueError(
-                "Use increasing, non-negative integers for previous time step values."
+                "Use increasing, non-negative integers for time step indices."
             )
 
     return out
@@ -363,16 +350,20 @@ def set_solution_values(
         name: Name of the quantity that is to be assigned values.
         values: The values that are set in the data dictionary.
         data: Data dictionary corresponding to the subdomain or interface in question.
-        time_step_index (optional): Determines the key of where `values` are to be
-            stored in `data[pp.TIME_STEP_SOLUTIONS][name]`. 0 means it is the most
-            recent set of values, 1 means the one time step back in time, 2 is two time
-            steps back, and so on.
-        iterate_index (optional): Determines the key of where `values` are to be
-            stored in `data[pp.ITERATE_SOLUTIONS][name]`. 0 means it is the most
-            recent set of values, 1 means the values one iteration back in time, 2 is
-            two iterations back, and so on.
-        additive: Flag to decide whether the values already stored in the data
-            dictionary should be added to or overwritten.
+        time_step_index: ``default=None``
+
+            Determines the key of where ``values`` are to be stored in
+            ``data[pp.TIME_STEP_SOLUTIONS][name]``.
+            0 is the first previous time step, 1 the one before that and so on.
+        iterate_index: ``default=None``
+
+            Determines the key of where ``values`` are to be stored in
+            ``data[pp.ITERATE_SOLUTIONS][name]``.
+            0 is the current iterate, 1 the previous iterate, and so on.
+        additive: ``default=False``
+
+            Flag to decide whether the values already stored in the data dictionary
+            should be added to or overwritten.
 
     Raises:
         ValueError: In the case of inconsistent usage of indices
@@ -409,21 +400,29 @@ def get_solution_values(
     """Function for fetching values stored in the data dictionary, for some
     time-dependent or iterative term.
 
+    Note:
+        Compared to :func:`get_solution_values` the getter works only for 1 defined
+        index, whereas the setter can take both a time and iterate index.
+
     Parameters:
         name: Name of the parameter whose values we are interested in.
         data: The data dictionary.
-        time_step_index: Which time step we want to get values for. 0 is current, 1 is
-            one time step back in time. This is only limited by how many time steps are
-            stored from before.
-        iterate_index: Which iterate we want to get values for. 0 is current, 1 is one
-            iterate back in time. This is only limited by how many iterates are stored
-            from before.
+        time_step_index: ``default=None``
+
+            Determines the key of where ``values`` are to be stored in
+            ``data[pp.TIME_STEP_SOLUTIONS][name]``.
+            0 is the first previous time step, 1 the one before that and so on.
+        iterate_index: ``default=None``
+
+            Determines the key of where ``values`` are to be stored in
+            ``data[pp.ITERATE_SOLUTIONS][name]``.
+            0 is the current iterate, 1 the previous iterate, and so on.
 
     Raises:
         ValueError: In the case of inconsistent usage of indices
             (both None or negative values).
-        AssertionError: If the user attempts to get iterate and time step values
-            simultanously. Only 1 index is permitted in getter
+        ValueError: If the user attempts to get multiple iterate and time step values
+            simultanously. Only 1 index is permitted in the getter.
         KeyError: If no values are stored for the passed index.
 
     Returns:
@@ -431,9 +430,10 @@ def get_solution_values(
 
     """
     loc_index = _validate_indices(time_step_index, iterate_index)
-    assert (
-        len(loc_index) == 1
-    ), "Cannot get value from both iterate and time step at once. Call separately."
+    if len(loc_index) != 1:
+        raise ValueError(
+            "Cannot get value from both iterate and time step at once. Call separately."
+        )
 
     loc, index = loc_index[0]
 
@@ -445,6 +445,50 @@ def get_solution_values(
         ) from err
 
     return value
+
+
+def shift_solution_values(
+    name: str,
+    data: dict,
+    location: Any,
+) -> None:
+    """Function to shift numerical values stored in the data dictionary.
+
+    The shift is implemented s.t. values at index ``i`` are copied to index ``i + 1``,
+    based on the number of stored values.
+
+    Note:
+        The data stored must have support for ``.copy()`` in order to avoid faulty
+        referencing (e.g., numpy arrays or sparse matrices).
+
+    Note:
+        After this operation, values at index 0 and 1 will be the same.
+        Use :meth:`set_solution_values` to update the latest values at index 0.
+
+    Parameters:
+        name: Key in ``data`` for which quantity the shift should be performed.
+        data: A grid data dictionary.
+        location: Either :data:`~porepy.utils.common_constants.TIME_STEP_SOLUTIONS`
+            or :data:`~porepy.utils.common_constants.ITERATE_SOLUTIONS`.
+
+    Raises:
+        ValueError: If unsupported ``location`` is passed.
+
+    """
+    if location not in [pp.ITERATE_SOLUTIONS, pp.TIME_STEP_SOLUTIONS]:
+        raise ValueError(f"Shifting values not implemented for location {location}")
+
+    # NOTE return because nothing to be shifted. Avoid confusion by introducing data
+    # dictionaries for values which were never set using pp.set_solution_values
+    if location not in data:
+        return
+    if name not in data[location]:
+        return
+
+    num_stored = len(data[location][name])
+
+    for i in range(num_stored, 0, -1):
+        data[location][name][i] = data[location][name][i - 1].copy()
 
 
 class MergedOperator(operators.Operator):
--- a/src/porepy/numerics/ad/equation_system.py
+++ b/src/porepy/numerics/ad/equation_system.py
@@ -569,22 +569,23 @@ class EquationSystem:
         index specified by the user. The global order is preserved and independent of
         the order of the argument.
 
+        See also:
+            :meth:`~porepy.numerics.ad._ad_utils.get_solution_values`.
+
         Parameters:
-            variables (optional): VariableType input for which the values are
-                requested. If None (default), the global vector of unknowns is returned.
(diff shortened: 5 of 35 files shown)


src/porepy/models/solution_strategy.py
```diff
@@ -264,7 +264,7 @@ def time_step_indices(self) -> np.ndarray:
             An array of the indices of which time step solutions will be stored.
```
IvarStefansson, line 264:
```suggestion
            An array of the indices of which time step solutions will be stored, counting
            from 0. Defaults to storing the most recently computed solution only.
```

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -363,16 +350,20 @@ def set_solution_values(
...
-            stored in `data[pp.ITERATE_SOLUTIONS][name]`. 0 means it is the most
-            recent set of values, 1 means the values one iteration back in time, 2 is
-            two iterations back, and so on.
-        additive: Flag to decide whether the values already stored in the data
-            dictionary should be added to or overwritten.
+        time_step_index: ``default=None``
+
+            Determines the key of where ``values`` are to be stored in
+            ``data[pp.TIME_STEP_SOLUTIONS][name]``.
+            0 is the first previous time step, 1 the one before that and so on.
```
IvarStefansson, line 357:
```suggestion
            0 is the most recent time step, 1 the one before that and so on.
```

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -409,31 +400,40 @@ def get_solution_values(
     """Function for fetching values stored in the data dictionary, for some
     time-dependent or iterative term.
 
+    Note:
+        Compared to :func:`get_solution_values` the getter works only for 1 defined
```
IvarStefansson, line 404:
```suggestion
        Compared to :func:`set_solution_values` the getter works only for 1 defined
```

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -409,31 +400,40 @@ def get_solution_values(
...
-            one time step back in time. This is only limited by how many time steps are
-            stored from before.
-        iterate_index: Which iterate we want to get values for. 0 is current, 1 is one
-            iterate back in time. This is only limited by how many iterates are stored
-            from before.
+        time_step_index: ``default=None``
+
+            Determines the key of where ``values`` are to be stored in
+            ``data[pp.TIME_STEP_SOLUTIONS][name]``.
+            0 is the first previous time step, 1 the one before that and so on.
```
IvarStefansson, line 414:
```suggestion
            0 is the most recent time step, 1 the one before that and so on.
```

src/porepy/numerics/ad/equation_system.py
```diff
@@ -638,15 +639,15 @@ def set_variable_values(
...
-                set.
-            time_step_index: Index of previous time step for which the values are
-                intended.
-            iterate_index: Iterate index for current time step for which the values are
-                intended.
-            additive (optional): Flag to write values additively. To be used in
-                iterative procedures.
+            variables: ``default=None``
+
+                VariableType input for which the values are requested.
```
IvarStefansson, line 644:
```suggestion
                VariableType input for which the values are prescribed.
```

src/porepy/numerics/ad/operators.py
```diff
@@ -1175,30 +1175,29 @@ def __init__(
...
-            value at the most recent iterate.
-
         """
-        return self._time_step_index
+        if self._time_step_index < 0:
+            return None
+        else:
+            return self._time_step_index
 
     def previous_timestep(
```
IvarStefansson, line 1202: What do you think about renaming these methods to the more precise `value_at_timestep/iteration(self, step)`?

src/porepy/numerics/ad/operators.py
```diff
@@ -1285,26 +1284,41 @@ def __init__(
 
         """
 
-        self._iterate_index: int = 0
+        self._iterate_index: int = -1
         """iterate index, starting with 0 (current iterate at current time) and
```
IvarStefansson, line 1288:
```suggestion
        """Iterate index, starting with 0 (current iterate at current time) and
```

src/porepy/numerics/ad/operators.py
```diff
@@ -1285,26 +1284,41 @@ def __init__(
...
     @property
-    def iterate_index(self) -> int:
+    def iterate_index(self) -> int | None:
         """Returns the iterate index this instance represents, at the current time.
 
-        - 0 represents the current iterate
-        - 1 represents the first previous iterate
-        - 2 represents the iterate before that
+        - None indicates this instance is at a previous time
+        - 0 represents the first previous iterate
```
IvarStefansson, line 1301:
```suggestion
        - 0 represents the most recently computed iterate.
```

src/porepy/numerics/ad/operators.py
```diff
@@ -1782,18 +1796,14 @@ def parse(self, mdg: pp.MixedDimensionalGrid) -> Any:
...
-                time_step_index=self.time_step_index,
-            )
-        else:
-            return pp.get_solution_values(
-                self.name,
-                data,
-                iterate_index=self.iterate_index,
-            )
+        # We can safely use both indices as arguments, without checking prev time,
+        # because iterate index is None if prev time, and vice versa
```
IvarStefansson, line 1800: Thanks for the comment!

src/porepy/numerics/ad/operators.py
```diff
@@ -1853,7 +1865,7 @@ def __init__(self, variables: list[Variable]) -> None:
                 len(set(time_indices)) == 1
             ), "Cannot create md-variable from variables at different time steps."
             assert (
-                len(set(iterate_indices)) == 1
+                len(set(iter_indices)) == 1 and len(set(current_iter)) == 1
```
IvarStefansson, line 1868: Why the check of current_iter?

src/porepy/numerics/ad/operators.py
```diff
@@ -1864,16 +1876,24 @@ def __init__(self, variables: list[Variable]) -> None:
...
         self._id = next(Variable._ids)
         # NOTE private time step index is -1 if public time step index of atomic
         # variables is None (current time)
         self._time_step_index = -1 if time_indices[0] is None else time_indices[0]
-        # NOTE private and public iterate indices are always integers
-        self._iterate_index = iterate_indices[0]
+
+        # If current time and iterate
+        if current_iter[0]:
+            self._iterate_index = -1  # current time and iter
```
IvarStefansson, line 1891: In this case, isn't iter_indices[0] None?

src/porepy/__init__.py
```diff
@@ -160,6 +160,7 @@
 from porepy.numerics.ad.equation_system import EquationSystem
 from porepy.numerics.ad._ad_utils import set_solution_values
```
IvarStefansson, line 161: Could we merge these three lines to one?

src/porepy/models/solution_strategy.py
```diff
@@ -440,9 +440,11 @@ def after_nonlinear_convergence(self) -> None:
 
         """
         solution = self.equation_system.get_variable_values(iterate_index=0)
-        self.equation_system.shift_time_step_values()
+        self.equation_system.shift_time_step_values(
+            max_index=len(self.time_step_indices)
```
vlipovac, line 444: Added this functionality to avoid flooding the memory.
Same in line 430

IvarStefansson (commented):
Nice work. Please have a look at my comments and questions!

src/porepy/models/solution_strategy.py
IvarStefansson: Seems like a very good idea!

src/porepy/numerics/ad/operators.py
vlipovac: If all sub variables are current iterate, then `iter_indices[0] == 0`, because they fetch their `.val` from that index during `.value_and_jacobian()`.

src/porepy/__init__.py
vlipovac: Can be merged into 1 import statement, but still spanning multiple lines because of the character limit.

src/porepy/numerics/ad/operators.py
IvarStefansson: I believe you had some opinions on this, @keileg and @jwboth. Could you please comment?

src/porepy/numerics/ad/operators.py
IvarStefansson: Thanks for the explanation. Could you please add one line of comment.

src/porepy/__init__.py
vlipovac: Merged into 1 import statement (for module `_ad_utils`) after discussion

src/porepy/numerics/ad/operators.py
vlipovac: Documentation added.

src/porepy/numerics/ad/operators.py
vlipovac: Discussed in person, added documentation why this is required and the if-else branch cannot be removed.

src/porepy/numerics/ad/operators.py
```diff
@@ -1852,8 +1879,11 @@ def __init__(self, variables: list[Variable]) -> None:
             assert (
                 len(set(time_indices)) == 1
             ), "Cannot create md-variable from variables at different time steps."
+            # NOTE both must be unique for all sub-variables, to avoid md-variables
+            # having sub-variables at different iterate states.
+            # Both current value, and most recent previous iterate have iterate index 0.
```
IvarStefansson, line 1884:
```suggestion
            # Both current value, and most recent previous iterate have iterate index 0,
            # hence the need to check the size of the current_iter set.
```

src/porepy/numerics/ad/operators.py
IvarStefansson: Works for me. What about the properties `is_previous_time` and `is_previous_iterate`? Do we get rid of "previous" there as well? Any suggestions for alternatives?

keileg: > Works for me. What about the properties `is_previous_time` and `is_previous_iterate`? Do we get rid of "previous" there as well? Any suggestions for alternatives?

This is more difficult. I like the `is` prefix, indicating the boolean return type, but modifying the rest is difficult. One possibility is `is_lagged_in_time` (resp. `_iterate`), but that is not clearly better than using `previous` (so, switching to a synonym does not work). 

I think I'm blank, so unless other suggestions come up, I vote for keeping `is_previous_` (for now).

jwboth: I have not much too add. I appreciate getting rid of previous when fetching values.

vlipovac: `.previous_timestep(steps: int = 1) -> .at_timestep(steps: int = 1)`

Using *previous* reflects that the implementation is going *backwards* in time, with corresponding number of steps.
The naming `at_timestep` indicates that the user could use a concrete index as an argument to "jump" between time steps.
That functionality would require some more safety checks.

Analogous for iterates.

vlipovac: Regarding the safety checks about stepping:

As of now, the user could create "holes" in the sequence of prev time and iter.
I.e. Operators for index 0, 1, 3,4 for example, but not for index 2.
Is this critical? Or does it fall in the category "The user should know what he is doing"?

src/porepy/models/boundary_condition.py
```diff
@@ -92,24 +95,22 @@ def update_boundary_condition(
                 # No previous time step exists. The method was called during
                 # the initialization.
                 vals = function(bg)
-            pp.set_solution_values(name=name, values=vals, data=data, time_step_index=1)
+
+            # Before setting the new, most recent time step, shift the stored values
+            # backwards in time
+            pp.shift_solution_values(
```
vlipovac, line 101: I enhanced the functionality of the BoundaryConditions mixin, to now support the full time step depth.
Removed the warning below.

src/porepy/numerics/ad/operators.py
IvarStefansson: I don't understand what you mean about the "holes". Please elaborate why this could be an issue, in person if you prefer.

vlipovac: Discussed in person, issue deemed not critical.

IvarStefansson: Jakub and I suggest `is_at_timestep` for the booleans.

vlipovac: I am not sure how that is interpreted as a boolean property. Are you inquiring for a specific time step?

vlipovac:
@IvarStefansson @keileg @jwboth
I had a discussion with Ivar and I would like to finish this PR as it is stalling progress in the chain of PRs regarding CF.

Essentially, I took the liberty to rename the stuff, having your comments in mind.
The common denominator I was able to filter out was the wish to rename it to something resembling proper language and to add the word *at*.
I strongly discourage the use of *value* anywhere, since per logic we have two representations of the mathematical model: One is more symbolic (Operators and Operator trees), the other numerical (parsing operators and getting arrays).
I think we should be aware of these parallel structures, which of course go hand in hand, but their internal workings and functionality are orthogonal in some sense. This should be reflected in the naming convention. I.e., keywords like *value* should indicate a switch from the symbolic representation to the numeric evaluation.
The keyword *previous* I kept to indicate the directional behavior of the functionality, backwards in time and the iterate sense.

So the members are renamed as follows:

Functions creating operators at previous X

1. `previous_timestep()` -> `at_previous_timestep()`
2. `previous_iteration()` -> `at_previous_iteration()`

Boolean flags checking the index

1. `is_previous_timestep` -> `is_at_previous_timestep`
2. `is_previous_iterate` -> `is_at_previous_iteration`
3. `is_current_iterate` -> `is_at_current_iteration`

@IvarStefansson If you agree with this, please resolve the respective conversation and let's finalize this, but vennligst have a final thought on the changes here so we don't rush into pushing something confusing like in the previous PR.

jwboth:
> @IvarStefansson @keileg @jwboth I had a discussion with Ivar and I would like to finish this PR as it is stalling progress in the chain of PRs regarding CF.
> 
> Essentially, I took the liberty to rename the stuff, having your comments in mind. The common denominator I was able to filter out was the wish to rename it to something resembling proper language and to add the word _at_. I strongly discourage the use of _value_ anywhere, since per logic we have two representations of the mathematical model: One is more symbolic (Operators and Operator trees), the other numerical (parsing operators and getting arrays). I think we should be aware of these parallel structures, which of course go hand in hand, but their internal workings and functionality are orthogonal in some sense. This should be reflected in the naming convention. I.e., keywords like _value_ should indicate a switch from the symbolic representation to the numeric evaluation. The keyword _previous_ I kept to indicate the directional behavior of the functionality, backwards in time and the iterate sense.
> 
> So the members are renamed as follows:
> 
> Functions creating operators at previous X
> 
> 1. `previous_timestep()` -> `at_previous_time_step()`
> 2. `previous_iteration()` -> `at_previous_iteration()`
> 
> Boolean flags checking the index
> 
> 1. `is_previous_timestep` -> `is_at_previous_time_step`
> 2. `is_previous_iterate` -> `is_at_previous_iteration`
> 3. `is_current_iterate` -> `is_at_current_iteration`
> 
> The switch from `timestep` to `time_step` is made to be consistent with the `time_step_index`, which a property but also an argument in the equation system and ad utility functions.
> 
> @IvarStefansson If you agree with this, please resolve the respective conversation and let's finalize this, but vennligst have a final thought on the changes here so we don't rush into pushing something confusing like in the previous PR.

I give all authority to @IvarStefansson - I will accept whatever will be decided here. 

But since the final decision is not made yet, and before it is, here is one more possibility for your inspiration, motivated by my aversion to too long names:
`at_time_step(lag=i)` and acordingly for iteration, where `i=0` gives the current and `i=1` gives the previous one. This allows to unify with current time steps/iterates as well, e.g., `is_at_iteration(lag=i)` removes the need for splitting up the boolean methods (I could not find those in the code - maybe I misunderstood and these are not methods...)

Maybe there is a better word than `lag`. Yet, I wanted to consisely indicate the backward nature of what we use to call `index`, which is counterintuitive considering how one would write time discretizations and iterative methods on paper. In general, now having looked at the code, and with the above. Such a change would strongly suggest also renaming `time_step_index` to `time_step_lag`.

To make it even more complicated, one could provide two possibilities to access the operators and also `get_solution_values`, using not only indices of type `int`, but a `Literal["current", "previous"], e. g., `at_iterate("current")`, `at_iterate("previous")`, `get_solution_values(time_step="current"` etc. Of course a nightmare to maintain. But as your main motivation is expressivity of the code, my answer would look like that.

I anticipate that this is not of interest. Nevertheless, since I was blank last time I was asked and these thoughts now popped into my head, here it is: my brain dump :)

keileg:
@vlipovac @IvarStefansson, @jwboth: It seems to me that, while we agree the current `previous_timestep`, `previous_iterate` should be renamed, we are not ready to conclude on a naming scheme. Still, Veljko is right we should finalize this process so as not to block the merges for compositional flow. I therefore suggest we revert the naming to what it was before this PR and revisit at a later stage. From what I understand, such a resolution will allow us to merge this PR with little extra effort.

Does this make sense to you?

IvarStefansson:
> @vlipovac @IvarStefansson, @jwboth: It seems to me that, while we agree the current `previous_timestep`, `previous_iterate` should be renamed, we are not ready to conclude on a naming scheme. Still, Veljko is right we should finalize this process so as not to block the merges for compositional flow. I therefore suggest we revert the naming to what it was before this PR and revisit at a later stage. From what I understand, such a resolution will allow us to merge this PR with little extra effort.
> 
> Does this make sense to you?

Yes, I think this is a good solution!

vlipovac:
> > @vlipovac @IvarStefansson, @jwboth: It seems to me that, while we agree the current `previous_timestep`, `previous_iterate` should be renamed, we are not ready to conclude on a naming scheme. Still, Veljko is right we should finalize this process so as not to block the merges for compositional flow. I therefore suggest we revert the naming to what it was before this PR and revisit at a later stage. From what I understand, such a resolution will allow us to merge this PR with little extra effort.
> > Does this make sense to you?
> 
> Yes, I think this is a good solution!

It's a bit of a pity that we failed to include this trivial change here, are you sure about this?
To my understanding, the target revision to return to would be [6016bf2](https://github.com/pmgbergen/porepy/pull/1172/commits/6016bf2fc329fc3e111deac96fdd14beac7eee20), correct?

Let me just clarify on thing regarding the booleans:
They are decorated with `@property`, hence not callable. The value is computed based on the index value, and it's `True` for *all* instances at some index, other than the current one.
Implementing a callable check by passing an index, makes to me little sense, primarily because this is a duplicate functionality (the user can simply check the value of the index, which is a public property of these operators)

keileg:
> It's a bit of a pity that we failed to include this trivial change here, are you sure about this?

I agree it would be better to make a change we are all happy with now and be done with it. Problem is, we have not yet managed to decide precisely what we want to do, despite a week of discussions. My experience in such situations is, we can go on until we reach a point of exhaustion and make a decision that has a high risk of being suboptimal. The better option usually is to cut the losses for now, and revisit the topic at a later point, so that is what I suggest we do.

> To my understanding, the target revision to return to would be [6016bf2](https://github.com/pmgbergen/porepy/pull/1172/commits/6016bf2fc329fc3e111deac96fdd14beac7eee20), correct?

Yes.

vlipovac:
> I agree it would be better to make a change we are all happy with now and be done with it. Problem is, we have not yet managed to decide precisely what we want to do, despite a week of discussions. My experience in such situations is, we can go on until we reach a point of exhaustion and make a decision that has a high risk of being suboptimal. The better option usually is to cut the losses for now, and revisit the topic at a later point, so that is what I suggest we do.

Reverted the changes two commits back.

**For future reference**

Reverted commits with changes in naming:

[051e645](https://github.com/pmgbergen/porepy/pull/1172/commits/051e6452b9368061132f6bcefb6cb5cf3d8c0752)
[7cc6c22](https://github.com/pmgbergen/porepy/pull/1172/commits/7cc6c2267e4f52b78488a78c30ec87cb3d528931)

Issue #1178, opened 2024-06-05
Robin conditions in Models


IvarStefansson:
Design suggestion for introduction of Robin conditions:

1. Add robin_operator parameter to _combine_boundary_operators
2. Refactor each call to _combine_boundary_operators into new methods combine_boundary_operators_darcy_flux etc.
3. Assuming for now that the Neumann and Dirichlet operators are fixed (e.g. "pressure"), create new methods robin_boundary_operator_darcy_flux etc., allowing for targeted tailoring.
4. Figure out the default for 3. - possibly simply passing zeros.

PR #1187, opened 2024-06-10
1178 robin conditions in models

IngridKJ:
## Proposed changes

Addresses issue #1178: Allow for usage of Robin type boundary conditions with the model class setups in PorePy.

This PR includes adding a robin operator parameter to the method _combine_boundary_operators(). In addition to this, each call to _combine_boundary_operators() is refactored into new methods specific for each physical problem, distinguished by their flux name:
* combine_boundary_operators_darcy_flux()
* combine_boundary_operators_fourier_flux()
* combine_boundary_operators_mechanical_stress()

Additionally: Fixes a problem of singular matrices related to nodes with only one subcell gradient but two or more boundary faces (for Robin conditions. This was already in place for Neumann). 

### Practicalities
The Robin conditions are implemented in the following form in PorePy: flux + alpha * potential = G. 
What is needed by the user to utilize such boundary conditions is to: 
In bc_type-methods: 
* Use the flag "rob" when setting boundary condition type, analogous to setting Dirichlet ("dir") and Neuman ("neu").
* Assign the Robin weight, alpha. The Robin weight is an attribute of the boundary condition class (.robin_weight). See the `__init__`-methods of BoundaryCondition/BoundaryConditionVectorial (porepy/params/bc.py) for an example of what the weight should look like.

In bc_values-methods:
* Assign the right hand side G. G has units of flux, and is set by bc_values_"flux_name"() - same as the Neumann values. As both BC type values are set by the same method, it is important to make sure the Neumann and Robin values are assigned to the correct boundary sides.

Why refactoring the combine operators call? 
Tailoring of each operator is now easily achieved by just overriding the wrapper method (combine_boundary_operators_"flux_name"). Note: The default Robin operator is the same as the Neumann operator.

## Types of changes

- [ ] Other:

## Checklist


OmarDuran (changes requested):
Thank you for your contribution. 

It is important to consider the following:

- Perform a test to ensure that boundary segments labeled with Robin type do not intersect with other types;

- Include a test for scalar and vector problem, for which Robin approximates the functionality of Dirichlet and Neumann;

- Include a tutorial that illustrates how the boundary condition is intended to be used in the models.

src/porepy/applications/test_utils/models.py
```diff
@@ -118,6 +118,87 @@ class Model(geometry, model_class):
     return model
 
 
+class ThreeBoundaryConditionTypes:
```
OmarDuran, line 121: Consider renaming it to `BoundaryConditionTypes`.

tests/models/test_thermoporomechanics.py
```diff
@@ -265,6 +265,77 @@ def test_positive_p_frac_positive_opening():
     assert np.allclose(t_frac, 8.3e-6, atol=1e-7)
 
 
+def test_robin_boundary_flux():
```
OmarDuran, line 268: For clarity, consider explicit assignment of the `robin_weight` parameter.

OmarDuran:
The question is general in nature. I believe that there is room for a unification of names and types, such as

Types:
- Dirichlet_type
- Neumann_type
- Robin_type

Data:
- Dirichlet_data
- Neumann_data
- Robin_data

IvarStefansson:
> Thank you for your contribution.
> 
> It is important to consider the following:
> 
> * Perform a test to ensure that boundary segments labeled with Robin type do not intersect with other types;
> * Included is a test for a scalar and vector problem, for which Robin approximates the functionality of Dirichlet and Neumann;
> * Include a tutorial that illustrates how the boundary condition are intended to be used in the models.

To your first point, @OmarDuran, isn't that more a test for the `BoundaryCondition` class, rather than this Model integration?

tests/models/test_boundary_condition.py
```diff
@@ -105,3 +105,204 @@ def test_boundary_condition_mixin(t_end: int):
...
+    def bc_values_displacement(self, bg: pp.BoundaryGrid) -> np.ndarray:
+        """Assigns displacement values in the x-direction of the west boundary."""
+        values = np.zeros((self.nd, bg.num_cells))
+        bounds = self.domain_boundary_sides(bg)
+        values[0][bounds.west] += np.ones(len(values[0][bounds.west]))
+        return values.ravel("F")
+
+    def bc_values_pressure(self, bg: pp.BoundaryGrid) -> np.ndarray:
+        """Assigns pressure values on the west boundary."""
+        values = np.zeros(bg.num_cells)
```
IvarStefansson, line 137: Wouldn't it be even better to test for nonzero values?

tests/models/test_boundary_condition.py
```diff
@@ -105,3 +105,204 @@ def test_boundary_condition_mixin(t_end: int):
         # Projecting the expected value to the subdomain.
         expected_val = bg.projection().T @ expected_val
         assert np.allclose(bc_val_prev_ts[bc_type.is_neu], expected_val[bc_type.is_neu])
+
+
+"""Here follows mixins related to testing of Robin limit cases, and eventually the test itself. """
+
+
+class Geometry:
```
IvarStefansson, line 113: I think we can use `SquareDomainOrthogonalFractures` with the parameter `fracture_indices` equal to `[]`.

tests/models/test_boundary_condition.py
IngridKJ: So, "more nonzero" than what is currently in place? It is technically nonzero now in the sense that I have unit values on the west boundary.

IvarStefansson: Yes, but that's only the Dirichlet one, not the ones you're actually testing, right?

IngridKJ: The values I am testing are values of primary the variables in all the domain cells. So even though the boundary values are zero, the values I am testing are not. But if nonzero also on the other boundaries make more sense, I can change it.

IngridKJ:
@IvarStefansson: PR is ready for (hopefully final) review now. I addressed your comment about nonzero boundary values, and after the meeting with you and Eirik today, I have also refactored the eliminate-asymmetric-stress-tensor functionality for Neumann/Robin.

src/porepy/models/constitutive_laws.py
```diff
@@ -1765,6 +1776,34 @@ def potential_trace(
         )
         return pressure_trace
 
+    def combine_boundary_operators_tpfa_flux(
```
IvarStefansson, line 1779: I think we need access to the separate methods for Darcy and Fourier, probably passed as arguments to diffusive_flux. Let's discuss if you disagree.

src/porepy/models/constitutive_laws.py
```diff
@@ -2715,6 +2745,29 @@ def fourier_flux(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
         flux.set_name("Fourier_flux")
         return flux
 
+    def combine_boundary_operators_fourier_flux(
+        self, subdomains: list[pp.Grid]
+    ) -> pp.ad.Operator:
+        """Combine fourier flux boundary operators.
```
IvarStefansson, line 2751: Please capitalize Fourier and Darcy throughout.

src/porepy/models/constitutive_laws.py
```diff
@@ -3428,6 +3474,30 @@ def mechanical_stress(self, domains: pp.SubdomainsOrBoundaries) -> pp.ad.Operato
         stress.set_name("mechanical_stress")
         return stress
 
+    def combine_boundary_operators_mechanical_stress(
+        self, subdomains: list[pp.Grid]
+    ) -> pp.ad.Operator:
+        """Combine mechanical stress boundary operators.
```
IvarStefansson, line 3480: Should we explain in the documentation of these methods that they can be used to define other Robin operators? I.e., not use `mechanical_stress` for both neu and rob.

src/porepy/models/energy_balance.py
```diff
@@ -355,6 +355,7 @@ def enthalpy_dirichlet(boundary_grids):
                 subdomains=subdomains,
                 dirichlet_operator=enthalpy_dirichlet,
                 neumann_operator=self.enthalpy_flux,
+                robin_operator=self.enthalpy_flux,
```
IvarStefansson, line 358: Are you sure this is relevant for advective fluxes? I don't think Upwind handles Robin.

src/porepy/numerics/fv/mpsa.py
```diff
@@ -832,8 +832,23 @@ def _create_inverse_gradient_matrix(
         # have at least one internal face connected to the node you are should be safe.
         # For the Neumann faces we eliminate the asymetic part this does in fact lead to
         # an inconsistency.
-        self._eliminate_ncasym_neumann(
-            ncasym, subcell_topology, bound_exclusion, cell_node_blocks, nd
+        self._eliminate_ncasym(
```
IvarStefansson, line 835: Great catch!

tests/models/test_boundary_condition.py
```diff
@@ -102,3 +108,205 @@ def test_boundary_condition_mixin(t_end: int):
...
+            "rob",
+        )
+        bc.is_rob[:, bounds.west] = False
+        bc.is_dir[:, bounds.west] = True
+
+        r_w = np.tile(np.eye(sd.dim), (1, sd.num_faces))
+        bc.robin_weight = np.reshape(r_w, (sd.dim, sd.dim, sd.num_faces), "F") * 0
+        return bc
+
+    def bc_type_scalar_rob_dir(self, sd: pp.Grid) -> pp.BoundaryCondition:
```
IvarStefansson, line 163: If you rename this method to e.g. `bc_type_scalar`, you can let the below class inherit from this, avoiding redefinition of `bc_type_x_flux` methods.

tests/models/test_boundary_condition.py
```diff
@@ -102,3 +108,205 @@ def test_boundary_condition_mixin(t_end: int):
...
+        "neu": run_model(MassAndEnergyBalanceNeu),
+    }
+    return models
+
+
+@pytest.mark.parametrize(
+    "balance_type", ["mass_balance", "momentum_balance", "mass_and_energy_balance"]
+)
+def test_robin_limit_case(run_models, balance_type):
+    """Test that Robin conditions are equivalent to Neumann with Robin weight = 0.
```
IvarStefansson, line 293: Nice documentation. Thanks!

src/porepy/applications/test_utils/models.py
```diff
@@ -118,6 +118,91 @@ class Model(geometry, model_class):
     return model
 
 
+class BoundaryConditionTypes:
```
IvarStefansson, line 121: Please comment on whether this class has overlap with existing ones.

IvarStefansson (commented):
Good work! Some relatively minor comments.

tests/models/test_boundary_condition.py
OmarDuran: In case I am not mistaken, you are testing one limit only, alpha -> 0 (Neuman). However, I believe that it would be relevant to include a test for alpha -> inf (Dirichlet), numerically large values {1e+12, 1e+14, 1e+16}.

tests/models/test_boundary_condition.py
IngridKJ: Maybe I misunderstand something here, but I do not see how that will help the case. The bc_type_scalar methods are different (one assigns Neumann, one assigns Robin), so redefinition is needed in some sense anyways? Or?

tests/models/test_boundary_condition.py
IngridKJ: Makes sense to include both. I am not sure if I will be able to finish it before I go off on holidays, but I have it on my to do list.

src/porepy/models/energy_balance.py
IngridKJ: Not sure, no. I do not know nearly enough about this, so when I went through the code after the Robin-relevant changes, I just adjusted all ``_combine_boundary_ooperators`` calls to match the new signature. If it is not relevant for certain fluxes, maybe the robin_operator in ``_combine_boundary_ooperators`` can be optional to pass?

src/porepy/applications/test_utils/models.py
IngridKJ: Not that I could find, no.

PR #1201, opened 2024-06-27
Improvements and fixes for the dynamic time stepper

Yuriyzabegaev:
## Proposed changes

Fixed a bug of dynamic time stepping when a time step failed right before meeting the schedule. Also improved documentation a bit.

Update: This also addresses the issue with resetting the iterate values after an unsuccessful time step.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/solution_strategy.py
+++ b/src/porepy/models/solution_strategy.py
@@ -498,6 +498,11 @@ class SolutionStrategy(abc.ABC):
             # Note: It will also raise a ValueError if the minimal time step is reached.
             self.time_manager.compute_time_step(recompute_solution=True)
 
+            # Reset the iterate values. This ensures that the initial guess for an
+            # unknown time step equals the known time step.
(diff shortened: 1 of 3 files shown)


Yuriyzabegaev:
I had to change the behavior of the `test_final_simulation_time`. Previously, it expected `None` which signalized that simulation is over if the simulation reached the end time, even if we asked to recompute solution. This behavior is incorrect, because it leads to the bug I'm fixing: We did not accept the solution at the last time step and want to recompute it, but the time manager stops the simulation. 

This opens another question: Should it return `None` if we ask to recompute the solution, and the simulation time for some reason extended the end time? Previously, it would stop the simulation. I believe that it's also not expected, so now it will recompute the solution if we ask for it, regardless of how close we are to the end time, or if we extended it. This shouldn't happen in a real simulation anyway.

Yuriyzabegaev:
The last commit aims to fix #1205.

keileg (approved):

PR #1202, opened 2024-06-28
1196-adaptivity-of-solutionstatistics

### LINKED ISSUE #1196: Adaptivity of SolutionStatistics
Currently, the statistics object is set in SolutionStrategy.__init__():
`self.nonlinear_solver_statistics = pp.SolverStatistics()`
The statistics to be stored are defined as class attributes. Thus, there is no support for expanding the set of attributes. I suggest changing to 
`self.nonlinear_solver_statistics = self.params.get("nonlinear_solver_statistics", pp.SolverStatistics)()`,
allowing me to pass 
```
class MySolverStatistics(pp.SolverStatistics):
   foo_error: field(default_factory=list)
   bar_attribute: field(default_factory=list)

   def log_error():
      super().log_error()
      self.foo_error.append(42)
```
and adding to `bar_attribute` in any other method with access to the model (e.g. when computing line search weights).

### LINKED ISSUE #1200: Lagging information about the number of Newton iterations
The information about how many Newton iterations were made is lagging one iteration behind. Consider [this line](https://github.com/pmgbergen/porepy/blob/9afd9001b45741b602ead379162d0b7fabf765fe/src/porepy/numerics/nonlinear/nonlinear_solvers.py#L121), we do the following:

1. `newton_step`
2. `after_nonlinear_convergence`, which inside updates the next time step value based on the number of iterations
3. `iteration_counter += 1`

If Newton converges after 1 linear system solved, the time manager will get the information that Newton did 0 iterations. It is easy to fix this by swapping the steps 2 and 3.

IvarStefansson:
## Proposed changes

This PR resolves #1196 by allowing passing of a solver statistics class, which can be a tailored version of `SolverStatistics`. Also resolves #1200 by swapping order of operations inside a Newton iteration.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/solution_strategy.py
+++ b/src/porepy/models/solution_strategy.py
@@ -68,6 +68,8 @@ class SolutionStrategy(abc.ABC):
     :class:`~porepy.viz.exporter.Exporter`.
 
     """
+    nonlinear_solver_statistics: pp.SolverStatistics
+    """Solver statistics for the nonlinear solver."""
     update_all_boundary_conditions: Callable[[], None]
     """Set the values of the boundary conditions for the new time step.
     Defined in :class:`~porepy.models.abstract_equations.BoundaryConditionsMixin`.
--- a/src/porepy/models/solution_strategy.py
+++ b/src/porepy/models/solution_strategy.py
@@ -12,7 +12,7 @@ import logging
 import time
 import warnings
 from pathlib import Path
-from typing import Any, Callable, Optional, Type
+from typing import Any, Callable, Optional
 
 import numpy as np
 import scipy.sparse as sps
--- a/src/porepy/models/solution_strategy.py
+++ b/src/porepy/models/solution_strategy.py
@@ -8,11 +8,12 @@ keep them separate, to avoid breaking existing code (legacy models).
 from __future__ import annotations
 
 import abc
+from dataclasses import dataclass
 import logging
 import time
 import warnings
 from pathlib import Path
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Optional, Type
 
 import numpy as np
 import scipy.sparse as sps
@@ -183,11 +184,7 @@ class SolutionStrategy(abc.ABC):
         is adjusted during simulation. See :meth:`before_nonlinear_loop`.
 
         """
-
-        self.nonlinear_solver_statistics = self.params.get(
-            "nonlinear_solver_statistics", pp.SolverStatistics
-        )()
-        """Statistics object for non-linear solver loop."""
+        self.set_solver_statistics()
 
     def prepare_simulation(self) -> None:
         """Run at the start of simulation. Used for initialization etc."""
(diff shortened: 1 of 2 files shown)


IvarStefansson:
I suggest @Yuriyzabegaev has the first look, before we involve @keileg for the code owner review.

src/porepy/models/solution_strategy.py
```diff
@@ -183,9 +183,7 @@ def __init__(self, params: Optional[dict] = None):
         is adjusted during simulation. See :meth:`before_nonlinear_loop`.
 
         """
-
-        self.nonlinear_solver_statistics = pp.SolverStatistics()
```
Yuriyzabegaev, line 187: When we remove it from the init, I believe, type checkers might not be able to figure out what type it is. Correct me if I'm wrong.

I suggest adding a type annotation for it on the class level in SolutionStrategy: `nonlinear_solver_statistics: pp.SolverStatistics` and a docstring below.

Yuriyzabegaev (approved):
Looks good to me, I just added one minor suggestion related to annotations. Good work!

src/porepy/models/solution_strategy.py
IvarStefansson: As a class attribute, you mean?

src/porepy/models/solution_strategy.py
```diff
@@ -228,6 +228,30 @@ def set_discretization_parameters(self) -> None:
...
+        may be added in the future.
+
+        """
+        # Retrieve the value with a default of pp.SolverStatistics
+        statistics_candidate = self.params.get(
+            "nonlinear_solver_statistics",
+        )
+        # Explicitly check if the retrieved value is a class and a subclass of
+        # pp.SolverStatistics for type checking.
+        if isinstance(statistics_candidate, type) and issubclass(
```
keileg, line 246: Two questions:
1. If I want to use a custom statistics object, this forces me to subclass from `pp.SolverStatistics` and explicitly disallows duck-typing, which I believe is the more pythonic approach. Comments?
2. Is there a reason for not raising an error instead of silently overriding a user's failed attempt at making a custom class? I can see this leading to a lot of confusion.

src/porepy/models/solution_strategy.py
```diff
@@ -228,6 +228,30 @@ def set_discretization_parameters(self) -> None:
 
         """
 
+    def set_solver_statistics(self) -> None:
+        """Set the solver statistics object.
+
+        This method is called at initialization. It is intended to be used to
+        set the solver statistics object(s). Currently, the solver statistics
+        object is related to nonlinearity only. Statistics on linear solvers
```
keileg, line 236:
```suggestion
        object is related to nonlinearity only. Statistics on other parts of the solution process, such as linear solvers,
```
No way that fits with the line length requirement, but then the whole paragraph probably would do with a reformatting.

keileg (commented):
I am not sure about the requirements set on a user-provided SolverStatistics, see questions.

src/porepy/models/solution_strategy.py
IvarStefansson: 1. I think allowing any (data)class will be difficult, since there is a contract about a save method called by the SolutionStrategy.
2. I think you're right. I'll do as you suggest.

src/porepy/models/solution_strategy.py
Yuriyzabegaev: Yes

keileg (approved):

PR #1204, opened 2024-07-01
Make the code compatible with numpy v2.0, scipy v1.14

keileg:
## Proposed changes
Though this PR changes a lot of lines, they are limited to the following cases:
* To convert a  `scipy.sparse.spmatrix` to a dense array, use `A.toarray()` instead of `A.A` (the latter is deprecated)
* The output from `numpy.unique(..., return_inverse=True)` can now be 2d. The code was updated to reflect this.
* Use `numpy.isin` rather than the deprecated `numpy.in1d`.
* Use `numpy.nan` instead of `numpy.NaN` (changed a single occurence)
* Updated documentation of a single function in `utils/setmembership.py`. This is the only change that was not directly linked to the updates of numpy and scipy versions.

Notes:
* The code should be compatible with both `numpy 1.X` and `2.X`
* While the test suite would benefit from `black` formatting, I did not do so in this PR, since this would have made it very difficult to get an overview of the changes. Instead I suggest to do this in a dedicated PR, and then probably require that the test folders are compatible with `black`  as part of the static tests.

## Types of changes

- [ ] Other:

## Checklist

- [ ] Static typing is included in the update. NOT RELEVANT

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,9 +4,9 @@ matplotlib
 meshio >= 5.0
 networkx
 numba >= 0.57
-numpy < 2.0
+numpy
 requests
-scipy < 1.14
+scipy
 seaborn
 shapely
 six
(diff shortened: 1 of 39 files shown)


IvarStefansson (approved):
Approving on faith, given type and volume of changes.

Issue #1179, opened 2024-06-07
Implement hash functions in Ad operators


keileg:
# Overview
We should implement hash functions for all Ad operators. The motivation is primarily to allow us to identify operators that are identical and thereby save time and memory during operator tree parsing. The hash will also allow us to store values of operators.

The implementation itself should be easy, see below. The potentially hard part will be to decide a way to define a suitable hash key for all types of operators.

# Prototype implementation
A few ideas are sketched below, but the code may well contain logical flaws

```python
class MergedOperator:

    def __key(self):
      tmp = [g.id for g in self.domain]
      tmp += [self.key]   # self.key and self.__key() is problematic
      tmp += [self.discr.keyword]
      return tuple(tmp)

    def __hash__(self):
        return hash(self.__key())

class Operator:

    def __key(self):
      tmp = [child for child in self.children]
      tmp += [self.operation]
      return tuple(tmp)

    def __hash__(self):
        return hash(self.__key())
```

jwboth:
Task: Identify a suitable hash function for all Ad operators, with identical values iff evaluation can be stored and reused. Implement. Design suggestions are available in #1179. Proper test coverage is critical. 

Outcome: Implementation of hash functions, with testing. Design principles are available for new classes of Ad operators.

PR #1215, opened 2024-08-16
FEAT: Hashing

Yuriyzabegaev:
## Proposed changes

This addresses #1179. Note that I have not implement hashing for the AD functions, since they are not used by the basic models. This however should not break anything, since right now nothing relies on the hashes.

Sidenote: I considered the original idea to name the new method `__key` with double underscore, however, it leads to the issue. Consider the following code: 
```
class Parent:

    def __hash__(self) -> int:
        return hash(self.__key())

    def __key(self):
        return 'hi from parent'
    

class Child(Parent):

    def __key(self):
        return 'hi from child'
    

p = Parent()
print(hash(p))

c = Child()
print(hash(c))
```

The hashes of the parent and the child will be the same (not as we want). This is because the method `__hash__` defined only for the parent always calls the parent's `__key` due to [name mangling in Python](https://stackoverflow.com/a/1301369). The workaround would be either to repeat the `__hash__` method for each child class or to replace the double underscore with the single one. I have chosen the latter. Please let me know if I understood the idea incorrectly.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/ad/operators.py
+++ b/src/porepy/numerics/ad/operators.py
@@ -1469,9 +1469,19 @@ class SparseArray(Operator):
     def _compute_spmatrix_hash(mat: sps.spmatrix) -> str:
         """Utility function that handles all the sparse formats to compute the hash."""
         # proper handling of all the formats like csr, coo, etc..
-        if isinstance(mat, (sps.csr_matrix, sps.csc_matrix, sps.bsr_matrix)):
+        if isinstance(
+            mat,
+            (
+                sps.csr_matrix,
+                sps.csr_array,
+                sps.csc_matrix,
+                sps.csc_array,
+                sps.bsr_matrix,
+                sps.bsr_array,
+            ),
+        ):
             properties = [mat.data, mat.indices, mat.indptr]
-        elif isinstance(mat, sps.coo_matrix):
+        elif isinstance(mat, (sps.coo_matrix, sps.coo_array)):
             properties = [mat.data, mat.row, mat.col]
         else:
             raise NotImplementedError("Hashing not provided for the format", type(mat))
--- a/src/porepy/numerics/ad/operators.py
+++ b/src/porepy/numerics/ad/operators.py
@@ -1470,9 +1470,9 @@ class SparseArray(Operator):
         """Utility function that handles all the sparse formats to compute the hash.
 
         The hash will match for two sparse arrays with the identical: sparse formats,
-        data arrays, rows and columns arrays. The identical matrices in the different
-        formats will have different hashes. Such behavior is expected, since if two
-        different formats are used, they are likely used for a good (efficiency) reason.
+        shapes, data arrays, rows and columns arrays. The identical matrices in the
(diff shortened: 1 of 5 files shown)


keileg (commented):
With regards to the key function, I see the point; how about _key?

src/porepy/numerics/ad/operators.py
```diff
@@ -1431,6 +1465,20 @@ def shape(self) -> tuple[int, int]:
         """Shape of the wrapped matrix."""
         return self._shape
 
+    @staticmethod
+    def _compute_spmatrix_hash(mat: sps.spmatrix) -> str:
+        """Utility function that handles all the sparse formats to compute the hash."""
+        # proper handling of all the formats like csr, coo, etc..
+        if isinstance(mat, (sps.csr_matrix, sps.csc_matrix, sps.bsr_matrix)):
```
keileg, line 1472: Scipy encourages gradual introduction of csr_array etc., see https://docs.scipy.org/doc/scipy/reference/sparse.html, which will break this logic. Please account for arrays as well as matrices.

keileg (commented):
Minor comments only. I see that you went for _key instead of __key (and key); this is fine with me.

src/porepy/numerics/ad/operators.py
Yuriyzabegaev: Done

src/porepy/numerics/ad/operators.py
```diff
@@ -1431,6 +1465,30 @@ def shape(self) -> tuple[int, int]:
...
         return self._shape
 
+    @staticmethod
+    def _compute_spmatrix_hash(mat: sps.spmatrix) -> str:
+        """Utility function that handles all the sparse formats to compute the hash."""
+        # proper handling of all the formats like csr, coo, etc..
+        if isinstance(
+            mat,
+            (
+                sps.csr_matrix,
```
keileg, line 1475: This will fail for dia arrays / matrices. Is there a way to do this without having to explicitly treat the various types matrices?

src/porepy/numerics/ad/operators.py
```diff
@@ -1431,6 +1465,30 @@ def shape(self) -> tuple[int, int]:
         """Shape of the wrapped matrix."""
         return self._shape
 
+    @staticmethod
+    def _compute_spmatrix_hash(mat: sps.spmatrix) -> str:
+        """Utility function that handles all the sparse formats to compute the hash."""
```
keileg, line 1470: As a service to the future us, perhaps give some reasoning behind the design of the hash for this class. It is less obvious than the other classes.

keileg (commented):
Minor comments only.

src/porepy/numerics/ad/operators.py
Yuriyzabegaev: Provided, please take a look

keileg (approved):
Looks good now; I don't think we will find any remaining issues before they show up in live code.

PR #1229, opened 2024-09-17
Turn off redundant residual computations if tolerance np.inf

jwboth:
## Proposed changes

This is a minor change. Currently, the residual is always computed as part of the convergence check. However, in the case when the tolerance for the residual is `np.inf`, which is also the default value, this computation is redundant. This PR suggests turning off the unnecessary computation. This leads to some design choice how to forward empty residuals and norms. There may be other solutions. I am open for comments.

## Types of changes

- [ ] Other:

## Checklist

- [X] If new skipped tests have been introduced in this PR, `pytest` was run with the `--run-skipped` flag.


src/porepy/models/solution_strategy.py
```diff
@@ -511,7 +511,7 @@ def after_simulation(self) -> None:
     def check_convergence(
         self,
         nonlinear_increment: np.ndarray,
-        residual: np.ndarray,
+        residual: Optional[np.ndarray],
```
IvarStefansson, line 514: I don't think optional arguments should appear before required arguments in a signature.

src/porepy/models/solution_strategy.py
```diff
@@ -520,7 +520,7 @@ def check_convergence(
         Parameters:
             nonlinear_increment: Newly obtained solution increment vector
             residual: Residual vector of non-linear system, evaluated at the newly
-            obtained solution vector.
+            obtained solution vector. Potentially None, if the residual is not needed
```
IvarStefansson, line 523:
```suggestion
            obtained solution vector. Potentially None, if the residual is not needed.
```
Isn't there also a lack of indentation for this line?

src/porepy/models/solution_strategy.py
```diff
@@ -593,7 +599,11 @@ def compute_residual_norm(
             float: Residual norm.
 
         """
-        residual_norm = np.linalg.norm(residual) / np.sqrt(residual.size)
+        residual_norm = (
```
IvarStefansson, line 602: In general, I think we have tried to avoid long single-expression conditional statements for readability. Is that correct, @keileg?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -87,6 +88,7 @@ def newton_step() -> None:
             nonlocal nonlinear_increment
             nonlocal reference_residual
             nonlocal residual_norm
+            nonlocal residual
```
IvarStefansson, line 91: Why is this needed now and not previously? Same question for line 73.

IvarStefansson (commented):
I like the idea to avoid computation. Please consider the comment about optional arguments in particular.

src/porepy/models/solution_strategy.py
jwboth: `residual` is still required and not optional. The typing just indicates that the argument could have value `None`, which is here indeed the case if the residual is not computed for the computation of the norm.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
jwboth: Now it may be the case, that the residual is not recomputed. In this case, the value outside of the internally defined `newton_step` needs to be forwarded, this would be then the default value. An alternative (maybe cleaner?) could be to remove line 73 and 91, and instead set residual to `None` when the computation is not needed, i.e., in an `else` case after line 112. I actually would opt for that solution. Any opinion?

src/porepy/models/solution_strategy.py
jwboth: You are right, I will fix this in the end.

src/porepy/models/solution_strategy.py
jwboth: No strong opinion here, and will follow your guideline after you settled this - this info in itself will be of interest to me. 

Indeed, writing out the if-else conditional, will not take more lines of code. I guess I am just somewhat a fan of these "one-liners".

keileg: Yes, the preferrence is for explicit if-elses in such cases; for some of us, resolving these nested expressions can be mentally taxing.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -103,9 +105,11 @@ def newton_step() -> None:
             nonlinear_increment = self.iteration(model)
 
             model.after_nonlinear_iteration(nonlinear_increment)
-            # Note: The residual is extracted after the solution has been updated by the
-            # after_nonlinear_iteration() method.
-            residual = model.equation_system.assemble(evaluate_jacobian=False)
+            if self.params["nl_convergence_tol_res"] is not np.inf:
+                # Note: The residual is extracted after the solution has been updated by the
+                # after_nonlinear_iteration() method. This is only required if the residual
```
keileg, line 110: Please fix line length.

src/porepy/models/solution_strategy.py
IvarStefansson: Sorry, you're quite right!

src/porepy/numerics/nonlinear/nonlinear_solvers.py
IvarStefansson: I see. Yes, I think I agree that the else option is better.

IvarStefansson: Just to check my understanding: Would it make sense to move the line 112 call to inside check_convergence?

IvarStefansson (commented):
With your clarification of Optional, the rest should be easy.

jwboth:
Sorry for the massive delay...I fixed the remaining issues and followed all of your suggestions.

keileg:
I think this should be fine now, but please have a short look as well @IvarStefansson.

IvarStefansson (approved):
Looks good to me. Sorry about the delay - I didn't notice the mention.

Issue #1219, opened 2024-08-30
Placement of the iteration counter in the Newton solver


mariusnevland:
At the moment, when performing Newton iterations, the iteration counter is updated after the convergence check. This leads to the final iteration not being counted, as you exit the Newton loop before the counter is updated.

Is there a specific reason for the iteration counter being after the convergence check? If not, I would suggest to have the counter before the check.

mariusnevland:
On second thought, I suppose the iteration_counter variable is now redundant, since this is now handled by SolverStatistics (which updates the iteration counter before the convergence check). Hence, we should probably remove iteration_counter and replace its usage everywhere by the num_iteration attribute of SolverStatistics.

Yuriyzabegaev:
#1200 may be related.

jwboth:
Tasks: Consider essentially the second comment by @mariusnevland. `NewtonSolver` should not keep track of iteration counts anymore. Instead, the nonlinear solution statistics object should be consulted. Thus, the task is to clean up the NewtonSolver.

Please also have a look at how the iteration counter is forwarded to `after_nonlinear_convergence`. My intuitive thought is that the explicit forwarding can be removed and again replaced by explicit reference to the `nonlinear_solution_strategy` .

PR #1232, opened 2024-09-27
Remove iteration counter and norms that are already covered by SolverStatistics

pschultzendorff:
Changes proposed by @IngridKJ and I. 

## Proposed changes

Adresses issues #1219 and #1200.

``model.nonlinar_solver_statistics`` already keeps track of the number of nonlinear iterations, residual norms, and nonlinear increment norms. Thus, any explicit passes of these values between ``SolutionStrategy.after_nonlinear_convergence``, ``SolutionStrategy.check_convergence``, and ``NewtonSolver.solve`` are removed. Conveniently, this fixes ``iteration_counter`` lagging behind by one step. ``NewtonSolver.solve`` only returns one boolean value ``converged`` now.

On an additional note, we noticed that the test suite does not cover the case of divergence due to nan values in ``SolutionStrategy.check_convergence``, i.e., lines 547-551 in ``solution_strategy.py``. We only noticed a mistake in our changes due to flake complaining, not the tests. Any opinions on that would be welcome :)

We added tests for ``SolutionStrategy.check_convergence`` and ``NewtonSolver.solve``.

## Types of changes

- [ ] Other:

Not quite sure how to categorize this, as it is only code cleanup, but any custom scripts depending on ``iteration_counter`` and ``check_convergence`` will break.

## Checklist

### DIFF
--- a/src/porepy/numerics/nonlinear/nonlinear_solvers.py
+++ b/src/porepy/numerics/nonlinear/nonlinear_solvers.py
@@ -58,19 +58,13 @@ class NewtonSolver:
             A 2-tuple containing:
 
             bool:
-
                 True if the solution is converged.
             int:
-
                 Number of iterations used.
 
         """
         model.before_nonlinear_loop()
 
-        iteration_counter = 0
-
-        residual_norm = np.inf
-        nonlinear_increment_norm = np.inf
         is_converged = False
         is_diverged = False
         nonlinear_increment = model.equation_system.get_variable_values(
@@ -86,14 +80,13 @@ class NewtonSolver:
             # Bind to variables in the outer function
             nonlocal nonlinear_increment
             nonlocal reference_residual
-            nonlocal residual_norm
-            nonlocal nonlinear_increment_norm
             nonlocal is_converged
             nonlocal is_diverged
 
             # Logging.
             logger.info(
-                f"Newton iteration number {iteration_counter}"
+                "Newton iteration number "
+                + f"{model.nonlinear_solver_statistics.num_iteration}"
                 + f" of {self.params['max_iterations']}"
             )
 
@@ -107,28 +100,24 @@ class NewtonSolver:
             # after_nonlinear_iteration() method.
             residual = model.equation_system.assemble(evaluate_jacobian=False)
 
-            residual_norm, nonlinear_increment_norm, is_converged, is_diverged = (
-                model.check_convergence(
-                    nonlinear_increment, residual, reference_residual, self.params
-                )
+            is_converged, is_diverged = model.check_convergence(
+                nonlinear_increment, residual, reference_residual, self.params
             )
 
         # Progressbars turned off or tqdm not installed:
         if not self.progress_bar or not _IS_TQDM_AVAILABLE:
             while (
-                iteration_counter <= self.params["max_iterations"] and not is_converged
+                model.nonlinear_solver_statistics.num_iteration
+                <= self.params["max_iterations"]
+                and not is_converged
             ):
                 newton_step()
-
-                iteration_counter += 1
 
                 if is_diverged:
                     # The nonlinear solver failure is handled after the loop.
                     break
                 elif is_converged:
-                    model.after_nonlinear_convergence(
-                        iteration_counter=iteration_counter
-                    )
+                    model.after_nonlinear_convergence()
                     break
 
         # Progressbars turned on:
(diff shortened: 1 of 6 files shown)


src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -145,19 +134,23 @@ def newton_step() -> None:
...
+                        + f"{model.nonlinear_solver_statistics.num_iteration + 1} of \t"
+                        + f"{self.params['max_iterations']}"
+                        ""
                     )
                     newton_step()
-                    iteration_counter += 1
 
                     solver_progressbar.update(n=1)
+                    # Ignore line being too long, because we would need an additional
+                    # variable to fix this.
```
IvarStefansson, line 151: Do you agree with this, @keileg?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -168,15 +161,13 @@ def newton_step() -> None:
...
-                            iteration_counter=iteration_counter
-                        )
+                        model.after_nonlinear_convergence()
                         break
 
         if not is_converged:
             model.after_nonlinear_failure()
 
-        return is_converged, iteration_counter
+        return is_converged, model.nonlinear_solver_statistics.num_iteration
```
IvarStefansson, line 170: Is it important to return the counter?

IvarStefansson (commented):
Thanks!
Would it make sense to add a test that runs for a fixed number of iterations and than e.g. checks that at the end the iteration count is as expected?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
pschultzendorff: The only uses of ``NewtonSolver.solve`` in PorePy are in ``run_models.py``, which
ignores both return values. We can freely choose what to return. Although, for a
function that is called solve, having some return value other than ``None`` is
probably good practice.

pschultzendorff:
> Would it make sense to add a test that runs for a fixed number of iterations and than e.g. checks that at the end the iteration count is as expected?

Since the issues were about the iteration count being off by one, I like the idea! Will add a test for this.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
IvarStefansson: Yes, I'd keep is_converged. But I think we should remove the counter, which can be retreived from the model's solution statistic object.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
keileg: Yes. Sorry for the late response.

pschultzendorff:
We addressed the comments and added tests.

tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -0,0 +1,61 @@
...
+            < self.expected_number_of_iterations
+        ):
+            converged = False
+            return converged, diverged
+        converged = True
+        return converged, diverged
+
+    def after_nonlinear_failure(self) -> None:
+        """Method which is called if the non-linear solver fails to converge."""
+        self.save_data_time_step()
```
IvarStefansson, line 41: Why is this needed?

tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -0,0 +1,61 @@
...
+from porepy.models.fluid_mass_balance import SinglePhaseFlow
+
+
+class NonlinearSinglePhaseFlow(SinglePhaseFlow):
+    """Class setup which forces a set number of nonlinear iterations to be run for
+    testing."""
+
+    def _is_nonlinear_problem(self):
+        """Method for stating that the problem is nonlinear."""
+        return True
```
IvarStefansson, line 14: Is this not true by default?

tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -0,0 +1,61 @@
...
+def test_nonlinear_iteration_count():
+    """Test for checking if the nonlinear iterations are counted as expected.
+
+    A pre set value of expected iterations is set, and the test checks that the
+    iteration count matches the pre set value after convergence is obtained.
+
+    """
+    params = {"max_iterations": 96}
+    model = NonlinearSinglePhaseFlow()
+    model.expected_number_of_iterations = 13
```
IvarStefansson, line 55: Can't we make this number smaller to avoid unnecessary test time?

IvarStefansson (commented):
Couple of small questions to the new test.

tests/numerics/nonlinear/test_nonlinear_solvers.py
IngridKJ: Hm, apparently it is not needed. I had some error messages when I made test, and including this method (seemingly) removed the problem. I see now that it is not needed, so I will delete it.

jwboth (approved):
Thanks both of you for removing this redundancy and adding a test!

PR #1244, opened 2024-10-21
Extending fluid concept to fit CF & Standard formulations

### LINKED ISSUE #1240: Unifying the fluid concept
With the introduction of `porepy.compositional`, PorePy allows the user to model general multi-component, multi-phase fluid mixtures, including singular cases of 1-component and/or 1-phase.
The fluid in the PorePy models so far is based on a dict-like data structure for implementing heuristic laws, without the notion of phases and components though. Since this case can be viewed as the 1-phase, 1-component fluid (mixture), unifying the concept of fluids and the respective API for PorePy-models would reduce duplicate code and make PorePy easier to learn by providing only one instead of two concepts.

### Current state

PorePy has the concept of `FluidConstants` included in the modelling framework, upon which various constitutive-law-classes are built to provide domain-based methods like `fluid_density: list[Grid] -> pp.ad.Operator`. The starting point is a model attribute `fluid: FluidConstants`, which is mixed into the model by the solution strategy.

#1186 introduced the compositional sub-package, as a preparation for a general compositional flow model coming in #1236 .
CF models have (as of now) an attribute `fluid_mixture: FluidMixture`, which is provided by a separate, mixed-in class, and which bundles relevant terms in the form of e.g., `model.fluid_mixture.density: list[Grid] -> pp.ad.Operator`.

The consequence of this is that the implementations of balance equations for the CF setting, and the existing fluid mass and energy balance, require to large degree duplicate implementations.

### Proposed change

Since the implementation of fluid-related term can be seen as a family of constitutive laws, it is natural to introduce a more general object as a model attribute, which represents the respective aspect of the model implementation-wise (like `model.mdg` provides in essence every grid-related functionality).

More precisely, an extension of the `model.fluid: FluidConstants` to `model.fluid: FluidMixture` and changing the access of fluid-related terms to use this attribute, i.e.,  `model.fluid.density: list[pp.Grid] -> pp.ad.Operator`.

- It obtains an own place in the model set-up (like the creation of the grid).
- It concentrates the jungle of mixed-in methods `model.fluid_*` in the `model.fluid`, and provides a clear interface, **independent of the phase and component configuration.**
- It takes a chunk out of `constitutive_laws.py` and eases the mixin framework and the coming protocols.
- Balance and other equations (like equilibrium) can be unified to handle an arbitrary fluid mixture.

The name `FluidMixture` should probably changed to `Fluid` for simplicity.

### Challenges

1. While the CF setting is more general, simpler heuristics and singular cases are more frequently used in 
mathematics-oriented research. There is a risk of "bloating" the implementation. For example, the density is in general $\sum_j s_j \rho_j$, which reduces in the 1-phase case to $\rho_0$. While we would like to inform the model that it indeed has a (by the user accessible) phase object in the fluid, the unnecessary expansion of the operator tree of `model.fluid.density` is a legitimate concern. This is relevant for every thermodynamic property. If you think that this is only a minor impact, I would appreciate some feedback since it would make this fix easier.

2. The `Phase` class calls for an EoS-oriented implementation, which also includes the usage of the `SurrogateOperator` and storage of 
   computed property values and derivative values in the memory. Since simpler heuristics do not require any of that and PorePy's AD 
   framework is fully capable to handle it, the full machinery as presented in #1236 is not required.

### Solution

The first step is to unify the class `FluidConstants` and the new class `porepy.compositional.ChemicalSpecies`. A result would be a data-class, with functionality handling unit conversion. The stored data includes:

- Identification/naming: a "name" or formula (f.e. `'H2O'), CASs registry number (international dictionary uniquely identifying any molecule, used by all chemistry-based libraries), ...
- thermodynamic constants: critical p/T, acentric factor, ...
- heuristic constants: compressibility, reference p/T/rho

A modification of the `Component` follows in order to handle this new class.

The second step is to adapt the class `FluidMixture`/`Fluid` as the general model attribute `model.fluid`, which already implements the notion of components and phases, and has extensive logic regarding the DOFs for any number of phases and components.

Third, to modify the model set-up in `prepare_simulation` to include a fluid set-up. Note that the creation of DOFS `create_variables` is already implemented and there is an existing mixin for arbitrary mixtures `CompositionalVariables`. It handles the singular cases without introducing redundant variables.

The final step is incorporate various classes in `constitutive_laws.py` and their concepts in the fluid, and this is the point where design becomes challenging.

1. We can either be mathematically and thermodynamically rigorous and implement "heuristic" phase classes, which represent the existing constitutive laws. To avoid the "bloating" of the equations, the `Fluid` class must be modified to handle the singular cases. The fluid is always clearly defined in terms of phases, components and carries in the singular cases some meta-data with it (like the definition of the physical state of the phase - liquid/gas) and leaves flexibility for future development.

2. We can define derived fluid classes which implement the singular case. It would be a "lightweight" version of the full formulation, but the tradeoff is more code/less unification.

In both cases, the existing solution strategies are fine to handle whatever `fluid` brings with it. But usage of EoS-based phases with external computations and `SurrogateOperator`s as phase properties requires an own solution strategy (see for example [composite-flow/compositional_flow.SolutionStrategyCF](https://github.com/pmgbergen/porepy/blob/7e886cef6e1782ed247aea3ec8d1c1716229ab21/src/porepy/models/compositional_flow.py#L2547)).

### Notes

1. The notion of mobility exists currently in the simple form in [develop/constitutive_laws.FluidMobility](https://github.com/pmgbergen/porepy/blob/7bed70b1159addc58fa7d22f61a2108f7bf1d68f/src/porepy/models/constitutive_laws.py#L729), and in the general form in [composite-flow/ compositional_flow.MobilityCF ](https://github.com/pmgbergen/porepy/blob/7e886cef6e1782ed247aea3ec8d1c1716229ab21/src/porepy/models/compositional_flow.py#L87). It is clear that this must remain a separate mixin implementing the mobility terms which access the `model.fluid`. This is strictly speaking neither a constitutive law, nor a fluid property, but related to the balance equations. It can be unified into a single mixin class though.
4. We will probably require a separate, more general solution for mD-specific terms like normal thermal conductivity.
It exists as a constant given in `FluidConstants` in [develop/constitutive_laws.ThermalConductivityLTE](https://github.com/pmgbergen/porepy/blob/7bed70b1159addc58fa7d22f61a2108f7bf1d68f/src/porepy/models/constitutive_laws.py#L2557), and as a projection of the regular fluid thm. conductivity from the higher-dimensional domain in [composite-flow/compositional_flow.ThermalConductivityCF](https://github.com/pmgbergen/porepy/blob/7bed70b1159addc58fa7d22f61a2108f7bf1d68f/src/porepy/models/constitutive_laws.py#L2557). This is indeed a constitutive law in its sense and will probably remain in two separate classes.
3. There are currently indeed many mixed-in model methods of form `.fluid_*` from various classes in `constitutive_laws.py`. There is room to simplify some implementation along the way since there are several methods which represent only 1 step in the construction of "higher" properties (f.e., `fluid_compressibility` is only used inside a heuristic for `fluid_density`).

### EDIT

Regarding the point about including ""heuristic" phases to accommodate existing heuristic laws in `constitutive_laws.py`, I just realized that `compostional.compositional_mixins.FluidMixtureMixin` provides hooks to define the thermodynamic properties of a phase. These hooks can be used to provide the existing constitutive laws like `DensityFromPressure`, or even better, to modify the `DensityFromPressure` to overwrite the hook via mixing. So I think this would be a great solution which is less invasive in the existing code.

### LINKED ISSUE #1247: Isolation of Units and conversion functionality
As identified in #1244, an isolation of the units concept would allow for persistent use instead of attaching it both to fluids and solid etc. This is thus requested independent of the discussion in #1244. 

The process shall be performed through a user project with the two tasks of 1. removing the functionality from MaterialConstants and moving to a separate Units location, and 2. updating the code base accordingly.

These changes are important for the user project #1150 on numerical constants.

vlipovac:
Closes #1240
Closes #1247 

## Proposed changes

Reworking the concept of `model.fluid` to be a general fluid with phases and components, instead of just a data collection of constants.

Breaking change with a multiple minor changes throughout the packages. The essential changes are summarized in a self-review.

`porepy.compositional.ChemicalSpecies` was merged into `FluidConstants`, including data class functionality. This led to some changes for `SolidConstants` as well for inheritance reasons.

## Types of changes

- [ ] Other:

## Checklist

- [X] `pytest` was run with the `--run-skipped` flag.


src/porepy/models/solution_strategy.py
```diff
@@ -203,6 +214,12 @@ def prepare_simulation(self) -> None:
         # Order of operations is important here.
         self.set_equation_system_manager()
         self.create_variables()
+        # After fluid and variables are defined, we can define the secondary quantities like
+        # fluid properties (which depend on variables).
+        # NOTE this is critical in the case where properties depend on some fractions.
+        # The callables for those are dynamically created during create_variables, as opposed
+        # to e.g., pressure or temperature.
+        self.assign_thermodynamic_properties_to_phases()
```
vlipovac, line 222: This step was added to the set-up.
It is critical that the Fluid is created in `set_materials`, before the variables, since the dynamic number of variables depends on the fluid definition.
The step `assign_thermodynamic_properties_to_phases` requires the variables to be already created, since fluid properties are in general secondary terms which depend on some variable (keyword SurrogateFactory)

src/porepy/models/fluid_property_library.py
```diff
@@ -0,0 +1,375 @@
+""""Module containing some constant heuristic fluid property implemenations.
```
vlipovac, line 1: This file takes fluid-related heuristics out of the `constitutive_laws.py` file.
Some methods are reworked to fit the general fluid concept, where properties of phases are dynamically created based on the number of modelled phases.

src/porepy/compositional/compositional_mixins.py
```diff
@@ -1095,25 +1129,37 @@ def create_mixture(self) -> None:
 
         self.set_components_in_phases(components, phases)
 
-        self.fluid_mixture = FluidMixture(components, phases)
+        self.fluid = Fluid(components, phases)
 
     def get_components(self) -> list[Component]:
```
vlipovac, line 1134: This method has now a default-behavior, creating a one-component fluid based on the fluid constants passed by the user. Important to mention is that the simulation set-up does not change for the standard 1-phase, 1-component case.
Users still use and pass a FluidConstants instance in params.

src/porepy/compositional/compositional_mixins.py
```diff
@@ -1128,17 +1174,14 @@ def get_phase_configuration(
             :class:`~porepy.compositional.base.Phase`.
 
         """
-        raise CompositionalModellingError(
-            "Call to mixin method. Configure phases by overriding this method."
-        )
+        return [(AbstractEoS(components), PhysicalState.liquid, "liquid")]
```
vlipovac, line 1177: The default behavior of get_phase_configuration was also changed to model a 1-phase fluid with a liquid-like phase.
The AbstractEoS was modified for this purpose, not being abstract anymore but playing the role of the "mathematician's" EoS, i.e. a dummy doing nothing for models using heuristic fluid properties (constitutive_laws.py)

src/porepy/compositional/compositional_mixins.py
```diff
@@ -1171,42 +1211,30 @@ def assign_thermodynamic_properties_to_mixture(self) -> None:
...
             for comp in phase:
                 phase.fugacity_coefficient_of[comp] = self.fugacity_coefficient(
                     comp, phase
                 )
 
-        self.fluid_mixture.density = self.fluid_density
-        self.fluid_mixture.specific_volume = self.fluid_specific_volume
-        self.fluid_mixture.specific_enthalpy = self.fluid_specific_enthalpy
-
     def dependencies_of_phase_properties(
```
vlipovac, line 1238: This is a bit tricky.

I hade to change the default behavior of this, because of the mechanics-only models, who inherit from SolutionStrategy.
Problems include the missing of the classical variable mixins like pressure and temperature.

The default return value is now an empty list, leading to fluid properties which raise an error if they were not overridden by some constitutive law.
It makes sense, but may be hard to digest for people digging deeper into the code.

src/porepy/compositional/base.py
```diff
@@ -957,3 +924,137 @@ def reference_component(self) -> Component:
         """Returns the reference component as designated by
         :meth:`reference_component_index`."""
         return self._components[self.reference_component_index]
+
+    def density(self, domains: pp.SubdomainsOrBoundaries) -> pp.ad.Operator:
```
vlipovac, line 928: The class `Fluid` (formerly FluidMixture) is now the new interface for other model mixin classes to access fluid properties.
It implements the fluid properties required by the PDE classes, while the framework handles the assignment of heuristic functions to the (single) phase properties under the hood.

tests/models/test_fluid_mass_balance.py
```diff
@@ -354,6 +352,12 @@ def test_tested_vs_testable_methods_single_phase_flow(
         ("well_fluid_flux", 0, 2),  # use dim_restriction=2 to ignore well flux
         ("well_flux_equation", 0, 2),  # use dim_restriction=2 to ignore well equation
         ("well_radius", 0.1, None),
+        # Testing of methods with deeper namespace
```
vlipovac, line 355: The testing of the fluid mass balance was modified and adapted to the new interface for fluid properties.
The test was extended to cover the standard model implemented (1 phase 1 component).

vlipovac (commented):
Here is a self-review with the essential changes.

All other changes are essentially due to the change in the interface to fluid properties:

`model.fluid_density()` -> `model.fluid.density` and similar.

Also, the access to FluidConstants and SolidConstants is not done by calling anymore, thanks to the data class framework.

keileg:
Thanks @vlipovac. I'll get started on this later today.

src/porepy/utils/porepy_types.py
```diff
@@ -35,3 +37,23 @@
     "pp.fracs.fracture_network_2d.FractureNetwork2d",
     "pp.fracs.fracture_network_3d.FractureNetwork3d",
 ]
+
```
vlipovac, line 40: I introduced these two types for sake of dynamic creation of compositional variables and phase properties.
Inside `fluid.density()`, the method `phase.density()` is called. While `fluid.density` is a static interface for all other model interfaces, the `phase.density` is created depending on phase context. The static checks and intellisense have now something to work with.

keileg (commented):
Various comments and questions. Some are marked as 'postpone' and can definitely be ignored while discussing the concepts, but that likely also applies to more methods. There may also be more such minor comments at a later stage (I tried to read with an eye on the big picture in this iteration).

My main impression is that the overall design is sound, I don't see any red flags. There are some questions regarding smaller design choices that should be addressed, or explained, but these should not be show stoppers.

I suggest you have a look at the comments @vlipovac and that we then talk through the changes as a second check that I have not overlooked something. If not, I think @IvarStefansson can have a look quite soon.

vlipovac:
@keileg  thanks for the effort!

From your comments I filtered 3 points which are on the more crucial side:

1. `mob_rho_h`: Yes, I would suggest to do that generalization as a next preparation for merger #1236 (to minimize duplicate code in the CF equations)
2. `convert_units`: I agree that the code is cumbersome, due to the attempt to make this merger a bit less breaking. Given your and Ivar's comments about units last Thursday (have a single `Units` instance per simulation and persistently use it), I suggest moving this functionality out of the material constants and into the class `Units`. Each model has a `model.units` attribute, and the usage will be more clear.
3. `fluid` in `MomentumBalance`: This is true, the whole thing works without a fluid and it does not need it. The reason it has it is due to the base `SolutionStrategy` executing `prepare_simulation() -> set_materials() -> create_fluid()`. The last method is mixed in by `FluidMixin` (BTW, also as of now, `MomentumBalance` has a `fluid: FluidConstants` attribute which it does not use). I haven't found a more elegant way without introducing logic, which might be even more conflicting when mixing momentum balance with flow/transport/energy.

Regarding point 3, the core problem is that the fluid creation (constants or general), is on the lowest level in modelling (every solution strategy inherits from the base solution strategy). Any local solution to *not* create a fluid leads to a pitfall when mixing multi-physics models. I.e., the mechanics model must be mixed in last, such that flow, transport & energy override that local solution.
I am not sure if there is an easy, understandable solution which can be programmed, or if it is something users have to be educated about.
Hence my (lazy) solution to just mix it into `MomentumBalance`.

src/porepy/compositional/materials.py
```diff
@@ -0,0 +1,536 @@
...
+
+# The dataclass is frozen, since material parameters are not supposed to change, as well
+# as the units they are given in.
+# By using this in combination with keyword_only arguments for the construction of
+# materials, the user is forced to instantiate the constants with the right names of
+# constants, otherwise errors are raised (no separate checks required).
+# By providing default values to the fields, not every constant is required. The user
+# is expected to be aware which physics are used in the model.
+@dataclass(frozen=True, kw_only=True)
+class MaterialConstants:
```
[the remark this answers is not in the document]
vlipovac: Yes. The original implementation performed a conversion when the constant was called, IF it had a physical dimension. Otherwise it just returned the constant.
Now the post init processing takes care of the conversion, IF a dimension is given in SI_units, and we can define constants simply using the dataclass mechanism.

src/porepy/models/solution_strategy.py
```diff
@@ -339,32 +356,48 @@ def reset_state_from_file(self) -> None:
...
+        solid = solid.to_units(self.units)
+
+        # Set the solid for the model
+        # NOTE this will change with the generalization of the solid
+        self.solid = solid
+
+        # Store the default fluid constants to be accessible by the compositional.Fluid
+        # TODO considering that there is a order in prepare_simulations, is there a better
+        # solution to transfer the fluid constants to the next step in prepare simulations?
+        self.params["_default_fluid_constants"] = fluid
```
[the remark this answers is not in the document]
vlipovac: Starting with _ to hint to user to not mess with it.
I put it in params such that it is accessible to the FluidMixin when performing create_fluid, and other methods called subsequently. Either this, or I change multiple signatures, but that change requires more code adaptation for the general CF case.

src/porepy/compositional/base.py
```diff
@@ -337,24 +340,23 @@ def fractions_from_molalities(self, *molalities: FloatType) -> list[FloatType]:
...
     """Abstract EoS class defining the interface between thermodynamic input
     and resulting structure containing thermodynamic properties of a phase.
 
     Component properties required for computations can be extracted in the constructor.
 
     Note:
-        This class is called 'abstract EoS'. Users can implement any correlations but
-        are encouraged to focus on thermodynamic consistency.
+        This class is called 'abstract EoS', but is in fact not strictly abstract and can be
+        instantiated.
```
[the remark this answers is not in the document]
vlipovac: I agree. Just EquationOfState?

src/porepy/models/fluid_mass_balance.py
```diff
@@ -473,7 +472,9 @@ def bc_values_pressure(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
             values on the provided boundary grid.
 
         """
-        return self.fluid.pressure() * np.ones(boundary_grid.num_cells)
+        return self.fluid.reference_component.pressure * np.ones(
```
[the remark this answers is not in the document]
vlipovac: Yes, I also have some troubles accepting pressure or temperature as a material propert. While density (or to be more precise - reference_density) is somehow acceptable, this one would fit better to a phase. But that would require some similar structure X for the phase as well.
FluidConstants -> Component
X -> Phase

tests/models/test_fluid_mass_balance.py
```diff
@@ -374,8 +378,15 @@ def test_ad_operator_methods_single_phase_flow(
             compactness, only tested in one dimension.
 
     """
-    # Get the method to be tested in callable form..
-    method: Callable = getattr(model_setup, method_name)
+    # processing name space
+    method_namespace = method_name.split('.')
+    owner = model_setup
+    # loop top to bottom through namespace to get to the actual method defined on some grids
+    # and returning an operator
```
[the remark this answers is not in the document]
vlipovac: Yes, the principle of the test stays the same, just dealing with the intermediate layer given by the Fluid class.

keileg:
@vlipovac 

> 1. ... Yes, I would suggest to do that generalization as a next preparation ...

Fine with me.

> 2.  ... I suggest moving this functionality out of the material constants and into the class Units.

This makes sense. Do you agree @IvarStefansson?

3. I see the dilemma, and agree that on balance it is better to have an inactive fluid attribute in the momentum balance than to introduce various workarounds. So let's keep the formulation as it is.

src/porepy/compositional/compositional_mixins.py
vlipovac: Improved documentation, here and at the default implementation of fluid properties in this class.

src/porepy/compositional/materials.py
```diff
@@ -0,0 +1,536 @@
...
+    Important:
+        Every derived class must have a class attribute :attr:`SI_units`, **annotated as
+        ClassVar**. This is to inform the base class about the used SI units.
+
+        For Examples, see :class:`FluidConstants`.
+        For instructions on how to write composed units, see :meth:`convert_units`
+
+    This class is intended for 1 species only. Different domains or fluids in the
+    mD-setting require each their own material constants instance in the case of
+    heterogenity.
```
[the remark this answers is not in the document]
vlipovac: Split and moved documentation to FluidConstants and Solidconstants respectively.

src/porepy/compositional/materials.py
```diff
@@ -0,0 +1,536 @@
...
+
+    """
+
+    # NOTE Annotating it as a ClassVar leads to the dataclasss decorator ignoring this in its
+    # machinery. The annotation must not be forgotten in derived classes.
```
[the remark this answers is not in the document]
vlipovac: Should be resolved throughout the changed files.
(I don't know though why flake8 is not complaining)

src/porepy/compositional/materials.py
```diff
@@ -0,0 +1,536 @@
...
+
+    specific_heat_capacity: number = 1
+
+    thermal_expansion: number = 0
+
+    viscosity: number = 1
+
+    thermal_conductivity: number = 1
+
+    normal_thermal_conductivity: number = 1
```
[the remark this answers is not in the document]
vlipovac: Ordered manually in alphabetical order.

src/porepy/compositional/compositional_mixins.py
```diff
@@ -1222,106 +1250,69 @@ def dependencies_of_phase_properties(
...
-        elif self._has_equilibrium:
-            dependencies += [
-                phase.partial_fraction_of[component]
-                for component in phase
-                if self.has_independent_partial_fraction(component, phase)
-            ]
-        else:
+        Important:
+            This method must be overwritten in every flow problem which does not rely
+            on heuristics. The base method returns an empty list for reasons of compatibility
```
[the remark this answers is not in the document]
vlipovac: Improved documentation to refer to heuristic laws for phase properties.

src/porepy/compositional/compositional_mixins.py
```diff
@@ -1222,106 +1250,69 @@ def dependencies_of_phase_properties(
...
-        Raises:
-            CompositeModellingError: If :attr:`equilibrium_type` is None, and hence no
-                phase fractions were introduced by :class:`CompositionalVariables`.
-                The consistent definition of the specific molar enthalpy of a fluid
-                mixture always depends on phase fractions.
+                        dependencies = [self.pressure]
+                        if self._has_unified_equilibrium:
 
-        """
+                            dependencies +=  [self.temperature] + [
```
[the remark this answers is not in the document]
vlipovac: Added documentation of the code snippet and intra-sphinx links to the implementation of the boolean properties.

src/porepy/compositional/base.py
```diff
@@ -337,24 +340,23 @@ def fractions_from_molalities(self, *molalities: FloatType) -> list[FloatType]:
...
 
-        Phase properties are defined as secondary expressions, and the framework is
-        able to pick up the dependencies and call :meth:`compute_phase_state` with
-        the right values.
+        This is by intention such that phases can be created with the abstract EoS, in a
+        simulation setting which uses heuristics. The method :meth:`compute_phase_properties`
```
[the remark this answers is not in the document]
vlipovac: Clarified docs.

src/porepy/models/protocol.py
```diff
@@ -0,0 +1,816 @@
...
+
+# Conditional importing ensures that the protocols do not mess with the runtime
+# definitions, i.e., the protocol empty method is accidentally called as a proper method
+# and returns None.
+if not TYPE_CHECKING:
+    # This branch is accessed in python runtime.
+    class PorePyModel(Protocol):
+        """This is an empty placeholder of the protocol, used mainly for type hints."""
+
+        def __init__(self, *args, **kwargs) -> None:
```
vlipovac, line 21: @keileg @Yuriyzabegaev Marking this change here, since it fixes something we do not know of as of now.

vlipovac:
Added changes as per #1247 

I rebased the branch onto the recent version of develop which includes Protocols.
A `models.protocol.FluidProtocol` is included now.
I think we can have a second round of review @keileg @IvarStefansson .
Also, most of the remarks made by Eirik should be fixed now.

src/porepy/models/protocol.py
```diff
@@ -0,0 +1,825 @@
+"""Contains the protocols that declare the methods and attributes that must be present
+in an instance of a PorePy model. The proper implementations of these methods can be
+found in various classes within the models directiory.
+
+Note that the protocol framework is accessed by static type checkes only!
+
+Warning:
```
vlipovac, line 7: Adding this warning to reflect the insights gained with @Yuriyzabegaev 
The runtime `PorePyModel` class does not inherit from `Protocol` anymore.

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -10,52 +10,24 @@
...
-    """Fluid constants for the simulation."""
-    nd: int
-    """Number of spatial dimensions for the simulation."""
-    nonlinear_solver_statistics: pp.SolverStatistics
-    """Non-linear solver statistics for nonlinear solver."""
-
     def save_data_time_step(self) -> None:
-        """Export the model state at a given time step, and log time.
-        The options for exporting times are:
-            * None: All time steps are exported
```
keileg, line 51: I don't think we want to delete the docstrings from these methods, even though they are also present in the protocol. Do you agree @Yuriyzabegaev?

keileg: This applies to several methods in this file.

src/porepy/models/fluid_property_library.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+# TODO this does not belong here, requires a solution for 1 and m-phase
+class FluidMobility(PorePyModel):
+    """Class for fluid mobility and its discretization in single-phase flow problems."""
+
+    mobility_keyword: str
+    """Keyword for the discretization of the mobility. Normally provided by a mixin of
+    instance :class:`~porepy.models.SolutionStrategy`.
+
```
keileg, line 209: Let's try not to introduce new todos unless they are strictly necessary.

keileg (commented):
Some minor comments before I had to run for the weekend. I think most of my concerns are taken care of, I'll try to resolve comments when I have the time.

Navigating the changes in difficult due to the protocol changes being included. I suggest you consider to formally close this PR (after we have an overview of which comments have been taken care of), and then open a new one before Ivar has a look - that should make things easier for him. does this make sense?

src/porepy/viz/data_saving_model_mixin.py
Yuriyzabegaev: Yes, the current convention is to duplicate docstrings. This is something we can mention on the next user meeting.

vlipovac: This change was not intended, may have been caused by the messy rebase. I will check for other missing docstrings.

src/porepy/compositional/base.py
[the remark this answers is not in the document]
keileg: Fine with me.

src/porepy/models/solution_strategy.py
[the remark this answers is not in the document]
keileg: I think I understand. This is why we went so heavily for mixins in the first place, then we don't have these issues with access of information. I believe the solution is fair provided you clearly mark the intended usage also on relevant places (I cannot find them at the moment)

src/porepy/compositional/compositional_mixins.py
```diff
@@ -1222,106 +1228,102 @@ def dependencies_of_phase_properties(
...
-                phase.partial_fraction_of[component]
-                for component in phase
-                if self.has_independent_partial_fraction(component, phase)
-            ]
-        else:
+        Important:
+            This method must be overwritten in every flow problem which relies on e.g.,
+            :class:`~porepy.numerics.ad.surrogate_operator.SurrogateFactory` for
+            externally computed values. The framework requires information about the
+            input arguments of the externally computes property values (primary
```
keileg, line 1235:
```suggestion
            input arguments of the externally computed property values (primary
```

src/porepy/compositional/compositional_mixins.py
```diff
@@ -1222,106 +1228,102 @@ def dependencies_of_phase_properties(
...
 
-        Raises:
-            CompositeModellingError: If :attr:`equilibrium_type` is None, and hence no
-                phase fractions were introduced by :class:`CompositionalVariables`.
-                The consistent definition of the specific molar enthalpy of a fluid
-                mixture always depends on phase fractions.
+            Let's assume a general mixture with multiple components and phases, which
+            uses the :class:`CompositionalVariables`, and either
+            :attr:`_MixtureDOFHandler._has_unified_equilibrium` or
+            :attr:`_MixtureDOFHandler._has_equilibrium` defined. I.e., the
```
keileg, line 1260:
```suggestion
            :attr:`_MixtureDOFHandler._has_equilibrium` defined. i.e., the
```

src/porepy/models/energy_balance.py
```diff
@@ -673,14 +675,13 @@ def reference_temperature(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
                 Operator representing the reference temperature.
 
         """
-        t_ref = self.fluid.temperature()
-        assert t_ref == self.solid.temperature()
+        t_ref = self.fluid.reference_component.temperature
```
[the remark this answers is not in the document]
keileg: Can be resolved after a short explanation has been added

src/porepy/compositional/compositional_mixins.py
keileg: Can be resolved for my sake.

src/porepy/compositional/compositional_mixins.py
keileg: Can be resolved for my sake

src/porepy/compositional/compositional_mixins.py
keileg: Can be resolved from my side

src/porepy/models/fluid_mass_balance.py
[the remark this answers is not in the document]
keileg: Can be resolved after an explanatory comment has been added

src/porepy/models/solution_strategy.py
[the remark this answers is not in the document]
keileg: Feel free to resolve when you feel you're don, @vlipovac

tests/models/test_fluid_mass_balance.py
keileg: This can be resolved from my side

vlipovac:
Closing this PR and re-submitting in #1252 

Reasons include a confusing git history due to a messy rebase.

PR #1245, opened 2024-10-22
MOD: Replacing usage of numba.guvectorize by numba.njit in compositional.utils

### LINKED ISSUE #1233: Avoid numba compilation during import
We have some functions that are compiled by `numba` during import, see for instance [utilities](https://github.com/pmgbergen/porepy/blob/31654ffd1c1de609bd138a2b6061051af6236816/src/porepy/compositional/utils.py) in the compositional flow branch. Note also the contrast to other numba calls that are not exposed during import, for instance [here](https://github.com/pmgbergen/porepy/blob/31654ffd1c1de609bd138a2b6061051af6236816/src/porepy/utils/setmembership.py#L197). 

Compiling significantly slows import and is a drag on development, particularly during testing. We should therefore explore if it is possible to refactor the code so that the functions are only compiled upon invocation.

vlipovac:
Closes #1233

## Proposed changes

Dropping the usage of `numba.guvectorize` and replacing it by a classical parallelized loop.

`guvectorize` seems to have some open issues involving import time. I was not able to find out exactly why, but most stackoverflow posts recommend dropping it and implementing simply `njit` with parallelization instructions.

- Scoping did not help.
- Dropping the signature did not help as well, since it accesses internally the scalar function and numba cannot compile it then without defining a signature.

The standard `njit` with parallel instructions works just as fine, but it is slightly more code and the resulting function is not a universal numpy function anymore.

The import time of porepy should now be almost the same when `NUMBA_CACHE == True` (by default).
@keileg  please check this and confirm with the example you mentiond in #1233.
I observed no noticeable increase in test execution when including the new version of `compositional.utils` in porepy. 

## Types of changes

- [X] Minor change (e.g., dependency bumps, broken links).
- [ ] Other:

### DIFF
--- a/src/porepy/compositional/utils.py
+++ b/src/porepy/compositional/utils.py
@@ -53,7 +53,7 @@ def safe_sum(x: Sequence[_Addable]) -> _Addable:
         return cast(_Addable, 0)
 
 
-@numba.njit("float64[:,:](float64[:,:])", fastmath=NUMBA_FAST_MATH, cache=True)
+@numba.njit("f8[:,:](f8[:,:])", fastmath=NUMBA_FAST_MATH, cache=True)
 def normalize_rows(x: np.ndarray) -> np.ndarray:
     """Takes a 2D array and normalizes it row-wise.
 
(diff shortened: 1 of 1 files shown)


keileg:
It seems this resolves the issue of import time. I suggest we merge this and cross our fingers the consequences for the compositional work are not too bad.

keileg (approved):
I did the timing, it all seems good now. Thanks for the effort!

Issue #1216, opened 2024-08-21
Benchmarks for measuring assembly time


keileg:
Define and implement a few benchmark models that can be used to measure (improvements in) model assembly time. 

Suggested criteria:
1. There should only be a few (say, 3-4) benchmarks.
2. The total setup time should be limited to facilitate frequent testing.
3. There should be a span in geometric complexity (2d and 3d, up to dozens of fractures)
4. Cover 2-3 sets of physics (ex: fluid flow, HM, possibly THM)
5. Setting up the models should not be a major effort. Reuse of setups from tests or similar should be feasible.

Other considerations to be made:
1. Should we place the code in a separate repository? My instinct says yes.
2. Will we base timing on logging, `line_profiler` or an external tool like https://github.com/plasma-umass/scalene?

EDIT
While there are nuances in how best to measure various aspects related to multiphysics, it seems clear we want a benchmark mainly dedicated to geometric complexity, keeping the physics simple (that is, mass balance only). The specification of this first step is roughly as follows (some critical thinking and interpretation should be applied):

1. Test cases: It is natural to use setups from [2d](https://doi.org/10.1016/j.advwatres.2020.103759) and [3d](https://doi.org/10.1016/j.advwatres.2020.103759) benchmarks for flow, though it is not a goal to cover all the cases. Some of the setups are already available, see tutorial `flow_benchmarks,` and more may be available through `fracture_sets.py`. Note that there are non-trivial aspects of the geometry and boundary conditions for some of the tutorials, EK can give more information (and point to gmsh files that partly resolve these issues).
2. For each geometry, there should be some flexibility in terms of mesh resolution, possibly also other parameters, but this should not be overdone.
3. (Partly) decoupled from the specific setups is where to put the code, how to structure it to allow for reuse and to fit with there being other benchmarks (I don't know exactly what this means) etc. The expectation is that this should be kept in mind but not optimized prematurely.
4. It is also of interest to consider solutions for tracking of performance, including monitoring over time. Again, this is something not to be overengineered at this early stage.

keileg:
A perhaps better approach to design is to identify parts of the framework that should be covered by the framework and use this to define the cases. My suggestion would be:

1. Operations connected to the md grid structure: Subdomain and mortar projections, matrices from discretizations of elemental differential operators (e.g. divergence and Darcy's law). We may want to cover scaling with the number of subdomains, grid size, and possibly spatial dimension.
[...]

PR #1261, opened 2024-11-08
1216 benchmark

pschultzendorff:
## Proposed changes
This adds a `run_profiler` script, which profiles a selected benchmark with `viztracer` and displays the results. This is a first implementation of such a feature and open to changes in the future. The plan is that people can try it out in the coming weeks and at some point we update based on our findings of what is actually needed in practice. Adresses #1216.

In addition, this adds a new benchmark case with 64 fractures based on [this article](https://www.sciencedirect.com/science/article/pii/S0309170817300143?casa_token=9yRY5edz5RAAAAAA:h7lQQW4cyCRm36vcoKRi5kzmUhQc5K0u9QgwGwiAd7ObugGeWGhN16Ef8o2u2izcRxmJJ31S6A).

Code written by @Yuriyzabegaev and me.

## Types of changes

- [ ] Other:

## Checklist


src/porepy/examples/run_profiling.py
```diff
@@ -0,0 +1,286 @@
...
+        choices=["flow", "poromechanics"],
+        help="Physics to run. Choices are single-phase flow or poromechanics.",
+    )
+    parser.add_argument(
+        "--geometry",
+        type=int,
+        default=0,
+        choices=[0, 1, 2, 3],
+        help=(
+            "0: 1st 2D case, 1: 2nd 2D case, 2: 2D case with 64 fracture, 3: 3D case."
```
pschultzendorff, line 251: plural

src/porepy/examples/run_profiling.py
```diff
@@ -0,0 +1,286 @@
...
+        if args.physics == "flow":
+            model = FlowBenchmark2dCase3aModel
+        elif args.physics == "poromechanics":
+            model = Case3aPoromech2D
+    elif args.geometry == 2:
+        if args.physics == "flow":
+            model = FlowBenchmark2dCase4Model
+        elif args.physics == "poromechanics":
+            model = Case4Poromech2D
+    elif args.geometry == 3:
```
pschultzendorff, line 163: Raise a warning on memory requirements if this is run with grid_refinement 2

keileg:
I removed the link to #1216 since that topic is bigger than what is covered in this PR

src/porepy/applications/md_grids/fracture_sets.py
```diff
@@ -126,6 +129,35 @@ def benchmark_2d_case_3(size: pp.number = 1) -> list[pp.LineFracture]:
...
+    The setup is composed of 64 fractures grouped in 13 different connected networks,
+    ranging from isolated fractures up to tens of fractures each.
+
+    References:
+        - [1] Flemisch, B., Berre, I., Boon, W., Fumagalli, A., Schwenck, N.,
+        Scotti, A., ... & Tatomir, A. (2018). Benchmarks for single-phase flow in
+        fractured porous media. Advances in Water Resources, 111, 239-258.
+
+    """
+    fracture_network_path = (
```
keileg, line 144: Please add file to repository?

keileg:
I will have a more thorough look at this, including running the benchmark, but it will not be before second half of next week.

src/porepy/applications/md_grids/fracture_sets.py
Yuriyzabegaev: Done

src/porepy/examples/run_profiling.py
keileg: Yes, that would be useful.

src/porepy/examples/flow_benchmark_2d_case_4.py
```diff
@@ -0,0 +1,86 @@
...
+        bounds = self.domain_boundary_sides(boundary_grid)
+        values = np.zeros(boundary_grid.num_cells)
+        return values
+
+
+# Ignore type errors inherent to the ``SinglePhaseFlow`` class.
+class FlowBenchmark2dCase4Model(  # type: ignore[misc]
+    Geometry,
+    BoundaryConditions,
+    pp.fluid_mass_balance.SinglePhaseFlow,
```
keileg, line 84: Please also add permeability values and aperture as specified in the benchmark.

src/porepy/examples/run_profiling.py
```diff
@@ -0,0 +1,286 @@
...
+            combination of geometry and physics is not supported.
+
+    """
+    # Set up fixed model parameters.
+    model_params = {
+        "material_constants": {
+            "solid": FractureSolidConstants(
+                residual_aperture=1e-1,
+                fracture_permeability=1e-1,
+                normal_permeability=1e-1,
```
keileg, line 118: Why override the standard parameter values as specified in the models?

src/porepy/examples/run_profiling.py
```diff
@@ -0,0 +1,286 @@
...
+    model.prepare_simulation()
+    print("Num dofs:", model.equation_system.num_dofs())
+    pp.run_time_dependent_model(
+        model,
+        {
+            "prepare_simulation": False,
+            "nl_divergence_tol": 1e8,
+            "max_iterations": 25,
+            "nl_convergence_tol": 1e-2,
+            "nl_convergence_tol_res": 1e-2,
```
keileg, line 223: These are very loose tolerances. Any reason to do so?

keileg: I see two dangers:
1. This will underestimate the time spent on solvers compared to everything else.
2. There is a risk that a loose tolerance may lead the solution to a weird part of the solution space, thus to profiling results that are biased in some sense.

keileg (commented):
Some comments based on my reading of the code. It is clear that this wlil be an iterative effort, but let's make sure we have a reasonable starting point before letting this loose into the wild. 

My attempt at viewing the results in a browser failed, so I'll need to ask for some assistance at some point.

src/porepy/examples/run_profiling.py
Yuriyzabegaev: I decided to make it like this because otherwise I could not make Newton converge for some of the models. With this, it takes 1-2 Newton iterations to accept the solution. The whole simulation consists of only one time step, so we don't use these unphysical solutions for the next time step. I'm also afraid of making the benchmark too time consuming if forcing it to make 10 or more Newton iterations.

I see your point about biased results, but I can't fully agree with it. As I understand, we are now interested in measuring the times of: 1) initialization 2) assembly 3) linear and nonlinear solver. 1 and 2 should be identical for any physical state and material constants. For a linear solver, it should be more or less the same for a direct solver. What matters for iterative linear solvers and Newton is the number of iterations, while each iteration should take the same time and also not depend on physical state.

Anyway, to make this work properly with the correct convergence tolerances, one should find the proper time step and the physical material constants and initial and boundary conditions, which I've not done, but I can do it if needed.

Yuriyzabegaev:
@keileg about viewing the results in the browser, I can't say that this solution is the most robust - sometimes it just fails with an error. What works for me is to reload the page until it works, it usually take 2-5 reloads. The usual time to "open" the data for me is not more than 10 seconds.

pschultzendorff:
@Yuriyzabegaev and @keileg: I haven't experienced any issues with the results not loading in Firefox so far. Perhaps you can try that and see if it works more reliable for you.

src/porepy/examples/run_profiling.py
keileg: I see your point in differing between simulations for benchmarking and actual example simplifications. Let's keep this for now, but please explain in comments what is being done and the rationale behind it.

IvarStefansson:
> @IvarStefansson we called the new runscript `run_profiler.py` and placed it in `src/porepy/examples`. Do you think these are appropriate choices or would you suggest something else?

Do you mean run_profiling.py? And I don't know about the placement without some more context/discussion.

src/porepy/examples/run_profiling.py
Yuriyzabegaev: Fixed

src/porepy/examples/run_profiling.py
Yuriyzabegaev: Added

src/porepy/examples/run_profiling.py
Yuriyzabegaev: Added a comment

src/porepy/examples/flow_benchmark_2d_case_4.py
Yuriyzabegaev: Added, however, not used in the profiling runscript.

src/porepy/examples/run_profiling.py
```diff
@@ -0,0 +1,290 @@
...
+
+    # Set cell_size/refinement_level model parameter based on choice of geometry and
+    # grid refinement.
+    if args.geometry in [0, 1, 2]:
+        if args.grid_refinement == 0:
+            cell_size = 0.1
+        elif args.grid_refinement == 1:
+            cell_size = 0.01
+        elif args.grid_refinement == 2:
+            cell_size = 0.005
```
keileg, line 137: This will fail completely for `args.geometry == 2`, which has a domain size of `600 x 700`.

Yuriyzabegaev: Nice catch, thanks!

src/porepy/examples/run_profiling.py
```diff
@@ -0,0 +1,306 @@
...
+>>> viztracer --min_duration 0.01ms --ignore_frozen --ignore_c_function your_script.py
+>>> vizviewer --port 9002 results.json
+
+Example (intended to show basic input and output functionality):
+    # If not already installed, get viztracer:
+    >>> pip install viztracer
+    # Run the script:
+    >>> python run_profiler.py --physics poromechanics --geometry 0 --grid_refinement 0
+    # This will run a single-phase poromechanics benchmark on the first 2D case with the
+    coarsest grid refinement.
```
jwboth, line 14:
```suggestion
    # coarsest grid refinement.
```

src/porepy/examples/run_profiling.py
```diff
@@ -0,0 +1,306 @@
...
+    >>> python run_profiler.py --physics flow --geometry 1 --grid_refinement 2
+    # This will run a single-phase flow benchmark on the second 2D case with the finest
+    grid refinement.
```
jwboth, line 17:
```suggestion
    # grid refinement.
```

keileg:
Regarding placement of `run_profiling.py`: Let's make a subfolder of `applications`, called `profiling` and put it there. It might be we decide to move it later on, but right now, this seems a fair solution.

keileg (approved):
Thanks for the effort

Issue #1150, opened 2024-04-25
Numerical parameters


IvarStefansson:
We might want to collect numerical parameters in some dedicated object, possibly of the `MaterialConstants` type. @keileg's comments copied from other discussion:

Agree! I can think of a few reasons for revisiting this:

1. Cleanliness.
2. Getting an overview of such numerical parameters in use.
3. (hopefully) introduce scaling of numerical parameters, just as with physical ones.

IvarStefansson:
Note that this may affect both model parameters and solver parameters as defined in #1210.

jwboth:
Task: Map out "all" numerical constant which are potentially in use. Provide a new object (my suggestion would be to follow the naming style and use `NumericalConstants`) - check out models/material_constants.py as well. In the process, we will determine how subclasses can be designed to better indicate the thematic grouping of constants. The priority for the first should be the two reasons mentioned by Ivar. Scaling will follow in a second step.

isakhammer:
# Work in progress of Overview of Potential Parameters classes PorePy  

Below is a list of parameters that are less frequently used but could be valuable to incorporate into the class.

1. **Where and what these parameters are used**  
   (Based on the source code)

2. **How these parameters are tested**  
   (Based on the testing framework, tutorials and run scripts)

#### Suggestions from Eirik:
- Search for `tol` to identify various tolerance parameters across different contexts.
- Maintain a high-level overview; avoid going too deep into specific cases.


### Candidates for contact mechanics 
**Todo**


### Newton solver
**Todo**

### Physical parameters 
**Todo**


### Candidates for mesh arguments og grids
**Todo**

1. [test_refinement 1](https://github.com/pmgbergen/porepy/blob/e10aeaa351181b9a71b6e890a50b32484386d8b6/tests/grids/test_refinement.py#L379)
    ```python
    mesh_args = {"mesh_size_frac": 1, "mesh_size_bound": 1, "mesh_size_min": 0.1}
    num_ref = 3
    params = {
        "mode": "nested",
        "num_refinements": num_ref,
        "mesh_param": mesh_args,
    }
    
    
        params = {
        "mode": "nested",
        "num_refinements": num_ref,
        "mesh_param": mesh_args,
        # The fracture is a constraint
        "grid_param": {"constraints": np.array([0])},
    }

    ```


[...]

PR #1262, opened 2024-11-14
New Numerical Constant Class

mariusnevland:
Addressing issue https://github.com/pmgbergen/porepy/issues/1150,

To separate the material and numerical constants, we introduce the class NumericalConstants. As a first go, we only added numerical values related to contact mechanics, but we plan to later consider newton tolerances and mesh parameters for the same class.

## Types of changes

- [ ] Other:

## Checklist


src/porepy/compositional/materials.py
```diff
@@ -446,6 +423,36 @@ class SolidConstants(Constants):
...
+
+    characteristic_displacement: number = 1.0
+
+    contact_mechanics_scaling: number = 1e-1
+    """Safety scaling factor, making fractures softer than the matrix"""
+
+    open_state_tolerance: number = 1e-5
+    """Tolerance parameter for the tangential characteristic contact mechanics.
+
+    FIXME: Revisit the tolerance.
```
jwboth, line 451: @IvarStefansson With characteristic scaling factors for tractions, is this FIXME now obsolete?

src/porepy/compositional/materials.py
```diff
@@ -446,6 +423,36 @@ class SolidConstants(Constants):
     well_radius: number = 0.1
 
 
+@dataclass(kw_only=True, eq=False)
+class NumericalConstants(Constants):
+    """Data class containing numerical method parameters,
+    including characteristic sizes.
+    """
```
jwboth, line 430:
```suggestion

    """
```

src/porepy/compositional/materials.py
```diff
@@ -446,6 +423,36 @@ class SolidConstants(Constants):
...
+    SI_units: ClassVar[dict[str, str]] = dict(
+        {
+            "characteristic_displacement": "m",
+            "characteristic_contact_traction": "Pa",
+            "open_state_tolerance": "-",
+            "contact_mechanics_scaling": "-",
+        }
+    )
+
+    characteristic_contact_traction: number = 1.0
```
jwboth, line 441:
```suggestion
    characteristic_contact_traction: number = 1.0
    """Characteristic traction used for scaling of contact mechanics."""
```

src/porepy/compositional/materials.py
```diff
@@ -446,6 +423,36 @@ class SolidConstants(Constants):
...
+
+    characteristic_displacement: number = 1.0
```
jwboth, line 443:
```suggestion
    characteristic_displacement: number = 1.0
    """Characteristic displacement used for scaling of contact mechanics."""
```

src/porepy/compositional/materials.py
```diff
@@ -446,6 +423,36 @@ class SolidConstants(Constants):
...
+
+    contact_mechanics_scaling: number = 1e-1
+    """Safety scaling factor, making fractures softer than the matrix"""
```
jwboth, line 446:
```suggestion
    """Safety scaling factor, making fractures softer than the matrix."""
```

src/porepy/models/solution_strategy.py
```diff
@@ -300,6 +300,10 @@ def set_materials(self):
         fluid = cast(
             pp.FluidComponent, constants.get("fluid", pp.FluidComponent(name="fluid"))
         )
+        numerical = cast(
```
jwboth, line 303: Please adapt also the comment in line 295 regarding the numerical constants.

tests/models/test_energy_balance.py
```diff
@@ -221,9 +221,11 @@ def bc_values_pressure(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
 
     solid_vals = pp.solid_values.extended_granite_values_for_testing
     fluid_vals = pp.fluid_values.extended_water_values_for_testing
+    numerical_vals = {"characteristic_displacement": 0.2}
```
jwboth, line 224: @IvarStefansson What do you think how the definition of numerical constants is handled in the tests? There has been two possibilities: 1. Explicit definition in all tests (like done here). 2. A central definition as used for the material constants. To me both is fine. Yet, if I count correctly, it is 5 instances in which the characteristic displacement needs explicit setting. Thus, centralizing it would also be reasonable.

jwboth (commented):
This looks good! I just left small comments, and requested some changes in the documentation. Please have a look. As it looks to me, the only issue which needs discussion concerns the handling of setting numerical constants in the tests.

src/porepy/compositional/materials.py
IvarStefansson: Almost. I think we should consider tightening it, given that it now compares values which we have scaled towards unity.

tests/models/test_energy_balance.py
IvarStefansson: Good question. I guess it depends on whether we expect to expand the set of numerical parameters and, if so, what they will contain. If we end up with a larger set which can readily be reused, I'd prefer the sentralised version.

jwboth: I agree. Let's go for a centralized variant already now. This will then remove the need to think about that again, when adding other values later.

src/porepy/compositional/materials.py
jwboth: I like the idea. Any preferred value? 1e-8?

tests/models/test_energy_balance.py
mariusnevland: Should be fixed now.

src/porepy/models/solution_strategy.py
mariusnevland: Done

jwboth (approved):
Thanks for the work!

PR #1266, opened 2024-11-21
Face error computation

IvarStefansson:
## Proposed changes

Improves the error computation for face values with a volume-like measure. Also introduces optional parameter for weighted norms.
Testing of l2_error function is improved.
Solves #1267 

## Types of changes

- [ ] Other:

## Checklist


tests/applications/test_convergence_analysis.py
```diff
@@ -788,85 +792,201 @@ def grids() -> list[pp.Grid, pp.MortarGrid]:
...
-        (True, True, False),  # subdomain vector cell-centered quantity
-        (True, False, False),  # subdomain vector face-centered quantity
-        (False, True, True),  # interface scalar cell-centered quantity
-        (False, True, False),  # interface vector cell-centered quantity
+        (True, True, True, None),
+        (True, False, True, None),
+        (True, True, False, None),
+        (True, False, False, None),
+        (False, True, True, None),
+        (False, True, False, 1.2),
```
jwboth, line 907: Shouldn't `parameter_weight` here just be `False` and `True`. Looking at the code below, I think this was possibly intendend.

tests/applications/test_convergence_analysis.py
```diff
@@ -788,85 +792,201 @@ def grids() -> list[pp.Grid, pp.MortarGrid]:
...
             face-centered quantity.
-        grids: List of grids. The first element is a two-dimensional subdomain grid,
-            and the second element is an interface grid. See the fixture grids().
+        is_scalar: Whether the array is corresponds to a scalar quantity. False
+            implies a vector quantity.
+        parameter_weight: Weight to be applied to the error. If a scalar is provided,
+            the same weight is applied to all cells or faces. If an array is provided,
+            the weight is broadcasted to each component of the vector quantity. While
+            the resulting error is the same, this design tests two different
+            implementations of the error computation.
```
jwboth, line 940: Documentation not consistent with the signature.

tests/functional/test_mandel.py
```diff
@@ -61,17 +61,17 @@ def results() -> list[MandelSaveData]:
     # t = 25 [s]
     DesiredError(
         error_pressure=0.02468899282842615,
-        error_flux=0.4277241964548815,
+        error_flux=0.362852,
         error_displacement=0.0007426275934204356,
-        error_force=0.007199561782640474,
+        error_force=0.006999,
```
jwboth, line 66: Interesting that most of the values do not differ too much after the change. I would have not expected that. But the math looks correct!

tests/applications/test_convergence_analysis.py
```diff
@@ -788,85 +792,201 @@ def grids() -> list[pp.Grid, pp.MortarGrid]:
     return [mdg.subdomains()[0], mdg.interfaces()[0]]
 
 
+@pytest.fixture(scope="module")
+def grids_3d() -> list[pp.Grid, pp.MortarGrid]:
+    """Create a mixed-dimensional grid on a unit cube with a single fracture.
+
+    Returns:
+        A list containing one subdomain grid (a Cartesian 2x2x2 grid) and one mortar
+        grid (a two dimensional mortar grid with 6 mortar cells).
```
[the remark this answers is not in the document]
jwboth: Since it is a simplex, it should be 8?

jwboth (approved):
Looks very good to me! I just have minor comments. If being very critical, I just do not like the name "error" while the actual meaning of the function is a "norm". But I can live with that. One alternative test that would come to my mind to test the sub-volume computations is to compute the distance between 1's and 0's both on volumes and faces. Both should give the same value, namely the volume of the geometry.

tests/applications/test_convergence_analysis.py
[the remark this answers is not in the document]
IvarStefansson: Sorry, copied documentation.

tests/applications/test_convergence_analysis.py
IvarStefansson: I updated this documentation, but apparantly forgot to push. Sorry!

tests/functional/test_mandel.py
IvarStefansson: Thanks for a thorough check! I believe the explanation is that we use relative errors with relatively homogeneous values, so the measures cancel between numerator and denominator.

tests/applications/test_convergence_analysis.py
```diff
@@ -788,85 +792,201 @@ def grids() -> list[pp.Grid, pp.MortarGrid]:
...
+    return [mdg.subdomains()[0], mdg.interfaces()[0]]
+
+
+def face_error(
+    sd: pp.GridLike,
+    true_array: np.ndarray,
+    approx_array: np.ndarray,
+    is_scalar: bool,
+    relative: bool = False,
+    parameter_weight: np.ndarray | None = None,
```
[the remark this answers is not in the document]
keileg: I suggest we let people be free to use.

PR #1280, opened 2024-12-01
Fix function to refine 1d grids. Improve tests.

### LINKED ISSUE #1277: Bug in refine_grid_1d
The grid produced by refine_grid_1d has only ones as its cell_face matrix. The implementation of the function must be improved and tests should be added.

keileg:
## Proposed changes
The function `pp.refinement.refine_grid_1d` returned grids with non-sensical topology (the cell-face map was clearly wrong), however this was not covered by the tests.

This PR fixes that issue, and also improves the implementation of the function so that it can be used for grids with non-trivial topology (ordering of cells, faces and nodes which are not linear). The new implementation is also much better documented.

The tests have also been improved. I decided to only test grids that are aligned with the x-axis. This is seemingly a poorer coverage than what used to be the case, however, I found the old tests' focus on geometry misguided; the chances of getting the geometry wrong are miniscule, while the topology is more complex.

Resolves #1277 

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/grids/refinement.py
+++ b/src/porepy/grids/refinement.py
@@ -108,7 +108,7 @@ def refine_grid_1d(g: pp.Grid, ratio: int = 2) -> pp.Grid:
 
     # When looping over cells in the old grid, we will hit upon all internal nodes
     # twice, but they should only be added to the refined grid the first time. We
-    # therefore need to keep track of whether an node in the old grid should be added
+    # therefore need to keep track of whether a node in the old grid should be added
     # and, if not, the node index in the new grid by which it is already present.
     #
     # Construct an array that indicates whether an item in the cell-node relation
@@ -172,7 +172,7 @@ def refine_grid_1d(g: pp.Grid, ratio: int = 2) -> pp.Grid:
             + g.nodes[:, end].reshape((-1, 1)) * theta
         )
 
-        # Register the indices of the new nodes and increasen the counter.
+        # Register the indices of the new nodes and increase the counter.
         loc_new_ind += list(node_counter + indices_template)
         node_counter += ratio - 1
 
--- a/src/porepy/grids/refinement.py
+++ b/src/porepy/grids/refinement.py
@@ -112,7 +112,10 @@ def refine_grid_1d(g: pp.Grid, ratio: int = 2) -> pp.Grid:
 
     # Template array of node indices for refined cells.
     indices_template = np.vstack((np.arange(ratio), np.arange(ratio) + 1)).ravel("F")
+    indices_template = np.repeat(np.arange(ratio), 2)
     nd = np.r_[np.diff(cell_nodes.indices)[1::2], 0]
+
+    old_2_new_nodes = {}
 
     # The refinement must work also for grids that are not equidistant. While it might
     # be possible to do the refinemnet in a vectorized fashion, looping over the cells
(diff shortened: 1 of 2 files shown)


src/porepy/grids/refinement.py
```diff
@@ -86,67 +86,134 @@ def refine_grid_1d(g: pp.Grid, ratio: int = 2) -> pp.Grid:
...
+    # Counter used to keep track of how many nodes have been added to the new grid.
+    node_counter = 0
+
+    # The refinement must work also for grids that are not equidistant. While it might
+    # be possible to do the refinemnet in a vectorized fashion, looping over the cells
+    # and refining them one by one is a simpler solution.
+
+    # When looping over cells in the old grid, we will hit upon all internal nodes
+    # twice, but they should only be added to the refined grid the first time. We
+    # therefore need to keep track of whether an node in the old grid should be added
```
jwboth, line 111:
```suggestion
    # therefore need to keep track of whether a node in the old grid should be added
```

src/porepy/grids/refinement.py
```diff
@@ -86,67 +86,134 @@ def refine_grid_1d(g: pp.Grid, ratio: int = 2) -> pp.Grid:
...
+        # Add coordinates of the internal nodes (those not present in the old grid). The
+        # coordinates added here run from closest to the start node to closest to the
+        # end node. Reshape is needed since we may add more than one node at a time (in
+        # contrast to the assignments in the if-statements above and below).
+        x[:, node_counter : (node_counter + ratio - 1)] = (
+            g.nodes[:, start].reshape((-1, 1)) * (1 - theta)
+            + g.nodes[:, end].reshape((-1, 1)) * theta
+        )
+
+        # Register the indices of the new nodes and increasen the counter.
```
jwboth, line 175:
```suggestion
        # Register the indices of the new nodes and increase the counter.
```

tests/grids/test_refinement.py
```diff
@@ -38,25 +40,88 @@ def test_grid_perturbation_1d_internal_nodes_fixed(self):
...
+
+        # The two grids should cover the same domain.
+        assert np.sum(g.cell_volumes) == np.sum(g_ref.cell_volumes)
+
+        # If the ratio is given, all cells should be refined by this ratio.
+        if ratio is not None:
+            assert np.allclose(g_ref.num_cells / g.num_cells, ratio)
+
+        # There should be as many faces as nodes in a 1d grid. Check that the face-node
+        # map is 1-1, and that there are as many node coordinates as there are nodes.
```
jwboth, line 75: The test for 1-1 tests only necessary but not sufficient conditions. As the grids tested here are small, checking whether `np.abs(np.linalg.det(g_ref_face_nodes))==1` should be feasible and do the job.

tests/grids/test_refinement.py
```diff
@@ -38,25 +40,88 @@ def test_grid_perturbation_1d_internal_nodes_fixed(self):
...
+        fn = sps.identity(5, format="csr")
+        cf = sps.csc_array(
+            np.array([[-1, 1, 0, 0, 0], [0, -1, 1, 0, 0], [0, 0, 0, -1, 1]]).T
         )
+        g_3 = pp.Grid(dim=1, nodes=x, face_nodes=fn, cell_faces=cf, name="")
+
+        for g in [g_0, g_1, g_2, g_3]:
+            g.compute_geometry()
+            g_ref = pp.refinement.refine_grid_1d(g, ratio)
+            self.compare_refined_grids(g, g_ref, ratio)
```
jwboth, line 124: Is this check really convincing that the refinement does what it should? I would consider adding one "bad" test for g_0, and see whether g_ref coincides with [0, 0.5, 1, 1.5, ...] as well as [0, 0.33, 0.67, 1, ...] - a bit more like the old tests, where at least nodes were compared. At the moment merely shapes are tested but not really topology.

jwboth (commented):
Looks good to me. Most critical comment relates to the test.

tests/grids/test_refinement.py
keileg: Agree, will implement.

tests/grids/test_refinement.py
keileg: Added a test that the node coordinates are as expected. 

I also modified the documentation somewhat to make the rational behind the testing clearer: It is not meaningful to do a detailed check of the topology of the refined grid, as no guarantees are given for this beyond its sanity. In particular the ordering of nodes and cells can differ significantly between the old and refined grid.

tests/grids/test_refinement.py
keileg: Implemented.

tests/grids/test_refinement.py
jwboth: Thanks for the clarification. I think the newly added test does nevertheless not hurt. Thanks for adding it.

jwboth (approved):

PR #1296, opened 2025-01-02
Avoid-plotting-to-same-figure

### LINKED ISSUE #1289: Update to matplotlib >= 3.10
After Matplotlib was updated to 3.10, there is a bug that makes the check of tutorials fail. The matplotlib version has temporarily been frozen, see [https://github.com/pmgbergen/porepy/pull/1288](https://github.com/pmgbergen/porepy/pull/1288). This issue is a reminder that we should find and fix the problem.

IvarStefansson:
## Proposed changes

The reason matplotlib 3.10 doesn't work has to do with something in matplotlib being affected by the call to pp.plot_grid. Turns out this only happens for 3d plots, not 2d, hence the fix. Alternative fix is to create a new figure. 
NOTE: While the symptom is fixed, please consider during review whether the cause should be understood fully.
Fixes #1289 and should enable #1290.

## Types of changes

- [ ] Other:

## Checklist


keileg:
This seems fair to me, but I have never been a fan of 3d plots in matplotlib anyhow. Should we perhaps switch the default of `plot_2d` to `True` and leave a comment that 3d plots may not work?

keileg:
I guess the control question is, do we use 3d plotting? I guess the answer is hardly, considering it is only this tutorial that failed.

IvarStefansson:
> I guess the control question is, do we use 3d plotting? I guess the answer is hardly, considering it is only this tutorial that failed.
Not exactly. Creating a wontfix issue describing what to do in the 3d case.

keileg (approved):

PR #1310, opened 2025-01-22
Fracture-damage

IvarStefansson:
## Proposed changes

Functionality for computing history-dependent damage for fracture mechanics.
Main functionality
* New variable and equation in models.fracture_damage.
* New constitutive laws for friction and shear dilation.
* Ad function for computing 1/val, where val may contain zeros. This function may need improvements.
* Example with complete classes, also used for testing.

Please consider naming of new files, classes and methods inside during review. Also search for new todos, these should be resolved during the PR.

Note that a skipped test (slow) was added. Thus, we must remember to run the full suite before merging.

## Types of changes

- [ ] Other:

## Checklist


src/porepy/numerics/ad/functions.py
```diff
@@ -141,6 +142,35 @@ def l2_norm(dim: int, var: pp.ad.AdArray) -> pp.ad.AdArray:
...
+
+    Parameters:
+        numerator: AdArray representing the numerator.
+        denominator: AdArray representing the denominator.
+
+    Returns:
+        AdArray representing the safe division.
+
+    """
+    if isinstance(var, np.ndarray):
```
keileg, line 160: The typing indicates this is not possible. There are of course plenty of cases where our implementation is more general than the typing.

src/porepy/numerics/ad/functions.py
```diff
@@ -141,6 +142,35 @@ def l2_norm(dim: int, var: pp.ad.AdArray) -> pp.ad.AdArray:
...
+        _val = var
+    else:
+        _val = var.val
+
+    nonzero_inds = np.abs(_val) > 1e-15
```
keileg, line 165: Is it possible to give a rational behind this number, besides 'what else can it be'? Is it meaningful to make this a user controlled parameter with a default?

src/porepy/models/fracture_damage.py
```diff
@@ -0,0 +1,326 @@
...
+        scalar_to_tangential = pp.ad.sum_projection_list(tangential_basis)
+        f_norm = pp.ad.Function(partial(pp.ad.l2_norm, self.nd - 1), "norm_function")
+        # TODO: Reconsider this function and its implementation.
+        f_power = pp.ad.Function(
+            partial(pp.ad.safe_power, -1, 1 / np.sqrt(self.nd - 1)), "safe power"
+        )
+        # Compute normalised tangential displacement increment. First, compute the norm
+        # of the increment.
+        norm_u_t = scalar_to_tangential @ f_norm(u_t)
+        # Then, normalise the increment by multiplying it by the inverse of the norm.
```
keileg, line 267: In total, m_t will be a unit vector in the direction of u_t, unless u_t is strictly zero, in which case it is zero. Correct?

src/porepy/models/fracture_damage.py
```diff
@@ -0,0 +1,326 @@
...
+        tangential_to_scalar = pp.ad.sum_projection_list(
+            [e_i.T for e_i in tangential_basis]
+        )
+
+        m_t = self._normalized_tangential_plastic_jump(subdomains)
+        num_steps = self.time_manager.time_index
+        # The first term in the equation, the implicit part.
+        current_inner = tangential_to_scalar @ (m_t * u_t)
+        # We are only interested in positive values of the inner product.
+        f_heaviside = pp.ad.Function(partial(pp.ad.functions.heaviside, 0.5), "max")
```
keileg, line 216: This will render 0.5 if u_t == 0?

src/porepy/models/fracture_damage.py
```diff
@@ -0,0 +1,326 @@
...
+            tangential_to_scalar @ (m_t * u_t_increment)
+        )
+        # Then add the explicit part, i.e., the sum of the inner product of m with the u_t
+        # increment from all previous time steps. The sum starts at 1 since the first
+        # increment is already included in the implicit part.
+        for i in range(1, num_steps):
+            u_t_i = u_t.previous_timestep(i)
+            u_t_increment_i = u_t_i - u_t.previous_timestep(i + 1)
+            if np.allclose(
+                self.equation_system.evaluate(u_t_increment_i), 0, atol=1e-12
```
keileg, line 228: I see the reasoning, and like the motivation, but I am not very happy with such assumptions on the absolute scaling of variables. I suppose you think it worth it, all things considered?

keileg: Assuming yes, can we do a similar check on u_t_i? Or does my question above regarding u_t==0 come into play?

src/porepy/models/fracture_damage.py
```diff
@@ -0,0 +1,326 @@
...
+
+        Parameters:
+            subdomains: List of subdomains where the damage history is defined. Should
+                be of co-dimension one, i.e. fractures.
+
+            Returns:
+                Operator for the anisotropic damage history equation.
+
+        """
+        # Get the tangential component of the plastic displacement jump.
```
keileg, line 198: General comment regarding this function: The implementation looks correct to me, but since it involves much constitutive modeling with tacit assumptions in edge cases (cf. Heaviside function below), I would say that rather extensive documentation is needed. Before doing that, I suggest we talk through this once more on a blackboard, just to make sure we agree on all what are the choices to be made. This also regards the convolution integral and possible shortcuts to take there.

src/porepy/models/fracture_damage.py
```diff
@@ -0,0 +1,326 @@
...
+
+        num_steps = self.time_manager.time_index
+        # The first term in the equation, the implicit part.
+        eq = self.damage_history(subdomains) - f_norm(u_t_increment)
+        # Then add the explicit part, i.e., the sum of the inner product of m with the u_t
+        # increment from all previous time steps. The sum starts at 1 since the first
+        # increment is already included in the implicit part.
+        for i in range(1, num_steps):
+            u_t_increment_i = u_t.previous_timestep(i) - u_t.previous_timestep(i + 1)
+            # if np.allclose(
```
keileg, line 318: Please update

src/porepy/models/fracture_damage.py
```diff
@@ -0,0 +1,326 @@
...
+        """
+        return self.equation_system.get_variables(
+            variables=[
+                self.interface_displacement_variable,
+                self.contact_traction_variable,
+            ]
+        )
+
+
+class DamageHistoryEquation(pp.PorePyModel):
```
keileg, line 121: Note, though likely not for this PR: This equation will be algebraic, with no spatial coupling terms, and can therefore be eliminated locally prior to the linear solve using the framework developed (being developed?) for multiphase multicomponent. Whether this is important is unclear, but it could be beneficial for iterative solver if we start using this framework in earnest.

tests/compositional/test_materials.py
```diff
@@ -17,7 +17,9 @@
 import pytest
 
 import porepy as pp
+
 from porepy.examples.flow_benchmark_2d_case_1 import FractureSolidConstants
+from porepy.compositional.materials import FractureDamageSolidConstants
```
keileg, line 22: This is an awkward placement for the materials objects. I guess we all agree on that one.

src/porepy/examples/fracture_damage.py
```diff
@@ -0,0 +1,446 @@
...
+    BoundaryConditionsMechanicsDirNorthSouth,
+)
+from porepy.applications.convergence_analysis import ConvergenceAnalysis
+from porepy.applications.test_utils.models import ContactMechanicsTester
+from porepy.compositional.materials import FractureDamageSolidConstants
+from porepy.models import fracture_damage as damage
+
+
+class TimeDependentDamageBCs(BoundaryConditionsMechanicsDirNorthSouth):
+    def bc_values_displacement(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
```
keileg, line 17: boundary_grid -> bg

src/porepy/models/fracture_damage.py
```diff
@@ -0,0 +1,326 @@
...
+        In that case, the method should be overridden.
+
+        Returns:
+            List of variables.
+
+        """
+        return self.equation_system.get_variables(
+            variables=[
+                self.interface_displacement_variable,
+                self.contact_traction_variable,
```
keileg, line 116: From the implementation in this file and the example of fracture damage, I don't see why the contact traction needs to be stored?

src/porepy/examples/fracture_damage.py
```diff
@@ -0,0 +1,446 @@
...
+            )
+            # Set the values of current time to most recent previous time.
+            pp.set_solution_values(name=name, values=vals, data=data, time_step_index=0)
+
+            # Set the unknown time step values.
+            vals = self.interface_displacement_parameter_values(intf).ravel(order="F")
+            pp.set_solution_values(name=name, values=vals, data=data, iterate_index=0)
+
+
+@dataclass
```
keileg, line 119: EK note to self: I stopped reading here.

src/porepy/models/fracture_damage.py
keileg: Got it, because this is a natural extension of the current models. Please document.

tests/models/test_fracture_damage.py
```diff
@@ -0,0 +1,115 @@
...
+    if isotropic:
+        # Use the isotropic damage model.
+        model_class = damage.IsotropicFractureDamage
+        # Change the exact solution to the isotropic version.
+        params_local["exact_solution"] = damage.ExactSolutionIsotropic
+    else:
+        # Use the anisotropic damage model.
+        model_class = damage.AnisotropicFractureDamage
+    # Add geometry mixin.
+    model_class = _add_mixin(geometry_mixins[dim - 2], model_class)
```
keileg, line 66: I prefer a dictionary over a list of geometries here. This is too clever for me.

keileg (commented):
I agree with the main design, but I am a bit unsure I have captured all nuances in the technical implementation. I suggest we talk through this before proceeding in earnest.

jwboth:
Sorry for late reply.

Anything will work for me as well. Pick a time, Enrico, which fits you best.

Jakub
________________________________

src/porepy/models/fracture_damage.py
```diff
@@ -0,0 +1,326 @@
...
+        h = \int{\tau} max(m_t \cdot \delta u_t, 0) \, d\tau,
+
+    where :math:`\delta u_t` is the tangential plastic displacement jump increment and
+    :math:`m_t` is the normalized tangential plastic displacement jump, while h is the
+    damage history variable.
+
+    """
+
+    damage_history: Callable[[list[pp.Grid]], pp.ad.Variable]
+    """Damage history variable on ad form."""
```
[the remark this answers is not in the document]
IvarStefansson: Not Ad. I think AD. Do you want me to add it to the user manual?

src/porepy/models/fracture_damage.py
```diff
@@ -0,0 +1,326 @@
...
+    r"""Anisotropic damage history equation.
+
+    The anisotropic damage history equation is given by
+
+    .. math::
+
+        h = \int{\tau} max(m_t \cdot \delta u_t, 0) \, d\tau,
+
+    where :math:`\delta u_t` is the tangential plastic displacement jump increment and
+    :math:`m_t` is the normalized tangential plastic displacement jump, while h is the
```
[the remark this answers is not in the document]
IvarStefansson: Yes. I updated the docstring.

src/porepy/models/fracture_damage.py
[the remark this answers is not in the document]
keileg: That would be good, thanks!

tests/compositional/test_materials.py
IvarStefansson: Indeed. Nothing to do here, though, right?

src/porepy/numerics/ad/functions.py
IvarStefansson: This goes for a lot of this module. I think this is necessary because of some fundamentals of the AD functions, but I'm not sure. Let's discuss if you don't recall this.

src/porepy/numerics/ad/functions.py
IvarStefansson: I'm not sure the default will work, since the AD functions are constructed using partial. But if you want me to, I can certainly set it as a mandatory third argument.

src/porepy/numerics/ad/forward_mode.py
```diff
@@ -96,7 +96,13 @@ def __init__(self, val: np.ndarray, jac: sps.spmatrix) -> None:
     def __repr__(self) -> str:
         s = f"Ad array of size {self.val.size}\n"
         s += f"Jacobian is of size {self.jac.shape} and has {self.jac.data.size}"
-        s += " elements"
+        s += " elements."
+        return s
+
+    def __str__(self) -> str:
```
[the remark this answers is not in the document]
IvarStefansson: (Of course) I'm not sure. But it met my needs during debugging. Do you have alternative suggestions/preferences?

src/porepy/examples/fracture_damage.py
IvarStefansson: Can this be resolved?

IvarStefansson (commented):
First round of responses to review. Thanks, @keileg.

src/porepy/models/constitutive_laws.py
```diff
@@ -3705,6 +3706,204 @@ def elastic_tangential_fracture_deformation(
...
+
+
+class DilationDamage(pp.PorePyModel):
+    """Dilation damage relations.
+
+
+    The dilation damage is computed from the history variable h as
+
+    .. math::
+        d = 1+(d_0-1)  exp⁡(-c h)
```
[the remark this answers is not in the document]
keileg: Yes, that placed in the class docstring should be sufficient.

src/porepy/models/constitutive_laws.py
```diff
@@ -3705,6 +3706,204 @@ def elastic_tangential_fracture_deformation(
         return u_t
 
 
+class FrictionDamage(pp.PorePyModel):
```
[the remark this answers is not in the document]
IvarStefansson: Sure, but I don't know what that would be. Any suggestions?

src/porepy/models/fracture_damage.py
```diff
@@ -0,0 +1,326 @@
...
+    def before_nonlinear_loop(self):
+        """Reset the damage history equation to include new term.
+
+        This needs to be done *after* time manager is updated, which happens in the
+        super class method.
+
+        """
+        super().before_nonlinear_loop()
+        fractures = self.mdg.subdomains(dim=self.nd - 1)
+        self.equation_system._equations[self.damage_history_equation_name] = (
```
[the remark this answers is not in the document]
keileg: Either that or make it a user project. You decide if you can make up your mind.

src/porepy/models/fracture_damage.py
keileg: Yes!

src/porepy/numerics/ad/forward_mode.py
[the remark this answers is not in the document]
keileg: As per https://stackoverflow.com/a/2626364 (which is the answer to most str vs repr questions), repr should provide more information than str. So I would say the two implementations should be switched, and both methods should give shape and nnz information for the Jacobian. does that make sense?

src/porepy/numerics/ad/functions.py
keileg: I think you are right. This may be fit for a maintenance project, but I'm also fine with you just resolving the comment.

src/porepy/numerics/ad/functions.py
keileg: Good point. My fear is that if this manifests as a bug at some point, it will be extremely hard to track down. Perhaps a mandatory argument is the safest option, so I think I lean in that direction.

tests/compositional/test_materials.py
keileg: Yes. User project? Resolve anyhow.

src/porepy/models/fracture_damage.py
IvarStefansson: Interesting thought.

src/porepy/numerics/ad/functions.py
IvarStefansson: What would you do in that project?

src/porepy/models/fracture_damage.py
IvarStefansson: Is it sufficiently documented now?

src/porepy/numerics/ad/forward_mode.py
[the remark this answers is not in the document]
IvarStefansson: I think I agree that it makes sense to swap. What do you mean about both methods giving shape and nnz? Do you want to add that to the current implementation of str?

src/porepy/numerics/ad/functions.py
keileg: That was a revealing question. Touche. I'll resolve.

src/porepy/models/constitutive_laws.py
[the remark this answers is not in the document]
keileg: I give up. Please ignore and resolve unless you get an diea.

keileg (approved):
Feel free to merge once the few remaining questions are resolved. Let me know if I have overlooked something and you need more input from my side.

PR #1317, opened 2025-01-30
Use proper mixins in concrete models

### LINKED ISSUE #1260: Replace base mixins by `PorePyModel` as base class in porepy-provided simulation set-ups
There are classes inside porepy which provide concrete set-ups for boundary conditions and md-grids for example.

As of now, some of them inherit directly from the respective mixins, which they should not in order to avoid MRO issues in the mixin framework.

Primarily affected are

- `porepy/applications`
- `porepy/examples`

A check whether other files/classes are affected should be carried out as well.

**Fix**:
Replace the mixins by the recently introduced `PorePyModel` protocol class, which suffices to provide an interface to the methods required from the mixins.
Make sure that no untyped code or code which cannot be picked up by linters, results from above changes.

vlipovac:
Closes #1260 

## Proposed changes

Models contained in porepy, excluding base models in `porepy.models`, use now proper mixins for

- geometry
- BC
- Variables and equations (not affected by changes in this PR, listing for completeness)

to eliminate circular paths in the inheritance tree of PorePy models.

Circular paths which I could not break up are the following two:

1. `DataSavingMixin` : Though all base models inherit from `DataSavingMixin`, models in `tests` and `porepy.examples` usually need `VerificationDataSaving` which inherits from `DataSaving`. To break those circular paths (if desired), the simplest solution would be to have all base models inherit from `VerificationDataSaving` instead of `DataSaving`. Then the individual verification setups can be turned into proper mixins. **Please comment if you wish to do so in this PR**.
2. Solution strategies: Currently affected are concrete models with custom IC and a verification setup including an exact solution, where both are included in the model set-up. There are also more complex cases like Biot-mechanics and Poro-mechanics where the design of solution strategies needs more thinking. My suggestion is to skip this for now and wait at least before the issue with `VerificationDataSaving` is resolved, and the abstraction of `InitialConditionMixin` from #1236 is included.

## Types of changes

The changes here aim to form the inheritance trees of porepy models into star-like graphs, to a degree possible considering above notes.

**Example: Mandel Biot**  (sorry for the messy graphs, the rendering engine is hard to configure)

Before:

class_chart_biot_old

After:

class_chart_biot_new

Optimal in theory (with "mixin" denoting a proper mixin):

optimal_mandel

Breaking change, since model components are now turned into proper mixins and do not work on their own (which was the case in some tests for custom geometries)

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/applications/test_utils/models.py
+++ b/src/porepy/applications/test_utils/models.py
@@ -112,7 +112,7 @@ def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
         pass
 
     # Create an instance of the combined class
-    model: pp.PorePyModel = cast(pp.PorePyModel, Model(params))
+    model = cast(pp.PorePyModel, Model(params))
 
     # Prepare the simulation
     # (create grids, variables, equations, discretize, etc.)
--- a/src/porepy/applications/test_utils/models.py
+++ b/src/porepy/applications/test_utils/models.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import inspect
-from typing import Any, Callable
+from typing import Any, Callable, cast
 
 import numpy as np
 
@@ -72,9 +72,7 @@ class Thermoporomechanics(  # type: ignore[misc]
     """Combine components needed for poromechanics simulation."""
 
 
-def model(
-    model_type: str, dim: int, num_fracs: int = 1
-) -> MassBalance | MomentumBalance | MassAndEnergyBalance | Poromechanics:
+def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
     """Setup for tests."""
     # Suppress output for tests
     fracture_indices = [i for i in range(num_fracs)]
@@ -114,7 +112,7 @@ def model(
         pass
 
     # Create an instance of the combined class
-    model = Model(params)
+    model: pp.PorePyModel = cast(pp.PorePyModel, Model(params))
 
     # Prepare the simulation
     # (create grids, variables, equations, discretize, etc.)
@@ -122,15 +120,9 @@ def model(
     return model
 
 
-class RobinDirichletNeumannConditions:
+class RobinDirichletNeumannConditions(pp.PorePyModel):
     """Mixin for applying Neumann, Dirichlet and Robin conditions for a
     thermoporomechanics model."""
-
-    params: dict
-
-    domain_boundary_sides: Callable[[pp.GridLike], pp.domain.DomainSides]
-
-    nd: int
 
     def bc_values_pressure(self, boundary_grid: pp.BoundaryGrid) -> np.ndarray:
         """Assigns pressure values on the north and south boundary."""
@@ -259,7 +251,7 @@ def subdomains_or_interfaces_from_method_name(
     return domains
 
 
-def _add_mixin(mixin, parent):
+def _add_mixin(mixin: type, parent: type) -> type:
     """Helper method to dynamically construct a class by adding a mixin.
 
     Multiple mixins can be added by nested calls to this method.
@@ -378,7 +370,7 @@ def compare_values(
         assert np.isclose(np.sum(values_0 - values_1), 0, atol=1e-10 + rtol)
 
 
-def get_model_methods_returning_ad_operator(model_setup) -> list[str]:
+def get_model_methods_returning_ad_operator(model_setup: pp.PorePyModel) -> list[str]:
     """Get all possible testable methods to be used in test_ad_operator_methods_xx.
 
     A testable method is one that:
--- a/src/porepy/examples/mandel_biot.py
+++ b/src/porepy/examples/mandel_biot.py
@@ -38,7 +38,6 @@ from porepy.applications.convergence_analysis import ConvergenceAnalysis
 from porepy.models.derived_models.biot import BiotPoromechanics
 from porepy.numerics.linalg.matrix_operations import sparse_array_to_row_col_data
 from porepy.utils.examples_utils import VerificationUtils
-from porepy.viz.data_saving_model_mixin import VerificationDataSaving
 
 # PorePy typings
 number = pp.number
@@ -116,7 +115,7 @@ class MandelSaveData:
     """Current simulation time."""
 
 
-class MandelDataSaving(VerificationDataSaving):
+class MandelDataSaving(pp.PorePyModel):
     """Mixin class to save relevant data."""
 
     darcy_flux: Callable[[list[pp.Grid]], pp.ad.Operator]
--- a/src/porepy/examples/mandel_biot.py
+++ b/src/porepy/examples/mandel_biot.py
@@ -1383,7 +1383,6 @@ class MandelSolutionStrategy(poromechanics.SolutionStrategyPoromechanics):
         """
         super().set_materials()
         self.exact_sol = MandelExactSolution(self)
-        self.results = []
 
         # Biot's coefficient must be one
         assert self.solid.biot_coefficient == 1
--- a/src/porepy/examples/mandel_biot.py
+++ b/src/porepy/examples/mandel_biot.py
@@ -33,7 +33,6 @@ import numpy as np
 import scipy.optimize as opt
 
 import porepy as pp
-import porepy.models.fluid_mass_balance as mass
 import porepy.models.poromechanics as poromechanics
 from porepy.applications.convergence_analysis import ConvergenceAnalysis
 from porepy.models.derived_models.biot import BiotPoromechanics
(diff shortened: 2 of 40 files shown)


src/porepy/applications/test_utils/models.py
```diff
@@ -114,24 +112,18 @@ class Model(geometry, model_class):
         pass
 
     # Create an instance of the combined class
-    model = Model(params)
+    model: pp.PorePyModel = cast(pp.PorePyModel, Model(params))
```
IvarStefansson, line 115: Is it necessary to both cast and declare?

src/porepy/examples/mandel_biot.py
```diff
@@ -1402,6 +1384,7 @@ def set_materials(self):
         """
         super().set_materials()
         self.exact_sol = MandelExactSolution(self)
+        self.results = []
```
IvarStefansson, line 1387: Are you sure this should go here? Init seems better than in set_materials, but other options could be possible. If a method does something which is not covered by its name, I try to check whether the operation can be moved or if more fundamental refactoring is necessary. Sometimes, pragmatism is called for, of course. Opinions, @keileg ?

IvarStefansson (commented):
Looks very good to me. One substantial comment, we can discuss in person if necessary. Also, let's run all tests for good measure. Withholding approval to remind us of that.

tests/functional/test_manu_thermoporomech_nofrac.py
```diff
@@ -253,6 +254,7 @@ def actual_ooc(material_constants: dict) -> list[list[dict[str, float]]]:
                     "meshing_arguments": {"cell_size": 0.25},
                     "perturbation": 0.3,
                     "heterogeneity": 10.0,
+                    "times_to_export": [],  # Suppress output for tests
```
IvarStefansson, line 257: Have you somehow tried to check whether you have caught all instances where this is needed? I.e., are any vtu files are written by the verification tests that were not written previously?

IvarStefansson (approved):
Great work. If you're confident about the one comment I made, please feel free to merge.

src/porepy/applications/test_utils/models.py
vlipovac: No, removing declaration.

src/porepy/examples/mandel_biot.py
vlipovac: I moved the initialization of `.results` to the base solution strategy after shifting the functionality of `VerificationDataSaving` into `DataSavingMixin` and removing the former (as discussed in person). Now all models have it and it will be filled if the user provides a custom `collect_data` method.

tests/functional/test_manu_thermoporomech_nofrac.py
vlipovac: I checked the contents of `tests/functional` and file IO is purged from there.
Started doing so in `tests/models` but some of the set-ups are quite nested, I did not manage to get every one.

vlipovac:
@IvarStefansson  I requested another review, due to some changes which should be inspected closer.
F.e., the Biot model had some intermediate parent classes for constitutive laws and something denoted as solution strategy.

Upon closer inspection, the code could be simplified, as well as the inheritance tree of the Terzaghi and Mandel models.

The main insight I gained from this (and from looking into the intermediate "solution strategies" for various benchmarks and _manu_ models) is that they are not actual solution strategies in the original sense. They just add some stuff and like exact solutions, initial conditions and some model assertions. Most of it does not deserve the status of a solution strategy, but should maybe in the future be shifted to the actual model class, since it is only valid and used there (in short: exact solution, model assumptions, are currently declared as "solution strategy" mixins, but the code would be cleaner and simpler if those things are with the actual model class).

src/porepy/examples/mandel_biot.py
IvarStefansson: Sounds very reasonable

src/porepy/examples/terzaghi_biot.py
```diff
@@ -542,8 +539,10 @@ def bc_type_mechanics(self, sd: pp.Grid) -> pp.BoundaryConditionVectorial:
             the South, and rollers on the sides.
 
         """
-        # Inherit bc from parent class. This sets all bc faces as Dirichlet.
-        bc = super().bc_type_mechanics(sd=sd)
+        # Start with all faces as Dirichlet faces, analogous to base mechanics set-up.
+        boundary_faces = self.domain_boundary_sides(sd).all_bf
+        bc = pp.BoundaryConditionVectorial(sd, boundary_faces, "dir")
+        bc.internal_to_dirichlet(sd)
```
IvarStefansson, line 545: Why do you prefer this?

tests/functional/setups/manu_flow_incomp_frac_2d.py
```diff
@@ -755,17 +754,6 @@ class ManuIncompSolutionStrategy2d(
     results: list[ManuIncompSaveData]
     """List of SaveData objects."""
 
-    def __init__(self, params: dict):
-        """Constructor for the class."""
-
-        super().__init__(params)
-
-        self.exact_sol: ManuIncompExactSolution2d
```
IvarStefansson, line 763: Just to be sure I understand: Is exact_sol now not declared anywhere?

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
```diff
@@ -1031,16 +1036,16 @@ class ManuThermoPoroMechSolutionStrategy2d(
 ):
     """Solution strategy for the verification setup."""
 
+    exact_sol: ManuThermoPoroMechExactSolution2d
+    """Exact solution object."""
+
+    results: list[ManuThermoPoroMechSaveData] = []
```
IvarStefansson, line 1042: Isn't this initialization superfluous, since it also happens in SolutionStrategy.__init__?

tests/functional/setups/manu_thermoporomech_nofrac_3d.py
```diff
@@ -873,18 +873,18 @@ class ManuThermoPoroMechSolutionStrategy3d(
 ):
     """Solution strategy for the verification setup."""
 
+    exact_sol: ManuThermoPoroMechExactSolution3d
+    """Exact solution object."""
+
+    results: list[ManuThermoPoroMechSaveData] = []
```
IvarStefansson, line 879: Not needed?

src/porepy/models/derived_models/biot.py
IvarStefansson, line 1: I'm not sure I agree on this one. I know setting materials isn't a major solution strategy operation. However, since we've accepted SolutionStrategy is whatever doesn't fit elsewhere, that's where that method "belongs". And I can see the advantage of keeping the division into solution strategy, constitutive laws (equations, variables...) also for derived models, mainly for consistency and making locating the various adjustments easier. Opinions, @keileg?

IvarStefansson (commented):
Some relatively minor questions.

src/porepy/models/derived_models/biot.py
keileg: Generally I agree, the `SolutionStrategy`'s status as an orphanage will have such consquences - and let's not introduce a new practice in derived models.

If you want me to look at something more specific, please indicate a line number or similar.

src/porepy/examples/terzaghi_biot.py
vlipovac: Eliminating the need to call super and the inheritance from `BoundaryConditionsMechanics`.

I also tried `BoundaryConditionsMechanics.bc_type_mechanics(self,..)` with `self` as an explicit argument, but mypy was giving me hell.

Then I wagered that 3 lines of duplicate code are worth resolving the loop in the inheritance.

tests/functional/setups/manu_flow_incomp_frac_2d.py
vlipovac: It's declared in the class attribute section, as per (our) mixin convention.

tests/functional/setups/manu_thermoporomech_nofrac_2d.py
vlipovac: You are right, removed it.

tests/functional/setups/manu_thermoporomech_nofrac_3d.py
vlipovac: Yep, removed it like above.

src/porepy/models/derived_models/biot.py
IvarStefansson: The comment was related to the changes in the file biot.py

tests/functional/test_manu_thermoporomech_nofrac.py
IvarStefansson: Forgive me if this is a stupid question: where does that leave us?

src/porepy/models/derived_models/biot.py
vlipovac: I reverted the removal of `SolutionStrategyBiot`.

tests/functional/test_manu_thermoporomech_nofrac.py
vlipovac: I guess I got that onto myself.
I was able to purge it everywhere after some search. It now happens only in `tests/viz` where it is explicitly tested.

PR #1083, opened 2023-11-30
Issue 1031 ad value eval

Yuriyzabegaev:
## Proposed changes

The `Operator.evaluate` method is transformed into two separate methods: `value` and `value_and_jacobian`. Credit to @vlipovac 

## Types of changes

What types of changes does this PR introduce to PorePy?
_Put an `x` in the boxes that apply._

- [ ] Minor change (e.g., dependency bumps, broken links).
- [ ] Bugfix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [x] Testing (contribution related to testing of existing or new functionality).
- [ ] Documentation (contribution related to adding, improving, or fixing documentation).
- [ ] Maintenance (e.g., improve logic and performance, remove obsolete code).
- [ ] Other:

## Checklist

_Put an `x` in the boxes that apply or explain briefly why the box is not relevant._

- [x] The documentation is up-to-date.
- [x] Static typing is included in the update.
- [x] This PR does not duplicate existing functionality.
- [x] The update is covered by the test suite (including tests added in the PR).
- [ ] If new skipped tests have been introduced in this PR, `pytest` was run with the `--run-skipped` flag.

PR #1318, opened 2025-01-30
Move evaluation of Ad operators to helper class

### LINKED ISSUE #1181: Implement an Ad parser class
## Background
In the current code, the evaluation of a `pp.ad.Operator` is implemented in the `Operator` class itself. The parsing of a tree of an operator (which is really a computational graph) is implemented as a depth-first traversal which identifies leaves in the tree, parses these to numerical values, and then propagates the values through the graph in a bottom-up fashion. While this works, there are shortcomings with the approach:
1. The graph traversal method is tailored for evaluation of numerical values. While other relevant graph operations, such as compatibility checks between operator sizes, or consistency checks of units, would require a similar traversal, the current code would necessitate the writing of a new, but very similar, traversal function for each new operation.
2. There is no way in which we can identify identical nodes, or subgraphs, and cache evaluated values to save computational time.
3. The top-level Operator class is now responsible both for representing combinations of operations and the evaluation of computational graphs, thus violating the single responsibility principle. While this is not a major problem by itself, it is a signal that the current design is not ideal.

## Suggested changes
The following large-scale changes are suggested - specifications are needed before implementation:
1. Implement a function for traversal of a computational graph. To EK, it seems natural to convert the tree into a priority queue, but the choice of data structure may not be a primary concern. A rough signature for the function wolud be `def graph_traversal(list[pp.ad.Operator]) -> Iterator[pp.ad.Operator]`. The motivation for accepting a list of operators is given in the outline of the Ad parser class below.
2. Implement a class, say `AdParser`, which takes over responsibility for parsing from the `Operator` class. This will use the graph traversal function to iterate over the computational graph and convert it to numerical values. Moreover, the parser should contain functionality for chaching of operator values to reduce evaluation time.

## Thoughts on the `AdParser` class
Below is an outline of the class:

```python
class AdParser:

    mdg: MixedDimensionalGrid
   # Parsing of Operator leaves will require access to the grid. This is currently obtained
   # by passing the EquationSystem, but direct access is preferable.

    _cache_table_for_current_value: dict[pp.ad.Operator, np.ndarray]
    _cache_table_for_current_value_and_jacobian: dict[pp.ad.Operator, pp.ad.AdArray]
    # Temporarily store computed values here during evaluation. Not sure if we need separate
    # tables for numpy and Ad arrays

     _longer_term_caching_of_values: dict[pp.ad.Operator, np.ndarray]
    ...
    # Operators that will not (are not likely) to change between evaluation steps, and which are
    # costly to compute, can be stored here inbetween evaluations. This should not be part of
    # the initial implementation.

    def value(self, list[pp.ad.Operator]) -> list[np.ndarray]:
        # Fetch a joint iterator for the operators. Evaluation has two steps: First check if this operator
        # has previously been evaluated (check if it is found in the cache tables, which will use the new
        # hash functions), compute if not available.
        # The signature is non-trivial: We need to accept a list to utilize caching between evaluation of
        # multiple equations, but when iterating, we also need to keep track of which operator tree the
        # individual operators are associated with. Achiving this may require tweaks to the outlined implementation.
        # The need to cache between equations also show that parsing must be moved out of the
        # individual operators.

    def value_and_jacobian():
        # Similar to above

```
## Access to the `AdParser`
While it is tempting to make the `AdParser` a singleton accessible as `pp.AdParser` or similar, there is no clear reason to do so. Instead, it is suggested that all models are equipped with an `AdParser`, just as they have an `EquationSystem`. It should also be possible to parse outside a model, by a call `pp.ad.AdParser(mdg).value_and_jacobian([operator])`, though such operations will not be very common.

## The role of the `EquationSystem`
While the functionality of the `AdParser` could have been integrated into the `EquationSystem`, it is preferable not to add more functionality into this (arguably already overloaded) class. Instead, the `EquationSystem` can be given the model-wide `AdParser` as an attribute and call this for evaluation of equations.

## Dependencies
This should not be started until completion of #1179 and #1180

keileg:
## Proposed changes
This PR introduces a helper object `AdParser` which takes over the implementation of numerical evaluation of an Ad `Operator` from the operator object itself. The frontend method for evaluation is moved to the `EquationSystem` class.

At the moment, code has mainly been moved and cleaned up, but further improvements of the `AdParser` are planned.

Notable changes:
1. Evaluation of an Ad operator now takes the form `equation_system.operator_value(op)`, instead of `op.value(equation_system)`. There are technical reasons, related to caching of evaluations, to why it is natural to let the `EquationSystem` take over the responsibility.
2. The methods `Operator.value()` and `Operator.value_and_jacobian()` are still functioning, but are marked as deprecated. The code base and tests have been updated.
3. The Operator class has been given a new method `as_graph` which returns the operator represented through a `networkx` graph. Though this is currently not being used, it will be central to the further improvements of the `AdParser`, and I therefore decided to keep it in this PR.

Clarifications needed:
1. The method names `EquationSystem.operator_value` and `operator_value_and_jacobian` are not ideal, but the shorter `value` is also not good, since we are asking the `EquationSystem` to evaluate an operator, not for the value of the `EquationSystem`. I am fine with renaming in the direction of `evaluate(operator, jacobian: bool=False)` (there are many more calls to `value` than to `value_and_jacobian`, hence the default), however, I do remember that the names `Operator.value` and `value_and_jacobian` were hard-fought compremises, so I am open to other ideas.
2. No tests have been added, though the `AdParser` is thoroughly tested through its use in the models. I think it is reasonable to postpone testing of the `AdParser` until a clearer need arises; let me know if you think differently.

Closes #1181.

## Types of changes

- [ ] Other:

## Checklist

NB: The tests should be run with --run-skipped, but only after method names have been clarified, see above.


src/porepy/numerics/ad/operators.py
```diff
@@ -241,20 +272,90 @@ def name(self) -> str:
...
+        Since networkx uses the object hash as the node id, and operators can share the
+        same hash, we cannot use the operator themselves as nodes. Instead, the nodes in
+        the graph are integers. To map the nodes back to the operators, a dictionary is
+        returned. For a given node, the actual operator is stored as an attribute, with
+        key "obj".
+
+        The graph will actually be that of a tree with self as root. The parameter
+        depth_first determines if the graph is constructed, and node ids assigned, by
+        depth-first or breadth-first traversal.
+        EK comment: At the moment, it is not clear whether this has any pratcital impact
```
jwboth, line 299:
```suggestion
        EK comment: At the moment, it is not clear whether this has any practical impact
```

src/porepy/numerics/ad/_ad_parser.py
```diff
@@ -0,0 +1,529 @@
...
+        state: np.ndarray | None,
+    ) -> (
+        float
+        | np.ndarray
+        | sps.spmatrix
+        | pp.ad.AdArray
+        | list[float | np.ndarray | sps.spmatrix]
+        | list[pp.ad.AdArray]
+    ):
+        """Evaluate the operator x and its derivative if requested.
```
jwboth, line 167: x?

src/porepy/numerics/ad/_ad_parser.py
```diff
@@ -0,0 +1,529 @@
...
+                    # Jacobian). If evalutaion ends up here, it is most likely that the
+                    # operator to be evaluated violates this assumption. However, it is
+                    # also possible to arrive here by creating a pp.ad.SparseMatrix, and
+                    # evaluting it and also requesting the Jacobian. Though this is in a
+                    # sense a reasonable request, the SparseMatrix can also be evaluated
+                    # by calling _parse() directly, and this is the recommended way to
+                    # handle this case.
+                    raise NotImplementedError(
+                        f"The Jacobian of {type(res)} is not implemented because it is "
+                        "multidimensional"
```
jwboth, line 227:
```suggestion
                        "multidimensional."
```

src/porepy/numerics/ad/_ad_parser.py
```diff
@@ -0,0 +1,529 @@
...
+        match operation:
+            case Operations.add | Operations.sub:
+                # Addition and subtraction can be handled rather straightforwardly,
+                # though with some tweaks for subtraction.
+                assert len(child_values) == 2  # These operations are binary
+
+                # Take note of whether the operands are flipped.
+                flipped = False
+                if isinstance(child_values[0], np.ndarray):
+                    # We should not do numpy_array {+,-,*} Ad_array, since numpy will
```
jwboth, line 322: Comment on "*" redundant here.

src/porepy/numerics/ad/_ad_parser.py
```diff
@@ -0,0 +1,529 @@
...
+        # This is not a leaf, but a composite operator. Parse the children and combine
+        # them according to the operator.
+        child_values = [
+            self._evaluate_single(child, ad_base, eq_sys) for child in op.children
+        ]
+
+        # Get the operation represented by op.
+        operation = op.operation
+        match operation:
+            case Operations.add | Operations.sub:
```
jwboth, line 314: I somehow get the classification in +/- and */div routines. Though, one could also consider symbol-based and special cases. add seems equally treated as mul. So why was it not possible to integrated sub as special case below as rmatmul?

jwboth (commented):
As written in the description, the new code is a lot if moving of code and change in api for evaluation of operators. Thus, only minor comments are posed. To me this looks very good.

No opinions regarding naming. I would find a short method name with the use of flags to invoke Jacobians most friendly. But all other options would also work.

src/porepy/numerics/ad/_ad_parser.py
keileg: Should be op. Fixed.

src/porepy/numerics/ad/_ad_parser.py
keileg: Thanks.

src/porepy/numerics/ad/_ad_parser.py
keileg: I originally had * together with + and -, but it did not work out with the reversion of child_values. This does no harm for addition, is necessary for subtraction, and is also for necessary for multiplication - but only if the second operand is an AdArray, for other cases it actually does harm. I forgot precisely what was the problem, but it surely was something. 

Hence, the structure ended up as it is. There are a few other structures that would have been functionally equivalent and roughly as logical as this one, but I could not find one superior structure - for that, there are too many special cases to handle.

src/porepy/numerics/ad/_ad_utils.py
```diff
@@ -571,6 +565,18 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         return f"{self._name}({self._physics_key}).{self._discretization_matrix_key}"
 
+    def _key(self) -> str:
+        if self._cached_key is None:
+            domain_ids = [domain.id for domain in self.domains]
+            s = f"(Merged_operator, name={self.name}, domains={domain_ids})"
```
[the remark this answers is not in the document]
keileg: Yes, there is a risk of that. But this will also lead to problems with non-unique parameters (for which the physics_key and domain is used) and storage of discretization matrices (which uses discretization_matrix_key, physics_key and possibly inner_physics_key). Being lazy here is a really bad idea.

src/porepy/numerics/ad/operators.py
```diff
@@ -93,16 +86,75 @@ def _get_previous_time_or_iterate(
...
             for child in op.children
         ]
-
         return new_op
 
 
+class Operations(Enum):
+    """Object representing all supported operations by the operator class.
+
+    Used to construct the operator tree and identify Operations.
```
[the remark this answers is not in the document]
keileg: Good question. Identical values of the enum members may lead to identical hashes (though chances are miniscule). Having equal symbols are fine, since these identifies mathematical operations as understood by the Python interpreter (hence mul and rmul are the same).

I added a comment on the requirement of unique string values below.

tests/functional/test_benchmark_2d_case_3.py
```diff
@@ -134,7 +135,7 @@ def test_effective_normal_permeability(model) -> None:
 
     """
     for intf in model.mdg.interfaces():
-        val = model.effective_normal_permeability([intf]).value(model.equation_system)
+        val = model.equation_system.operator_value([intf])
```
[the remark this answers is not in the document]
keileg: Thanks. This reminds me I need to run the skipped tests before the merge - that should have picked up this one. Will do that once I have pushed all changes.

keileg:
@jwboth Thanks for the review. There are two comments left for you to have a look at. I have also ran the skipped tests successfully.

What we have not tackled is the name of the evaluation methods in `EquationSystem`, see clarification no 1 in the original post. Any suggestions on this? @IvarStefansson, you may also want a say here.

src/porepy/numerics/ad/_ad_utils.py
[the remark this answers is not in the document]
jwboth: Thanks for the feedback. Feel free to resolve.

IvarStefansson:
I also remember us having a very thorough discussion resulting in moving from evaluate to value + value_and_jacobian. My instinct is to trust the decision made then rather than reverting it. If we need to reconstruct the arguments, we could 1. look at the old issue/PR and 2. consult other people involved in the discussion, which I believe includes Veljko and Yury.

keileg:
@Yuriyzabegaev @vlipovac: Frontend responsibility for evaluation of an Ad operator is being moved from `Operator` to `EquationSystem` (in practice, the evaluation is moved to a new backend class). In that connection, it is tempting to merge the two methods `value` and `value_and_jacobian` back to `evaluate`. 

Do the two of you recall the reason we did the opposite split some time ago? Specifically, was the reason we did not want to deal with typing issues related to complex method signatures and return types, or was it something more?

vlipovac:
> @Yuriyzabegaev @vlipovac: Frontend responsibility for evaluation of an Ad operator is being moved from `Operator` to `EquationSystem` (in practice, the evaluation is moved to a new backend class). In that connection, it is tempting to merge the two methods `value` and `value_and_jacobian` back to `evaluate`.
> 
> Do the two of you recall the reason we did the opposite split some time ago? Specifically, was the reason we did not want to deal with typing issues related to complex method signatures and return types, or was it something more?

I dug up the respective PR:
https://github.com/pmgbergen/porepy/pull/1083

It does not say much unfortunately, but if I remember correctly there was a discussion on clarity, since the functionality to evaluate residuals only was introduced. The preference to have two separate methods instead of one with an boolean argument was voiced.
We also solved typing issues and reduced signature complexity on the go.
Also, reminder of the code was easy to type and mypy was happier, because the return type of each method was clear.

keileg:
> It does not say much unfortunately, but if I remember correctly there was a discussion on clarity, since the functionality to evaluate residuals only was introduced. The preference to have two separate methods instead of one with an boolean argument was voiced.
> We also solved typing issues and reduced signature complexity on the go.
> Also, other code (outside of porepy.ad) was easy to type and mypy was happier, because the return type of each method was clear.

This also fits with my recollection. I'll have a go at unifying the methods again, and see what mypy says. Thanks for the feedback @vlipovac.

keileg:
@IvarStefansson @jwboth: I unified operator evaluation into 
`EquationSystem.evaluate(op: Operators|list[Operator], derivative:bool=False, ...)`, with the default value for the derivative motivated by the many places where we only need the value (e.g. multiple places in `constitutive_laws`). I think the simpler method name is worth it, please let me know what you think.

The minor downside is that type overloading of `EquationSystem.value` is less than intuitive, see ~line 1900-> in `equation_system.py`. In all, I think this is worth it.

jwboth:
> @IvarStefansson @jwboth: I unified operator evaluation into `EquationSystem.evaluate(op: Operators|list[Operator], derivative:bool=False, ...)`, with the default value for the derivative motivated by the many places where we only need the value (e.g. multiple places in `constitutive_laws`). I think the simpler method name is worth it, please let me know what you think.
> 
> The minor downside is that type overloading of `EquationSystem.value` is less than intuitive, see ~line 1900-> in `equation_system.py`. In all, I think this is worth it.

This solution would also align with my opinion.

IvarStefansson (approved):

PR #1321, opened 2025-02-05
FEAT: Add remove_variables method to EquationSystem

IvarStefansson:
## Proposed changes

The new method updates the attributes set during create_variable, removing everything related to the variables to be removed.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/ad/equation_system.py
+++ b/src/porepy/numerics/ad/equation_system.py
@@ -477,6 +477,40 @@ class EquationSystem:
 
         return merged_variable
 
+    def remove_variables(self, variables: VariableList) -> None:
+        """Removes variables from the system.
+
+        Parameters:
+            variables: List of variables to remove. Variables can be given as a list of
+                variables, mixed-dimensional variables, or variable names (strings).
+
+        Raises:
+            ValueError: If a variable is not known to the system.
+
+        """
+        variables = self._parse_variable_type(variables)
+        for var in variables:
+            if var.id not in self._variables:
+                raise ValueError(f"Variable {var} not known to the system.")
+            # Remove the variable from the system.
+            # _variables and _variable_dof_type are indexed by variable id.
+            del self._variables[var.id]
+            self._variable_dof_type.pop(var.id)
+            # _variable_num_dofs is indexed by variable number, which is the value in
+            # _variable_numbers corresponding to the key var.id.
+            num = self._variable_numbers[var.id]
+            self._variable_num_dofs = np.delete(self._variable_num_dofs, num)
+            self._variable_numbers = {
+                key: val - 1 if val > num else val
+                for key, val in self._variable_numbers.items()
+            }
+            # Now we have used the number of the variable to update the numbers of all
+            # variables with higher numbers. We can remove the variable from the
+            # variable numbers map.
+            self._variable_numbers.pop(var.id)
+            # Update the variable clustering.
+            self._cluster_dofs_gridwise()
+
     def update_variable_tags(
         self,
         tags: dict[str, Any],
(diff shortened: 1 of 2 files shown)


src/porepy/numerics/ad/equation_system.py
```diff
@@ -477,6 +477,42 @@ def create_variables(
 
         return merged_variable
 
+    def remove_variables(self, variables: VariableList) -> None:
+        """Removes variables from the system.
```
keileg, line 481: Should the method also delete any stored values in data[pp.TIME_STEP_SOLUTIONS]? Same for iterate.

The answer is not clear to me, it depends on the use case.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -477,6 +477,42 @@ def create_variables(
...
+            num = self._variable_numbers[var.id]
+            self._variable_num_dofs = np.delete(self._variable_num_dofs, num)
+            self._variable_numbers = {
+                key: val - 1 if val > num else val
+                for key, val in self._variable_numbers.items()
+            }
+            # Now we have used the number of the variable to update the numbers of all
+            # variables with higher numbers. We can remove the variable from the
+            # variable numbers map.
+            self._variable_numbers.pop(var.id)
```
keileg, line 512: I am pretty sure that all of this is also done in the below call to self._cluster_dofs_gridwise.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -477,6 +477,42 @@ def create_variables(
 
         return merged_variable
 
+    def remove_variables(self, variables: VariableList) -> None:
```
keileg, line 480: The method will change the counting of the dofs, but not update the variable state. Hence it is not safe to use it after the variables have been initialized, not even with zero values (the vector size will be unexpectedly large). I think a clarification of the intended usage is in place, together with either clear warnings when to use the method or extended functionality (and tests).

tests/numerics/ad/test_equation_system.py
```diff
@@ -196,6 +199,77 @@ def test_variable_creation():
     )
 
 
+@pytest.mark.parametrize("variable_to_be_removed", ["var_1", "var_2", "var_3", None])
+def test_remove_variables(variable_to_be_removed):
```
keileg, line 203: See comment to method implementation regarding modification of variable state. If the method is extended, so should the test be.

keileg (commented):
Main question: What happens to the stored variable state when a variable is removed? I think this needs to be updated, or else all kind of things will break. Am I missing something?

src/porepy/numerics/ad/equation_system.py
IvarStefansson: I think not. I think of it as removing variables from the ES, not deleting them entirely.

src/porepy/numerics/ad/equation_system.py
IvarStefansson: Which vector size are you referring to?

src/porepy/numerics/ad/equation_system.py
IvarStefansson: So, to be clear, it should be safe to do self._variable_numbers.pop(var.id) and skip lines 503 through 508?

keileg: Yes.

src/porepy/numerics/ad/equation_system.py
keileg: Variable state vectors, so basically my comment is the same as above.

src/porepy/numerics/ad/equation_system.py
keileg: Let us discuss.

keileg: Conclusion: All will be fine, provided the user remembers to update any global vectors of system states etc. (e.g., the result of `equaiton_system.get_variable_state()` obtained prior to the removal of the variable.

keileg (approved):

Issue #1328, opened 2025-02-11
Intended save path for `write_time_information`


pschultzendorff:
Hei,

when running a model with the default `DataSavingMixin`, `save_data_time_step` calls `write_pvd_and_vtu`, which in turn calls `time_manager.write_time_information` **without** passing a value for `path`. Thus, the time file gets written to `visualization/times.json` relative to wherever the `cwd` is when one runs the simulation. This is quite annoying, as all other simulation output gets written to `params["folder_name"]`. The only way to change the output location is to rewrite `DataSavingMixin.save_data_time_step` in a mixin/subclass.

Is this intended behavior and I am overlooking something, or has this just not bothered anyone so far? :D

If the latter, I can write a quick PR to change the default save location to `params["folder_name"]` or add a parameter specifying the save location.

IvarStefansson:
@jwboth, I believe you know the times.json the best.

jwboth:
I think the answer is, "nobody bothered". I noticed the same, but put low priority to that. The times.json file was intended for those who crash their valuable PorePy simulations and wished to restart. I do not think this has ever been the case after this issue was filed once, so nobody was either looking for these files. 

I think a fix would be great. As indeed, the intended overwrite for the current state of the code (I do not recall if it was the same when writing the functionality for exporting times.json - this is maybe 2 years ago), is too heavy as to expect the user to do it. To me it sounds the easiest to remove the option of providing a path to write_time_information and instead always use params["folder_name"]/times.json where the remainder of the exported data also lies / or keep path but just replace visualization with params["folder_name"].

PR #1330, opened 2025-02-12
Two small fixes

pschultzendorff:
## Proposed changes
This fixes two tiny issues:
1. Every run used to save time steps to `visualization/times.json` relative to the `cwd`. This goes now to `params["folder_name"]/times.json"`. Solves #1328.
2. Progressbars used to not work correctly with adaptive time stepping. When a time step failed and was restarted with smaller time step size, the bar would nonetheless move to the right. Now, it only updates when the solver converged. I don't know when a time step counts as converged for the iterative models, so I kept that part of the code unchanged. My understanding is that this is barely used anyways.

## Types of changes
- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/viz/data_saving_model_mixin.py
+++ b/src/porepy/viz/data_saving_model_mixin.py
@@ -222,6 +222,9 @@ class DataSavingMixin(pp.PorePyModel):
 
         Parameters:
             vtu_files: Path(s) to vtu file(s).
+            time_index: Index of the time step to be loaded.
+            times_file: Path to json file storing history of time and time step size. If
+                ``None``, the same default path as in :meth:`write_pvd_and_vtu` is used.
             keys: Keywords addressing cell data to be transferred. If ``None``, the
                 mixed-dimensional grid is checked for keywords corresponding to primary
                 variables identified through ``pp.TIME_STEP_SOLUTIONS``.
@@ -232,13 +235,16 @@ class DataSavingMixin(pp.PorePyModel):
             ValueError: If incompatible file types are provided.
 
         """
-
         # Sanity check
         if not (
             isinstance(vtu_files, list)
             and all([vtu_file.suffix == ".vtu" for vtu_file in vtu_files])
         ) and not (isinstance(vtu_files, Path) and vtu_files.suffix == ".vtu"):
             raise ValueError
+
+        # Default path for times_file.
+        if times_file is None:
+            times_file = Path(self.params["folder_name"]) / "times.json"
 
         # Load states and read time index, connecting data and time history.
         self.exporter.import_state_from_vtu(vtu_files, keys, **kwargs)
@@ -261,7 +267,8 @@ class DataSavingMixin(pp.PorePyModel):
             pvd_file: Path to pvd file with exported vtu files.
             is_mdg_pvd: Flag controlling whether pvd file is a mdg file, i.e., generated
                 with ``Exporter._export_mdg_pvd()`` or ``Exporter.write_pvd()``.
-            times_file: Path to json file storing history of time and time step size.
+            times_file: Path to json file storing history of time and time step size. If
+                ``None``, the same default path as in :meth:`write_pvd_and_vtu` is used.
             keys: Keywords addressing cell data to be transferred. If ``None``, the
                 mixed-dimensional grid is checked for keywords corresponding to primary
                 variables identified through ``pp.TIME_STEP_SOLUTIONS``.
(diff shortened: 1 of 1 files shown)


src/porepy/numerics/time_step_control.py
```diff
@@ -745,7 +745,7 @@ def write_time_information(self, path: Optional[Path] = None) -> None:
 
         # Storing as json
         if path is None:
-            path = Path("visualization") / Path("times.json")
+            path = Path("visualization") / "times.json"
```
jwboth, line 748: My sense is, currently, there is two defaults provided, which is unfortunate. My feeling would be to remove the default use here. Then one possibility would be to say, that if no path is provided, no times.json file should be exported (as anyhow already implemented below).

src/porepy/models/run_models.py
```diff
@@ -101,7 +101,7 @@ def time_step() -> None:
             + f" of {model.time_manager.time_final:.1e}"
             + f" with time step {model.time_manager.dt:.1e}"
         )
-        solver.solve(model)
+        return solver.solve(model)
```
jwboth, line 104: I am in general a fan of collecting the performance status. This can be done in different ways, with different purposes. I see, how you want to use here. To me this looks OK. I would suggest adding a short comment mentioning the content, i.e., convergence-status, so one does not have to scroll to look for the purpose.

jwboth (commented):
To me this looks overall fine. I am just not so sure regarding having two proposed paths for where to store times.json. I wonder if it wouldn't be better to remove one.

src/porepy/numerics/time_step_control.py
pschultzendorff: I see your point. Unfortunately, `TimeManager` is not able to access `model.params`, hence `DataSavingMixin.write_pvd_and_vtu` has to pass a path and we cannot just have one default in `write_time_information`. Your suggestion sounds sensible. I will also remove the default path in `load_time_information` then.

src/porepy/models/run_models.py
pschultzendorff: Done.

src/porepy/numerics/time_step_control.py
pschultzendorff: Pushed the update :)

src/porepy/numerics/time_step_control.py
```diff
@@ -722,16 +722,15 @@ def is_schedule_in_simulated_times(
         return schedule.size == np.sum(in1d)
 
     # I/O
-    def write_time_information(self, path: Optional[Path] = None) -> None:
```
jwboth, line 725: I think this can remain. The "default" case of a None path will trigger no export.

src/porepy/numerics/time_step_control.py
```diff
@@ -742,27 +741,20 @@ def write_time_information(self, path: Optional[Path] = None) -> None:
...
-
-        # Storing as json
-        if path is None:
-            path = Path("visualization") / Path("times.json")
-
         path.parent.mkdir(parents=True, exist_ok=True)
         with open(path, "w") as out_file:
             json.dump({"time": self.exported_times, "dt": self.exported_dt}, out_file)
 
-    def load_time_information(self, path: Optional[Path] = None) -> None:
```
jwboth, line 754: Same here.

jwboth (commented):
The small disadvantage currently is that since no "default" case is treated in the loading part, one has to adapt the paths when really restarting. But I feel that this should be OK. It is only a matter of convenience and not functionality.

keileg:
@pschultzendorff Be advised that we have changed the formatter from `black` to `ruff`, and also use `ruff` instead of `flake8`. You will need to do the following steps in order to pass the static tests:

`pip install ruff`
`ruff format src` (equivalent to black)
`ruff check src` (equivalent to flake8)

The changes on your files should be minimal, if any.

pschultzendorff:
> @pschultzendorff Be advised that we have changed the formatter from `black` to `ruff`, and also use `ruff` instead of `flake8`. You will need to do the following steps in order to pass the static tests:

Good to know, thanks!

pschultzendorff:
> The small disadvantage currently is that since no "default" case is treated in the loading part, one has to adapt the paths when really restarting. But I feel that this should be OK. It is only a matter of convenience and not functionality.

I think if the correct default path is passed to `TimeManager.load_time_information` in `load_data_from_vtu` and `load_data_from_pvd`, this should work. Or are there any other relevant methods for restarting?

jwboth:
No, I think you are right. My message sounded too strict maybe. I did not want to stress any issue there. Just an observation.

src/porepy/numerics/time_step_control.py
jwboth: Sorry, I somehow scrolled to far below, and just saw the ignoring behavior for the load when path is None. I am OK with your change then!

pschultzendorff:
> No, I think you are right. My message sounded too strict maybe. I did not want to stress any issue there. Just an observation.

No, it makes sense. I added the default behavior for loading now. Not sure about `ruff`, locally everything passes but previously there seemed to be some error.

IvarStefansson (approved):
Looks good to me. Thanks,  @pschultzendorff!

PR #1333, opened 2025-02-13
Make restrict to subgrid a method of tensor

IngridKJ:
## Proposed changes

This PR addresses something encountered when trying to use more than only the two default constitutive parameters (lambda and mu) of the momentum balance model class. The "discovery" is related to when (often large) grids are divided into subgrids for overcoming big memory usage upon discretization. In that case, the tensors (second order/fourth order) are restricted to the subgrids. The restriction-functionality was implemented as a function in fvutils, making adaptations in the case of several more constitutive parameters cumbersome. In this PR the restriction-functionality is moved to be part of the (new) Tensor class, which is intended to keep common parts between the second and fourth order tensors.

Implementation kudos to @IvarStefansson.

Currently only a draft PR as documentation is still a WIP.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/tests/params/test_tensor.py
+++ b/tests/params/test_tensor.py
@@ -11,6 +11,8 @@ def fourth_order_tensor_for_testing(
     """Construct a fourth order tensor with custom fields for testing.
 
     The tensor which is constructed here is for an anisotropic and homogeneous medium.
+    Specifically, the medium is represented by five material parameters and represents a
+    transversely isotropic medium.
 
     Parameters:
         num_cells: The number of cells the tensor is constructed for.
@@ -24,7 +26,10 @@ def fourth_order_tensor_for_testing(
     # parameters. The cell-wise parameter value for each basis is determined by a 1D
     # array, referred to as a "field". The matrices that form the basis for different
     # parameters, along with their corresponding fields, are organized as tuples. These
-    # tuples are later provided to FourthOrderTensor when constructing the tensor.
+    # tuples are later provided to FourthOrderTensor when constructing the tensor. As
+    # the stiffness tensor for the transversely isotropic media we consider here is
+    # determined by five independent material parameters, we need five matrices and five
+    # fields:
     matrix_and_field_1 = (
         np.array(
             [


keileg:
Related to this PR, it could also be of interest to add methods 'is_isotropic' and `is_diagonal` to `SecondOrderTensor`. The motivation is to speed up discretization, as `mpfa` can potentially be made significantly faster based on such information.

Knowledge of the structure of the tensor is available at initialization, based on which input parameters are given (`kxx`, `kyy` etc.). 

It is less relevant to do a similar modification of the `FourthOrderTensor` class, since it is much less clear how to exploit such information in `mpsa`.

IMPORTANT: This suggestion should be seen as an add-on to the original project, and should be dropped or postponed if it makes the PR to complex.

IngridKJ:
Current status: Tests now run. It was only a matter of changing the default value of `other_fields` in the `__init__` of `pp.FourthOrderTensor` and relabel a method from `@staticmethod` to `@property`.

Additionally, one (or multiple) tests should be made.

src/porepy/params/tensor.py
```diff
@@ -135,6 +135,57 @@ def __init__(
...
+
+        if nd == 1:
+            return bool(np.unique(kxx).size == 1)
+
+        # Extract relevant terms
+        kyy = np.array(values[1, 1, ::])
+        kzz = np.array(values[2, 2, ::]) if nd == 3 else np.array(0)
+
+        # Check if diagonal elements are equal
+        if not bool(np.all(kxx == kyy)):
```
IvarStefansson, line 182: Is the bool really needed?

src/porepy/params/tensor.py
```diff
@@ -135,6 +135,57 @@ def __init__(
...
+
+        Returns:
+            True if the tensor represents an isotropic medium, False if not.
+
+        """
+        values = self.values
+        kxx = np.array(values[0, 0, ::])
+
+        if nd == 1:
+            return bool(np.unique(kxx).size == 1)
```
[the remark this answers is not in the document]
IngridKJ: Yes you're right, thanks!

src/porepy/params/tensor.py
IngridKJ: I wrote a simple "test"-file and got some weird behavior without the bool, but when checking again now I see that it is not needed. Probably messed up the "test" I performed. I will remove the bools, thanks!

tests/params/test_tensor.py
```diff
@@ -0,0 +1,86 @@
...
+    kxx = np.random.uniform(0.5, 1.0, 16)
+    kyy = np.random.uniform(0.5, 1.0, 16)
+    kxy = np.random.uniform(0.1, 0.5, 16)
+
+    k = pp.SecondOrderTensor(kxx=kxx, kyy=kyy, kxy=kxy)
+
+    # Define some cells which we want to restrict the tensor k to:
+    active_cells = np.array([2, 1, 14, 10, 15])
+
+    # First copy k, and then restrict the copied k to the cell we chose above.
```
IvarStefansson, line 214:
```suggestion
    # First copy k, and then restrict the copied k to the cells we chose above.
```

src/porepy/numerics/fv/biot.py
```diff
@@ -341,14 +341,13 @@ def discretize(self, sd: pp.Grid, sd_data: dict) -> None:
             sd, active_cells
         )
         # Constitutive law and boundary condition for the active grid
-        active_constit: pp.FourthOrderTensor = (
-            pp.fvutils.restrict_fourth_order_tensor_to_subgrid(constit, active_cells)
-        )
+        active_constit = constit.copy()
```
keileg, line 344: How about giving `restrict_to_cells` a parameter `return_copy` instead?

src/porepy/params/tensor.py
```diff
@@ -158,7 +245,10 @@ class FourthOrderTensor(object):
     """
 
     def __init__(
-        self, mu: np.ndarray, lmbda: np.ndarray, phi: Optional[np.ndarray] = None
+        self,
+        mu: np.ndarray,
+        lmbda: np.ndarray,
+        other_fields: dict[str, tuple[np.ndarray, np.ndarray]] = {},
```
keileg, line 251: Don't initialize mutable optional arguments like that. See [here](https://web.archive.org/web/20200221224620id_/http://effbot.org/zone/default-values.htm) for an explanation.

src/porepy/params/tensor.py
```diff
@@ -101,6 +135,60 @@ def __init__(
...
+            True if the tensor represents an isotropic medium, False if not.
+
+        """
+        values = self.values
+        kxx = np.array(values[0, 0, ::])
+
+        if nd == 1:
+            return True
+
+        # Extract relevant terms
```
keileg, line 180: Proper punctuation, please.

src/porepy/params/tensor.py
```diff
@@ -5,12 +5,46 @@
...
+        """Restrict constitutive parameters to cells.
+
+        Used in the case of problems being split into several sub-problems.
+
+        Parameters:
+            cells: Indices of the cells which the constitutive parameters should be
+                restricted to.
+
+        """
+        # Restrict all constitutive parameters.
```
keileg, line 38: Please document this better (it is needed for the fourth order tensor etc.)

keileg (commented):
Some comments after a quick read through the code (not the test, though)

tests/params/test_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+
+def fourth_order_tensor_for_testing(
+    num_cells: np.array, return_tensor_and_matrix: bool
+):
+    """Construct a fourth order tensor with custom fields for testing.
+
+    The tensor which is constructed here is for an anisotropic and homogeneous medium.
+
+    Parameters:
+        return_tensor_and_matrix: Flag telling whether only the tensor should be
```
keileg, line 15: Please document `num_cells`

tests/params/test_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+            returned (False) or both the tensor and the matrix representation of the
+            tensor for one cell (True).
+
+    Returns:
+        The fourth order tensor with custom fields, or the tensor and the matrix
+        representation of the tensor for one cell.
```
keileg, line 21:
```suggestion
        The fourth order tensor with custom fields (if return_tensor_and_matrix is False), or the tensor and the matrix
        representation of the tensor for one cell.
```

tests/params/test_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+
+    """
+    # The various matrices serve as the foundation for building a tensor for specific
+    # parameters. The cell-wise parameter value for each basis is determined by a 1D
+    # array, referred to as a "field". The matrices that form the basis for different
+    # parameters, along with their corresponding fields, are organized as tuples. These
+    # tuples are later provided to FourthOrderTensor when constructing the tensor.
+    matrix_and_field_1 = (
```
keileg, line 29: Any reason we need five of these? I would think that one or two would suffice.

keileg: Also, since someone will wonder about this in a few years time: Is there any rational behind the patterns in these matrices, or are they essentially meaningless?

tests/params/test_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+    The methods return whether the tensor is diagonal or isotropic. This test tests that
+    the methods return the expected result for different second order permeability
+    tensors. Specifically we consider here isotropic, anisotropic, diagonal,
+    non-diagonal, homogeneous and heterogeneous tensors. All three spatial dimensions
+    are tested.
+
+    Parameters:
+        kxx: Array with cell-wise values of kxx permeability.
+        kyy: Array of kyy.
+        kzz: Array of kzz.
```
keileg, line 166: Not present among the parameters. Please update.

tests/params/test_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+        tensor = pp.SecondOrderTensor(kxx=kxx, kyy=kyy, kxy=kxy)
+    else:
+        tensor = fourth_order_tensor_for_testing(
+            num_cells=num_cells, return_tensor_and_matrix=False
+        )
+    # Define some cells which we want to restrict the tensor k to:
+    active_cells = np.array([2, 1, 14, 10, 15])
+
+    # First copy k, and then restrict the copied k to the cells we chose above.
+    tensor_copy = tensor.copy()
```
keileg, line 215: Remember to update this when the restrict method includes a copy.

tests/params/test_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+    custom fields. The tensor is assumed to be homogeneous, and the function returns
+    both the tensor and a hardcoded matrix representation of the tensor for one cell.
+    The matrix representation is used as a blueprint for checking if the tensor is
+    constructed correctly.
+
+    """
+    tensor, matrix = fourth_order_tensor_for_testing(
+        num_cells=5, return_tensor_and_matrix=True
+    )
+    num_cells = tensor.values.shape[2]
```
keileg, line 240: Rather make `num_cells` a parameter, feed it to the factory method, and check that  `shape[2]` is as expected. It will always be, but why not test it.

tests/params/test_tensor.py
keileg, line 15: We can also always return both arguments, and dump the second one if not needed. This is arguably simpler code, but feel free to fix as you see fit.

tests/params/test_tensor.py
IngridKJ: I wanted to keep the tensor meaningful, and the easiest choice I had for that was those five matrices (as I use them in my own work and already have them easily accessible). They correspond to a transversely isotropic media, so yes, there is a rationale behind them.  I can adapt it to fewer matrices, but then the tensor might be meaningless.

tests/params/test_tensor.py
IngridKJ: It is not clear to me what you mean here. Can you please elaborate?

IngridKJ: Never mind, I got it, I just misread your comment. Thanks!

tests/params/test_tensor.py
keileg: I suggest to make a comment of the physical interpretation and leave it at that.

keileg (approved):
Thanks.

Issue #1312, opened 2025-01-23
Make application geometries with non-matching grids


keileg:
Currently the available model geometries used for testing (e.g., applications/model_geometries.py) exclusively apply matching geometries. We should make a few cases available.

Suggested approach for the initial steps:
1. Take as a starting point the mixin SquareDomainOrthogonalFractures (2d geometry, up to fractures), and create a mixin that inherits from this class.
2. Make simple functionality to convert the grid covering this geometry to a non-matching grid. This will likely involve overriding set_geometry to (after call to super) to modify `self.mdg`. 
3. We will only refine the fracture and (later) mortar grid; the 2d grid will stay. The level of sophistication in the refinement (and possibly coarsening) will be specified under way. This should also make it work for Cartesian grids.
4. We need to find a test case to which we want to apply the geometry. To be specified.
5. A 3d case might be included later. To be decided.

It is an open question how many of the tests, and PorePy functionality in general, will be able to deal with non-matching grids, but that is a secondary concern.

jhabriel:
Just to give my unsolicited two cents since I'm currently working with non-matching grids:

Another option, instead of focusing on individual geometries (such as `SquareDomainOrthogonalFractures`) could be to produce non-matching grids starting from matching mdgs via a mixing `NonMatchingGridGenerator`. This can be catered to several levels of flexibility, but for testing sake, an easy take could be to pass global refinement targets for, say, subdomains of dimension n-1, and interfaces of dimension n-1 and then use `replace_subdomains_and_interfces()`.

keileg:
I agree that your suggestion is more useful in a general setting @jhabriel. If we end up applying non-matching grids more regularly, we will likely do something in the direction you sketch, or better, rely on functionality that you implement. 

The goal of this project is however limited to enabling testing on non-matching grids at all, and then (later) to do some diagnostics of which models are ready for non-matching grids. I expect this to be an efficient generator of GH issues.

keileg:
@IngridKJ @zhangyh0713 I took a short peak at the branch you are working on, and want to offload my thoughts here: it seems you are well under way; just be aware that the domain you are working with (`SquareDomainOrthogonalFractures`) can have 0, 1, or 2 fractures; steered by `self.params` (see implementation). You should be careful not to replace a fracture which is not present.

[...]

PR #1334, opened 2025-02-17
Nonmatching grid tests

zhangyh0713:
## Proposed changes

Issue: #1312
We created a new class to generate non-matching grids, which can handle cases with 0 to 2 orthogonal fractures. It returns a refined fracture set and a refined interface map. Its functionality is tested in test_fluid_mass_balance.py. The fluid mass problem, previously tested only on matching grids, is also tested with non-matching grids.

## Types of changes

- [ ] Other:

## Checklist


src/porepy/applications/md_grids/model_geometries.py
```diff
@@ -155,3 +155,96 @@ def meshing_arguments(self) -> dict:
...
+              the "recipient" mixed-dimensional grid. Fractures must be located at the
+              same place and the physical dimension of the grids must be the same.
+
+        """
+        super().set_geometry()
+
+        # Refine and replace fracture grids:
+        old_fracture_grids = self.mdg.subdomains(dim=1)
+
+        # Ratios which we want to refine the fracture grids with.
```
keileg, line 193: Does it make sense to make this a user-controlled parameter?

src/porepy/applications/md_grids/model_geometries.py
```diff
@@ -155,3 +155,96 @@ def meshing_arguments(self) -> dict:
...
+        new_fracture_grids = [
+            pp.refinement.refine_grid_1d(g=old_grid, ratio=ratio)
+            for old_grid, ratio in zip(old_fracture_grids, ratios)
+        ]
+
+        grid_map = dict(zip(old_fracture_grids, new_fracture_grids))
+
+        # Refine and replace interface grids:
+        # We first create a new and more refined mixed-dimensional grid.
+        def mdg_func(self, nx=2, ny=2) -> pp.MixedDimensionalGrid:
```
keileg, line 207: Why default values here?

src/porepy/applications/md_grids/model_geometries.py
```diff
@@ -155,3 +155,96 @@ def meshing_arguments(self) -> dict:
...
+
+            """
+            fracs = [self.fractures[i].pts for i in range(len(self.fractures))]
+            domain = self.domain.bounding_box
+            md_grid = pp.meshing.cart_grid(
+                fracs, np.array([nx, ny]), physdims=[domain["xmax"], domain["ymax"]]
+            )
+            return md_grid
+
+        mdg_new = mdg_func(self, nx=6, ny=6)
```
keileg, line 222: Two questions: 
1) Why not formulate this as refinement (so a ratio), rather than cell number? This will align better with the approach taken for the fracture grids.
2) Can the refinement level become a user controlled parameter?

src/porepy/applications/md_grids/model_geometries.py
```diff
@@ -155,3 +155,96 @@ def meshing_arguments(self) -> dict:
...
+        g_new = mdg_new.subdomains(dim=2)[0]
+        g_new.compute_geometry()
```
keileg, line 224: Can these be deleted?

src/porepy/applications/md_grids/model_geometries.py
```diff
@@ -155,3 +155,96 @@ def meshing_arguments(self) -> dict:
...
+                  update the fracture numbering such that the new fracture grids have
+                  the same fracture number as the old ones. This is because the fracture
+                  numbering is used to identify the correct interface grids later.
+                  TODO: Make sure that the refine-grid-functions also copy frac_num
+                  attribute such that the abovementioned is no longer a problem.
+            * Ensure that the "donor" mixed-dimensional grid is physically the same as
+              the "recipient" mixed-dimensional grid. Fractures must be located at the
+              same place and the physical dimension of the grids must be the same.
+
+        """
```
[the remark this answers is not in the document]
keileg: Generally, we try to avoid such comments that reflect on the development processes rather than on the resulting code, as only the latter is of interest later on. I tried to add the most important of these comments below, feel free to add if I have missed anything.

keileg (commented):
Nice work. I added some comments, and suggest to delete most of the documentation that reflect the development, rather than the result.

Main request: This function will be much more useful (also outside tests) if the user can controll the refinement process in fractures and interfaces. Please consider how this can be controlled by using parameters.

keileg:
We should also make sure the geometry setup is directly tested, not just through the inclusion in higher-order tests.

IngridKJ:
Note to selves after randomly meeting @keileg in the hallway: Make a test for testing the geometry itself.

src/porepy/applications/md_grids/model_geometries.py
IngridKJ: I addressed these two things in a commit just now. Would you have a quick look whenever you have time to see if you agree with closing this conversation? My sketch of 1. feels a bit brute force... 

Also addressed the fracture refinement ratios being controlled by a user parameter, so that conversation might be closed soon too.

keileg:
This look fairly good now. Let's try to do a final change in today's session: We see if we can move the actual mesh generation to `mdg_library,` and let the class `NonMatchingSquareDomainOrthhhogonalFractures` fetch from the library.

src/porepy/applications/md_grids/mdg_library.py
```diff
@@ -70,6 +80,89 @@ def square_with_orthogonal_fractures(
...
+            "fracture_refinement_ratio", 2
+        )
+        interface_refinement_ratio: pp.number = meshing_kwargs.get(
+            "interface_refinement_ratio", 2
+        )
+
+        # For the fracture grids, we can use the pp.refinement.refine_grid_1d function
+        # and impose the ratio directly. For the interface grids, the simplest way of
+        # refining is to create a new mixed-dimensional grid with the same geometry, but
+        # with refined rates, and then map the new grids to the old ones.
```
IvarStefansson, line 98: What are "rates"?

src/porepy/applications/md_grids/model_geometries.py
```diff
@@ -155,3 +159,97 @@ def meshing_arguments(self) -> dict:
...
+
+                # Checking the fracture number of the secondary grid in the recipient
+                # mdg. If they are the same, i.e., if the fractures are the same ones,
+                # we update the interface map.
+                if g_sec_coarse.frac_num == g_sec.frac_num:
+                    intf_map.update({intf_coarse: intf})
+
+        # Finally replace the subdomains and interfaces in the original
+        # mixed-dimensional grid.
+        self.mdg.replace_subdomains_and_interfaces(sd_map=grid_map, intf_map=intf_map)
```
IvarStefansson, line 252: Aren't lines 192 through 252 just a repetition of what happens during call to pp.mdg_library.square_with_orthogonal_fractures?

tests/applications/md_grids/test_mdg_library.py
```diff
@@ -113,6 +115,67 @@ def test_two_intersecting_custom_values(self):
...
+        # Loop over the fractures, fetch the projections to the 2d grid, and count the
+        # number of non-zero entries in the projection matrix.
+        non_zero_projection_primary = 0
+        for mg in self.mdg.interfaces(dim=1):
+            proj_primary = mg.mortar_to_primary_avg()
+            non_zero_projection_primary += proj_primary.nnz
+        # A matching grid would have equality here. We have refined the mortar grid,
+        # hence there should be more items in the projection matrix than there are
+        # fracture faces in the matrix grid.
+        assert non_zero_projection_primary > num_fracture_faces_from_matrix
```
IvarStefansson, line 169: For cartesian grids, we could check that these match the refinement ratios, right?

tests/applications/md_grids/test_model_geometries.py
```diff
@@ -0,0 +1,68 @@
...
+        expected_fracture_cell_num: The expected number of fracture cells after
+            refinement by the ratio(s) found in `fracture_refinement_ratios`.
+        expected_interface_cell_num: The expected number of interface cells
+            refinement by the ratio found in `interface_refinement_ratio`.
+
+    """
+    params = {
+        "fracture_indices": fracture_indices,
+        "fracture_refinement_ratios": fracture_refinement_ratios,
+        "interface_refinement_ratio": interface_refinement_ratio,
```
IvarStefansson, line 58: Why is there a single interface ratio and multiple fracture ratios?

IvarStefansson (commented):
Some minor comments.

tests/applications/md_grids/test_mdg_library.py
keileg: Good point, this applies also the mortar-fracture ratio below. Implemented.

src/porepy/applications/md_grids/model_geometries.py
```diff
@@ -155,3 +155,53 @@ def meshing_arguments(self) -> dict:
...
+
+        # The fracture grids are replaced by a refined version of the fractures which
+        # are already present in the grid. The interface grids are replaced by first
+        # creating a new mixed-dimensional grid with a higher resolution. The new
+        # mixed-dimensional grid will "donate" its interface grids to the original
+        # mixed-dimensional grid.
+
+        # First set the geometry and create a matching mixed-dimensional grid.
+        assert isinstance(self, pp.ModelGeometry)
+        super().set_geometry()  # type:ignore[safe-super]
```
keileg, line 185: @IvarStefansson please consider the following in particular: The call to set_geometry will generate a grid, which we then disregard in the below call. This can become costly if the grid is fine, while for standard testing, the cost will be negligible. I see three solutions:
1. Ignore the extra cost.
2. Replace the call to super().set_geometry() with the code contained within that method (call self.set_domain(), etc.). This is in effect a reimplementation of set_geometry(), and is risky if we ever change the super method.
3. Introduce a new method in model.geometry, create_grid(), and change the present method to reimplement only that one.

What do you think?

IngridKJ:
@IvarStefansson: Thoughts about whether the test in test_model_geometries.py should live?

keileg:
@IvarStefansson This should be ready for re-review now.

src/porepy/applications/md_grids/model_geometries.py
IvarStefansson: I'm inclined towards 3.

tests/applications/md_grids/test_mdg_library.py
```diff
@@ -113,6 +115,67 @@ def test_two_intersecting_custom_values(self):
...
+
+        # For the mortar grid vs fracture grids, we can simply do a cell count. We do
```
IvarStefansson, line 184: What does "mortar grid vs fracture grids" mean?

IvarStefansson:
> @IvarStefansson: Thoughts about whether the test in test_model_geometries.py should live?

You've added that test for the new function, right? Is there any reason why you wouldn't want to keep it?

src/porepy/applications/md_grids/model_geometries.py
keileg: Agree. I will implement.

tests/applications/md_grids/test_mdg_library.py
keileg: Improved comment.

IngridKJ:
> > @IvarStefansson: Thoughts about whether the test in test_model_geometries.py should live?
> 
> You've added that test for the new function, right? Is there any reason why you wouldn't want to keep it?

Mostly due to discussions between Yuhe, Eirik and I earlier, and Eirik suggested me to ask you here. A reason to delete it is that we are essentially just calling a method which is tested elsewhere. As it is tested elsewhere, maybe the test for the model geometry (which is essentially just a function call) can be deleted.

IvarStefansson:
> > > @IvarStefansson: Thoughts about whether the test in test_model_geometries.py should live?
> > 
> > 
> > You've added that test for the new function, right? Is there any reason why you wouldn't want to keep it?
> 
> Mostly due to discussions between Yuhe, Eirik and I earlier, and Eirik suggested me to ask you here. A reason to delete it is that we are essentially just calling a method which is tested elsewhere. As it is tested elsewhere, maybe the test for the model geometry (which is essentially just a function call) can be deleted.

In that case, I agree.

keileg:
@IvarStefansson I introduced methods `create_fracture_network` and `create_mdg` as we discussed, and rewrote the non-matching model so that it only overrides `create_mdg`. I also had a look at the other model geometries (in applications and examples), but found no cases where it was natural to take the new methods into use. 

I think the PR is ready for re-review, in due course.

src/porepy/applications/md_grids/model_geometries.py
```diff
@@ -165,25 +165,13 @@ class NonMatchingSquareDomainOrthogonalFractures(SquareDomainOrthogonalFractures
     geometry allows for non-matching grids and different resolution for each grid.
     """
 
-    def set_geometry(self) -> None:
-        """Define geometry and create a non-matching mixed-dimensional grid.
+    def set_mdg(self) -> None:
```
IvarStefansson, line 168: Shouldn't this be create_mdg?

IvarStefansson (approved):
One critical comment, otherwise good!

src/porepy/applications/md_grids/model_geometries.py
keileg: Yes.

PR #1348, opened 2025-02-25
MAINT: Added keyword to numpy reshape calls

keileg:
When juggling with different numpy versions, I ran into some issues where calls of the type `array.reshape((n_row, n_dim), "F")` were not permitted. This PR changes these few cases to  `array.reshape((n_row, n_dim), order="F")`.

I did a search through the code base, and found no other cases where we used `reshape` with an order argument but without specifying the `order` keyword.

## Proposed changes

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/params/bc.py
+++ b/src/porepy/params/bc.py
@@ -314,9 +314,9 @@ class BoundaryConditionVectorial(AbstractBoundaryCondition):
 
         #  Default robin weights
         r_w = np.tile(np.eye(sd.dim), (1, sd.num_faces))
-        self.robin_weight = np.reshape(r_w, (sd.dim, sd.dim, sd.num_faces), "F")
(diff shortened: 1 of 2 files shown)


IvarStefansson (commented):
Is there any danger the same applies to the order parameter of the ravel method?

keileg:
> Is there any danger the same applies to the order parameter of the ravel method?

Numpy did not complain, so I guess no. I am also not sure whether the issue was the missing keyword or something else, but since adding the keyword fixed it (and was the natural interpretation of the error message, from what I recall), it seems like a fair option. 

I am fine with gradually introducing `order=..` in reshape as well, to be be more explicit, but it does not seem urgent.

IvarStefansson (approved):

PR #1349, opened 2025-02-25
Minor change: Improved time manager

Yuriyzabegaev:
## Proposed changes

This is a small change that improves (at least my) experience of using PorePy. When using dynamic time stepping, the lower limit of a time step is by default inferred to be `Δt_min = 0.001 * t_max`. This will raise a mistake if I set the initial time step to be smaller, which is usually desired. Therefore, I must always pass an additional argument to change this limit.

My suggestion is to change this rule for the lower limit to the following: `Δt_min = min(0.001 * t_max, Δt_init)`, because if the user passes a small time step, it is obviously considered within the limits. I don't think this change can harm anything, please let me know if I'm wrong!

Upd: I also updated the corresponding tests to the new behavior.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/time_step_control.py
+++ b/src/porepy/numerics/time_step_control.py
@@ -132,9 +132,10 @@ class TimeManager:
             `schedule` are compatible.
         dt_min_max: Minimum and maximum permissible time steps.
             If None, then the minimum time step is set to 0.1% of the final simulation
-            time and the maximum time step is set to 10% of the final simulation time.
-            If given, then the first and second elements of the tuple corresponds to the
-            minimum and maximum time steps, respectively.
+            time or the initial time step if it is smaller than that. The maximum time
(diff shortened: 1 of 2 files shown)


keileg:
I agree that the code must be updated, but wonder if the lower limit should not be some fraction (say, 0.1) of the initial time step. There will anyhow be something random about this, but having some safety margin to the smallest time step seems preferable to me.

IvarStefansson (approved):
Thanks for fixing. Regarding @keileg's comment, I think @Yuriyzabegaev's  solution seems fair, since it is the minimal adjustment for the given situation.

keileg:
> Thanks for fixing. Regarding @keileg's comment, I think @Yuriyzabegaev's solution seems fair, since it is the minimal adjustment for the given situation.

Fine with me.

Issue #1151, opened 2024-04-25
Contact mechanics softening factor


IvarStefansson:
In the contact mechanics equations, we currently compare tractions to c_num times (increments of) displacement jumps. C_num is by default computed as 
c_num = shear modulus / (solid.residual_aperture() + solid.fracture_gap()) * SOFTENING FACTOR.

From a physical intuition, this resembles a Hooke's law, thus ensuring that the tractions and scaled displacements have similar magnitudes. However, the justification of the softening factor is truly heuristic ("making the fracture softer than the matrix " for the default value of 0.1). We should consider to either purge it or replace the shear modulus with the fracture's elastic modulus, i.e. the fracture stiffness parameter (which typically takes values roughly one order of magnitude smaller than the Lame parameters).

IvarStefansson:
This is partly related to #957

IvarStefansson:
I think we can solve this based on the now scaled version as well as some improved understanding of the problem. Let's discuss in person, @keileg and @jwboth.

IvarStefansson:
Suggested action:
- Remove softening factor
- Write down mechanical equations with all scalings
- Improve documentation

jwboth:
Some preliminary observations/conclusions - still WIP and not ready for discussion but as a reminder for later discussion:

- `characteristic_contact_traction` among the material parameters is not used in the code and can be removed.
- `characteristic_displacement` in exact sense does not enter longer the effective definition of c_num, as indicated in the issue description. When rescaling the contact equations with characteristic_contact_traction as implemented in the core, c_num has the interpretation c_num = shear modulus / L * SOFTENING FACTOR, where L is the max length of the geometry. Numerically, the performance of Newton's method remains depening on the choice. Yet, not too strongly. One guess is conditioning of the contact condition.
- If `characteristic_displacement` is chosen as the normal jump, it indeed is balancing the arguments of the normal contact condition, and choosing softeningfactor equals to 1, is natural. There is no guarantee that the tangential increment is of the same order though, which has to be kept in mind. Guessing the right sizes, may be possible in academic cases, but cannot be done though global rescaling.

IvarStefansson:
[...]

PR #1377, opened 2025-03-20
Contact mechanics scaling

mariusnevland:
## Proposed changes

Regarding issue #1151. We remove the contact mechanics softening factor, and introduce two new mixins CharacteristicTractionFromDisplacement and CharacteristicDisplacementFromTraction. The former mixin fetches the characteristic displacement from the numerical constants and computes the characteristic traction by Hooke's law. The latter mixin does the opposite. Note that this is a breaking change, as the default value of the contact mechanics numerical constant has changed.

Note: test_time_dependent_bc() needed some quite significant changes, as the removal of the softening factor caused the Newton solver to not converge. This test is very sensitive to the value of the contact mechanics constant, and we might want to redesign the test. Also, a tiny relaxation of some absolute tolerances in test_contact_mechanics() was needed.

## Types of changes

- [ ] Other:

## Checklist


jwboth (commented):
To me the changes look good. What about the failing tests? Any insight @mariusnevland ?

As of now, the code does not change behavior compared to the previous code. Experience will show which of the two provided mixins for determining characteristic quantities is more often used.

src/porepy/models/contact_mechanics.py
```diff
@@ -493,13 +497,7 @@ def contact_mechanics_numerical_constant(
...
 
         """
-        # Interpretation (EK):
-        # The scaling factor should not be too large, otherwise the contact problem
-        # may be discretized wrongly. I therefore introduce a safety factor here; its
-        # value is somewhat arbitrary.
-        softening_factor = pp.ad.Scalar(self.numerical.contact_mechanics_scaling)
-
-        constant = softening_factor / self.characteristic_displacement(subdomains)
+        constant = pp.ad.Scalar(1) / self.characteristic_displacement(subdomains)
```
jwboth, line 500:
```suggestion
        constant = pp.ad.Scalar(1.0) / self.characteristic_displacement(subdomains)
```

tests/models/test_momentum_balance.py
```diff
@@ -675,12 +679,12 @@ def test_time_dependent_bc():
 
     # Fixed values from a previous run. Both normal value (u_y=0) and ratio of
     # tangential displacements (1/50, see BC class) are correct.
-    u_e = np.array([0.50754939, 0.0, 25.37756547])
```
keileg, line 678: I trust that the change of values is reasonable here, but it seems that documentation on the surrounding lines (both above and below) should be updated. Please consider this.

IvarStefansson (approved):
Thanks for the contribution!

tests/models/test_momentum_balance.py
mariusnevland: The documentation is updated now.

PR #1380, opened 2025-03-26
Apply gravity onto the bulk in poromechanics models.

jwboth:
## Proposed changes

Typically, the body force in poromechanics models applies gravity to the bulk. Currently, it is just the solid density. This PR suggests the extension of gravity forces to the bulk (combining fluid and solid through convex combination) and the respective modification of the body force in poromechanics equations.

As this PR changes the model equations for pp.Poromechanics, and results would differ, this small change is categorized as breaking change.

## Types of changes

- [ ] Other:

## Checklist


keileg:
Is it correct we have no tests that fail because of this change? That is not necessarily bad, but it surely is information of sorts.

src/porepy/models/constitutive_laws.py
```diff
@@ -2747,6 +2751,11 @@ def gravity_force(
             rho = self.fluid.density(cast(pp.SubdomainsOrBoundaries, grids))
         elif material == "solid":
             rho = self.solid_density(grids)
+        elif material == "bulk":
+            phi = self.porosity(grids)
+            rho_f = self.fluid.density(cast(pp.SubdomainsOrBoundaries, grids))
```
IvarStefansson, line 2756: Is the typing of grids meaningful? I suspect the signature type does not make sense.

src/porepy/models/constitutive_laws.py
```diff
@@ -2768,14 +2777,14 @@ class ZeroGravityForce(pp.PorePyModel):
...
-        material: Literal["fluid", "solid"],
+        material: Literal["fluid", "solid", "bulk"],
     ) -> pp.ad.Operator:
         """Gravity force term on either subdomains or interfaces.
 
         Parameters:
             grids: List of subdomain or interface grids where the vector source is
                 defined.
-            material: Name of the material. Could be either "fluid" or "solid".
+            material: Name of the material. Could be either "fluid", "solid", "bulk".
```
IvarStefansson, line 2787:
```suggestion
            material: Name of the material. Could be either "fluid", "solid", or "bulk".
```

src/porepy/models/poromechanics.py
```diff
@@ -71,6 +71,20 @@ class EquationsPoromechanics(
 ):
     """Combines mass and momentum balance and contact mechanics equations."""
 
+    def body_force(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
```
IvarStefansson, line 74: Have you considered THM?

IvarStefansson (commented):
Looks good in the main. Thanks! A few relatively minor comments, though.

src/porepy/models/poromechanics.py
jwboth: You are the expert. I am not to well familiar with THM modeling. But in the end, it is still fluid and solid, i.e., the bulk, on which gravity acts. Temperature implicitly enters through temperature dependent porosity and density laws. Is there anything other to consider?

src/porepy/models/constitutive_laws.py
jwboth: I believe I do not understand. Are you referring to line 2731 or 2756?

src/porepy/models/poromechanics.py
keileg: I would guess @IvarStefansson's point is that a similar change would be needed in the THM model.

IvarStefansson: Yes, that was what I was thinking. Unless THM inherits this change from HM.

src/porepy/models/constitutive_laws.py
IvarStefansson: 2731, really.

src/porepy/models/poromechanics.py
jwboth: Indeed. I did not realize that pp.ThermoPoromechanics does not inherit its equations from pp.Poromechanics. I made the same adaptation. BTW, I just used the Equations class for overwriting what has been previously defined in MomentumBalanceEquations. If you think this is the wrong place, feel free to suggest alternatives.

src/porepy/models/constitutive_laws.py
jwboth: I did not make changes here. I am not sure. Does gravity enter Darcy's law on the mortar grid? Then interface grids are relevant. Or what are you concerned about? Maybe I am missing your point.

IvarStefansson: This will be addressed as a separate issue.

PR #1381, opened 2025-03-26
Add normals to boundary grids.

jwboth:
## Proposed changes

For boundary grids it is natural to request their normals (to e.g. apply stress tensors to a face by multiplication of the tensor with a normal). This PR adds cell normals to boundary grids by copying from their parent domain grid.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/grids/boundary_grid.py
+++ b/src/porepy/grids/boundary_grid.py
@@ -93,12 +93,13 @@ class BoundaryGrid:
 
         Compute the cell centers and volumes of the boundary grid. By default, the
         boundary grid cell information is constructed from the domain boundary faces of
-        theparent grid.
+        the parent grid.
 
         """
         parent_boundary = self._parent.tags["domain_boundary_faces"]
         self.cell_centers = self._parent.face_centers[:, parent_boundary]
         self.cell_volumes = self._parent.face_areas[parent_boundary]
+        self.cell_normals = self._parent.face_normals[:, parent_boundary]
 
     def set_projections(self) -> None:
         """Set projections from the parent grid and set the corresponding attributes.
--- a/src/porepy/grids/boundary_grid.py
+++ b/src/porepy/grids/boundary_grid.py
@@ -88,6 +88,14 @@ class BoundaryGrid:
 
         """
 
+        self.cell_normals: np.ndarray
+        """Normals of cells of the boundary grid. Remember that boundary grid cells are
+        faces of the parent grid. Thus, it stores normals of boundary faces.
+
+        Initialized in :meth:`~compute_geometry`.
+
+        """
+
     def compute_geometry(self) -> None:
         """Compute the geometry of the boundary grid.
 


IvarStefansson:
Before delving into this: Have you considered the option of projecting from the parent subdomain?

jwboth:
You are right, this works (although I needed to twist my head twice...

normals = np.transpose(bg.projection() @ np.transpose(bg.parent.face_normals))

Just for curiosity: Where is the difference to bg.cell_centers? Why not using projections here?

jwboth:
Feel free to pull the trigger. I do not care.

IvarStefansson:
> You are right, this works (although I needed to twist my head twice...
> 
> normals = np.transpose(bg.projection() @ np.transpose(bg.parent.face_normals))
> 
> Just for curiosity: Where is the difference to bg.cell_centers? Why not using projections here?

My thinking was that cell centers make sense on the boundary grid. But a normal of a grid is usually defined on a face, which for the bg would be different from the normals on the parent. However, I don't have strong feelings. Do you have an opinion, @keileg?

keileg:
> > You are right, this works (although I needed to twist my head twice...
> > normals = np.transpose(bg.projection() @ np.transpose(bg.parent.face_normals))
> > Just for curiosity: Where is the difference to bg.cell_centers? Why not using projections here?
> 
> My thinking was that cell centers make sense on the boundary grid. But a normal of a grid is usually defined on a face, which for the bg would be different from the normals on the parent. However, I don't have strong feelings. Do you have an opinion, @keileg?

Not really. Introducing cell normals would be a break with how we do things for other objects of co-dimension 1 (read fractures). However, the boundary grid is special, so if this becomes a commonly invoked attribute, I am not dead set against it.

jwboth:
I see your points. The question how often surface integrals are of interest. If the answer is "almost never" a hack like the above is fine with me. The PR was mostly meant as a readable utility.

To provide some motivation: My concrete use case is the evaluation of a stress onto a boundary (e.g. a Norne like geometry, for which north does not mean anything). Given a stress tensor, one simply needs to evaluate the stress times the normal of the boundary grid during evaluation of boundary conditions. 

Even more concretely, this approach allows for a fairly clean setup routine for initializing models with a given background stress.

keileg:
I think I'm leaning towards agreeing that this is something we want.

IvarStefansson:
> I think I'm leaning towards agreeing that this is something we want.

Sure, let's go for user-friendliness!

src/porepy/grids/boundary_grid.py
```diff
@@ -88,17 +88,26 @@ def __init__(self, g: pp.Grid, name: Optional[str] = None) -> None:
...
         Compute the cell centers and volumes of the boundary grid. By default, the
         boundary grid cell information is constructed from the domain boundary faces of
-        theparent grid.
+        the parent grid.
 
         """
         parent_boundary = self._parent.tags["domain_boundary_faces"]
         self.cell_centers = self._parent.face_centers[:, parent_boundary]
         self.cell_volumes = self._parent.face_areas[parent_boundary]
+        self.cell_normals = self._parent.face_normals[:, parent_boundary]
```
IvarStefansson, line 110: Should the normals be flipped to outwards?

jwboth: Is it ensured that `face_normals` point inwards? I am not too familiar with that part of the code, but after a quick look at their computation, I do not notice any part ensuring any sort of direction. Please correct me if I am wrong.

If it is indeed not the case, I also do not think it is important that the normals point outwards, but that they are consistent with how the Divergence operator is set up based on a fixed orientation of normals. This should ensure consistent application of BC in particular.

keileg: This is a minefield, let's tread carefully. There is no convention on the direction of the normal vector on boundaries, it can point inwards and outwards. The treatment of BCs is currently set up so that the user sets the traction at boundary faces in global coordinates, that is, with no regard to the direction of the normal vector. The necessary adjustments to the convention in the divergence operator under the hood (I can't say precisely where that happens off the top of my head, but it is somewhere).  

I *think* this implies that to allow for setting the boundary condition by projection of a stress tensor, the normal vector should be defined without regard to the direction of the face normal. Whether it should point inward or outwards is a secondary concern, but outwards is consistent with the divergence theorem, assuming positive tensile stress as we do. If the face normal is allowed to change direction, I think the adjustments under the hood (cf previous paragraph) will make for very strange de facto boundary conditions. To check this, one can set up a simplex grid, for which the normal vectors will change directions at random (on Cartesian grids there is a system).

Does this make sense? Do you need more input?

jwboth: I think this is sufficient. Just a short question regarding the first paragraph to be clear. If you say, the BC are defined in global coordinates and I e.g. have a stress only with negative z-z component (gravity), my BC in global coordinates to a face being tilted still is set [0,0,-F] in PP? Or would I still have to take into account the tilt which would correspond to [[0,0,0], [0,0,0], [0,0,-F]] * n? I did not consider how the boundary condition operator works. But I was assuming it also uses the orientation for faces adopted from the grid. From your explanation it sounds like there may be flipping involved here internally as the user was not required to think about this using global coordinates and orientations.

I will have to think about this.

Intuitively, I am still confident that with a consistent use of face normals, it does not matter. This is consistent with the divergence operator containing "-1" values - the "classical" divergence operator, and use of outward normals would require "+1" values. Explicit ensurance that outward normals face outwards would on the other hand be crucial to allow that a manual component wise definition of BC is correct.

I agree, tests will help. Will see when I have time for that.

keileg: Tilt: I am not sure I fully get what you mean. I suggest we discuss in person if possible, to avoid miscommunication. 

Agree, if all face normals on global boundaries pointed outwards, there would have been less need for adjustments elsewhere. I guess I would have implemented this differently if I saw the consequences during intitial development, but by now there is a lot of code, and tests, that need to change if we were to fix this.

jwboth:
Converted to issue #1389

PR #1385, opened 2025-03-27
Tutorial fluid modelling

### LINKED ISSUE #1254: Tutorial on how to create general `fluid` objects for porepy models.
Related to #1240 , the `fluid` attribute in pp models will be a more general object, involving a phase and component context (see PR #1252 )

A tutorial is required explaining the respective `FluidMixin` and the main methods which user can override to create a custom fluid:

- `get_components`
- `get_phase_configuration`

as well as the respective objects `Component`, `Phase`, `PhysicalState` and `EquationOfState`. The latter one should be covered briefly because it requires an own tutorial anyways.

The second part of the tutorial should also cover a deeper dive into setting up complex fluids for CF simulations involving correlations or equilibrium calculations. It involves the mixed-in methods:

- `set_components_in_phases`
- `dependencies_of_phase_properties`
- and various factories for phase properties like `density_of_phase`

vlipovac:
Closes #1254 

## Proposed changes

Adding a tutorial for modeling of multiphase and multicomponent fluids.

Changing the configuration of phases to omit the default `EquationOfState` and hence simplifying the setup for models not relying on EoS computations.

## Types of changes

- [X] Other: Tutorial

## Checklist


mariusnevland:
Some high-level comments after the first readthrough:

- As a complete newbie to multiphase, multicomponent flow, I do not know e.g. the difference between overall fractions and partial fractions, as well as some other technical terms. Maybe it is a good idea to state some "prerequisites" for the tutorial?
 
- It is unclear to me how to use the heuristic approach in the case of multiple phases and/or components. Did I understand correctly that the default constitutive laws will stop working if I add more phases/components? If so, how can they be redefined?

- Regarding the EoS classes, I do not understand how the return_dummy_properties() function should look like in the case of non-trivial phase properties. But otherwise, this code block was very helpful in understanding how to use the EoS framework.

- It would be very nice to have an actual runable model in the very end, but would this significantly increase the complexity of the tutorial? Is it possible to construct some "minimal working example" that still introduces most/all of the important functionality, e.g. some relatively simple two-phase, two-component setup?

keileg:
Thanks for the feedback @mariusnevland. I'll have a look and keep your comments in mind.

tutorials/fluid_modeling.ipynb
```diff
@@ -0,0 +1,935 @@
+{
```
keileg, line 1: <p>I think a bit more context will be useful here. Rough ideas (not complete):</p><ul><li>What do you mean by a component?</li><li>Why do we need to care about the fluid properties (they are part of the transport equations etc.)</li></ul><p>This should not take much space, and giving references to more information is certainly fair, but give a bit more context could be very useful for readers.</p><p> </p><p>Would it be an idea to introduce one physical system, say, two-phase two-component water-CO2 that can be followed throughout the tutorial? Sorry if the answer is obviously not.</p><p> </p><p>Question: What is meant by 'respective functionalities'?</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pmgbergen/porepy/pull/1385/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

keileg, line 1: <p>If it is important that the user understands the notion of reference components/phases, more information is needed. If not, can it go?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pmgbergen/porepy/pull/1385/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

keileg, line 1: <p>Perhaps move some of this towards the top, to gather such information when natural. Also 'overall fraction' must be defined somewhere.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pmgbergen/porepy/pull/1385/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

keileg, line 1: <p>How is this class meant to be used?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pmgbergen/porepy/pull/1385/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

keileg, line 1: <p>The first paragraph here is a partial answer to the question 'how can we simulate such systems', or 'what happens under the hood in such simulations' - but not necessary to answer how a non-trivial fluid can be modeled in PorePy. Both types of information is relevant, and both can be fit here, but the plan for when to communicate which type is not clear to me.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pmgbergen/porepy/pull/1385/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

keileg:
Most important: This will become extremely useful for everyone as we introduce more advanced fluid modeling. 

I only managed to read half the tutorial before I had to run, will try to cover the second half no later than Monday. I tried to keep the comments on a high level, it seems to me we should focus on getting the context and general disposition of the tutorial right before getting into details.

tutorials/fluid_modeling.ipynb
keileg, line 1: <p>Here it should be clearer that the 'EoS approach' refers to the particular implementation in PorePy that wraps the cubic EoS and similar, and not the modeling choice of using an EoS.</p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pmgbergen/porepy/pull/1385/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

keileg, line 1: <p>This cell gives a broad answer to the question 'how can I can invoke the EoS framework'? It does not so much answer the related question 'what do I need to do to invoke the EoS framework?' The difference to me is that there are a few contracts I have to fulfill and concepts I have to deal with (some keywords are 'dependencies_of_phase_properties', 'PhysicalState', 'PhaseProperties', and a few others), that are not really introduced, let alone explained. I am not asking for an explanation of the full architecture of the representation of fluids, even if that turns out to be needed it should not be done her. On the other hand, is there a way to explain on a very high lever what are the main components involved in the above code, and how do they interact?</p><p> </p><p>This is also a place where a concrete example would be good to explain how to go from a bunch of zero arrays (cf. the last return statement) to something representing a fluid. However, I realize now that such an example risks being contrived - let's discuss.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pmgbergen/porepy/pull/1385/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

keileg:
I left a few more comments. @vlipovac I suggest we discuss how to approach this PR at some point.

tutorials/fluid_modeling.ipynb
vlipovac: <p>I have amended this section keeping with regards to your comments.</p><p> </p><p>About the accompanying model, that would require to introduce quite a bit which is not yet part of the core code, essentially I would need a thermodynamic model (either an EoS, or couple of new mixins as some heuristic, simple fluid properties).</p>

vlipovac: <p>I tried to explain that in the following sentences.</p>

vlipovac: <p>I have introduced links to the respective quantities implemented in the code. The docstring contain sufficient information to make it clear what it is.</p><p>Is that fine?</p><p> </p><p>Since this tutorial is a bit extensive, I am trying to not burden it further with formulas and definitions, but focus on the application of the code and link to the respective parts. The documentation of the compositional framework is very extensive, I want to make use of it.</p>

vlipovac: <p>As a pure mixin. Added some extra explanation, here and for <code>MyFluid</code>.</p>

vlipovac: <p>Let's discuss in person.</p>

vlipovac: <p>I am not sure what you mean, because it is made for the choice of using an EoS.</p><p> </p><p>Should I emphasize that in theory it can be used for any computations?</p>

keileg: <p>I agree with the overall idea, we are not Wikipedia. Since the links just bring me to the top of the file, perhaps it is better to give this as an instruction, such as 'an explanation can be found in the documentation, see here'. </p>

vlipovac: The link should lead directly to the respective line. Have you tried that in vs code or online?
On my machine, locally in VS code, it leads to the correct line.

tutorials/fluid_modeling.ipynb
```diff
@@ -0,0 +1,1267 @@
...
```
keileg, line 1: <p>Please use punctuation in comments.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pmgbergen/porepy/pull/1385/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

keileg, line 1: <pre class="ql-syntax" spellcheck="false">Line #106.                                f"pseudo_k_value_equation_{component.name}_{phase.name}"
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pmgbergen/porepy/pull/1385/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

vlipovac: <p>Fixed.</p>

vlipovac: <p>Checked all comments and fixed punctuation.</p>

keileg (approved):

Issue #1350, opened 2025-02-25
Energy balance equation is mistakenly treated as linear


Yuriyzabegaev:
Applies to the `MassAndEnergyBalance` and the `THM` models. There is no override to `_is_nonlinear_problem` method for the energy equation, so the fallback method states that the model is linear. It is nonlinear due to advection. The fix is to add the corresponding override.

IvarStefansson:
Is there a danger this applies to others as well? E.g. mass balance, which also uses upwinding.

Yuriyzabegaev:
As I understand, mass balance uses only the mixed-dimensional unwinding, so it is linear without fractures.

IvarStefansson:
Do you know if this has been fixed, @Yuriyzabegaev?

Yuriyzabegaev:
> Do you know if this has been fixed, [@Yuriyzabegaev](https://github.com/Yuriyzabegaev)?

No, not yet

PR #1406, opened 2025-04-11
Fix is nonlinear

mikeljordan:
## Proposed changes

This PR addresses Issue #1350 by introducing an explicit mechanism to declare and track the nonlinearity of individual equation, and also system of equations within the EquationSystem class, which is then used by the SolutionStrategy class to appropriately determine the nonlinearity of the resulting problem. 

### Key changes:

1. **New API flag:** Introduced a boolean `is_nonlinear` argument to EquationSystem.set_equation() to allow users to explicitly declare if an equation is nonlinear. 
2. **Better bookkeeping:** Added a `equation_is_nonlinear` property, which maintains a dictionary of equation names mapped to their declared nonlinearity in the EquationSystem class.
3. **Improved model logic:** Updated the `_is_nonlinear_problem()` method in the SolutionStrategy class to use the new bookkeeping to determine the nonlinearity of the problem resulting from the system of ACTIVE equations. It is recommended that models that are explicitly linear, inheriting from base classes with nonlinear equations can override this method to avoid unnecessary code duplication.
4. **Tests added:** Included a parameterized test to verify correct handling of linear, nonlinear, and mixed equation types — both at the individual equation level and at the system level.

## Types of changes

- [x ] Bugfix (non-breaking change which fixes an issue).
- [x ] Testing (contribution related to testing of existing or new functionality).
- [x ] Documentation (contribution related to adding, improving, or fixing documentation).
- [x ] Maintenance (e.g., improve logic and performance, remove obsolete code).
- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/applications/test_utils/models.py
+++ b/src/porepy/applications/test_utils/models.py
@@ -126,7 +126,7 @@ def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
     # them in a new class.
 
     # Identify the geometry class
-    geometry: Any = None
+    geometry: type
     if dim == 2:
         geometry = RectangularDomainThreeFractures
     elif dim == 3:
@@ -135,7 +135,7 @@ def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
         raise ValueError(f"Unknown dimension {dim}")
 
     # Identify the physics class
-    model_class: Any = None
+    model_class: type
     if model_type == "mass_balance":
         model_class = pp.SinglePhaseFlow
     elif model_type == "momentum_balance":
@@ -155,7 +155,7 @@ def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
         pass
 
     # Create an instance of the combined class
-    model = cast(pp.PorePyModel, Model(params))
+    model: pp.PorePyModel = Model(params)
 
     # Prepare the simulation
     # (create grids, variables, equations, discretize, etc.)
--- a/src/porepy/applications/test_utils/models.py
+++ b/src/porepy/applications/test_utils/models.py
@@ -119,7 +119,11 @@ def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
     """Model for tests."""
     # Suppress output for tests
     fracture_indices = [i for i in range(num_fracs)]
-    params = {"times_to_export": [], "fracture_indices": fracture_indices}
+    params = {
+        "times_to_export": [],
+        "fracture_indices": fracture_indices,
+        "cartesian": True,
+    }
 
     # To define the model we comine a geometry with a physics class.
     # First identify the two component classes from the user input, and then combine
--- a/src/porepy/applications/test_utils/models.py
+++ b/src/porepy/applications/test_utils/models.py
@@ -150,7 +150,7 @@ def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
         model_class = pp.Poromechanics
     elif model_type == "thermoporomechanics":
         model_class = pp.Thermoporomechanics
-    elif model_type == 'contact_mechanics':
+    elif model_type == "contact_mechanics":
         model_class = pp.ContactMechanics
     else:
         # To add a new model, insert an elif clause here, and a new class above.
--- a/src/porepy/applications/test_utils/models.py
+++ b/src/porepy/applications/test_utils/models.py
@@ -117,9 +117,9 @@ class ContactMechanicsTester(ContactMechanics):
 
 def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
     """Model for tests."""
-    # Suppress output for tests.
     fracture_indices = [i for i in range(num_fracs)]
     params = {
+        # Suppress output for tests.
         "times_to_export": [],
         "fracture_indices": fracture_indices,
         "cartesian": True,
--- a/src/porepy/applications/test_utils/models.py
+++ b/src/porepy/applications/test_utils/models.py
@@ -117,7 +117,7 @@ class ContactMechanicsTester(ContactMechanics):
 
 def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
     """Model for tests."""
-    # Suppress output for tests
+    # Suppress output for tests.
     fracture_indices = [i for i in range(num_fracs)]
     params = {
         "times_to_export": [],
@@ -129,7 +129,7 @@ def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
     # First identify the two component classes from the user input, and then combine
     # them in a new class.
 
-    # Identify the geometry class
+    # Identify the geometry class.
     geometry: type
     if dim == 2:
         geometry = RectangularDomainThreeFractures
@@ -138,7 +138,7 @@ def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
     else:
         raise ValueError(f"Unknown dimension {dim}")
 
-    # Identify the physics class
+    # Identify the physics class.
     model_class: type
     if model_type == "mass_balance":
         model_class = pp.SinglePhaseFlow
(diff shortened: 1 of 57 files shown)


src/porepy/numerics/ad/equation_system.py
```diff
@@ -168,6 +168,10 @@ def __init__(self, mdg: pp.MixedDimensionalGrid) -> None:
 
         """
 
+        self._equation_is_nonlinear: dict[str, bool] = dict()
```
keileg, line 171: I suggest to rename to 'is_nonlinear_equation' to signify that this is a boolean property

src/porepy/models/momentum_balance.py
```diff
@@ -63,9 +63,11 @@ def set_equations(self) -> None:
         matrix_eq = self.momentum_balance_equation(matrix_subdomains)
         intf_eq = self.interface_force_balance_equation(interfaces)
         self.equation_system.set_equation(
-            matrix_eq, matrix_subdomains, {"cells": self.nd}
+            matrix_eq, matrix_subdomains, {"cells": self.nd}, is_nonlinear=False
+        )
+        self.equation_system.set_equation(
+            intf_eq, interfaces, {"cells": self.nd}, is_nonlinear=True
```
keileg, line 69: I don't see why this is nonlinear?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -316,6 +320,14 @@ def variable_domains(self) -> list[pp.GridLike]:
             domains.add(var.domain)
         return list(domains)
 
+    @property
```
keileg, line 323: Have you considered the alternative signature is_nonlinear_equation(self, eq_name:str) -> bool? This seems to be a cleaner approach.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1094,6 +1107,8 @@ def set_equation(
                 operators are able to provide information on their image space.
                 The dictionary must contain the number of equations per grid entity
                 (cells, faces, nodes) for the operator.
+            is_nonlinear: Contains a flag for every equation name (key) to indicate that
+                the equation is nonlinear.
```
keileg, line 1111:
```suggestion
            is_nonlinear: Contains a flag to indicate that the equation is nonlinear.
```

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1129,6 +1144,8 @@ def set_equation(
             self._equation_image_size_info.update({name: equations_per_grid_entity})
             # Store the equation itself.
             self._equations.update({name: equation})
+            # Store the linear/nonlinear flag.
```
keileg, line 1147: Can't we do this at the top of the method to avoid the double registration (here and towards the end)?

tests/models/test_solution_strategy.py
```diff
@@ -400,6 +400,9 @@ def __init__(self, params: dict):
         super().__init__(params)
         self.nonlinear_solver_statistics = pp.SolverStatistics()
 
+    def _is_nonlinear_problem(self) -> bool:
```
keileg, line 403: Why was this needed?

tests/models/test_solution_strategy.py
```diff
@@ -439,3 +442,29 @@ def test_check_convergence(
...
+        # There was a bug here once #1350.
+        dict(model_name="thermoporomechanics", num_fracs=0, is_nonlinear=True),
+        dict(model_name="thermoporomechanics", num_fracs=1, is_nonlinear=True),
+    ],
+)
+def test_linear_or_nonlinear_model(params: dict):
+    """Tests that the base models are properly tagged as linear or nonlinear."""
+    model_name: str = params["model_name"]
+    num_fracs: int = params["num_fracs"]
+    is_nonlinear: bool = params["is_nonlinear"]
```
keileg, line 467: Please add meshing parameters to ensure this model has no fractures and is discretized with a Cartesian grid (this will substantially limit the cost of this test)

tests/functional/setups/linear_tracer.py
```diff
@@ -702,7 +703,9 @@ def _set_enthalpy_closure(self) -> None:
         subdomains = self.mdg.subdomains()
         op = self.enthalpy(subdomains) - self.fluid.specific_enthalpy(subdomains)
         op.set_name("enthalpy_closure")
-        self.equation_system.set_equation(op, subdomains, {"cells": 1})
+        self.equation_system.set_equation(
+            op, subdomains, {"cells": 1}, is_nonlinear=False
```
keileg, line 707: Check: Have you conferred with @vlipovac to verify that the equations defined in the methods below are indeed linear?

keileg (commented):
Mainly looks good, I have a few comments and questions.

tests/functional/setups/linear_tracer.py
mikeljordan: Yes, the equation system is linear, as all its coefficients are constant

tests/models/test_solution_strategy.py
mikeljordan: Good question! The base implementation of _is_nonlinear_problem() accesses self.equation_system, which doesn't exist in the test class. Overriding it avoids an AttributeError and lets us manually control the nonlinear flag for isolated testing.

src/porepy/numerics/ad/equation_system.py
mikeljordan: Now that I thought of this, the name sounds like a flag for a single equation, which should be a collective name.

src/porepy/numerics/ad/equation_system.py
mikeljordan: This is a good point, we didn't think of this. `is_nonlinear_equation_system(self, eq_name:str) -> bool`.

src/porepy/models/momentum_balance.py
Yuriyzabegaev: My bad, it's linear.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: I agree, will change

tests/models/test_solution_strategy.py
Yuriyzabegaev: I marked the model as Cartesian explicitly, it did not make any existing test fail.

keileg (commented):
I added a (final, I hope) comment about Cartesian grids. Almost ready.

src/porepy/examples/mandel_biot.py
```diff
@@ -1430,6 +1430,9 @@ class MandelModel(  # type: ignore[misc]
         - units (pp.Units): Object containing scaling of base magnitudes. No scaling
           applied by default.
 
+    The model is set to be linear, make sure to change this behavior if needed when
```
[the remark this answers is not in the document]
Yuriyzabegaev: The mass balance equation is tagged to be nonlinear due to some stuff like density. Mandel model overrides these quantities and makes them de facto linear, but it doesn't override `set_equations` just to change the nonlinearity flag. Therefore using such a hack. We can discuss how to make it better

src/porepy/applications/test_utils/models.py
```diff
@@ -119,14 +119,18 @@ def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
     """Model for tests."""
     # Suppress output for tests
     fracture_indices = [i for i in range(num_fracs)]
-    params = {"times_to_export": [], "fracture_indices": fracture_indices}
+    params = {
+        "times_to_export": [],
+        "fracture_indices": fracture_indices,
+        "cartesian": True,
```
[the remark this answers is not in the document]
keileg: I see the point now.

IvarStefansson (commented):
Some minor comments

src/porepy/applications/test_utils/models.py
```diff
@@ -117,25 +117,29 @@ def interface_displacement_parameter_values(
 
 def model(model_type: str, dim: int, num_fracs: int = 1) -> pp.PorePyModel:
     """Model for tests."""
-    # Suppress output for tests
+    # Suppress output for tests.
```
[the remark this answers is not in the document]
Yuriyzabegaev: Moved it there

src/porepy/models/contact_mechanics.py
```diff
@@ -564,8 +564,14 @@ def contact_mechanics_open_state_characteristic(
...
-        """The contact mechanics problem is nonlinear."""
-        return True
+        """The contact mechanics problem is nonlinear, but it may happen that this
+        method is called from a model with no fractures. In this case, we fall back to
+        the default.
+
+        """
+        if self.mdg.dim_min() < self.nd:
+            return True
+        return super()._is_nonlinear_problem()
```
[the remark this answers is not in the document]
IvarStefansson: Good! I was confused by the next file in the PR being protocol, where the method is empty

IvarStefansson (commented):
I'm happy. Please approve if you are as well, @keileg.

keileg (approved):

Yuriyzabegaev:
Note for future self. We decided to step back from the idea proposed here, since it opens up multiple harmful edge cases. Mainly, when the equation is nonlinear, but the derived model (such as Biot) makes it linear. This leads to even more inconsistency and confusion than it was before. If the proper solution is once introduced, it should be based on the actual traversal of the AD tree to actually figure out whether the underlying AD object is linear or not.

PR #1407, opened 2025-04-13
Experiment: Attempt to run tests in parallel in github actions. 

jwboth:
## Proposed changes

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/.github/workflows/run-pytest-all.yml
+++ b/.github/workflows/run-pytest-all.yml
@@ -67,4 +67,4 @@ jobs:
 
     - name: all tests
       if: ${{always()}}
-      run: pytest --run-skipped -m "not tutorials"
+      run: pytest --run-skipped -m "not tutorials" -n auto
--- a/.github/workflows/run-pytest.yml
+++ b/.github/workflows/run-pytest.yml
@@ -71,4 +71,4 @@ jobs:
 
     - name: Run tests
       if: ${{always()}}
-      run: pytest tests -m "not tutorials"
+      run: pytest tests -m "not tutorials" -n auto
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,19 +40,9 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-development = [
-    "isort",
-    "ruff",
-    "mypy",
-    "mypy-extensions",
-    "traitlets"
-]
-testing = [
-    "pytest >= 4.6",
-    "pytest-cov",
-    "pytest-runner"
-]
+testing = ["pytest >= 4.6", "pytest-cov", "pytest-runner", "pytest-xdist"]
 
+development = ["isort", "ruff", "mypy", "mypy-extensions", "traitlets"]
 # Information for the build system.
 [tool.setuptools.packages.find]
 where = ["src"]


keileg:
@Yuriyzabegaev and I looked into this some months ago. The problem is all tests that interact with gmsh, where the common file name `gmsh_frac_file` leads to all kind of trouble. A solution may be to assign random names to files during testing. I'll do some experiments now.

jwboth:
> @Yuriyzabegaev and I looked into this some months ago. The problem is all tests that interact with gmsh, where the common file name `gmsh_frac_file` leads to all kind of trouble. A solution may be to assign random names to files during testing. I'll do some experiments now.

Cool. I am looking forward to the outcome of the currently running tests!

It all makes sense. This is consistent with my experience in running many simulations side-by-side. Seems like classical race conditions - in my simple cases there was a simple fix of including some seconds waiting times (I also used dedicated file names for gmsh files, but if I remember correctly this was not sufficient - I did not dig too deep as it worked in the end). Here, it is not so trivial I find.

jwboth:
> I pushed a change that will assign random names to gmsh files during testing, but not regular use. The good news is that the time for testing is almost cut in half locally (which in a sense is disappointing from four workers, but I'll take the improvement still). The not so good news are:
> 
> 1. There are occasional random errors from tests that seem to trace back to conflicting gmsh file names. The file names are assigned a random number between 1 and 10^6 (or was it ^8), so the chance of collision should be miniscule, but it could be that the workers all start with the same random seed.
> 2. There are also issues with some vtk exporting. I'll leave these to you, @jwboth.
> 
> It remains to be seen whether we get a meaningful speedup also on GH actions, but local speedups will also help.
> 
> In total, I would say this is worth pursuing a bit more, but not necessarily to be assigned very high priority right now. Thoughts?

I would suggest one more attempt. Look here: https://pytest-xdist.readthedocs.io/en/stable/distribution.html

It is possible to control the distribution. Without having done any deeper analysis, my intuition would say that the conflicts are if certain tests are run in parallel. The likelihood that these tests come from the same file may be quite high. It is possible to request that tests from the same file are run by the same worker. This can be done again by a simple flag in the run command.

.github/workflows/check_tutorials.yml
```diff
@@ -77,4 +77,4 @@ jobs:
 
     - name: Check tutorials
       if: ${{always()}}
-      run: pytest tests -m tutorials
+      run: pytest tests -m tutorials -n auto
```
jwboth, line 80:
```suggestion
      run: pytest tests -m tutorials -n auto -dist loadfile
```

.github/workflows/run-pytest-all.yml
```diff
@@ -67,4 +67,4 @@ jobs:
 
     - name: all tests
       if: ${{always()}}
-      run: pytest --run-skipped -m "not tutorials"
\ No newline at end of file
+      run: pytest --run-skipped -m "not tutorials" -n auto
```
jwboth, line 70:
```suggestion
      run: pytest --run-skipped -m "not tutorials" -n auto -dist loadfile
```

.github/workflows/run-pytest.yml
```diff
@@ -71,4 +71,4 @@ jobs:
 
     - name: Run tests
       if: ${{always()}}
-      run: pytest tests -m "not tutorials"
+      run: pytest tests -m "not tutorials" -n auto
```
jwboth, line 74:
```suggestion
      run: pytest tests -m "not tutorials" -n auto -dist loadfile
```

jwboth (commented):
Distribute same test files among same workers - should hopefully fix vtk tests.

.github/workflows/run-pytest-all.yml
jwboth, line 70:
```suggestion
      run: pytest --run-skipped -m "not tutorials" -n auto -dist loadfile
```

.github/workflows/check_tutorials.yml
```diff
@@ -77,4 +77,4 @@ jobs:
 
     - name: Check tutorials
       if: ${{always()}}
-      run: pytest tests -m tutorials
+      run: pytest tests -m tutorials -n auto -dist loadfile
```
jwboth, line 80:
```suggestion
      run: pytest tests -m tutorials -n auto --dist loadfile
```

.github/workflows/run-pytest-all.yml
```diff
@@ -67,4 +67,4 @@ jobs:
 
     - name: all tests
       if: ${{always()}}
-      run: pytest --run-skipped -m "not tutorials"
\ No newline at end of file
+      run: pytest --run-skipped -m "not tutorials" -n auto -dist loadfile
```
jwboth, line 70:
```suggestion
      run: pytest --run-skipped -m "not tutorials" -n auto --dist loadfile
```

.github/workflows/run-pytest.yml
```diff
@@ -71,4 +71,4 @@ jobs:
 
     - name: Run tests
       if: ${{always()}}
-      run: pytest tests -m "not tutorials"
+      run: pytest tests -m "not tutorials" -n auto -dist loadfile
```
jwboth, line 74:
```suggestion
      run: pytest tests -m "not tutorials" -n auto --dist loadfile
```

.github/workflows/check_tutorials.yml
```diff
@@ -77,4 +77,4 @@ jobs:
 
     - name: Check tutorials
       if: ${{always()}}
-      run: pytest tests -m tutorials
+      run: pytest tests -m tutorials -n auto --dist loadfile
```
jwboth, line 80:
```suggestion
      run: pytest tests -m tutorials -n auto --dist worksteal
```

jwboth:
> Distribute same test files among same workers - should hopefully fix vtk tests.

The last test run succeeded. Indeed, checking the tutorials does not seem to be more effective than before - it also just has available 2 workers. I am doing a final test with aggressive distribution of tests. The standard test suite has no groundbreaking speed up but still 2-3 minutes.

.github/workflows/run-pytest-all.yml
```diff
@@ -67,4 +67,4 @@ jobs:
 
     - name: all tests
       if: ${{always()}}
-      run: pytest --run-skipped -m "not tutorials"
\ No newline at end of file
+      run: pytest --run-skipped -m "not tutorials" -n auto --dist loadfile
```
jwboth, line 70:
```suggestion
      run: pytest --run-skipped -m "not tutorials" -n auto --dist worksteal
```

.github/workflows/run-pytest.yml
```diff
@@ -71,4 +71,4 @@ jobs:
 
     - name: Run tests
       if: ${{always()}}
-      run: pytest tests -m "not tutorials"
+      run: pytest tests -m "not tutorials" -n auto --dist loadfile
```
jwboth, line 74:
```suggestion
      run: pytest tests -m "not tutorials" -n auto --dist worksteal
```

Yuriyzabegaev:
> I pushed a change that will assign random names to gmsh files during testing, but not regular use. The good news is that the time for testing is almost cut in half locally (which in a sense is disappointing from four workers, but I'll take the improvement still). The not so good news are:
> 
> 1. There are occasional random errors from tests that seem to trace back to conflicting gmsh file names. The file names are assigned a random number between 1 and 10^6 (or was it ^8), so the chance of collision should be miniscule, but it could be that the workers all start with the same random seed.
> 2. There are also issues with some vtk exporting. I'll leave these to you, @jwboth.
> 
> It remains to be seen whether we get a meaningful speedup also on GH actions, but local speedups will also help.
> 
> In total, I would say this is worth pursuing a bit more, but not necessarily to be assigned very high priority right now. Thoughts?

I was also thinking towards doing this when we were investigating this. As a sidenote, could it avoid the random collisions if naming files after the test name, not at random? The test name can be accessed from the same env variable `PYTEST_CURRENT_TEST`

keileg:
> I was also thinking towards doing this when we were investigating this. As a sidenote, could it avoid the random collisions if naming files after the test name, not at random? The test name can be accessed from the same env variable `PYTEST_CURRENT_TEST`

If so, the file name would also need to include information on test parametrization. Better then to generate a temporary file using techniques like [this](https://docs.python.org/3/library/tempfile.html).

IvarStefansson:
When reading through the discussion, I wondered whether the times.json files might also interfere with each other. Just a thought.

jwboth:
I did not investigate the latest suggestions as the current setup seems to have worked (for the few test runs above). The latest failing tests were merely due to mypy which are fixed now. By simple comparison to a current PR: Indeed the gains in time are not too large. Tutorials have no true speed up (<1 min possibly of 6 minutes). The remaining tests are ca. 2-3 mins faster of 10-11 minutes for reference.

Issue #1417, opened 2025-04-25
Update the method assemble_schur_complement_system in EquationSystem


keileg:
In class `EquationSystem`, there is a method `assemble_schur_complement_system` that is intended for problems where some blocks can be inverted by local elimination (think local thermodynamic equilibrium). While the implementation should be more or less correct, it has two major shortcomings:
1. The inversion is based on a direct inverse of the full matrix, that is, without exploiting any block structure in the part to be eliminated. While this is formally okay, it is not the intended usage. The standard implementation should be changed to use the block inverter from `matrix_operations` instead.
2. The method relies on projection operations to bring the system to a block diagonal form before inversion, and then back again. Here the new `ArraySlicer` class can be expected to be much more efficient.

As reflected in the above description a third possible change is:
3. The method name is possibly confusing, in that the intention is to provide functionality for elimination of block diagonal parts only. If an explicit construction of the Schur complement is actually desired (really?) this should be considered part of the linear solver strategy, not assembly.

@vlipovac has a prototype that does some of these operations and that may form a starting point for this project.

mikeljordan:
I understand that the main advantage of using the Schur complement lies in the relative ease of inverting the secondary block (A_ss), particularly due to its sparse structure. Currently, it has a non-diagonal block structure hence, maybe be computationally expensive to invert. Because of this, we seek to project it to a block diagonal using suitable projections. From my inspection, I found that we can get this cheaply by doing a rework on the EquationSystem on how the (secondary/primary)equations/variables are ordered! without a need for this projections, Unless we have other hidden treasures i am not aware of, hence I suggest we have a more mathematical discussion about this today. Also, the prototype provided by @Vlipovac is brilliant—great work! That said, we might not fully achieve our goal if the secondary block remains connected, as this could limit efficiency of the current approach.

Issue #1327, opened 2025-02-10
Wrong order of routines before nonlinear iteration for Darcy flow


vlipovac:
In the `SinglePhaseFlow` setup there is an inconsistency in the order of updated and stored values of discretizations and permeabilities.

Looking at the method `before_nonlinear_iteration` we have the following order code executed:

1. `SinglePhaseSolutionStrategy.before_nonlinear_iteration`
    a. Updates stored flux values **<---**
    b. Super call to base SolutionStrategy
 2. `SolutionStrategy.before_nonlinear_iteration`
    a. Calls `set_discretization_parameters`
3. `SinglePhaseSolutionStrategy.set_discretization_parameters`
    a. Updates stored permeabilities in a `SecondOrderTensor` **<---**
4. Back to `SolutionStrategy.before_nonlinear_iterationan`
    a. Calls rediscretize (which rediscretizes Upwinding from `SinglePhaseFlowSolutionStrategy`)

The two lines with arrows show that the flux values for Upwinding are evaluated and stored before permabilities are updated.

Now, for setups without non-linearities in the permeability tensor, this does not matter.
But for setups which include advanced mechanical effects, or setups which use a pressure equation with mobility terms in the tensor (frational flow in CF framework), this does matter.

The order of execution needs to be revisited, or possibly some methods need to be split to make a correct order possible given PorePy's mixin approach.

**TODO once solved**
Adapt `SolutionStrategyNonlinearMPFA` which was merged in #1236 .

vlipovac:
## Proposed solution:

This issue is critical for compositional flow and (thermo) poromechanics.

It can be solved by breaking the `before_nonlinear_iteration` routine into 4 different sub-routines, with each sub-routine having the respective method defined in the base sol. strategy.

The 4 sub-routines are:

1. Setting/ updating discretization parameters: already exists as `SolutionStrategy.set_discretization_parameters`
2. re-discretizing fluxes: A method `SolutionStrategy.rediscretize_fluxes` (or maybe more general `SolutionStrategy.rediscretize_elliptic_discretizations`?)
3. Copy results from first discretization to where it is required for the next step
4. re-discretize upwinding: already exists as `SolutionStrategy.rediscretize`, but should in this case be renamed to sth. like `SolutionStrategy.rediscretize_hyperbolic_discretizations`?

With these sub-routines, the base method `SolutionStrategy.before_nonlinear_iteration` would have the following *unchanging* form

```python

class SolutionStrategy:

[...]

PR #1438, opened 2025-05-30
Updates related to mypy v 1.16

keileg:
## Proposed changes

Notable changes:
1. I had to make a minor adjustment to the GH action for static tests to allow for a mypy update. The new version of the file seems reasonable. For explanation, see documentation in the action setup.
2. Mypy found a bug related to update of Biot discretizations. In my understanding, this also indicates that test coverage of this part is not good enough (as expected). I added a warning statement related to the update option in mpfa, mpsa, and biot. Though we have not prioritized to fix this yet, it may become a more pressing issue in the future.

Besides this, the changes are rather small.

There is one remaining mypy issue for which I may need your assistance, @vlipovac: The `_MixtureDofHandler` defines a read-only property `overall_fraction_variables`, which is also declared (for typing purposes) in the class `SolutionStrategySchurComplement` (line 1719 in `compositional_flow.py`). This results in an error when the classes are mixed down in line 1995 in the same file (note that mypy has been updated, hence the pain). I guess the reason you use a read-only property is that the variable names returned by this method must be correct, or else everything breaks?

I really hope we can fix this with minor adjustments of the code - do you see a way do achieve this? We may need @IvarStefansson's input on this matter at some point as well.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/.github/workflows/run-static-checks.yml
+++ b/.github/workflows/run-static-checks.yml
@@ -96,7 +96,9 @@ jobs:
     # for the install (hence the installed packages are sure to be available also after
     # the build), while --no-deps will only install the specified packages (in this
     # case, the dependencies of PorePy). The third line similarly installs the
-    # development dependencies.
+    # development dependencies. Note that we do not use the --no-deps option for the
+    # development dependencies, since we need fully functioning versions of mypy, ruff,
+    # and isort to run the static checks.
     #
     # Thanks, copilot!
     - name: Install requirements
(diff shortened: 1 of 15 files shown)


vlipovac:
@keileg  You can add `tyoe:ignore` to any issue caused by `SolutionStrategySchurComplement` and `SolutionStrategyNonlinearMPFA`, as both classes will be **eliminated** with the additions of #1417 and #1327 respectively.

IvarStefansson (approved):

PR #1458, opened 2025-06-23
BUG: Grid.cell_nodes returns a CSC matrix

keileg:
This was not necessarily the case after the update to scipy v1.16..

## Proposed changes

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "networkx",
     "numba",
     "numpy < 2.2",
-    "scipy",
+    "scipy < 1.16",
     "seaborn",
     "shapely",
     "six",


IvarStefansson (approved):
Thanks for fixing!

keileg:
The tutorial `fluid_modeling.ipynb` still fails. Specifically, the Newton iterations fail to converge for what should be a linear advection problem. 

It seems the following conditions are needed for failure to occur:
1. `scipy` must have version 1.16.0 (with 1.15.3, everything works, though I am a bit surprised by the number of non-linear iterations needed for the simulation.
2. `mkl` must not be present (`pip install mkl` resolves the issue).

From my (incomplete) reading and understanding of the changes in `scipy.sparse` between versions 1.15 and 1.16, there is no clear reason why this all happens. The fact that installing `mkl` resolves the matter points to an error which is potentially very deep.

Suggested way forward:
1. @vlipovac Can you please confirm that the problem is as simple as it looks like, and comment on the reason why the number of non-linear iterations is set as high as 50? I should know this, but no.
2. Depending on the answer to 1., the path of least resistance is either to install `mkl` or to increase the maximum number of non-linear iterations. 

Most important, @vlipovac, please let me know if you do not have time to look into this now. If so, we will postpone the update of `scipy` for the time being. There will be no harm in this.

vlipovac:
Local runs in my environment showed that it requires more than 50 iterations with `linear_solver='scipy_sparse'`, while with `pypardiso` it was able to converge within the 50 iterations.
I am not sure if there are some changes regarding that (solution strategy says pypardiso is default`, the new example model parameters indicate scipy sparse), but it's worth trying for which linear solver this is fine, without resorting to installing mkl?

keileg:
Thanks for the update, @vlipovac, and sorry for my poor memory of the nature of the model (it won't be the last time). I observe the same behavior as you, introducing either `mkl` or `pardiso` solves the issue, but there is something new in scipy version 1.16 that has lead to this increase in the number of Newton iterations needed.

As you say, there are several possible fixes here, introducing either of `mkl`, `pardiso` or simply increasing the maximum number of permissible iterations should do. I would like to understand the issue a bit better though, in case it tells us something about the system we are solving. Hence, we will freeze the scipy version to 1.15 for now, and return to the issue later, either by applying one of the fixes, or by digging some more.

PR #1466, opened 2025-08-04
Optimize get_variable_values method of EquationSystem class

mikeljordan:
… data fetching

## Proposed changes

This PR contains a targeted optimization to the EquationSystem.get_variable_values() method, which can significantly reduce simulation time in cases with many variables and shared subdomains. Motivated by compositional simulations.

The updated method avoids redundant operations by:

- Using a set for faster lookup of variable IDs.
- Caching domain data to prevent repeated self._get_data() calls.
- Returning an empty vector via np.empty(0) instead of np.array([]).

These changes reduce unnecessary memory operations and domain-level lookups, which previously contributed substantial overhead during per-iteration routines such as thermodynamic property updates and auxiliary computations / evaluate() method.

Benchmarking: On a water–salt multiphase system with ~3222 mdg cells, the internal time spent in get_variable_values() was reduced by over ~42%.

## Types of changes

- [ x] Maintenance (e.g., improve logic and performance, remove obsolete code).
- [ ] Other:

## Checklist

- [x ] The documentation is up-to-date.
- [ x] Static typing is included in the update.
- [x ] This PR does not duplicate existing functionality.
- [ x] The update is covered by the test suite (including tests added in the PR).

### DIFF
--- a/src/porepy/numerics/ad/equation_system.py
+++ b/src/porepy/numerics/ad/equation_system.py
@@ -631,7 +631,7 @@ class EquationSystem:
             :meth:`num_dofs`.
 
         """
-        # Normalize the variable input
+        # Normalize the variable input.
         variables = self._parse_variable_type(variables)
         var_ids = {var.id for var in variables}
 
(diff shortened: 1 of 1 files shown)


mikeljordan:
> Looks good.
> 
> For curiosity, do you have estimates on how much time you gain by this? And whether it is the caching or using a set for fast lookup that makes the most difference?

From my test, the set-based lookup gave the biggest impact than the caching.

keileg (approved):
Nice work. Thanks.

PR #1475, opened 2025-08-22
Speedup of grid cell diameter calculation

keileg:
## Proposed changes
The current calculation of cell diameters of a grid can take surprisingly long time. Specifically, it may be a bottleneck when running grid convergence studies. This PR reduces the cost significantly (the calculation is no longer a bottleneck, according to my simulations), by the following two changes:

1. Instead of essentially looping over all cells, then over all nodes pairs, the new implementation constructs a global set of node pairs much more efficiently. This has the downside of not providing a cell-wise diameter. It may be possible to map back to cell-wise quantities, but this is non-trivial for general grids. Since the PR is motivated by convergence studies, where a global quantity is needed, I decided to request that the user provide a function (e.g. `np.min`, `np.max`) that is used to calculate a such global quantity with no cell-wise considerations.
2. The grid diameter calculation is stored in an lru_cache to avoid recomputing numbers (this may happen with some setups for convergence studies).

## Types of changes

What types of changes does this PR introduce to PorePy?
_Put an `x` in the boxes that apply._

- [ ] Minor change (e.g., dependency bumps, broken links).
- [ ] Bugfix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] Testing (contribution related to testing of existing or new functionality).
- [ ] Documentation (contribution related to adding, improving, or fixing documentation).
- [x] Maintenance (e.g., improve logic and performance, remove obsolete code).
- [ ] Other:

## Checklist

_Put an `x` in the boxes that apply or explain briefly why the box is not relevant._

- [x] The documentation is up-to-date.
- [x] Static typing is included in the update.
- [x] This PR does not duplicate existing functionality.
- [x] The update is covered by the test suite (including tests added in the PR).
- [ ] If new skipped tests have been introduced in this PR, `pytest` was run with the `--run-skipped` flag.

PR #1484, opened 2025-09-03
BUG: Logical error in grid diameter computation

keileg:
## Proposed changes
This fixes an error introduced in #1475: When calculating cell diameters it is necessary, first to take the maximum of all diameters (node-node distances) within a cell, and then apply a user-defined function (say, `np.min`) to get the quantity (say, minimum diameter) in the entire grid. In #1475, the step of calculating cell-wise diameter was skipped; hence the code in some cases gave wrong results for non-simplicial grids.

I am starting to think that the new implementation may be sufficiently general to replace the old implementation, which is activated when setting `cell_wise=True`. However, I am not fully confident that this is really the case, so I would prefer not to do that generalization now, but return to the matter if the cell diameter calculation is taken into much more frequent use than what is  now the case.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/grids/grid.py
+++ b/src/porepy/grids/grid.py
@@ -993,19 +993,35 @@ class Grid:
                 [ind[ptr[i] + template[sz]] for i, sz in enumerate(num_nodes)]
             )
 
-            # In-place sort of the columns.
-            all_node_pairs.sort(axis=0)
-
-            # We only need the unique pairs of nodes.
-            unique_pairs = np.unique(all_node_pairs, axis=1)
             # Compute the diameter of each unique pair of nodes.
             dist = np.linalg.norm(
-                self.nodes[:, unique_pairs[0]] - self.nodes[:, unique_pairs[1]], axis=0
+                self.nodes[:, all_node_pairs[0]] - self.nodes[:, all_node_pairs[1]],
+                axis=0,
             )
+            # The cell diameter is the maximum of the distances associated to a cell. To
+            # find this, we first make an array of cell indices for all elements in
+            # dist. Spelled out: For each element in num_nodes (the number of nodes in
+            # each cell), create an array of cell indices (ci). The size of this array
+            # should be the number of node-node combinations within that cell (given by
+            # the number of columns in the corresponding template). Concatenate the
+            # arrays created for all cells by using np.hstack.
+            cell_index = np.hstack(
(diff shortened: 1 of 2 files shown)


src/porepy/grids/grid.py
```diff
@@ -993,19 +993,35 @@ def diam(n):
                 [ind[ptr[i] + template[sz]] for i, sz in enumerate(num_nodes)]
             )
 
-            # In-place sort of the columns.
-            all_node_pairs.sort(axis=0)
-
-            # We only need the unique pairs of nodes.
-            unique_pairs = np.unique(all_node_pairs, axis=1)
             # Compute the diameter of each unique pair of nodes.
```
IvarStefansson, line 996: Are the pairs unique or should the comment be updated?

src/porepy/grids/grid.py
```diff
@@ -993,19 +993,35 @@ def diam(n):
...
+            # the number of columns in the corresponding template). Concatenate the
+            # arrays created for all cells by using np.hstack.
+            cell_index = np.hstack(
+                [np.full(template[sz].shape[1], ci) for ci, sz in enumerate(num_nodes)]
             )
+            # Take the maximum over all distances associated with a cell. We will store
+            # the maximum in cell_diam.
+            cell_diam = np.zeros(self.num_cells)
+            # By numpy magic and copilot cleverness, the following two lines efficiently
+            # takes the cell-wise maximum. It is equivalent to (but much faster than)
```
IvarStefansson, line 1015:
```suggestion
            # take the cell-wise maximum. It is equivalent to (but much faster than)
```

tests/grids/test_grid.py
```diff
@@ -18,24 +20,61 @@
...
+    # Move the last node of the grid (it is easier to do this here than to modify the
+    # grid in the test parametrization). Since the grid is expected to be Cartesian,
+    # this will impact the diameter of cell 1, but not cell 0. Do not perturb the
+    # z-coordinate, so that a 2d grid stays 2d.
+    g.nodes[:2, -1] = 2
+
     # The test is run for a 2d grid and a 3d grid.
-    cell_diameters = grid.cell_diameters(cell_wise=cell_wise, func=np.max)
+    cell_diameters = g.cell_diameters(cell_wise=cell_wise, func=func)
     if cell_wise:
```
IvarStefansson, line 69: Should we check that the warning is raised?

IvarStefansson (approved):

Issue #1476, opened 2025-08-24
Immature pp.SolverStatitics


jwboth:
The pp.SolverStatistics shows currently several shortcomings - this issue aims at collecting some with potential solutions:
- In the wish of a lighter pp.SolverStatistics object, no time information is available. To nevertheless log the success over time steps, the `save` method appends the statistics to the previously printed statistics. However, when rerunning the simulation, the file is not removed. (A similar comment holds for vtk files). One solution could be to not read from file, but remembering the solution strategy in a dictionary with a persistent key, e.g., connected to attempted time steps and thus number of calls to `NewtonSolver.solve()`.
- No information on "success" or "failure" is stored to file. It would be simple to include `ConvergenceStatus` from #1448 into the logging. By extension of `ConvergenceStatus` to include other statuses, e.g, `MAX_NUMBER_ITERATIONS_REACHED`, `CYCLED`, `NAN` etc. one could provide even more information for later analysis. This would require an additional attribute `convergence_status`.
- When time steps need to be repeated, also failed attempts of running the time step are stored. It is then difficult to assign the different logs to the different attempts. By using persistently the `time_manager.time_index` in addition with the above `ConvergenceStatus` and even if wished storing `time_manager.time` an `time_manager.dt` one could quite simply store all information and still be able to post-analyze e.g. only successful time steps,  identify number of attempted or wasted time step computations, total number of nonlinear iterations etc.  This would require additional attributes `time_index`, `time`, `dt`.
- Related to the last point: If a simulation ends prematuraly because of a minimal `dt` reached during time step cutting, the final time is not reached. To ease the identification of such a case just from reading the solver statistics file, it may be an idea to add the attribute `final_time_reached`.
- Currently SolverStatistics.save() is invoked during exporting simulation results. It may be an option to decouple these. While storing all efforts connected to the numerical solution, it may be not always advisable to export all attempted (and potentially failed) field values, cf. #1421

pschultzendorff:
The second point is addressed by #1471 , currently without the functionalities introduced by #1448 though

PR #1486, opened 2025-09-04
Add comment about inconsistent recomputation count

IngridKJ:
## Proposed changes

The _recomp_num-attribute of the TimeManager counts either for the time-step or for the entire simulation. Added a comment about this. 

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/time_step_control.py
+++ b/src/porepy/numerics/time_step_control.py
@@ -157,9 +157,8 @@ class TimeManager:
         recomp_factor: Failed-to-converge solution recomputation factor.
             Factor by which the time step will be multiplied in case the solution must
             be recomputed. We require `recomp_factor` to be strictly less than one.
-        recomp_max: Failed-to-converge maximum recomputation attempts. The maximum
-        allowable
-            number of consecutive recomputation attempts.
+        recomp_max: Maximum recomputation attempts. The maximum allowable
+            number of recomputation attempts.
         print_info. Whether to print on-the-fly time-stepping information or not.
         rtol: Relative tolerance parameter for float point equality.
         atol: Absolute tolerance parameter for float point equality.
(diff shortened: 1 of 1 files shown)


jhabriel:
Oh, I see. From a quick glance at the code, it looks like your intended usage is not covered. The thing is, say you solve one time step with `recomp_max=5` and the flag `recompute_solution = True`. For some reason, your time step is too large, so the first attempt fails (this internally sets `_recomp_num=1`) so the algorithm proposes a smaller dt (say `dt/2`). In this attempt, you achieve convergence, you move on with your simulation. Now, if the `recompute_solution` flag is still `True`, then nothing really changes the value of `_recomp_num` (it remains `1`). The intended usage is to externally set `recompute_solution=False`, since now you have a `working_dt`. This will try to calculate a new `dt` (and even if it fails, it will there reset `_recomp_num` to 0). 


Now, if you always use `recompute_solution=True`, then the algorithm will never reset `._recomp_count` to `0`.


That being said, I'm not against modifications of the algorithm. But I don't see an easy way to "tell" the algorithm to reset the recomputation counter. One possibility might be to pass an extra flag "has_converged: bool", to `compute_time_step()` so the algorithm knows that the previous time step reduction was successful.

jwboth:
While I remain silent regarding logic in the `pp.TimeManager`, I just want to hint at the possibility to log the convergence status etc in a central place. See #1476 . I believe it is simpler in terms of synchronization and bookkeeping.

jhabriel:
Hi @jwboth. Agree 100%. Now this opens at least two possibilities:

1. Passing the instance of `pp.SolverStatistics` to `compute_time_step`, or
2. Passing only the convergence flag.

Option 1 is a strong candidate, since this unlocks other criteria for time-control (residual-based time control, for example).

jwboth:
Agree to option 1 pretty much. More to come hopefully soon ;)

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

Issue #1469, opened 2025-08-11
Explore the use of Gmsh for processing of fracture network geometries


keileg:
# Background
As part of meshing of domains with fractures, we do a lot of geometry processing, such as identification of fracture intersection etc. This is handled by the two classes `FractureNetwork2d`, `FractureNetwork3d` together with helper functions, mainly in the geometry subpackage. In total, this involves thousands of lines of code. The code is robust, in the sense that we very rarely run into trouble, but whenever we do, maintenance is a nightmare.

Since the code for geometry processing was written, `Gmsh` has added functionality to do this type of processing through a backport to `OpenCascade`. This functionality is likely faster and more robust than our current functionality. We should therefore explore the use of Gmsh, to see if we can do a replacement.

# Tasks

## Short term
The initial exploration can be based on the following steps (all assumed to be based on the `Gmsh` python API, as documented in various [tutorials], see also current usage in the PorePy file `gmsh_interface.py`):

- [ ] Find out how to represent the three relevant classes of fractures (1d lines, 2d polygons, 2d ellipses/circles) in `Gmsh.`
- [ ] Find out how to represent a domain (closed box) in `Gmsh`.
- [ ] Use `Gmsh` to compute the intersection between two objects. Store the return values, understand and document the returned information.

Output format: My best suggestion for now is to document the findings in the form of a Jupyter notebook. Other suggestions are much welcome.

## Intermediate term
Assuming the results of the short term investigation are positive, the following steps should be considered (note that these relate the `Gmsh` processing to the full meshing process, hence they may require a temporary hack of the `mesh` method in the `FractureNetwork` classes - it should be specified what this means before implementation proceeds).

- [ ] Generate meshes of domains with no intersecting fractures, using `Gmsh` geometry processing. 
- [ ] Figure out which of the intersection information from `Gmsh` must be used, and how, to generate the hierarchy of subdomain grids used in a `MixedDimensionalGrid` (ex: the intersection of two 2d fractures will be a 1d line; how ).
- [ ] Generate meshes of domains with intersecting fractures, using `Gmsh` geometry processing. 
- [ ] Add the possibility to disregard or truncate fractures by imposing the domain boundary.

Output format: A branch which allows for meshing using the new functions, either by replacing the `mesh` methods in `FractureNetwork` classes, or by letting the old and new method live side by side. Or a different format, if this is decided upon later.

## Longer term
Preliminary for now. To be specified as the project progresses.
[...]

PR #1392, opened 2025-04-10
1325 translate physical names from gmsh meshes to porepy tags

pschultzendorff:
## Proposed changes
Writing this on my phone, apologies for anything I miss and any name errors.

This PR introduces functionality to translate physical names from ‘’.msh’’ files to tags in the sense of ‘’pp.Grid.tags’’. This allows, e.g., to easily recreate the geometry of [SPE11](https://github.com/Simulation-Benchmarks/11thSPE-CSP/tree/main/geometries) by loading the grid and assigning permeability and porosity based on tags.

Porepy offers functionality to import from ‘.msh’ files via ‘’pp.fracs.fracture_importer. dfm_from_gmsh’’. At some point in the process, ‘’pp.fracs.msh_2_grid.create_nd_grids’’ is called to create nd subdomain grids.

We introduce a helper function ‘’tag_grids’’ which takes the information available to ‘’create_nd_grids’’ and the created subdomain grid and updates ‘‘grid.tags’’ accordingly. The function is called by every ‘’create_nd_grids’’ at the very end.

So far done:
- Write ‘’tag_grids’’
- Include ‘’tag_grids’’ in ‘’create_nd_grids’’
- Write a test for ‘’tag_grids’’
- Write tests for ‘’create_nd_grids’’. These test only the tagging

ToDo:
- Pass the correct input to ‘’create_nd_grids’’. The way we try to do this is by meshing with ‘’fracture_network.mesh’’ and reading the created file. However this does not include physical names nor does it assign any physical names to mesh entities (vertices, lines, triangles, tetrahedra). The way we planned to do this is to loop through the file and add the correct lines manually. This is quite challenging. Possibly this whole approach has to be changed.

Hope this gives a good overview, let us know if you need more info

## Types of changes

- [ ] Minor change (e.g., dependency bumps, broken links).
- [ ] Bugfix (non-breaking change which fixes an issue).
- [x] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] Testing (contribution related to testing of existing or new functionality).
- [ ] Documentation (contribution related to adding, improving, or fixing documentation).
- [ ] Maintenance (e.g., improve logic and performance, remove obsolete code).
- [ ] Other:

## Checklist

- [x] The documentation is up-to-date.
- [x] Static typing is included in the update.
- [x] This PR does not duplicate existing functionality.
- [x] The update is covered by the test suite (including tests added in the PR).
[...]

PR #1507, opened 2025-10-15
Fix consistency errors in 2d simplex grids

### LINKED ISSUE #1501: Negative cell volumes for 2d grids
When meshing 2d geometries, I occasionally get errors from a check in the geometry computation in the grid class that the [cell volumes are all non-negative](https://github.com/pmgbergen/porepy/blob/b3d019f491cd5cb1dbe73c52477708b9d0b8eba0/src/porepy/grids/grid.py#L522). The issue can be traced through an [if statement on the grid orientation](https://github.com/pmgbergen/porepy/blob/b3d019f491cd5cb1dbe73c52477708b9d0b8eba0/src/porepy/grids/grid.py#L500) back to the [calculation of the grid orientation](https://github.com/pmgbergen/porepy/blob/b3d019f491cd5cb1dbe73c52477708b9d0b8eba0/src/porepy/grids/grid.py#L452). Essentially, the grid is considered oriented if the face-node and cell-face mappings are organized such that, for a cell with nodes `a,b,c` , the three faces have the nodes ordered in a pattern `(a,b), (b, c), (c, a)`, that is, each node occurs first for one face and second for a different face.

To me, it seems the problem is the with the definition of orientation, which currently reads 

```python
is_oriented = (fn_orient @ self.cell_faces).nnz == 0
```
The problem is that `self.cell_faces` contains positive and negative items, hence `is_oriented` can be `True` even for unoriented cells. It seems to me that this can be fixed by 
```python
cell_faces_abs = np.abs(self.cell_faces)
is_oriented = (fn_orient @ cell_faces_abs).nnz == 0
```
@alessiofumagalli (or perhaps @wmboon): I believe it was you guys who introduced this check in the first place. Have I understood the logic correctly, and is this a reasonably fix?

For clarity: The issue turned up during work with #1469, which seems to break some patterns in how nodes from gmsh grids are ordered (this is a long story). It is difficult to imagine that there is a deeper underlying issue here, so all previous simulations should be fine.

keileg:
## Proposed changes
This PR resolves two issues related to the construction of simplex grids in 2d:
1. The construction of the face-node relations erroneously used a trick with sorting which, for some special grids, lead to negative cell volumes, as reported in #1501. 
2. Debugging of point 1. revealed that, for 2d simplex grids, it was actually possible to define the cell-node relation in a way that created a cell-face array with rows of two positive or negative numbers (hence the cell-face matrix can no longer be used as a divergence, which would have given all sort of problems).

AFAIK, the first of these issues rarely showed up in practice, see #1501 for some examples, while the second issue has never materialized for grids used in any sort of practical simulation. The reason is likely that the structure of the grid information generated by gmsh, or by the `StructuredTriangleGrid` class, did not trigger the errors (though we started running into these when changing the way we interact with gmsh, in project #1469). 

Changes made in this PR:
1. Both of the errors have been fixed, see changes to init of TriangleGrid.
2. To ensure there is no danger of a similar bug for the TetrehedralGrid, the code for generating face-node and cell-face relation was revisited, with documentation much improved.
3. A test was added to init of the Grid class, so that, if we are ever given a grid with two positive or negative elements in a row of the cell-face relation, an error will be raised.
4. A test of msh_2_grid, introduced in #1392, that previously struggled with negative volumes in geometry computations (due to issue no 1 in the first list of this PR) no longer has any problems. The logic has therefore been simplified to use standard geometry computations.

I did make a test for the 2d case that would have failed with the previous code, but decided against it, since creating grids with such errors will anyhow raise an error in the init-method of the grid class, or in geometry computation.

Big thanks to @wmboon and @pschultzendorff for helping out on this one.

Resolves #1501.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/grids/simplex.py
+++ b/src/porepy/grids/simplex.py
@@ -127,7 +127,8 @@ class TriangleGrid(Grid):
 
         # Cell-face relation. This can be constructed from the mapping back to the
         # cell-wise face-node relation, recalling that the cell-nodes were stacked so
-        # that the faces of all first cells came first, etc.
+        # that the first face of each cell came first, then the second face of each cell
+        # etc.
         num_faces_per_cell = 3
         # Reshape and ravel in Fortran order to get the faces of the first cells first.
         cell_face_indices = cell_face_mapping.reshape(
@@ -258,6 +259,9 @@ class TetrahedralGrid(Grid):
 
             Name of grid type. If None, ``'TetrahedralGrid'`` will be assigned.
 
+    Raises:
+        ValueError: If not enough points are provided to construct the grid.
+
     """
 
     def __init__(
@@ -277,10 +281,10 @@ class TetrahedralGrid(Grid):
             name = "TetrahedralGrid"
 
         num_nodes = p.shape[1]
+        if num_nodes <= 3:
+            raise ValueError("Not enough points to construct tetrahedral grid.")
 
         nodes = p
-        assert num_nodes > 3  # Check of transposes of point array.
-
         num_cells = tet.shape[1]
 
         # As a preparatory step to construct the face-node and cell-face relations,
@@ -299,9 +303,9 @@ class TetrahedralGrid(Grid):
         # Define face-nodes so that the first column contains fn of cell 0, etc. Due to
         # the permutation of the nodes in the previous step, and the order in which the
         # nodes are listed in the definition of face_nodes, the nodes of each face are
-        # ordered so that the normal vector formed by the cross product of the vector from
-        # node 0 to node 1 and the vector from node 0 to node 2 points points in the
-        # same direction as the vector from node 0 to node 3 (to see this, draw an
+        # ordered so that the normal vector formed by the cross product of the vector
+        # from node 0 to node 1 and the vector from node 0 to node 2 points points in
+        # the same direction as the vector from node 0 to node 3 (to see this, draw an
         # example and verify). This implies that the two cells sharing a face will have
         # that face represented with opposite ordering of the nodes.
         face_nodes = np.vstack(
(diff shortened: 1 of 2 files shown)


src/porepy/grids/simplex.py
```diff
@@ -252,35 +279,53 @@ def __init__(
...
 
         num_cells = tet.shape[1]
+
+        # As a preparatory step to construct the face-node and cell-face relations,
+        # permute the nodes for all tetrahedra. After this step, the nodes in each cell
+        # will be ordered so that, if taking the cross product of the vector from node 0
+        # to node 1 and the vector from node 0 to node 2, this will point in the
+        # opposite direction to the vector from node 0 to node 3 (to see this, read the
+        # code in _permute_nodes carefully). In EK's understanding, the point is to get
+        # a systematic ordering of the nodes, hereunder a system that let us deal with
```
IvarStefansson:
```suggestion
        # a systematic ordering of the nodes, including a system that lets us deal with
```

src/porepy/grids/simplex.py
```diff
@@ -294,14 +339,25 @@ def __init__(
             (data, face_nodes, indptr), shape=(num_nodes, num_faces)
         )
 
-        # Cell face relation
+        # Cell-face relation. Index pointers are straightforward, since we know that
+        # each cell has exactly 4 faces.
```
IvarStefansson:
```suggestion
        # each cell has exactly four faces.
```

IvarStefansson (commented):
Looks good to me, but not extreme confidence in my own job as a reviewer.

IvarStefansson (approved):

Issue #1410, opened 2025-04-13
parameter_dictionaries.py


IvarStefansson:
Downstream of #1409. I expect the entire file to be removed if #1409 is effectuated.

keileg:
Agreed, from a brief glance of the code, it seems this can be deleted.

Issue #1409, opened 2025-04-13
Parameters class


IvarStefansson:
This class could be replaced with a built-in, e.g. dict or defaultdict. That would probably remove several other functions in data.py. We might want to keep some (modified) utility functionality. Let's discuss.

keileg:
Agree, let's map this out and then implement.

PR #1521, opened 2025-10-28
Remove parameters

Yuriyzabegaev:
## Proposed changes

This addresses #1410, #1409. `Parameters` class is replaced to `dict`. 

Signature of `pp.initialize_data` is changed - it does not accept its first argument "grid". This will affect user scripts and will become a breaking change. So far, we emit a warning suggesting to replace the signature. I also covered this function with tests.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/fracture_damage.py
+++ b/src/porepy/models/fracture_damage.py
@@ -159,8 +159,11 @@ class DamageHistoryEquation(pp.PorePyModel):
         """
         super().before_nonlinear_loop()
         fractures = self.mdg.subdomains(dim=self.nd - 1)
-        self.equation_system._equations[self.damage_history_equation_name] = (
-            self.damage_history_equation(fractures)
+        self.equation_system.update_equation(
+            equation_name=self.damage_history_equation_name,
+            new_equation=self.damage_history_equation(fractures),
+            grids=fractures,
+            equations_per_grid_entity={"cells": 1},
         )
 
     def damage_history_equation(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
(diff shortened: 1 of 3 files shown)


src/porepy/params/data.py
```diff
@@ -38,340 +61,161 @@
...
-class Parameters(Dict):
-    """Class to store all physical parameters used by solvers.
+        return decorator
 
-    The intention is to provide a unified way of passing around parameters, and
-    also circumvent the issue of using a solver for multiple physical
-    processes (e.g. different types of boundary conditions in multi-physics
-    applications). The keyword assigned to parameters and discretization operators
-    ensures that the right data is used: An operator will always use the parameters
-    stored with under the keyword it has been assigned.
```
IvarStefansson:
```suggestion
The only signature the users should use, and the only one that will be preserved in the future.
```

tests/numerics/fv/test_biot.py
```diff
@@ -27,13 +27,12 @@ def discretization_matrices(flow_keyword, mechanics_keyword):
...
     bnd = pp.BoundaryConditionVectorial(g)
-    data = pp.initialize_default_data(
-        g,
+    data = pp.initialize_data(
         {},
         mechanics_keyword,
         specified_parameters=partial_update_parameters(stiffness, bnd, g.num_cells),
     )
-    data = pp.initialize_default_data(g, data, flow_keyword)
+    data = pp.initialize_data(data, flow_keyword)
```
IvarStefansson: Is there any need to call initialize twice?

IvarStefansson (approved):
Excellent work! Feel free to merge once you have addressed my minor comments.

tests/numerics/fv/test_biot.py
Yuriyzabegaev: Yes, this is just bad naming. What it does is initializes a sub-dictionary corresponding to the passed keyword. So we first initialize the "flow" subdictionary and then the "mechanics" subdictionary in data.

Yuriyzabegaev: I also don't think so, especially since we agree it's a user-facing api

Issue #1544, opened 2025-11-25
Revise setup of Docker containers


keileg:
The docker containers for PorePy have the following structure: The base image is not meant to be used, but provide common parts for the images based on the develop and main branches (the images called, respectively, dev and stable). In addition, there is an image with extensions (PETSc etc.), soon to be added (see #1543). 

Once #1543 is ready, the following changes are needed:
1. The standard user in the docker containers should no longer be root, but a sudo-enabled user, say, porepy
2. Installation should be by uv, not pip.
3. Installation should be in a virtual environment (uv will more or less force us to take care of this).
4. As indicated in #1537, we would like to change the dockerfiles to enable parametrization of the branch names on which they are built. It could be natural to include this in the same PR. The check on whether we succeed in this would be to run the GH action [build-test-stable-docker_image.yml](https://github.com/pmgbergen/porepy/blob/develop/.github/workflows/build-test-stable-docker_image.yml).

See also the checklist [here](https://github.com/pmgbergen/porepy/pull/1553#discussion_r2590832360)

PR #1553, opened 2025-12-04
Codespace

### LINKED ISSUE #1546: Make PorePy available through GH codespaces
We should make a PorePy installation available as a GitHub code space to enable easy exploration of the software. If technically feasible, the solution should include the possibility to run tutorials in the code space so that these can be explored interactively.

vlipovac:
Closes #1546 

## Proposed changes

Introduces devcontainer setup to build an image used for creating codespaces.
The image is based on the `porepy\stable` image.

Adds basic python setup for vscode in the codespace lowering entry barrier.
Installs everything needed to run jupyter notebooks, narrows down available python interpreters to highest version (3.13.9), which is the default python in the image (this was done to avoid users accidently chosing an intepreter where the tutorial execution will fail. Consider removing 2 redundant python installations from image)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,31 @@
+{
+  "name": "porepy-stable",
+  "build": {
+    "dockerfile": "../dockerfiles/stable/Dockerfile"
+  },
+  "remoteUser": "root",
+  "workspaceFolder": "/workdir/porepy",
+  "runArgs": ["-it"],
+  "waitFor": "updateContentCommand",
+  "updateContentCommand": "python -m pip install jupyter ipykernel notebook",
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "ms-python.python",
+        "ms-python.vscode-pylance",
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "porepy-stable",
+  "build": {
+    "dockerfile": "../dockerfiles/stable/Dockerfile"
+  },
+  "remoteUser": "root",
```
vlipovac, line 6: The following settings need to be adapted if they change in the porepy image:

- `remoteUser` (I think you discussed adding some default user besides root)
- `workspaceFolder` (in case you change the directory of the porepy code)
- `defaultInterpreterPath` and `kernels.ExcludePythonEnvironments` whenever that changes. But please consider keeping the number of Python installation in the image at 1, I had a hard time to make this user-friendly.

keileg: Referred to in #1544.

keileg (approved):

PR #1555, opened 2025-12-09
Model-exporting-refactor

IvarStefansson:
## Proposed changes

Creates three new model exporting classes:
- Equation residuals
- Iterations (promoted from tutorial
- Fracture quantities (some new fields, some refactored from base DataSaving). This makes the PR BREAKING, since apertures and specific volumes are no longer exported by default.

Main questions:
Do the current models seem like valuable additions striking reasonable balance between convenience and cost (size of exported files)?
How do we want to test these classes? 

## Types of changes

- [ ] Other:

## Checklist


src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -315,3 +289,172 @@ def load_data_from_pvd(
...
+            # - regular values are positive for fractures in contact,
+            # - negative slip tendency is non-physical, except for
+            # - zero normal traction.
+            slip_tendency = (
+                -np.linalg.norm(traction_loc[:-1], axis=0) / traction_loc[-1]
+            )
+
+            data.append((sd, "slip_tendency", slip_tendency))
+
+            data.append((sd, "contact_traction_in_Pa", traction_loc.ravel("F") * char))
```
jwboth: Why not just removing the non-scaled contact traction? Is there a reason to keep it?

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -315,3 +289,172 @@ def load_data_from_pvd(
...
+        """Returns data for exporting.
+
+        Returns:
+            A list of tuples (subdomain, variable name, variable values).
+        """
+        # Start with data from super class. This includes standard variables.
+        data = super().data_to_export()  # type: ignore[misc]
+        # Add the three fracture-specific vector quantities.
+        sds = self.mdg.subdomains(dim=self.nd - 1)
+        cell_offsets_nd = np.cumsum([0] + [sd.num_cells * self.nd for sd in sds])
```
jwboth, line 428: Is there a reason why the evaluations are not pulled inside the for loop? Then one would not need to keep track of the indices and I believe the code would be slightly simpler to read (and 5 lines shorter).

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -315,3 +289,172 @@ def load_data_from_pvd(
...
+        This method is typically called by after_nonlinear_iteration.
+        """
+        # To make sure the nonlinear iteration index does not interfere with the time
+        # part, we multiply the latter by the next power of ten above the maximum number
+        # of nonlinear iterations. Default value set to 10 in accordance with the
+        # default value used in NewtonSolver.
+        n = self.params.get("max_iterations", 10)
+        p = round(np.log10(n))
+        r = 10**p
+        if r <= n:
```
jwboth: Would it hurt to remove these lines? Just a quick control question - I do not care too much.

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -315,3 +289,172 @@ def load_data_from_pvd(
...
+        data = super().data_to_export()  # type: ignore[misc]
+
+        # Add residuals. Loop over equations in the equation system.
+        for name, operator in self.equation_system.equations.items():
+            residuals = cast(NDArray, self.equation_system.evaluate(operator))
+
+            # Get image_info as dict[GridEntity, int], where
+            # GridEntity = Literal["cells", "faces", "nodes"]
+            image_info = self.equation_system._equation_image_size_info[name]
+            dof_start, dof_end = 0, 0
```
jwboth, line 388: Similar to the DeformationExporting, why not pulling the evaluation of operator over [g] inside the for-loop, and getting rid of the explicit dof management?

jwboth (approved):
Looks overall very good to me. I only have comments with potential improvements.

src/porepy/viz/data_saving_model_mixin.py
IvarStefansson: Yes, this is faster. Each evaluation may have a non-negligible overhead.

src/porepy/viz/data_saving_model_mixin.py
IvarStefansson: Same answer:)

src/porepy/viz/data_saving_model_mixin.py
jwboth: Thanks for the explanation.

src/porepy/viz/data_saving_model_mixin.py
IvarStefansson: The case r==n will occur if n is a power of 10, right? This is not good, since we need an additional digit to avoid confusion between time steps. The case r<n should not occur regardless of the value of n. The solution `r = math.ceil(math.log10(n + 1))` may be slightly more elegant. What do you think?

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -315,3 +289,190 @@ def load_data_from_pvd(
...
+    def save_data_iteration(self):
+        """Export current solution to vtu files.
+
+        This method is typically called by after_nonlinear_iteration.
+        """
+        # To make sure the nonlinear iteration index does not interfere with the time
+        # part, we multiply the latter by the next power of ten above the maximum number
+        # of nonlinear iterations. Default value set to 10 in accordance with the
+        # default value used in NewtonSolver.
+        n = self.params.get("max_iterations", 10)
```
keileg, line 329: This type of doubly hard-coded default values is bound to fail at some point, though the consequences here should not be severe. I'm fine with leaving it for now, but it would be better to have a single location for all such default values. TODO, I guess?

src/porepy/viz/data_saving_model_mixin.py
keileg: Using ceil sounds fine.

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -315,3 +289,190 @@ def load_data_from_pvd(
...
+        """
+        data = super().data_to_export()  # type: ignore[misc]
+
+        # Add residuals. Loop over equations in the equation system.
+        for name, operator in self.equation_system.equations.items():
+            residuals = cast(NDArray, self.equation_system.evaluate(operator))
+
+            # Get image_info as dict[GridEntity, int], where
+            # GridEntity = Literal["cells", "faces", "nodes"]
+            image_info = self.equation_system._equation_image_size_info[name]
```
keileg, line 387: FYI: https://github.com/pmgbergen/porepy/issues/1561 

Feel free to close this one.

src/porepy/viz/data_saving_model_mixin.py
IvarStefansson: I agree

src/porepy/viz/data_saving_model_mixin.py
IvarStefansson: I'll paste a note in that issue.

src/porepy/viz/data_saving_model_mixin.py
IvarStefansson: I don't have strong feelings. I would have to actively remove, though, introducing an exception to the rule that all primary variables are exported. Opinions, @keileg?

src/porepy/viz/data_saving_model_mixin.py
```diff
@@ -315,3 +289,190 @@ def load_data_from_pvd(
...
+            # set it to 1.
+            zero_inds = np.isclose(traction_loc[-1], 0)
+            traction_loc[-1, zero_inds] = 1
+            # Minus to follow convention that positive slip tendency indicates slip and
+            # compressive normal traction is negative. Summing up:
+            # - regular values are positive for fractures in contact,
+            # - negative slip tendency is non-physical, except for
+            # - zero normal traction.
+            slip_tendency = (
+                -np.linalg.norm(traction_loc[:-1], axis=0) / traction_loc[-1]
```
IvarStefansson: Do you want me to scale by friction coefficient, making the slip tendency immediately interpretable and reside in [0,1]?

keileg: Yes, that sounds like a good idea.

src/porepy/viz/data_saving_model_mixin.py
keileg: Not really, though I lean towards uniformity in what is exported (i.e. all primary variables and then some more).

src/porepy/viz/data_saving_model_mixin.py
keileg: https://github.com/pmgbergen/porepy/issues/1562

keileg (approved):

PR #1571, opened 2026-01-09
Make tests of HM and THM models work for more general grids. 

keileg:
## Proposed changes
The test `test_poromechanics.py::test_positive_p_frac_positive_opening` is driven by a source term placed in the fracture. This term takes a fixed value for all fracture cells, without accounting for the cell volume. The computed solution is compared to a reference value. When changing the number of fracture cells, which I happened to do as part of the meshing project, the total source term increases, and hence the solutions cannot be expected to match the reference. The same carries over to the very similar `test_thermoporomechanics.py::test_positive_p_frac_positive_opening`.

This PR introduces two changes:
1. A scaling of the source term that allows for a varying number of the cells: Assuming that the solution scales linearly with the total source term (that is with the number of fracture cells), the reference value is adjusted accordingly.
2. An adjustment of the accepted deviations to make the tests pass with the new grids in the soon-to-be-placed meshing PR. In my understanding, these adjustments are needed because the solutions do not scale linearly with the source term. As expected, the pressure needs only a small adjustment, while the temperature sees a bigger change.

Comments:
1. The solution presented here is hacky, but I believe the alternative is to change the reference values whenever the grid undergoes even minor changes (in the upcoming meshing PR, with the current settings the cell count is doubled, but I can see this changing in the months ahead as we may make minor adjustments to the new meshing algorithm). I am far from 100% happy with the solution and open to discuss other approaches.
2. The actual change of the deviations logically belong to the meshing PR, it was just mentally and technically (shared git commits) easier to include them here. I would prefer to keep them here, but I am open to removing them again.

In addition, the tests of HM and THM with TPSA now runs only with Cartesian grids (where tpsa is consistent). This change was previously introduced for the test of momentum balance models, but for some reason, we (presumably be) did not follow through for the derived models.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/tests/models/test_poromechanics.py
+++ b/tests/models/test_poromechanics.py
@@ -216,6 +216,11 @@ def create_model_with_fracture(
         "u_north": [0.0, uy_north],  # Note: List of length nd. Extend if used in 3d.
         "max_iterations": 20,
     }
+
+    if issubclass(model_class, TailoredPoromechanicsTpsa):
+        # Tpsa is only consistent with Cartesian grids.
+        model_params["cartesian"] = True
+
     model = model_class(model_params)
     return model
 
--- a/tests/models/test_poromechanics.py
+++ b/tests/models/test_poromechanics.py
@@ -152,7 +152,8 @@ class NonzeroFractureGapPoromechanics(pp.PorePyModel):
                 val = self.units.convert_units(
                     self.params["fracture_source_value"], "kg * s ^ -1"
                 )
-                vals.append(val * np.ones(sd.num_cells))
+                # Distribute source term over cells based on cell volumes.
+                vals.append(val * sd.cell_volumes / np.sum(sd.cell_volumes))
         fracture_source = pp.wrap_as_dense_ad_array(
             np.hstack(vals), name="fracture_fluid_source"
         )
@@ -483,7 +484,7 @@ def test_push_north_zero_opening():
 )
 def test_positive_p_frac_positive_opening(model_class):
     model = create_model_with_fracture({}, {}, {}, 0.0, model_class)
-    model.params["fracture_source_value"] = 0.001
+    model.params["fracture_source_value"] = 0.004
     pp.run_time_dependent_model(model)
     _, _, p_frac, jump, traction = get_variables(model)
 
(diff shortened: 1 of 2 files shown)


jwboth:
To the degree, I understand the change from hardcoded values to more general values, I do not see any difficulty in accepting your changes. I also trust you that this is currently the best functioning solution for this check. I cannot see another alternative so quickly.

tests/models/test_poromechanics.py
```diff
@@ -491,14 +496,24 @@ def test_positive_p_frac_positive_opening(model_class):
...
+    # Fracture pressure is positive.
+    #
+    # EK: The original reference values are for mesh with 4 fracture cells. Since the
+    # fracture source is set constant per fracture cell, effectively acting as an
+    # intensive quantity, the total source term and hence the fracture pressure scales
+    # with the number of fracture cells. We do a (perhaps somewhat rough) correction
+    # here by scaling with the ratio between number of fracture cells in the current
+    # model and the original number of fracture cells (4).
+    orig_mean_pressure = 4.8e-4
+    orig_deviation = 2e-5
```
IvarStefansson: Wouldn't it be more elegant to scale fracture_source_value by number of cells?

tests/models/test_thermoporomechanics.py
```diff
@@ -294,9 +298,32 @@ def test_positive_p_frac_positive_opening(model_class):
...
+    # intensive quantity, the total source term and hence the fracture pressure scales
+    # with the number of fracture cells. We do a (perhaps somewhat rough) correction
+    # here by scaling with the ratio between number of fracture cells in the current
+    # model and the original number of fracture cells (4). For the temperature we also
+    # need to increase the tolerance to account for the non-linear relation between
+    # the pressure and temperature.
+    orig_mean_pressure = 4.8e-4
+    orig_pressure_deviation = 2e-5
+    orig_mean_temperature = 8.3e-6
+    orig_temperature_deviation = 1e-7
```
IvarStefansson: Why isn't the factor 10 uced on line 325 backed into this value?

IvarStefansson (commented):
One trivial question and one slightly more involved one. The latter is a suggestion to scale the source term instead of the reference solution. I think that'd be more elegant but I'm not adamant on this - open for discussion.

tests/models/test_poromechanics.py
keileg: Arguably yes. Though the number of cells in not know a priory, so we would need to first run `model.prepare_simulation()`, then set the source term and then run. Still, I think I agree, so I'll implement.

keileg: Correction: The fluid source is set as a `DenseArray`. To make the update, the equations must be set once more, but that is not the intended usage. Or I can make the source term into a time dependent array. But still, I feel the elegancy disappearing. 

Opinions?

IvarStefansson: Or, if it doesn't break other tests, replace
`vals.append(val * np.ones(sd.num_cells))`
in line 155 by
`vals.append(val * sd.cell_volumes)`. 
I think that should work.

tests/models/test_thermoporomechanics.py
keileg: Fixed.

tests/models/test_poromechanics.py
keileg: Implemented (with some modifications).

IvarStefansson (approved):

PR #1594, opened 2026-02-11
1582 refactor simulation flowchart wrt time step control

### LINKED ISSUE #1582: Refactor simulation flowchart wrt. time step control
Aim. Refactor the logical "time stepping -> nonlinear solution -> linear solution" pipeline to make the simulation flow more robust, maintainable, and easier to extend, while reducing the risk of subtle bugs in time stepping and solution control.

The current design for running non-stationary simulations in PorePy distributes the responsibility for time stepping and control logic across several classes and methods, primarily:
- `time_step` (called within `run_time_dependent_model`)
- `SolutionStratgey.before_nonlinear_loop` (called within a (nonlinear) solver) - note its well attempted but wrongfully executed docstring.
- `SolutionStrategy.after_nonlinear_convergence` (called within a (nonlinear) solver).
- `Solution.Strategy.after_nonlinear_failure` (called within a (nonlinear) solver).

Additionally, saving the status via `SolverStatistics` and using the `Exporter` to export solutions are tightly coupled to this flow. See in particular:
- the method `save_data_time_step` (called as part of `SolutionStrategy.after_nonlinear_convergence` and `Solution.Strategy.after_nonlinear_failure`).

There have been several issues related to unexpected behavior in time stepping, especially when nonlinear solution fails (see #1421, #1422, PR #1471, and related closed issues).

This issue aims at a wholistic approach to provide more solid ground for better communication and coordination between the aforementioned classes. The primary goal is to align and clarify the responsibilities related to time stepping, so that the hierarchy is: "time stepping -> nonlinear solution -> linear solution". Each level should only be responsible for its own purpose. In particular, nonlinear solvers should not be responsible for reacting to time step failure or resetting time steps. 

Several steps are suggested as a guideline (pre-planning will be nevertheless needed and is central in an aligned plan):

- [ ] Define a flowchart for that follows a hierarchical design "time stepping/time step control -> nonlinear solution -> linear solution" with clear responsibilities per level. Include exporting and logging, recomputation of time steps, and restarting failed time steps.
- [ ] Reevaluate the aforementioned methods (and their use) and reorganize them. Initial guess is that `time_step`, `SolutionStrategy.after_nonlinear_convergence` and `SolutionStrategy.after_nonlinear_failure` are most crucial.
- [ ] Split exporting solutions and saving `SolverStatistics`. Use only the latter for success and failure and the former in case of success.
- [ ] Refactor `NewtonSolver` and `LinearSolver` to take sole responsibility for solving a given time step (and not resetting it). Time step control is to be moved to a suitable location outside the "nonlinear solution -> linear solution" levels. Observation: Solvers report convergence success in form of a bool.
- [ ] Adapt and double check: does `run_stationary_model` still come across any time step control?

Milestones of the project:
1. Flowchart is presented to and accepted by core dev group.
2. Code is reorganized and current tests run.
3. Evaluate and implement any new unit tests required by the refactored design.

Friendly warning:
This issue is orthogonal but strongly connected to PR #1448. In this PR the `NewtonSolver` and `LinearSolver` are refactored and define status objects carrying information on success and failure. Merging the above time step control and the herein mention solver refactoring should be likely not a bit deal.

### LINKED ISSUE #1586: Refactor run_models
Currently, different logic and bookkeeping regarding time stepping, logging, progressbars, and solvers is crammed into ``run_time_dependent_model``. This makes it hard to see the time stepping and solver logic.

Analog to the rewrite of ``NewtonSolver`` in #1448, we want to introduce a base class ``ModelRunner``, with childs ``StationaryModelRunner``, ``TimeDependentModelRunner``, and ``IterativeModelRunner`` that clearly separate the logic into different methods.

### LINKED ISSUE #1587: Separate ``ModelSolverInterface`` and ``SolutionStrategy``
``SolutionStrategy`` crams many things together. This issue introduces ``ModelSolverInterface`` that has ``nonlinear_solver_statistics`` and ``time_manager`` as attributes and contains all ``before``/``after`` methods for interaction of the model and the solver and model runner.

vlipovac:
Closes #1582 
Closes #1586 
Closes #1587 

## Proposed changes

Refactoring the interface between models, solvers, time step control and run-methods.

Includes renaming of functionality for clarity, moving functionality, and consolidating hierarchies in the objects required to run a simulation.

Another change I want to propose within this PR:
Moving the `time_manager` out of the model and into the new runner. The communication regarding time stepping control is happening mostly in the runner. The model does not need to know about the time manager, except that it has a time step (AD time step) whose size is set in the outer time loop. If am currently investigating how invasive that change would be. Largest concern are the export mechanisms, which deal with time schedule, time indices and other information provided by the time manager. But technically, that information can be provided "from above" by the runner.

## Types of changes

- [ ] Other:

## Checklist


src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+        Increases the time and sets the model's AD time step value.
+        Executes :meth:`~porepy.models.solution_strategy.ModelSolverInterface.
+        before_time_step` and logs the progress.
+
+        """
+        # Increase the simulation time.
+        self.model.time_manager.increase_time()
+        self.model.time_manager.increase_time_index()
+        # Update the model's AD time step object.
+        self.model.ad_time_step.set_value(self.model.time_manager.dt)
```
pschultzendorff: Once it is decided where the time manager will live, we should consider moving this back into model.before_time_step. I think it avoids confusion to keep all changes to ad objects inside the model class.

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+                self.model.time_manager.compute_time_step(
+                    iterations=self.model.nonlinear_solver_statistics.num_iteration
+                )
+
+            # Update progressbar length.
+            self.time_progressbar.update(n=self.model.time_manager.dt / self._dt_0)
+            self.model.after_time_step_convergence()
+        else:
+            if self.model.time_manager.is_constant:
+                raise pp.TimeSteppingError(
```
vlipovac: Consider moving this exception into the time manager (which currently only raises a warning if compute_time_step is called with recompute=True and constant=True)

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+        self._is_nonlinear = self.model.is_nonlinear_problem()
+        """Flag indicating whether the problem is nonlinear, set at initialization."""
+
+        self._is_time_dependent = self.model.is_time_dependent()
+        """Flag indicating whether the problem is time-dependent, set at
+        initialization."""
+
+        self._dt_0: float = model.time_manager.dt
+        """Initial time step size, used for progress bar updates."""
+
```
jwboth: My sense would be to define this as part of `init_time_progressbar`. Or is there a good reason why not?

src/porepy/models/run_models.py
```diff
@@ -1,265 +0,0 @@
...
-
-    solver = _choose_solver(model, params)
-
-    solver.solve(model)
-
-    model.after_simulation()
-
-
-def run_time_dependent_model(model, params: Optional[dict] = None) -> None:
-    """Run a time dependent model.
```
jwboth, line 51: To be less disruptive, one could consider keeping this file and just reduce the call to create a ModelRunner and call its method run. The decision depends rather on strategy regarding deprecation. Postpone the decision for one round (easy to fix).

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+        if time_step_converged:
+            # Update the time step magnitude if the dynamic scheme is used.
+            if not self.model.time_manager.is_constant:
+                self.model.time_manager.compute_time_step(
+                    iterations=self.model.nonlinear_solver_statistics.num_iteration
+                )
+
+            # Update progressbar length.
+            self.time_progressbar.update(n=self.model.time_manager.dt / self._dt_0)
+            self.model.after_time_step_convergence()
```
jwboth: Since this is the main action here, I sort of would like to see this line of code at its ideal place. Wouldn't it be equivalent to call it just as the first action in this if-block? Then I would suggest moving it up.

src/porepy/models/solution_strategy.py
```diff
@@ -21,211 +21,29 @@
...
 
-        # Export initial condition
-        self.save_data_time_step()
+    """
 
-    def initialize_previous_iterate_and_time_step_values(self) -> None:
-        """Method to be called after initial values are set at ``iterate_index=0`` in
-        the mixins for initial conditions.
+    def is_nonlinear_problem(self) -> bool:
+        """Specifies whether the Model problem is nonlinear.
```
jwboth:
```suggestion
        """Specifies whether the problem is nonlinear.
```

src/porepy/models/protocol.py
```diff
@@ -498,13 +498,278 @@ def isotropic_second_order_tensor(
...
+        def after_solver_failure(self) -> None:
+            """Called if the solver fails to converge or diverges."""
+
+        def after_time_step_convergence(self) -> None:
+            """Called after a new time step solution has been achieved.
+
+            The base method does the following:
+
+            1. Calls :meth:`update_solution` with the current global iterate vector.
+            3. Calls :meth:`save_data_time_step`.
```
jwboth:
```suggestion
            2. Calls :meth:`save_data_time_step`.
```

src/porepy/models/solution_strategy.py
```diff
@@ -330,383 +150,164 @@ def schur_complement_primary_variables(self, names: list[str]) -> None:
...
 
         """
         solution = self.equation_system.get_variable_values(iterate_index=0)
-
-        # Update the time step magnitude if the dynamic scheme is used.
-        if not self.time_manager.is_constant:
-            self.time_manager.compute_time_step(
-                iterations=self.nonlinear_solver_statistics.num_iteration
-            )
         self.update_solution(solution)
```
jwboth: Any reason why solution `should` not be fetched within `update_solution`. Seems to me the more principled distribution of responsibilities, while after_time_step_convergence is only the manager that instructs the different operations. Might oversee why one would want to change a converged solution before updating it. I quickly checked locally and it should work.

src/porepy/models/solution_strategy.py
```diff
@@ -330,383 +150,164 @@ def schur_complement_primary_variables(self, names: list[str]) -> None:
...
+    def after_time_step_failure(self) -> None:
+        """Called after a time step has failed to converge.
+
+        The base method does the following:
+
+        1. Resets the current iterate values to the previous time step solution.
+
+        """
+        prev_solution = self.equation_system.get_variable_values(time_step_index=0)
+        self.equation_system.set_variable_values(prev_solution, iterate_index=0)
```
jwboth: This is just food for thought and no true request - is thus maybe mostly addressed towards core dev group:

Why don't we call methods by their actual acttions, e.g., here `initialize_iterate`? It would be much more understandable what is done when reading the code calling the method.

Also, this method implements the most simple initialization routine (using the latest time step). But one could use many more sophisticated guesses here (using momentum methods etc.). I agree that it is sort of stupid to call this method, while the iterate is already set correctly based on the nolinear solution during the previous time step, but would it really hurt to do this operation at each start of an attempted time step instead of at the end of each failed one?

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+
+        if self._is_time_dependent:
+            # Redirect the root logger, to avoid logger-progressbars interference.
+            with logging_redirect_tqdm([logging.root]):
+                # Time loop.
+                while not self.model.time_manager.final_time_reached():
+                    self.before_time_step()
+                    time_step_converged = self.solver.solve(self.model)
+                    self.after_time_step(time_step_converged)
+        else:
```
jwboth, line 228: Most serious concern of the first review. I do not see how the time manager is correctly instructed after a failed time step. Instead after a failed time step, if I understood `after_time_step` correctly, a new dt is computed and a new time step is attempted. But then in `before_time_step` time is increased again by the new dt, the time step index is increased. These two should not happen. Did I missunderstand?

jwboth (commented):
Just had an initial read and need clarification regarding the logic in the main method `run` in case of a failed time step. I am wondering whether I am not understanding or the test suite is not extensive enough to actually test the original run_model routines (might be!).

I will continue reading but to me it also seemed the rest was less critical and is going into the right direction!

src/porepy/models/model_runner.py
pschultzendorff: Agree

src/porepy/models/solution_strategy.py
pschultzendorff: Make more sense to me as well! Move this to `before_time_step` or `before_nonlinear_guess`. As I see it, the current code relies on not touching the iterate solution between `after_nonlinear_convergence` and `before_nonlinear_loop` to ensure it serves as an initial guess for the next Newton solve.

src/porepy/models/model_runner.py
pschultzendorff: Currently, the time manager will raise a warning, but no error if a recomputation is attempted with a constant time step. I don't really see why we need both, but perhaps there is a usecase? Ansonsten weg mit diesem Doppel-Gemoppel.

src/porepy/models/model_runner.py
pschultzendorff: That's just how `TimeManager` works. In `TimeManager._adaptation_based_on_recomputation`, the current `dt` is subtracted from `time`.

src/porepy/models/model_runner.py
jwboth: Before I forget. I think it would be beneficial (at least for some more time) to let the model keep control over the `time_manager`. Conceptually, I see how in principle the general model does only require `dt` but just consider source terms that depend on `time`, consider that `SolverStatistics` will actually track also some time information and waisted number of time steps etc. Let us postpone the discussion of a possible change in responsibility over the `time_manager` here.

src/porepy/models/model_runner.py
jwboth: Fully agree, and again surprised how deep one has to dig to understand what is going on, although I had read that part of the code.

I agree with your suggestion to not touch the `time_manager` in this PR. But could we add more documentation to `after_time_step` that indicates these modifications. I do not think one needs to write much here, but needs to address all actions that have been chosen to be applied on the level of a time step within `run` which essentially is everything within `before_time_step`. Since `before_time_step` calls `increase_time` and `increase_time_index`, I suggest adding a short comment in `after_time_step` (ca. line 208 where dealing with failure) that the `time_manager` also manages the reverse of these two steps on the inner level.

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+ModelInstance = TypeVar("ModelInstance", bound=pp.PorePyModel)
+"""Type variable for objects inheriting from the PorePy model protocol."""
+
+
+class ModelRunner:
+    """Class for running PorePy models according to their configurations.
+
+    Sets the outer solver (nonlinear or linear), which in the nonlinear case can
+    be customized by providing a solver type as ``params["nonlinear_solver"]``.
+
```
jwboth, line 100: Maybe add a comment on the implications of return values for `model.is_nonlinear` and particularly `model.is_time_dependent`

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+        self._dt_0: float = model.time_manager.dt
+        """Initial time step size, used for progress bar updates."""
+
+        self.set_solver()
+
+        if self._is_time_dependent:
+            self.init_time_progressbar()
+        else:
+            self.time_progressbar = DummyProgressBar()
```
jwboth: Consolidate this within `init_time_progressbar`.

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+        # etc.
+        # When ``NewtonSolver`` is called inside ``run``, the
+        # ``_nl_progress_bar_position`` parameter specifying the position of the
+        # ``NewtonSolver`` progressbar is passed.
+        self.params.update({"_nl_progress_bar_position": 1})
+
+        # Check if the user wants a progress bar. Initialize an instance of the
+        # progressbar_class, which is either :class:`~tqdm.trange` or
+        # :class:`~DummyProgressbar` in case `tqdm` is not installed.
+        if self.params.get("progressbars", False):
```
jwboth:
```suggestion
        if use_progress_bar:
```

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+            )
+            self.time_progressbar = progressbar_class(
+                range(expected_time_steps),
+                desc="time loop",
+                position=0,
+                dynamic_ncols=True,
+            )
+        # Otherwise, use a dummy progress bar.
+        else:
+            self.time_progressbar = DummyProgressBar()
```
jwboth, line 209: No strict suggestion:

When moving the creation of `DummyProgressBar` inside the init method for the case `_is_time_dependent == False` one could also consider using at the top:

```
        if not self._is_time_dependent or not use_progress_bar:
            self.time_progressbar = DummyProgressBar()
            return
```

Maybe this is short and clean?

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+            f"Time step {self.model.time_manager.time_index + 1}"  # Why +1? Consistent?
+        )
+
+    def after_time_step(self, time_step_converged: bool) -> None:
+        if time_step_converged:
+            # Update the time step magnitude if the dynamic scheme is used.
+            if not self.model.time_manager.is_constant:
+                self.model.time_manager.compute_time_step(
+                    iterations=self.model.nonlinear_solver_statistics.num_iteration
+                )
```
jwboth, line 293: Following the comment on consolidating lines 203-206, I also think this could be consolidated (but maybe at later point in time after the PR).

src/porepy/models/model_runner.py
jwboth: Let's discuss. Although I agree with you, here the raise of the error on this level clearly controls the simulation workflow. And I would argue for that it controls it at the right level. But I am open to listen to your arguments.

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # Logging and progressbar update.
+        logger.info(
+            f"\nTime step {self.model.time_manager.time_index} at time"
+            + f" {self.model.time_manager.time:.1e}"
+            + f" of {self.model.time_manager.time_final:.1e}"
+            + f" with time step {self.model.time_manager.dt:.1e}"
+        )
+        self.time_progressbar.set_description_str(
+            f"Time step {self.model.time_manager.time_index + 1}"  # Why +1? Consistent?
```
jwboth: Did you add the comment here?

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,211 @@
...
+    def after_time_step(self, time_step_converged: bool) -> None:
+        if time_step_converged:
+            # Update the time step magnitude if the dynamic scheme is used.
+            if not self.model.time_manager.is_constant:
+                self.model.time_manager.compute_time_step(
+                    iterations=self.model.nonlinear_solver_statistics.num_iteration
+                )
+
+            # Update progressbar length.
+            self.time_progressbar.update(n=self.model.time_manager.dt / self._dt_0)
```
jwboth, line 296: Can be moved inside the if block above.

src/porepy/models/solution_strategy.py
jwboth: Regarding a potential renaming, note the presence of `update_solution`, it would be the analogue to provide a routine for `init_iterate` or so and call it wherever suitable.

jwboth (commented):
Some more comments. To summarize, the main changes are the `ModelRunner`, a lighter `NewtonSolver` and the introduction of an additional interface in form of the `ModelSolverInterface`. Hope to get this confirmed during our discussion.

src/porepy/models/model_runner.py
jwboth: I am aware of being schizophrenic here...let's discuss what a transparent structure would be here.

src/porepy/models/model_runner.py
pschultzendorff: I included self._is_time_dependent in use_progress_bar and added a comment.

src/porepy/models/model_runner.py
pschultzendorff: The if-block? It should be updated in any case.

src/porepy/models/model_runner.py
pschultzendorff: I don't see anything in VSCode

src/porepy/models/solution_strategy.py
pschultzendorff: Ohh I thought you meant renaming the variable. I don't favor renaming the method to
``init_iterate`` as I think keeping the naming scheme of `before_`/`after_` helps
understanding the interaction between `ModelSolverAPI` and `nonlinear_solver` or `ModelRunner`.

pschultzendorff: Perhaps add an additional method called ``init_iterate`` that is called by ``before_nonlinear_loop``? But we already have a lot of methods that do nothing in the base case.

src/porepy/models/model_runner.py
pschultzendorff: I moved the ad update back to `model.before_time_step`

src/porepy/models/model_runner.py
pschultzendorff: Removed and logging and progressbar report the same time index now.

src/porepy/models/solution_strategy.py
```diff
@@ -330,383 +150,166 @@ def schur_complement_primary_variables(self, names: list[str]) -> None:
...
+    def before_solver_iteration(self) -> None:
+        """Called before a solver performs an iteration (calling the linear solver)."""
 
-    def after_nonlinear_iteration(self, nonlinear_increment: np.ndarray) -> None:
-        """Method to be called after every non-linear iteration.
+    def after_solver_iteration(self, nonlinear_increment: np.ndarray) -> None:
+        """Called after a solver performed an iteration and the linear solver computed
+        an increment.
 
         The base method does the following:
```
pschultzendorff, line 237: The `_solver_` methods are somewhat confusing now, because it's not immediately clear
that they refer to the nonlinear solver.

src/porepy/models/solution_strategy.py
pschultzendorff: I implemented `initialize_nonlinear_solution`.

src/porepy/models/model_runner.py
pschultzendorff: I modified the error message and will create an issue collecting all the ``TimeManager``
problems we found

src/porepy/models/model_runner.py
pschultzendorff: Will include in ``TimeManager`` issue.

src/porepy/models/solution_strategy.py
jwboth: Thanks for cleaning up the comments @pschultzendorff!

Could I possibly as a last action request a suggestion for how to deal with the above comment @vlipovac  and @pschultzendorff ? Since you suggested the new interface, maybe you have the best intuition here. If not, I can also think about how to deal with it.

pschultzendorff: I think we had a discussion about the names, but I don't remember what we agreed on. When fixing stuff I had to read the docstring three times to understand what was going on. I think rewording the docstring would also help.

jwboth: So to summarize, your suggestion would be in addition to the restructuring the methods (depending on the responsibility) a renaming, but currently without concrete suggestions? Would be a fair-enough conclusion. I just want to know.

pschultzendorff: We renamed them from `before_nonlinear_loop` etc. to `before_solver_loop` etc. I just realized that this is confusing after two weeks of not looking at the code. I think it is best to just be very explicit in the docstrings about the interplay with both nonlinear and linear solvers and `model.is_nonlinear_problem`. I can add the changes later today.

src/porepy/numerics/linear_solvers.py
jwboth: Reminder to myself. Revisit as part of this PR?

src/porepy/models/solution_strategy.py
pschultzendorff: I forgot about this yesterday. Added it to my ToDo list now.

src/porepy/numerics/linear_solvers.py
```diff
@@ -80,6 +83,10 @@ def solve(self, model: SolutionStrategy) -> SimulationStatus:
         # Perform a single (Newton) iteration.
         model.assemble_linear_system()
         nonlinear_increment = model.solve_linear_system()
+        model.after_solver_iteration(nonlinear_increment)
+        # NOTE: The linear solver performs only one iteration.
+        # FIXME: Consider renaming the solver statistics to just "solver statistics".
+        model.nonlinear_solver_statistics.num_iteration = 1
```
jwboth: Reminder to myself. Revisit. Needed? The workflow make sure this is true. (see model.before_* routines)

src/porepy/models/solution_strategy.py
jwboth: Wait for my go. I am cleaning up from the attempt to resolve the conflicts - of course I failed in some places. But should be done hopefully soon (maybe tomorrow).

jwboth:
@copilot do not build, what is the reason that no checks are performed?

src/porepy/models/solution_strategy.py
vlipovac: Sorry for the late reply.
That name is inspired by a todo note I found as an inline comment in the linear solver. In Short, the linear solver called model methods with "nonlinear" in their name to store the solution of the linear problem using the AD framework. The original authors didn't want to duplicate code, which is reasonable. I just resolved that todo by renaming the method to avoid confusion, all solvers should without worry access that method.
The only difference now is that the Linear solver does not do += 1 to iteration but =1 for iteration, which is mathematically sound.

src/porepy/numerics/linear_solvers.py
vlipovac: Related to this
https://github.com/pmgbergen/porepy/pull/1594#discussion_r2894420336

src/porepy/numerics/linear_solvers.py
vlipovac: Also related to
https://github.com/pmgbergen/porepy/pull/1594#discussion_r2894420336

src/porepy/models/solution_strategy.py
pschultzendorff: Just let me know when I should have a look @jwboth

src/porepy/numerics/linear_solvers.py
jwboth: Now it is not needed anymore, and commented out. During upcoming review, I would like to instantiate a general discussion whether most of the long comments in the `LinearSolver` class could be removed. They feel outdated and connected to old "aha" moments. Unless, one sees a strong reason for keeping them.

src/porepy/numerics/linear_solvers.py
jwboth: See comment below. Remove this?

src/porepy/models/solution_strategy.py
jwboth: Took some time... It would be great if you could now have a final look at the method names @pschultzendorff (possibly in collaboration with @vlipovac if you had a previous discussion). It is also OK, if you just explain your suggestions here in the comment section, and I will do the replacement.

And no other changes are required. I will do a final round of cleaning up in a few places. There is some noisy code here and there which felt necessary to make the merge of develop and your code work (at least as quick as possible). 

Thanks!

src/porepy/applications/convergence_analysis.py
```diff
@@ -179,18 +179,16 @@ def run_analysis(self) -> list:
...
                 # Run time-dependent model
-                pp.run_time_dependent_model(model)
-                # Complement information in results
-                setattr(model.results[-1], "cell_diameter", model.mdg.diameter())
+                pp.ModelRunner(model).run()
                 setattr(model.results[-1], "dt", model.time_manager.dt)
+            else:
+                # Run stationary model
+                pp.ModelRunner(model, deepcopy(self.model_params[level])).run()
+
```
pschultzendorff: Simplify this (and possibly remove a bug because both `run` calls differ?). `ModelRunner` checks whether `model` is time dependent.

src/porepy/models/compositional_flow.py
```diff
@@ -1680,7 +1680,7 @@ def after_nonlinear_convergence(self) -> None:
         assert isinstance(self, pp.SolutionStrategy), (
             "This is a mixin. Require SolutionStrategy as base."
         )
-        super().after_nonlinear_convergence()  # type:ignore[safe-super]
+        super().after_solver_convergence()  # type:ignore[safe-super]
 
         subdomains = self.mdg.subdomains()
         nt = self.time_step_indices.size
```
pschultzendorff, line 1686: Should this be moved to `after_time_step`?

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,371 @@
...
+        """Flag indicating whether the problem is nonlinear, set at initialization."""
+
+        self._is_time_dependent = self.model.is_time_dependent()
+        """Flag indicating whether the problem is time-dependent, set at
+        initialization."""
+
+        self.set_solver()
+
+        self.init_time_progressbar()
+
```
pschultzendorff: Do this only when `self._is_time_dependent`. `init_time_progressbar` accesses
`model.time_manager`. I think this could lead to bugs.

pschultzendorff:
```suggestion
        if self._is_time_dependent:
            self.init_time_progressbar()

```

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,371 @@
...
+            # Need to log before updating the time step size.
+            self.logging(simulation_status)
+
+            # Update the time step magnitude if the dynamic scheme is used.
+            if not self.model.time_manager.is_constant:
+                assert isinstance(
+                    self.model.nonlinear_solver_statistics, pp.NonlinearSolverStatistics
+                )
+                self.model.time_manager.compute_time_step(
```
pschultzendorff, line 291: If this is needed, consider adding a short comment why this has to be checked.

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,371 @@
...
+            self.time_progressbar.update(n=self.model.time_manager.dt / self._dt_0)
+
+        elif solver_status.is_failed() or solver_status.is_stopped():
+            if self.model.time_manager.is_constant:
+                logger.warning(
+                    """Solver failed to converge but time step size is constant and """
+                    """cannot be reduced."""
+                )
+                simulation_status = pp.SimulationStatus.STOPPED
+                self.logging(simulation_status)
```
pschultzendorff, line 305: Why not use `pp.SimulationStatus.FAILED`?

src/porepy/models/model_runner.py
```diff
@@ -0,0 +1,371 @@
...
+                # This calls
+                # ``time_manager._adaptation_based_on_recomputation``, which substracts
+                # the current ``dt`` from the simulation time, computes a shorter
+                # ``dt``, and adds the updated ``dt`` to the simulation time again.
+                # It will also raise a TimeSteppingError if the minimal time step
+                # is reached.
+                try:
+                    simulation_status = pp.SimulationStatus.FAILED
+                    self.model.after_time_step_failure()
+                    # Need to log before updating the time step size.yy
```
pschultzendorff:
```suggestion
                    # Need to log before updating the time step size.
```

src/porepy/models/model_runner.py
pschultzendorff: Okay, I see the difference now. I think the naming could be more clear. I would expect
`pp.SimulationStatus.FAILED` to correspond to a status that cannot be fixed and
`pp.SimulationStatus.STOPPED` to a temporary break in running the simulation.

src/porepy/models/solution_strategy.py
pschultzendorff: I misunderstood your comment and thought we were supposed to do a more broad review of
the changes. Luckily I haven't spent more than 20 minutes so far. Feel free to ignore my other comments.

pschultzendorff: I like the choice of names!

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -30,7 +30,7 @@
 logger = logging.getLogger(__name__)
 
 
-class NewtonSolver:
+class NewtonSolver(LinearSolver):
     """Nonlinear solver class implementing the Newton-Raphson method.
```
pschultzendorff, line 34: 😮

src/porepy/numerics/linear_solvers.py
```diff
@@ -59,9 +74,13 @@ def __init__(self, params: Optional[dict] = None) -> None:
         )
         """Divergence criterion used in the convergence check."""
```
pschultzendorff, line 76: This has the potential to be very confusing. Maybe it's better having a solver base class (possibly `ABC`)
that both `LinearSolver` and `NewtonSolver` are derived from? Or move this
`nl_divergence_criteria` initialization to `NewtonSolver`.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -389,7 +388,7 @@ def iteration(self, model: SolutionStrategy) -> np.ndarray:
         return nonlinear_increment
```
pschultzendorff, line 393: Consider renaming the second function to avoid confusion. Possibly
`linearization_iteration`?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
pschultzendorff: If it is kept this way, I think it should be well documented why this inheritance is done and that `NewtonSolver` implements "its own" linear solver instead of calling `LinearSolver`.

vlipovac: The code was such that they in principle have the same interface, because the interface is used by run models, now model runners. The suggestion here is to formalize this interface, including a hierarchy.

src/porepy/models/compositional_flow.py
vlipovac: I did not follow all the changes introduced by you and Jakub after I got reassigned. Just making clear that this needs to happen after convergence success only.

src/porepy/numerics/linear_solvers.py
jwboth: I fully agree (before and after) your suggestion to introduce inheritance. Their functionality has gotten quite close now. I suggest to wait with the discussion on the introduction of such for a separate PR.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
jwboth: I understood the intention behind your suggestions just as you explain. In view of Peter's comment below, I currently do not see a reason for more documentation. But would rather prioritize to design a abstract solver class that defines the connectors to the new model-solver interface.

src/porepy/models/model_runner.py
jwboth: Thanks for the comment! It seems we will introduce further status objects in the very near future (linear solvers, etc.) I suggest postponing renaming efforts until later. Let's keep the comment for now to not forget, and I will move this later to an issue.

src/porepy/models/model_runner.py
jwboth: as always, mypy... but OK, can do.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
jwboth: I partially agree. Here it is indeed a full linearization step. In the subclass `LineSearch` class suddenly it is a relaxed linearization iteration. Therefore, I suggest `iteration` or `nonlinear_iteration` in the sense of a general iterative solver. Keeping the comment for now to wait for more comments from others and votes.

src/porepy/applications/convergence_analysis.py
jwboth: Done

src/porepy/models/compositional_flow.py
jwboth: It should be consistent. So I think, no change is needed here.

src/porepy/models/model_runner.py
jwboth: Done.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
jwboth: I added a sentence.

keileg:
Parts of this PR was cherry-picked into a smaller PR #1648. 

The ideas underlying some of the other changes will be picked up at a later point, though we will need to take a broader look at the model-numerics interface before we can define a reasonable API for that one.

Closing this one now.

Issue #1184, opened 2024-06-07
Consider ignoring equations with empty domains of definition


keileg:

mikeljordan:
**Tentative tasks:**

- [x] Profile first — based on the above, write a small flow/mechanics benchmark script comparing assembly time with/without fracture equations on a fracture-free domain, different grid cell sizes.
- [ ] Prototype a skip — add a check in e.g, assemble() method that skips equations whose total DOF count is zero.
- [ ] Test carefully — provided, we safe a lot of cost based on the second task, run the existing test suite to see what breaks, which will reveal the "workflow trouble."

Implementations are pushed to this branch [here](https://github.com/pmgbergen/porepy/tree/ignore_equations_on_empty_grids_%231184)

mikeljordan:
This is going to really improve efficiency of the assembly pipeline. See a quick profiling result.

<img width="611" height="545" alt="Image" src="https://github.com/user-attachments/assets/a1485b8f-fe7e-4de1-8201-fdf4aca4a8b4" />

keileg:
Resolved by #1605

PR #1605, opened 2026-03-09
Ignore equations on empty grids #1184

mikeljordan:
## Proposed changes

Equations defined on empty domains (e.g., fracture-related equations in a non-fractured geometry) were previously assembled even though they contribute zero DOFs to the system. This PR filters out such equations in `_parse_equations()` before they enter the assembly pipeline, avoiding unnecessary AD operator evaluation.

### Changes

- `EquationSystem._parse_equations()`: When no explicit equation list is provided (the default path), equations whose total DOFs (from `_equation_image_space_composition`) are zero are excluded. This is the path used by `assemble()` when called without arguments, which is the standard usage during time-stepping and nonlinear iterations. When an explicit equation list is provided, all requested equations are returned without filtering, under the assumption that the caller has intentionally selected those equations.
- EquationBasedLebesgueMetric.__call__(): Skip equations not present in assembled_equation_indices to handle empty-domain equations that are filtered out during assembly.

### New tests:

- Updated` test_parse_equations`, to verify that `_parse_equations()` excludes the empty equation while _equations still contains it.
- `test_assemble_ignores_empty_equations`: Verifies that adding an empty-domain equation does not change the assembled Jacobian or residual vector.
- `test_poromechanics_empty_equation_filter`: Integration test on a poromechanics model without fractures, confirming that empty equations are filtered and assembly results are identical.

Co-authored-by @RonghaoCui

## Types of changes

- [ x] Testing (contribution related to testing of existing or new functionality).
- [ x] Documentation (contribution related to adding, improving, or fixing documentation).
- [ x] Maintenance (e.g., improve logic and performance, remove obsolete code).
- [ ] Other:

## Checklist

- [ x] The documentation is up-to-date.


src/porepy/numerics/ad/equation_system.py
```diff
@@ -1367,9 +1367,22 @@ def _parse_equations(
...
         if equations is None:
-            return dict((name, None) for name in self._equations)
+            # Precompute equations on non-empty domain. This is to avoid
+            # injecting such equations into the assembly pipeline
+            non_empty_equations = [
+                name
+                for name in self._equations
+                if sum(
+                    len(indices)
+                    for indices in self._equation_image_space_composition[name].values()
```
keileg, line 1381: Why not use the property instead of the private variable?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1367,9 +1367,22 @@ def _parse_equations(
             equation rows) as values. If no restriction is given, the value is None.
 
         """
+
         # The default return value is all equations with no grid restrictions.
         if equations is None:
-            return dict((name, None) for name in self._equations)
+            # Precompute equations on non-empty domain. This is to avoid
```
keileg, line 1374: Please update the method documentation as well.

tests/models/test_poromechanics.py
```diff
@@ -621,3 +624,59 @@ def test_poromechanics_well():
...
+        "material_constants": {"fluid": fluid, "solid": solid},
+        "u_north": [0.0, 0.001],
+        "grid_type": "cartesian",
+        "times_to_export": [],
+    }
+    model = model_class(params)
+    model.prepare_simulation()
+    equation_system = model.equation_system
+
+    all_equations = list(equation_system._equations.keys())
```
keileg:
```suggestion
    all_equations = list(equation_system.equations.keys())
```

tests/models/test_poromechanics.py
```diff
@@ -621,3 +624,59 @@ def test_poromechanics_well():
...
+    parsed_equations = list(equation_system._parse_equations().keys())
```
keileg, line 663: Please document what happens here. I don't understand how this returns something different from the previous line (I'm sure it does, but it should not be necessary to digest the called method to get the point).

tests/models/test_poromechanics.py
```diff
@@ -621,3 +624,59 @@ def test_poromechanics_well():
...
+@pytest.mark.parametrize(
+    "model_class", [TailoredPoromechanics, TailoredPoromechanicsTpsa]
+)
+def test_poromechanics_empty_equation_filter(model_class):
+    """Test that empty domain equations in poromechanics models exist and are
+    filtered before assembly.
+
+    For poromechanics models without fractures, the fracture-related equations
+    can still exist in the equation system. These empty domain equations should
+    be filtered as they donot contribute to the assembly pipline.
```
keileg:
```suggestion
    be filtered as they do not contribute to the assembly pipline.
```

tests/models/test_poromechanics.py
```diff
@@ -621,3 +624,59 @@ def test_poromechanics_well():
...
+    model = model_class(params)
+    model.prepare_simulation()
+    equation_system = model.equation_system
+
+    all_equations = list(equation_system._equations.keys())
+    parsed_equations = list(equation_system._parse_equations().keys())
+
+    # Check empty domain equations exist.
+    empty_equations = []
+    for name in equation_system._equations:
```
keileg:
```suggestion
    for name in equation_system.equations:
```
Please don't access private attributes when public ones are available.

tests/models/test_poromechanics.py
```diff
@@ -621,3 +624,59 @@ def test_poromechanics_well():
...
+        total = sum(
+            len(indices)
+            for indices in equation_system._equation_image_space_composition[
```
keileg: Same comment as above.

tests/numerics/ad/test_equation_system.py
```diff
@@ -1203,6 +1216,15 @@ def test_parse_equations(model: EquationSystemMockModel):
         received_equations_3[all_equation_names[0]], np.arange(model.sd_top.num_cells)
     )
 
+    # Add an equation on an empty domain to the equation system.
+    model.add_equation_on_empty_domain()
+
+    # Check that the empty equation is included in the equation system.
+    assert "empty_equation" in equation_system._equations
```
keileg: Private methods..

keileg: Please correct also other places in the PR. Where possible, of course

keileg (commented):
Mainly looks good. A few minor comments, though.

tests/models/test_poromechanics.py
RonghaoCui: done

tests/models/test_poromechanics.py
RonghaoCui: done

tests/models/test_poromechanics.py
RonghaoCui: done

tests/models/test_poromechanics.py
RonghaoCui: Now the only public methods are used here.

src/porepy/numerics/ad/equation_system.py
mikeljordan: This one is fine as a class should have access to its own internals. But outside the class, yes we should allow the callers to use the public access property instead

tests/models/test_poromechanics.py
RonghaoCui: I documented here to explain what are the differences.

tests/numerics/ad/test_equation_system.py
RonghaoCui: All added tests have been checked to ensure that no unnecessary private methods are used.

keileg (approved):

PR #1617, opened 2026-03-18
Avoid unnecessary memory use in discretization matrices

### LINKED ISSUE #1611: Unnecessary memory consumption in discretization matrices
Context: The following observations are based on a 3d THM setup with a varying number of cells. I have not investigated other types of simulations, problem may be more or less severe there as well.

Findings from memory measurements around the point of an iterative linear solve, which is presumably the point of maximum memory consumption:
1. The size of a multiphysics model is ridiculously high: I get 1.2GB for a grid with <9K 3d cells. Similar observations apply for other grid sizes, same model.
2. It seems that about 25-30% of the memory is spent on the full dicsretization matrix. As we cannot do matrix-free methods, this is a cost that we have to bear somehow. We should make sure the matrix is not represented twice (once in block form, once in PETSc or similar), and that it is deleted when no longer needed. That should be a relatively minor fix over in the iterative solver repository.
3. Of the remaining memory usage, almost all is spent on storing discretization matrices for the various terms in the equations. To some degree, this is a consequence of legacy decisions on the fundamental design of PorePy (and to some degree, this again reflects this desire not to use loop-based assembly in Python). Even given that, there are substantial savings to be made.
4. Inspection (by pympler, which measures the actual memory consumption of an array, not the nominal size given by arr.nbytes) of the big individual matrices shows that storing the data requires about twice as much data as expected. According to an AI, this may be a result of scipy deciding to make a bigger view of the data arrays, related to indixing, slicing etc.  The way to fix it is first to verify that `matrix.data.flags.owndata` is `False` (this seems to be an indication extra space has been added) and then do `matrix.data` = matrix.data.copy()`, which forces `matrix.data` do own its data, and thereby drop the extra space. The cost of an extra copy should be miniscule in this context. Note that I have not tested where in the workflow to do this reduction, and whether its effect persists or not (speculatively, it could be that the extension happens inside the ad parsing, which will make this a recurring issue; but it could also be that a single pruning of the matrices is sufficient). In my ruff estimates, fixing this can cut memory consumption of the bad matrices with about 1/3. If this applies to all discretization matrices in the case that I investigated (no idea if all matrices are bad in this case), that should be about 2-300 MB of the 1.2GB size of the model.
5. We systematically use int64 for representing indices in CSR/CSC matrices, even though we almost always can do int32, or even int16 if the number of rows is less than 65K or thereabout. For sparse matrices with a significant number of non-zeros, reducing to int32 can reduce the memory footprint by a factor ~25% (assuming point 4 has been implemented). Though int16 is tempting, it also puts us at risk for having to check matrix sizes systematically, but int32 should in practice be safe in that regard.
6. For reference, I also tried to measure the memory consumption of the PETSc linear solver. Though I don't fully trust that the measurements are accurate, the observation is that this is not a major drag on memory. 

Suggested actions: 
* Fixing point 4 should be given high priority. If we can make a function `prune_matrices` and run all discretization matrices through this, it could be a way to do both 4 and 5. We also need to know whether the matrices must be pruned regularly.
* As a lower priority (and strictly speaking a task over in the [linear solver repository](https://github.com/pmgbergen/porepy-iterative-solvers)), we should do further measurements of the memory footprint of the PETSc solver.

keileg:
## Proposed changes
Introduce a pruning method that i) makes sure that the data of a sparse matrix does not claim unnecessarily much space, ii) use i16/i32 data types to represent indices and index pointers where relevant.

Only csc/csr matrices are covered, since these are the ones that we struggle with.

I am sure we can do similar optimizations elsewhere, in particular on the index/index pointer side, but profiling indicated that the current fixes handle the excessive memory usage. 

Note that I did not include testing when the matrix pruning is called from the EquationSystem. The actual functionality should be covered by the lower-level methods which are tested, while the tests of models confirm that the method does not break. That leaves untested that the pruning methods actually prune when called from EquationSystem, but the probability of a failure specifically there seems so low I opted not to introduce extra tests.

Resolves #1611. 

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/ad/equation_system.py
+++ b/src/porepy/numerics/ad/equation_system.py
@@ -1590,6 +1590,8 @@ class EquationSystem:
         # Uniquify to save computational time, then discretize.
         unique_discr = pp.ad.uniquify_discretization_list(discr)
         pp.ad.discretize_from_list(unique_discr, self.mdg)
+        # Reduce the memory footprint of discretization matrices.
+        pp.matrix_operations.prune_discretization_matrices(self.mdg)
 
     @overload
     def assemble(
--- a/src/porepy/numerics/linalg/matrix_operations.py
+++ b/src/porepy/numerics/linalg/matrix_operations.py
@@ -1505,6 +1505,82 @@ def invert_diagonal_blocks(
     return ia
 
 
+def prune_matrix(A: sps.spmatrix) -> None:
+    """Prune sparse matrix to reduce memory usage.
+
+    Two changes are considered:
+        1. If the matrix does not own its own data, as can happen if scipy has created
+           a view of the matrix, the data array is copied; a byproduct of this is that
+           the data array will not occupy more space than necessary.
+        2. The data types for row and column indices/index pointers are adjusted to the
+           size of the matrix.
+
+    Parameters:
+        A: The matrix to be pruned. Should be of csr or csc format. The matrix is
+           modified in place.
+
+    """
+    MAX_I16 = np.iinfo(np.int16).max
+    MAX_I32 = np.iinfo(np.int32).max
+
+    # A small buffer is used to avoid overflow when the size is close to the limit.
+    buffer = 10
+
+    num_rows, num_cols = A.shape
+    if A.format == "csr" or A.format == "csc":
+        if not A.data.flags.owndata:
+            # We could probably do this with matrix types beyond csr/csc as well, but
+            # there are hardly any practical use of this (except from diagonal matrices,
+            # where the utility of this fix is questionable), and we don't have
+            # sufficient test coverage of those cases to know it is safe. Hence, we
+            # limit to csr/csc.
+            A.data = A.data.copy()
+        if A.format == "csr":
+            if num_cols < MAX_I16:
+                A.indices = A.indices.astype(np.int16)
+            elif num_cols <= MAX_I32:
+                A.indices = A.indices.astype(np.int32)
+        else:  # A.format == 'csc'
+            if num_rows < MAX_I16 - 10:
+                A.indices = A.indices.astype(np.int16)
+            elif num_rows <= MAX_I32 - 10:
+                A.indices = A.indices.astype(np.int32)
+        if A.data.size < MAX_I16 - buffer:
+            A.indptr = A.indptr.astype(np.int16)
+        elif A.data.size <= MAX_I32 - buffer:
+            A.indptr = A.indptr.astype(np.int32)
+
+
+def prune_matrices_in_dict(data: dict[str, sps.spmatrix]) -> None:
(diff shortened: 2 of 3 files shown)


src/porepy/numerics/linalg/matrix_operations.py
```diff
@@ -1505,6 +1505,83 @@ def inv_python(indptr, ind, data, sz):
     return ia
 
 
+def prune_matrix(A: sps.spmatrix) -> None:
```
IvarStefansson, line 1552: Is there any risk of the indices being restricted causing trouble during stacking or similar operations?

tests/utils/test_matrix_operations.py
```diff
@@ -680,25 +680,96 @@ def test_diagonal_matrix_from_sparse_blocks():
...
+            assert optimized.getformat() == "csr"
+
+    def mat(self):
+        """Helper method to construct a sparse matrix with specific properties."""
+        mat = sps.csr_matrix(np.array([[1.1, 0], [3, 2]]))
+        mat.indices = mat.indices.astype(np.int64)
+        mat.indptr = mat.indptr.astype(np.int64)
+        mat.data[:] = mat.data[:] + 2.0
+        # By inspection, the matrix has been confirmed not to own its own data at this
+        # point, hence the pruning has work to do.
```
IvarStefansson: Would it make sense to add "assert not mat.data.flags.owndata"?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1590,6 +1590,8 @@ def discretize(
         # Uniquify to save computational time, then discretize.
         unique_discr = pp.ad.uniquify_discretization_list(discr)
         pp.ad.discretize_from_list(unique_discr, self.mdg)
+        # Reduce the memory footprint of discretization matrices.
+        pp.matrix_operations.prune_discretization_matrices(self.mdg)
```
IvarStefansson, line 1594: Have you timed this?

IvarStefansson (approved):
Thanks for fixing! Some minor comments, then merge.

src/porepy/numerics/linalg/matrix_operations.py
keileg: Good point. It should be safe for pure scipy stacking, but when we do the stacks manually, as is done in some matrix_operations functions, we may not be safe. I'll add some conversions there.

tests/utils/test_matrix_operations.py
keileg: Yes, will do.

src/porepy/numerics/ad/equation_system.py
keileg: Not systematically, but it should not be noticeable compared to other operations (anything done on 1d numpy arrays is cheap compared to the rest of what we are doing).

PR #1620, opened 2026-03-20
FV discretizations do not offer reconstruction of primary variables on internal faces

### LINKED ISSUE #1612: Reduce size of FV face field reconstructions
For mpfa/tpfa and mpsa, we produce reconstruction matrices for face pressures and displacements, respectively. These are called `bound_pressure_{cell,face}` and `bound_displacement_{cell,face}`. These seem to be constructed for both internal and external faces, but I believe they are only ever needed for external faces. If true (should be verified), we can save a lot of memory by not storing these numbers.

This probably applies to the tpfa/mpfa 'bound_pressure_Vector_source' as well. 

Action: 
1. Verify usage (likely by a search through the code base). 
2.  Assuming the answer to 1. is not intimidating, make it a default option wipe most of the matrix elements at the time of construction. We probably want an option to keep the reconstruction on internal faces.
3. Resulting from this, some tests may need an update as well.

keileg:
## Proposed changes
The mpsa, mpfa and tpfa discretization used to construct matrices for reconstruction of the primary variables (displacement, pressure), on all faces. In practice we only use this for external faces, hence storing the quantities needed for the computation on internal faces is a not needed. This PR does away with these extra fields.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/fv/_fvutils.py
+++ b/src/porepy/numerics/fv/_fvutils.py
@@ -355,6 +355,33 @@ def find_active_indices(
     return active_cells, active_faces
 
 
+def boundary_face_mask(sd: pp.Grid, internal_face_active: bool) -> sps.spmatrix:
+    """Construct a mask to be applied to the discretization matrices, to eliminate
+    contributions from internal faces if these are not active.
+
+    Parameters:
+        sd: Grid to be discretized.
+        internal_face_active: Whether the discretization should include contributions
+            from internal faces. If False, contributions from internal faces will be
+            eliminated by the mask.
+
+    Returns:
+        sps.spmatrix: Mask to be applied to the discretization matrices of shape
+            (sd.num_faces, sd.num_faces).
+
+    """
+
+    if internal_face_active:
+        filter_array = np.ones(sd.num_faces, dtype=bool)
+    else:
+        filter_array = np.zeros(sd.num_faces, dtype=bool)
+        filter_array[sd.get_all_boundary_faces()] = True
+    boundary_face_mask = sps.dia_matrix(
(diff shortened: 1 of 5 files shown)


src/porepy/numerics/fv/_fvutils.py
```diff
@@ -355,6 +355,32 @@ def find_active_indices(
...
+    contributions from internal faces if these are not active.
+
+    Parameters:
+        sd: Grid to be discretized.
+        internal_face_active: Whether the discretization should include contributions
+            from internal faces. If False, contributions from internal faces will be
+            eliminated by the mask.
+
+    Returns:
+        sps.spmatrix: Mask to be applied to the discretization matrices.
```
IvarStefansson: Specify shape?

src/porepy/numerics/fv/mpfa.py
```diff
@@ -155,6 +158,9 @@ def discretize(self, sd: pp.Grid, data: dict) -> None:
         inverter: Literal["numba", "python"] = parameter_dictionary.get(
             "mpfa_inverter", "numba"
         )
+        internal_face_reconstruction = parameter_dictionary.get(
```
IvarStefansson: Can we limit to at most two names among the three current ones, please?

IvarStefansson: It could be that we could do with with a single reconstruct_on_internal_faces

IvarStefansson (approved):
Naming question, otherwise good. Thanks!

src/porepy/numerics/fv/mpfa.py
keileg: Will do.

PR #1630, opened 2026-03-30
BUG: Remove `skipped` tag from fixture

pschultzendorff:
## Proposed changes
Fix tests not collecting in VSCode due to `pytest.mark.skipped` marking a fixture.

## Types of changes

- [ ] Other:

## Checklist


IvarStefansson:
As far as I understand, skipping the fixture means skipping all tests depending on it, in this case test_order_of_convergence of the same file. Since that test is itself skipped, this PR doesn't change behaviour of current code and testing. Do you agree with this, @pschultzendorff? If yes, I still think it's a change we should make to clean up and avoid future confusion.

pschultzendorff:
I cannot find proper documentation for this, but from what I can gather from some google searches is that fixtures should be skipped with `pytest.skip` inside the function body instead of the `pytest.mark.skipped` decorator. I can change the code to do that or add a comment that this behavior should be considered if further tests that depend on the fixture are added.

jhabriel:
Hi @pschultzendorff, the `skipped` flag is a custom flag I've created back in the days to skip pytest tests; here's the implementation: https://github.com/pmgbergen/porepy/blob/develop/conftest.py.

Sadly, I do not remember why we did not use the standard `skip` option from pytest, but I vaguely remember that it had some limitations (maybe it was just we wanted a custom -- option ). 

All the best.

pschultzendorff:
Thanks for the explanation @jhabriel ! I was already confused about the naming. Reading through `conftest.py`, all `pytest.mark.skipped` decorators are internally transferred to `pytest.mark.skip`, which still is not correct for fixtures.

IvarStefansson (approved):

PR #1631, opened 2026-03-31
SPEED: Caching of Ad operators for buoyancy flux computations

keileg:
## Proposed changes
These are really minor changes, but they give a notable speedup for buoyancy-related simulations.

NOTE: Slow tests must pass before this is merged.
EDIT: OK, test passed.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/fluid_property_library.py
+++ b/src/porepy/models/fluid_property_library.py
@@ -769,8 +769,8 @@ class FluidBuoyancy(pp.PorePyModel):
 
         """
         normals = self.outwards_internal_boundary_normals(interfaces, unitary=True)
-
         subdomain_neighbors = self.interfaces_to_subdomains(interfaces)
+
         projection = pp.ad.MortarProjections(
             self.mdg, subdomain_neighbors, interfaces, dim=self.nd
         )
@@ -797,6 +797,7 @@ class FluidBuoyancy(pp.PorePyModel):
         w_flux.set_name("interface_density_driven_flux_" + density_metric.name)
         return w_flux
 
+    @pp.ad.cached_method
     def __entity_buoyancy_flux(
         self,
         advected_gamma_quantity: pp.ad.Operator,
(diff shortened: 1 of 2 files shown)


src/porepy/models/fluid_property_library.py
```diff
@@ -751,6 +751,21 @@ def density_driven_flux(
         w_flux.set_name("density_driven_flux_" + density_metric.name)
         return w_flux
 
+    @pp.ad.cached_method
+    def __interface_gravity_projection_operator(self, interfaces) -> pp.ad.Operator:
+        """Construct a mortar projection mapping used in interface_density_driven_flux.
+
```
jwboth: Why the method name? I see that this project shall be used in a specific context. But the method itself is sensible in a general context. Would this projection not be needed in other situations as well? Would it make sense to apply the idea of providing a cached projection more globally (or at least also in other models)?

jwboth (approved):
Looks good. See my comment on the naming and wider use of the cached projection.

src/porepy/models/fluid_property_library.py
keileg: Good question. The background is that constructing a mortar projection is somewhat costly. When this is done as part of `set_equations`, the cost is usually negligible in the bigger scheme of things. In this particular case, the calling method is invoked before each Newton iteration (as part of updating fluxes used in upwinding) and this becomes a notable drag on performance. The usual work around is to use the decorator pp.ad.cached_method, but for some reason I cannot understand, this cannot be applied to `interface_density_driven_flux` without impacting the behavior of the simulation (specifically, some tests in `test_buoyancy_flow` broke when I tried to do so).

I agree that the current solution is not optimal. A better option would be to make caching work, possibly by figuring out how the method is cached (note to self: is it something with the hash of the ad operator representing the density metric?). I have done some work on operator equality in an unrelated project, will make a PR later today (sorry!), and will return to the current PR after having incorporated those changes, to see if that improves the matter. Hence, this PR will be put on hold.

The better solution would likely be to make the `MortarProjection` class persistent outside the calling methods where they are invoked. This will however require a redesign of that part of the models, which will take more of an effort.

keileg: I ended up reverting this part of the PR. The issue is still there, but we should not introduce a temporary fix like the one I suggested.

I will merge once the tests are completed.

PR #1632, opened 2026-03-31
Ad speedups

keileg:
## Proposed changes
This is a collection of minor changes to the Ad framework, mainly related to surrogate operators. 

My profiling (based on limited data) indicate that the time for evaluation of a SurrogateOperator can be halved by the changes here. No guarantees this holds in general.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/ad/ad_utils.py
+++ b/src/porepy/numerics/ad/ad_utils.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 
 import warnings
 from abc import ABCMeta
+from functools import lru_cache
 from typing import Any, Optional
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


jwboth (approved):
As far as I understand, looks good! I can see that you considered the case that global dofs can change during the course of a simulation and emptied the cache in that situation. Whether you managed to catch all such situations in the code, I am not confident in assessing.

keileg:
> As far as I understand, looks good! I can see that you considered the case that global dofs can change during the course of a simulation and emptied the cache in that situation. Whether you managed to catch all such situations in the code, I am not confident in assessing.

Should be fine for now - I did a rather careful search. We don't do such dynamic updates of ordering, and if we find a good reason to do so, it will anyhow require a careful rethink of the treatment of dofs, and of testing. 

My preference would rather be to strip the `EquationSystem` of some of this reordering functionality, or at least verify that the current approach in practice gives us savings that warrant the extra complexity (I agree it seems reasonable in theory). But that will be another day.

PR #1634, opened 2026-04-07
Update anderson_acceleration.py

MrShuffle:
## Proposed changes

Addition of an Anderson acceleration class, which removes bad columns in the resulting least-squares problem based on the condition number of the residual matrix. It prohibits the mixing coefficients from blowing up, leading to a more robust implementation. Due to the history matrices going from a cyclic update to a shift, ensuring that the newest data is in the last column, it is written as its own class since the ordering is important when removing old history.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality).
- [ ] Other:

## Checklist

- [ x] The documentation is up-to-date.
- [ x] Static typing is included in the update.
- [x ] This PR does not duplicate existing functionality.
- [ ] The update is covered by the test suite (including tests added in the PR). No current test for Anderson acceleration exists

### DIFF
--- a/src/porepy/numerics/nonlinear/anderson_acceleration.py
+++ b/src/porepy/numerics/nonlinear/anderson_acceleration.py
@@ -25,9 +25,9 @@ class AndersonAcceleration:
         # Initialize arrays for iterates.
         self.reset()
         self._fkm1: np.ndarray = self._Fk.copy()
-        """Previous residual (or increment) used for building the Fk matrix."""
+        """Previous residual (or increment) used for building Fk."""
         self._gkm1: np.ndarray = self._Gk.copy()
-        """Previous application of the fixed point iteration used for building the Gk matrix."""
+        """Previous application of the fixed point iteration used for building Gk."""
 
     def reset(self) -> None:
         """Reset the history of Anderson acceleration."""
--- a/src/porepy/numerics/nonlinear/anderson_acceleration.py
+++ b/src/porepy/numerics/nonlinear/anderson_acceleration.py
@@ -5,29 +5,38 @@ from scipy.linalg import lstsq
 class AndersonAcceleration:
     """Anderson acceleration Algorithm 4 as described by An and Jia and Walker in
     doi.org/10.1016/j.jcp.2017.06.031.
+
+    NOTE: This code is not well tested and needs to be used with care.
+
     """
 
     def __init__(
-        self, dimension, depth, filtering: bool = True, drop_tol: float = 1e10
+        self, dimension, depth, filtering: bool = False, drop_tol: float = 1e10
     ) -> None:
         self._dimension = dimension
+        """Dimension of the algebraic problem."""
         self._depth = depth
+        """Depth of the history to be used in Anderson acceleration."""
         self._filtering = filtering
-        self._drop_tol = drop_tol  # Tolerance for condition number of Fk
+        """Whether to drop columns of Fk if the condition number is too large."""
+        self._drop_tol = drop_tol
+        """Tolerance for condition number of Fk when filtering is enabled."""
 
         # Initialize arrays for iterates.
         self.reset()
         self._fkm1: np.ndarray = self._Fk.copy()
+        """Previous residual (or increment) used for building the Fk matrix."""
         self._gkm1: np.ndarray = self._Gk.copy()
+        """Previous application of the fixed point iteration used for building the Gk matrix."""
 
     def reset(self) -> None:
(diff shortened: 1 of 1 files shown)


src/porepy/numerics/nonlinear/anderson_acceleration.py
```diff
@@ -65,3 +65,80 @@ def apply(self, gk: np.ndarray, fk: np.ndarray, iteration: int) -> np.ndarray:
         self._gkm1 = gk.copy()
 
         return x_k_plus_1
+
+
+class AndersonAccelerationFiltering:
+    """Anderson acceleration Algorithm 4 as described by An and Jia and Walker in doi.org/10.1016/j.jcp.2017.06.031.
+    """
```
jwboth: This codes reimplements many pieces already provided above. Could you try to make use of inheritance?

```suggestion
class AndersonAccelerationFiltering(AndersonAcceleration):
    """Anderson acceleration Algorithm 4 as described by An and Jia and Walker in doi.org/10.1016/j.jcp.2017.06.031.
    """
```

src/porepy/numerics/nonlinear/anderson_acceleration.py
```diff
@@ -65,3 +65,80 @@ def apply(self, gk: np.ndarray, fk: np.ndarray, iteration: int) -> np.ndarray:
...
+        self._fkm1: np.ndarray = self._Fk.copy()
+        self._gkm1: np.ndarray = self._Gk.copy()
+
+    def reset(self) -> None:
+        self._Fk: np.ndarray = np.zeros(
+            (self._dimension, self._depth)
+        )  # changes in increments
+        self._Gk: np.ndarray = np.zeros(
+            (self._dimension, self._depth)
+        )  # changes in fixed point applications
```
jwboth: To follow the standard behavior make use of `super()`. Here, and in other places.

```suggestion
          super().reset()
```

src/porepy/numerics/nonlinear/anderson_acceleration.py
jwboth: Feel free to use `reset` here, just as in your suggested code below.

src/porepy/numerics/nonlinear/anderson_acceleration.py
```diff
@@ -65,3 +65,80 @@ def apply(self, gk: np.ndarray, fk: np.ndarray, iteration: int) -> np.ndarray:
...
+        self._mk = 0 # Tracks current depth of the history
```
jwboth: Feel free to introduce `self._mk` in the `AndersonAcceleration` main class if it makes it easier.

src/porepy/numerics/nonlinear/anderson_acceleration.py
```diff
@@ -65,3 +65,80 @@ def apply(self, gk: np.ndarray, fk: np.ndarray, iteration: int) -> np.ndarray:
...
+            self._Fk[:, col] = fk - self._fkm1
+            self._Gk[:, col] = gk - self._gkm1
+            
+            # Drop the oldest columns if the condition number is too large.
+            if self._filtering:
+                while self._mk > 1 and np.linalg.cond(self._Fk[:, :self._mk]) > self._drop_tol:
+                    # Drop the oldest column (index 0) and shift the rest left
+                    self._Fk[:, :self._mk - 1] = self._Fk[:, 1:self._mk]
+                    self._Gk[:, :self._mk - 1] = self._Gk[:, 1:self._mk]
+                    self._mk -= 1
```
jwboth, line 81: Is this then the only code that is different? If yes, then there is no need for two AA classes, but a single one (the more general class) providing filtering on demand.

jwboth (commented):
Thanks @MrShuffle ! This looks like a valuable extension. I just wonder about code duplication. Lies the only actual difference in the active filtering in apply, and some additional resets? I believe, one can merge both classes, or at least make stronger use of inheritance.

src/porepy/numerics/nonlinear/anderson_acceleration.py
MrShuffle: Yes, this is the only difference. I will write this into a single general class. This will also make the other comments obsolete, so I'll resolve those now.

MrShuffle (commented):
A single Anderson acceleration class with filtering. If filtering is turned off, it defaults to previously existing code.

jwboth (approved):
Thanks @MrShuffle. I have mildly modified the changes and mainly added (already previously) missing documentation. Most importantly, I have changed the default value for the filtering to `False` to keep the original form for now.

PR #1647, opened 2026-04-14
Fix bug in differentiable Tpfa with mpfa as base discretization

### LINKED ISSUE #1556: Differentiable TPFA with MPFA as base discretization
Said combination produces weird results at the boundaries, at least if gravity is present. See below for an illustration. ATM, I have done no further investigation and, unfortunately, not produced anything close to an MRE (believe me, the simulation I ran is not minimal).

Image

keileg:
## Proposed changes
This PR fixes two minor bugs related to the differentiable two-point flux approximation. In both cases, terms related to the discretization of boundary conditions were missing, with the scheme using mpfa as base discretization was mainly suffering.

The bugfix revealed a mistaken assumption in a test related to the sparsity of Jacobian matrices on faces with Neumann boundary conditions. The tested condition is only true when using tpfa as a base discretization, though the bug had made the test pass also for mpfa. The test now only runs for tpfa.

Also added is a test that verifies that the number of Newton iterations with the differentiable scheme is not larger than the number with standard discretizations. In practice, the number of iterations is the same for most if not all of the test parametrizations. I made some attempts at finding a permeability function that gave a strict improvement with the differentiable scheme, but did not succeed. This does not imply that such a function does not exist though, so I am open for a discussion on whether to try once more. _Please consider this question in particular in the review._

Resolves #1556.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/tests/numerics/fv/test_tpfa.py
+++ b/tests/numerics/fv/test_tpfa.py
@@ -6,9 +6,11 @@ The tests fall into two categories:
 
 """
 
+from typing import Literal
 import numpy as np
 import pytest
 import scipy.sparse as sps
+from copy import deepcopy
 
 import porepy as pp
 from porepy.applications.discretizations.flux_discretization import FluxDiscretization
@@ -16,6 +18,7 @@ from porepy.applications.md_grids import model_geometries
 from porepy.applications.md_grids.model_geometries import CubeDomainOrthogonalFractures
 from porepy.applications.test_utils import common_xpfa_tests as xpfa_tests
 from porepy.applications.test_utils import well_models
+from porepy.applications.test_utils.models import add_mixin
 
 """Local utility functions."""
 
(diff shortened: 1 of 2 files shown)


tests/numerics/fv/test_tpfa.py
```diff
@@ -768,6 +772,146 @@ def test_diff_tpfa_and_standard_tpfa_give_same_linear_system(base_discr: str):
...
+@pytest.mark.parametrize("constant_permeability", [True, False])
+def test_diff_tpfa_newton_performance(
+    base_scheme: Literal["mpfa", "tpfa"],
+    gravity: bool,
+    include_fracture: bool,
+    constant_permeability: bool,
+):
+    """Verify that the solutions computed with and without diff-tpfa are consistent,
+    and that inclusion of the diffenetiable tpfa does not increase the number of
+    Newton iteration.
```
IvarStefansson:
```suggestion
    Newton iterations.
```

IvarStefansson (approved):
I don't think it's worth it atm to spend more time finding a "better" case to test.

PR #1651, opened 2026-04-16
1602 introduce tests for logging and progressbars

### LINKED ISSUE #1602: Introduce tests for logging and progressbars
Issues with progressbars have periodically popped up since the first implementation. While I hope that we discovered everything by now, this issue proposes a simple extension to the test suite to make progressbar and logging functionality future-proof. It may also act as a template to test logging more extensively in the future.

**Implementation plan:**
- [x] Set up a simple model that runs one time step with one Newton iteration
- [x] Run with logging on/off and tqdm on/off
- [x] [Capture sys.stdout](https://docs.pytest.org/en/stable/how-to/capture-stdout-stderr.html) and compare number of lines and/or to hardcoded strings

pschultzendorff:
## Proposed changes
Introduces tests checking that logging messages and progressbars are displayed according to the desired settings. Fixes a bug in the new `NewtonSolver` that showed the Newton progressbar only on the first time step. Closes #1602.

## Types of changes
- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/run_models.py
+++ b/src/porepy/models/run_models.py
@@ -17,10 +17,11 @@ import numpy as np
 
 import porepy as pp
 from porepy.models.model_runner import ModelRunner
-from porepy.utils.ui_and_logging import DummyProgressBar, progressbar_class
+from porepy.utils.ui_and_logging import DummyProgressBar
 from porepy.utils.ui_and_logging import (
     logging_redirect_tqdm_with_level as logging_redirect_tqdm,
 )
+from porepy.utils.ui_and_logging import progressbar_class
 
 # Module-wide logger
 logger = logging.getLogger(__name__)
--- a/src/porepy/models/run_models.py
+++ b/src/porepy/models/run_models.py
@@ -17,11 +17,10 @@ import numpy as np
 
 import porepy as pp
 from porepy.models.model_runner import ModelRunner
-from porepy.utils.ui_and_logging import DummyProgressBar
+from porepy.utils.ui_and_logging import DummyProgressBar, progressbar_class
 from porepy.utils.ui_and_logging import (
     logging_redirect_tqdm_with_level as logging_redirect_tqdm,
 )
-from porepy.utils.ui_and_logging import progressbar_class
 
 # Module-wide logger
 logger = logging.getLogger(__name__)
@@ -173,7 +172,7 @@ def _run_iterative_model(model, params: dict) -> None:
             )
             time_progressbar = progressbar_class(
                 range(expected_time_steps),
-                desc="time loop",
+                desc="Time loop",
                 position=0,
                 dynamic_ncols=True,
             )
@@ -183,8 +182,8 @@ def _run_iterative_model(model, params: dict) -> None:
 
         # Time loop.
         while not model.time_manager.final_time_reached():
-            time_progressbar.set_description_str(
-                f"Time step {model.time_manager.time_index + 1}"
+            time_progressbar.set_postfix_str(
+                f"Time step size {model.time_manager.dt:.2e}"
             )
             time_step()
             # Update progressbar length. Currently, there is no convergence check
--- a/tests/numerics/nonlinear/test_nonlinear_solvers.py
+++ b/tests/numerics/nonlinear/test_nonlinear_solvers.py
@@ -16,6 +16,7 @@ from porepy.numerics.nonlinear.convergence_check import (
     ConvergenceStatusCollection,
     SimulationStatus,
 )
+from porepy.utils.ui_and_logging import DummyProgressBar
 
 # ! ---- Auxiliary fixtures and classes ---- ! #
 
@@ -917,6 +918,10 @@ def test_after_nonlinear_loop(
     model = MockModel()
     solver = default_newton_solver
 
+    # Mock the solver progressbar. Usually it is initialized in
+    # NewtonSolver.before_nonlinear_loop, which is never called in this test.
+    solver.solver_progressbar = DummyProgressBar()
+
     # Minimal mimicking of loop.
     model.nonlinear_solver_statistics.simulation_status_history = [
         SimulationStatus.SUCCESSFUL
@@ -935,6 +940,10 @@ def test_before_nonlinear_iteration(default_newton_solver):
     # Init model and solver.
     model = MockModel(nonlinear_increment_history=[2.0], residual_history=[1.0])
     solver = default_newton_solver
+
+    # Mock the solver progressbar. Usually it is initialized in
+    # NewtonSolver.before_nonlinear_loop, which is never called in this test.
+    solver.solver_progressbar = DummyProgressBar()
 
     # Check initial iteration index.
     assert solver.iteration_index == 0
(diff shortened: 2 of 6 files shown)


tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -917,6 +918,10 @@ def test_after_nonlinear_loop(
     model = MockModel()
     solver = default_newton_solver
 
+    # Mock the solver progressbar. Usually it is initialized in
+    # NewtonSolver.before_nonlinear_loop, which is never called in this test.
+    solver.solver_progressbar = DummyProgressBar()
+
```
jwboth, line 924: What is the point of including the progress bar in these tests? I do not see specific tests addressing the progress bar. Or is there an implicit testing as the progress bar is performing something during after_nonlinear_loop? (and similar for the other tests where a solver_progressbar is added.

tests/utils/test_ui_and_logging.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def before_nonlinear_iteration(self) -> None:
+        # IMPLEMENTATION NOTE Write logging messages of all levels. This way we can
+        # test in test_logging_and_progressbars that
+        # pp.utils.ui_and_logging.logging_redirect_tqdm_with_level correctly redirects
+        # messages of all levels through tqdm.
+        mock_logger.debug(f"Starting Newton step")
+        mock_logger.info(f"Starting Newton step")
+        mock_logger.warning(f"Starting Newton step")
+        mock_logger.error(f"Starting Newton step")
```
jwboth, line 83: I appreciate the effort made in testing logging. Yet, do I understand it correct? In essence, it is a test of the mock logger and thus only checks whether logging as a module changes its behavior and the compatibility of logging and tqdm. It does not test any PorePy related functionality or logging specifically within PorePy.

If so, I would conclude to remove the logging related tests.

tests/utils/test_ui_and_logging.py
```diff
@@ -0,0 +1,234 @@
...
+    # Count the number of "Starting Newton step" messages originating from
+    # MockModel.before_nonlinear_iteration of each level.
+    num_before_newton_step_debug_records: int = 0
+    num_before_newton_step_info_records: int = 0
+    num_before_newton_step_warning_records: int = 0
+    num_before_newton_step_error_records: int = 0
+
+    captured_logging_records = caplog.records
+    for record in captured_logging_records:
+        if record.msg == "Starting Newton step":
```
jwboth: To complement the above comment: Here, it looks to me simply that the logging is captured and checked whether the expected logging messages defined in the MockModel (not anywhere else in the core) are put through.

src/porepy/models/run_models.py
```diff
@@ -243,8 +243,8 @@ def time_step() -> None:
 
         # Time loop.
         while not model.time_manager.final_time_reached():
-            time_progressbar.set_description_str(
-                f"Time step {model.time_manager.time_index + 1}"
+            time_progressbar.set_postfix_str(
```
jwboth: Make sure that the updates are correctly migrated to the newly added ModelRunner.

jwboth (commented):
Thanks, @pschultzendorff! As you will see, recently the ModelRunner has been merged into develop. Can you make sure to resolve the conflicts? Two comments regarding the tests. Let me know if you want to discuss.

tests/utils/test_ui_and_logging.py
pschultzendorff: That's correct. This test just checks that logging messages are correctly passed, i.e., it tests that `logging_redirect_tqdm_with_level` works correctly. If this is considered unnecessary, I can remove it.

tests/numerics/nonlinear/test_nonlinear_solvers.py
pschultzendorff: `solver.solver_progressbar` is assumed to exist in all tested functions where I added these lines. However, because I moved `init_progressbar` from `solver.__init__` to `solver.before_nonlinear_loop`, it doesn't in these tests. Thus, I add a mock object.

tests/utils/test_ui_and_logging.py
pschultzendorff: See above

tests/numerics/nonlinear/test_nonlinear_solvers.py
pschultzendorff: See what I wrote on Teams two weeks ago for a more thorough explanation.

pschultzendorff:
Should be mergable now. Let me know how I shall handle your comments :)

tests/numerics/nonlinear/test_nonlinear_solvers.py
jwboth: Thanks for the explanation. Makes sense now.

tests/utils/test_ui_and_logging.py
jwboth: Ah, now I see. I also took a look at `logging_redirect_tqdm_with_level`. Initially, I thought you are purely testing compatibility of tqdm and logging, which should be rather tested in tqdm than PorePy. But you are essentially testing an own implementation of enhancing tqdm with logging levels. Then I think it is fine. But I would suggest you split the single test in two. Then it is much cleaner in terms of unit testing and once a test fails, it is also clearer what part breaks. What do you think?

src/porepy/models/run_models.py
jwboth: Thanks for updating to main. I think you need to mildly upgrade your MockModel and then the tests will also go through.

tests/utils/test_ui_and_logging.py
jwboth: And if you could add a short docstring for each of the tests describing what is tested (minimally more concrete than "works") and when it is expected to fail, this would be super!

pschultzendorff: I tried splitting, but then the fixtures have to run twice because `capsys` and `caplog` are function-scoped fixtures.

I could skip the cases "progressbars: off; logging: off" and "progressbars: on; logging: off" for the logging test. On the other hand, I would like to keep "progressbars: on; logging: off" and "progressbars: on; logging: on" for the progressbar test to check that `logging_redirect_tqdm_with_level` works both on the logging and on the progressbar side.

That would leave us with 6 mock model runs instead of the current 4.

src/porepy/models/run_models.py
pschultzendorff: Just noticed myself :D Should be fixed.

tests/utils/test_ui_and_logging.py
pschultzendorff: I updated the docstring

pschultzendorff: Just pushed an updated version with two separate tests

jwboth: Thanks. Also for the docstrings explaining all tests and notes. I think this is useful!

jwboth (approved):
Thanks, Peter!

PR #1670, opened 2026-05-06
Refactor solver status

jwboth:
## Proposed changes

Minor clean up on the solver level:
- Consistent naming of solver_status (not simulation_status which should be handled one level above). NB: inconsistency remains at statistics level and requires a clearer distinction between solver and simulation status (suggested for future).
- Cleaner method design with separate responsibility for providing the solver status as a result of the convergence status.
- (Attempt of) identical logics and structure in linear and nonlinear solver.

Not done:
- No further tests were added (one has been renamed after introducing more granularity but essentially testing the same) - this PR does not introduce new functionality.
- Clean distinction between solver and simulation status (in particular no new classes). Some NOTE and TODO comments have been modified to indicate the status.
- Further processing on the time stepping level.
- Further streamlining with a common mother class for LinearSolver and NonlinearSolver exploiting the similar structure.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/linear_solvers.py
+++ b/src/porepy/numerics/linear_solvers.py
@@ -8,6 +8,7 @@ case, see numerics.nonlinear.nonlinear_solvers.
 
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
 from warnings import warn
 
 from porepy.models.solution_strategy import SolutionStrategy
@@ -21,6 +22,9 @@ from porepy.numerics.nonlinear.convergence_check import (
     SimulationStatus,
 )
 from porepy.viz.solver_statistics import TimeStatistics
+
+if TYPE_CHECKING:
+    import numpy as np
 
 
 class LinearSolver:
@@ -115,10 +119,10 @@ class LinearSolver:
 
         Returns:
             Tuple containing:
-            - ConvergenceStatusCollection: Convergence status collection at the end of the
-              nonlinear iteration.
-            - ConvergenceStatusCollection: Divergence status collection at the end of the
-              nonlinear iteration.
+            - ConvergenceStatusCollection: Convergence status collection at the end of
+                the nonlinear iteration.
+            - ConvergenceStatusCollection: Divergence status collection at the end of
+                the nonlinear iteration.
 
         """
         # Combines "before, actual and after" iteration steps.
@@ -209,12 +213,12 @@ class LinearSolver:
         Returns:
             Tuple containing:
 
-            - ConvergenceStatusCollection: Convergence status collection at the end of the
-              nonlinear iteration.
-            - ConvergenceStatusCollection: Divergence status collection at the end of the
-              nonlinear iteration.
-            - ConvergenceInfoCollection: Convergence information collection at the end of
-              the nonlinear iteration.
+            - ConvergenceStatusCollection: Convergence status collection at the end of
+                the nonlinear iteration.
+            - ConvergenceStatusCollection: Divergence status collection at the end of
+                the nonlinear iteration.
+            - ConvergenceInfoCollection: Convergence information collection at the end
+                of the nonlinear iteration.
 
         """
         # Fetch the residual.
--- a/src/porepy/numerics/linear_solvers.py
+++ b/src/porepy/numerics/linear_solvers.py
@@ -70,6 +70,31 @@ class LinearSolver:
             SimulationStatus: The status of the simulation.
 
         """
+        # Prepare solving.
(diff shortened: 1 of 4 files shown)


src/porepy/numerics/linear_solvers.py
```diff
@@ -70,53 +74,151 @@ def solve(self, model: SolutionStrategy) -> SimulationStatus:
...
+        Parameters:
+            model: The model instance specifying the problem to be solved.
+            convergence_status: Convergence statuses.
+            divergence_status: Divergence statuses.
+
+        Returns:
+            SimulationStatus: The overall status of the nonlinear solver.
+
+        """
+        if convergence_status.is_converged():
```
keileg, line 157: Is there a risk we have convergence and divergence decleared at the same time?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -308,44 +311,64 @@ def nonlinear_loop(
...
+                # NOTE: FAILED on solver level, and FAILED on simulation level.
+                # Should possible be handled on the level above the solver.
+                # NOTE: Currently, if a simulation fully stops, this is not logged in
+                # SolverStatistics. For this, better coordination between solver and
+                # time stepping is needed.
+                # TODO: Get back to this when reimplementing time stepping.
+
+                # Declare total failure which shall result in stopping the simulation.
+                warn(
+                    "Failed to solve linear system for the linear problem. "
```
keileg: There should be at least one non-linear here, right?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -541,7 +566,7 @@ def update_solver_statistics(
             model.nonlinear_solver_statistics.log_convergence_info(convergence_info)
 
         # Basic discretization-related information and overall simulation status.
-        if simulation_status is not None:
+        if solver_status is not None:
             pp.LinearSolver.update_solver_statistics(
-                cast(pp.LinearSolver, self), model, simulation_status
+                cast(pp.LinearSolver, self), model, solver_status
```
keileg, line 571: For reference,  I really don't like such calling methods of different classes, but let's not fix that now.

keileg (approved):
Minor questions only. Feel free to merge when you feel done, or let me know if you want to discuss something.

keileg (approved):
I posted some minor questions. Feel free to merge when you feel you are finished.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
jwboth: OK! This slipped through before unnoticed... I think a solver base class would solve that issue. With the almost parallel logic in LinearSolver and NewtonSolver, one should be able to streamline a bit more here, while also introducing an API.

src/porepy/numerics/linear_solvers.py
jwboth: By design yes (and I was not happy about this yesterday when already also happily merging convergence_status and divergence_status here such that the test would be simply convergence_status.is_converged() or convergence_status.is_diverged() - I reverted these changes again but would be more than happy to have a simpler structure here).

The problem is: The test for maximum number of iterations is designed as a DivergenceCriterion. To allow for the cases that the NewtonSolver just converges in the last feasible iteration. Since all criteria are evaluated after computing the iterate and to omit one unnecessary extra iteration, the check displays divergence to indicate the fastest possible exit out of the loop without extra cost.
(and I think this was the only relevant case of simultaneous convergence and divergence I could come up with that required something like this)

I see conceptually three ways out:
- Do not allow such overlap but accept the extra computational cost (this would require an update of the MaxIterationsCriterion and corresponding test)
- Actually, not using criteria to control max number of iterations (later also attempts to retry adaptive time cutting etc) but use a classical `for iter in range(max_iter)` structure instead of `while True: ... stop if is_diverged()`. (this would require additional booking, different initialization in Newton, mild stripping down of divergence criteria and tests.) If going for this solution, one could also leverage a `for: ... else: ....` structure.
- Accept this as is.

jwboth: left as is for now, but should be revisited, and on my radar.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
jwboth: yes, thanks. fixed.

PR #1676, opened 2026-05-14
Update helper method to create 1d grids (not necessarily aligned with the x-axis) 

keileg:
## Proposed changes
Simplified the implementation of the function `create_embedded_line_grid`, which generates a 1d grid from a set of points. The old implementation forced a sorting of the nodes, and thereby required that the nodes were collinear. The new implementation keeps the option to sort (and the requirement of collinearity), but also allows for points that are not aligned (to the price of no sorting). This allows for a future application to multilinear grids, e.g. along multi-segment wells.

Also introduced a test of the helper function and took the function into use in other parts of the code. For the latter, I tried to make judgements on where this gave a real improvement, and where the existing code was best left as is.

**Question for review:** Neither the name of the helper function nor its placement is ideal. My best suggestion for name is `create_1d_grid_from_nodes`, but I am open for suggestions. Placement: Perhaps move to `grids/_structured.py`?

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/fracs/msh_2_grid.py
+++ b/src/porepy/fracs/msh_2_grid.py
@@ -457,41 +457,38 @@ def create_0d_grids(
 
 
 def create_embedded_line_grid(
-    loc_coord: np.ndarray, glob_id: np.ndarray, tol: float = 1e-4
+    loc_coord: np.ndarray,
+    glob_id: np.ndarray | None = None,
+    sort: bool = True,
+    tol: float = 1e-4,
 ) -> pp.Grid:
     """Create a 1d grid embedded in a higher dimensional space.
 
     Parameters:
         loc_coord: Coordinates of points to be used in the grid.
         glob_id : Global indexes of the points. Typically refers to a global mesh, where
-            the points of this grid is a subset.
+            the points of this grid is a subset. If provided, the returned grid will
+            have an attribute ``global_point_ind`` that maps from local to global point
+            numbering.
+        sort: Whether to sort the points. The sorting assumes the points are colinear.
         tol: Tolerance used for check of collinearity of the points.
 
     Returns:
         g: 1d grid.
 
     """
-    loc_center = np.mean(loc_coord, axis=1).reshape((-1, 1))
-    (
-        sorted_coord,
-        rot,
-        active_dimension,
-        sort_ind,
-    ) = pp.map_geometry.project_points_to_line(loc_coord, tol)
-    g = pp.TensorGrid(sorted_coord)
+    if sort:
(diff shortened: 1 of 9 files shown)


IvarStefansson:
Regarding multi-segment wells: The current implementation uses non-collinear 1d grids, I think. That is what causes inconsistent discretization, as far as I understand. So let's consider very carefully before going for that option instead of splitting into one grid for each linear segment.

keileg:
> Regarding multi-segment wells: The current implementation uses non-collinear 1d grids, I think. That is what causes inconsistent discretization, as far as I understand. So let's consider very carefully before going for that option instead of splitting into one grid for each linear segment.

Though I see the point, I would very much like to avoid the complexities related to indexing. I'll keep it in mind, though.

src/porepy/fracs/msh_2_grid.py
```diff
@@ -457,41 +457,38 @@ def create_0d_grids(
...
     """Create a 1d grid embedded in a higher dimensional space.
 
     Parameters:
         loc_coord: Coordinates of points to be used in the grid.
         glob_id : Global indexes of the points. Typically refers to a global mesh, where
-            the points of this grid is a subset.
+            the points of this grid is a subset. If provided, the returned grid will
+            have an attribute ``global_point_ind`` that maps from local to global point
+            numbering.
+        sort: Whether to sort the points. The sorting assumes the points are colinear.
```
IvarStefansson: Let's decide on one or two l-s in collinear. Seems both are accepted.

tests/fracs/test_msh_2_grid.py
```diff
@@ -352,3 +353,42 @@ def test_create_grids_with_high_dim_inclusion(
...
+    Parameters:
+        nodes: The nodes to be used in the grid. Should be of shape (3, num_nodes).
+        is_linear: Whether the points are colinear.
+        shuffle: Whether to shuffle the order of the points.
+
+    """
+    if shuffle:
+        nodes = nodes[:, [2, 0, 1]]
+
+    # Only shuffle if the points are linear.
```
IvarStefansson: This seems to be a comment on the parametrization, not on the logic of the test itself. Move?

IvarStefansson (approved):
Looks good to me, although I maintain that this should be used with care.

src/porepy/fracs/msh_2_grid.py
keileg: OED says collinear, so we'll go with that. I'll update throughout the code.

tests/fracs/test_msh_2_grid.py
keileg: Should be only sort.

Issue #1657, opened 2026-04-23
Failed simulations should throw error


IvarStefansson:
Non-convergence with pp.run_time_dependent_model currently raises a warning instead of an error. Thus, the user may fail to notice that no (final) solution was computed. The suggested fix is to catch this by inspecting `SimulationStatus` at the end of simulation.

Warning:
`/workdir/porepy/src/porepy/models/solution_strategy.py:682: UserWarning: Failed to solve the nonlinear problem.  warn("Failed to solve the nonlinear problem.")`

PR #1682, opened 2026-05-27
Trial time

IvarStefansson:
## Proposed changes

## Types of changes

- [ ] Other:

## Checklist


Yuriyzabegaev:
I've ran through the changes and can say, it's a solid step in the right direction! Some thoughts here.

I like how we distinguish between statuses `ACCEPTED`, `REJECTED`, and `STOPPED`, but I feel like we need to document what the latter is used for. REJECTED applies to both cases of divergence and iteration limit. Was the intention of STOPPED to indicate a critical failure, when we should not attempt to decrease the time step and retry? E.g., when both state and iterate arrays somehow became nans. Do we even need to consider this situation?

The old `TimeManager` now has 2 responsibilities: (i) to be a configuration data-structure, (ii) to compute the new time step based on feedback. Could we go even further and decouple it into 2 objects? Also, it's a part of the model, but it's probably natural to move it to be a part of the time stepper?.

The time stepper class is composed of 3 components. `TimeStepAcceptanceCriteria`, `TimeStepRejectionCriteria` and `TimeManager` (logically). What I think would be very nice to have is the ability to customize these components individually via dependency injection, i.e., the user passes their own `TimeStepAcceptanceCriteria` in the model runner parameters, and they are passed to `TimeManager.__init__`. This would make it flexible and testable. The same applies to the `TimeStepper` itself: make the `TimeStepperFactory` provide a custom instance of a `TimeStepper`, if the user passes it.

So to say, I really like where this project is going, good job @jwboth! Will add a few minor notes in the code below.

src/porepy/time/time_stepper.py
```diff
@@ -0,0 +1,446 @@
...
+            rejection_status: Rejection statuses.
+            info: Diagnostic information.
+        """
+        # TODO: Need wholistic upgrade - check out SolverStatistics.
+        # if acceptance_status is not None:
+        #     self.acceptance_status_history.append(acceptance_status)
+        # if rejection_status is not None:
+        #     self.rejection_status_history.append(rejection_status)
+        # if info is not None:
+        #     self.info_history.append(info)
```
Yuriyzabegaev: I see it's not done yet, but would it be easier to have a single timeline of acceptance / rejection statuses? Imagine you have rejection_status_history with two entries: time step 1 and time step 400, and acceptance_status_history with every other time step, and good luck transforming this data to some dashboard for analysis

src/porepy/time/time_step_acceptance.py
```diff
@@ -0,0 +1,381 @@
...
+
+        return TimeStepStatus.ACCEPTED
+
+
+# ============================================================================
+# Criterion Collections
+# ============================================================================
+
+
+class TimeStepAcceptanceCriteria(dict[str, TimeStepAcceptanceCriterion]):
```
Yuriyzabegaev: Regarding the dependency injection idea from my comment above. If we go there, it's probably better keep a collection of TimeStepAcceptanceCriterion (that a user passes in parameters) just a dict, and not its subclass. And the check method over the collection can be a part of a non-polymorphic wrapper class, or just a function.

src/porepy/time/time_stepper.py
```diff
@@ -0,0 +1,446 @@
...
+        time_step_status = self.perform_trial_time_step(model, solver)
+
+        # Advance time if accepted
+        if time_step_status == TimeStepStatus.ACCEPTED:
+            logger.debug(f"Time advanced to {self.time_manager.time}")
+        else:
+            logger.warning("Trial rejected in direct time stepper; stopping.")
+            time_step_status = TimeStepStatus.STOPPED
+
+        return time_step_status
```
Yuriyzabegaev: DirectTimeStepper does not roll back the time if the simulation fails. Consider we have a schedule [0, 1] and dt=1. We make a time step and fail. Now the simulation time is 1, and `final_time_reached` will return True

jwboth: This is true for both time steppers. I tried thinking about it. What would be the main reason to roll back the time? See my other answer regarding STOPPED status. The simulation will be aborted no mater what.

src/porepy/time/time_stepper.py
jwboth: Yes, should have not been here. Will focus on the status logging as next step.

jwboth:
Thanks @Yuriyzabegaev for having a look. I was not aware of the fact that this code snippet was ready :) The idea here was just to provide the orchestration skeleton underlining the integration of the time manager.

> I like how we distinguish between statuses ACCEPTED, REJECTED, and STOPPED, but I feel like we need to document what the latter is used for. REJECTED applies to both cases of divergence and iteration limit. Was the intention of STOPPED to indicate a critical failure, when we should not attempt to decrease the time step and retry? E.g., when both state and iterate arrays somehow became nans. Do we even need to consider this situation?

STOPPED indeed associates with critical failure. The idea has been to use the same style as in NewtonSolver. It will be possible to distinguish which criterion failed from an info object. But I may want to discuss it, as the current solution may be too verbose (it stores essentially everything).

> The old TimeManager now has 2 responsibilities: (i) to be a configuration data-structure, (ii) to compute the new time step based on feedback. Could we go even further and decouple it into 2 objects? Also, it's a part of the model, but it's probably natural to move it to be a part of the time stepper?.

I all pro chopping it up into smaller pieces. Before doing that, one has to decide how time steps are cut. Should there be always 1 rule or allow we for multiple rules? Or will this be the responsibility/design choice of a class that shall suggest time step sizes? The model does only need to know time and dt, nothing more. 

> The time stepper class is composed of 3 components. TimeStepAcceptanceCriteria, TimeStepRejectionCriteria and TimeManager (logically). What I think would be very nice to have is the ability to customize these components individually via dependency injection, i.e., the user passes their own TimeStepAcceptanceCriteria in the model runner parameters, and they are passed to TimeManager.__init__. This would make it flexible and testable. The same applies to the TimeStepper itself: make the TimeStepperFactory provide a custom instance of a TimeStepper, if the user passes it.

Fully agreed. We can talk about it. Current idea was to end up with a similar design as for NewtonSolver, where it is also possible to customize criteria from solver_params.

I will continue once #1684 is merged as it also touched the model runner - the main idea was actually to throw away this branch... but now its here.

src/porepy/models/model_runner.py
```diff
@@ -153,6 +158,14 @@ def set_solver(self) -> None:
         else:
             self.solver = pp.LinearSolver(self.params)
 
+    def set_time_stepper(self) -> None:
+        """Set the time stepper for time-dependent problems."""
+        self.time_stepper = TimeStepperFactory.create_time_stepper(
+            self.model.time_manager,
+            params=self.params,
```
IvarStefansson: I suggest wrapping time stepper params in a separate dictionary, self.params.get("time_step_parameters", None), say.

src/porepy/models/model_runner.py
```diff
@@ -211,162 +221,87 @@ def init_time_progressbar(self) -> None:
...
-            + f" with time step {self.model.time_manager.dt:.1e}"
-        )
-        self.time_progressbar.set_postfix_str(
-            f"Time step size {self.model.time_manager.dt:.2e}"
-        )
-
-    def after_time_step(self, solver_status: SimulationStatus) -> SimulationStatus:
-        """Method to be executed at the end of each time step.
+        # Conclude.
+        if simulation_status.is_failed():
```
IvarStefansson: I'm not sure it's obvious to a user (or me, at the time of writing) what the difference between failed and stopped is.

src/porepy/time/time_step_acceptance.py
```diff
@@ -0,0 +1,384 @@
...
+# ============================================================================
+# Evaluation Context
+# ============================================================================
+
+
+@dataclass
+class TimeStepEvaluationContext:
+    """Unified context for time-step acceptance/rejection evaluation.
+
+    This is the **single source of truth** for all data available to criteria.
```
IvarStefansson: What is gained by introducing this with model as one of the fields? Currently, model contains "everything", doesn't it? Why not, then, let model be the context and access model.nonlinear_solver_statistics etc.?

src/porepy/time/time_step_acceptance.py
```diff
@@ -0,0 +1,384 @@
...
+            tuple: (status, diagnostic_info)
+        """
+        pass
+
+    def reset(self) -> None:
+        """Reset any internal state (e.g., reference values)."""
+        pass
+
+    @property
+    def description(self) -> str:
```
IvarStefansson: How does description relate to repr and str?

src/porepy/time/time_step_acceptance.py
```diff
@@ -0,0 +1,384 @@
...
+
+    @abstractmethod
+    def check(self, context: TimeStepEvaluationContext) -> TimeStepStatus:
+        """Check the rejection criterion.
+
+        Parameters:
+            context: TimeStepEvaluationContext with all available data.
+
+        Returns:
+            TimeStepStatus: ACCEPTED or REJECTED.
```
IvarStefansson: Should a rejection criterion return ACCEPTED?

src/porepy/time/time_step_acceptance.py
```diff
@@ -0,0 +1,384 @@
...
+    """
+
+    @abstractmethod
+    def check(self, context: TimeStepEvaluationContext) -> TimeStepStatus:
+        """Check the rejection criterion.
+
+        Parameters:
+            context: TimeStepEvaluationContext with all available data.
+
+        Returns:
```
IvarStefansson: Why no info return here when it is returned in AcceptanceCriterion?

src/porepy/time/time_step_acceptance.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+        is_accepted = context.solver_status == SimulationStatus.SUCCESSFUL
+        status = TimeStepStatus.ACCEPTED if is_accepted else TimeStepStatus.REJECTED
+        info = f"Solver status: {context.solver_status}"
+
+        return status, info
+
+
+class IterationCountCriterion(TimeStepAcceptanceCriterion):
+    """Accept if iteration count is within acceptable range.
```
IvarStefansson: When would you want to throw away a solution even if/because it was expensive?

src/porepy/time/time_step_acceptance.py
```diff
@@ -0,0 +1,384 @@
...
+    Parameters:
+        min_iterations: Minimum iterations (advisory).
+        max_iterations: Maximum iterations (rejection threshold).
+    """
+
+    def __init__(self, min_iterations: int = 2, max_iterations: int = 15) -> None:
+        self.min_iterations = min_iterations
+        self.max_iterations = max_iterations
+
+    def check(
```
IvarStefansson: I don't get why we need separate Acceptance and Rejection for this.

src/porepy/time/time_step_status.py
```diff
@@ -0,0 +1,127 @@
+"""Collection of objects and functions related to time-step acceptance checking.
+
+This includes:
+- Status classes for time-step acceptance.
+- Information classes for time-step checks.
+- Base time-step criterion classes.
+- Concrete acceptance and rejection criteria.
```
IvarStefansson: Update

src/porepy/time/time_stepper.py
```diff
@@ -0,0 +1,423 @@
...
+
+    def perform_trial_time_step(
+        self,
+        model,  #: pp.SolutionStrategy,
+        solver,  #: pp.LinearSolver | pp.NewtonSolver,
+    ) -> TimeStepStatus:
+        """Perform a single time step and evaluate acceptance/rejection criteria."""
+
+        # Execute trial time step.
+        model.before_time_step()
```
IvarStefansson: It seems these methods may be called multiple times during a time step. Are they more precisely thought of as "before trial time step" etc.?

src/porepy/time/time_stepper.py
```diff
@@ -0,0 +1,423 @@
...
+            logger.debug("Trial rejected by rejection criteria.")
+        else:
+            raise ValueError(
+                f"Invalid time step acceptance status {acceptance_status} and "
+                f"rejection status {rejection_status}."
+            )
+
+        return time_step_status
+
+    def after_time_step(self, time_step_status: TimeStepStatus, model) -> None:
```
IvarStefansson: I like this logic being contained here. Could there be any reason to mirror it in the before_time_step case?

src/porepy/time/time_stepper.py
```diff
@@ -0,0 +1,423 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+# ============================================================================
+# Base Time Stepper
+# ============================================================================
+
+
+class TimeStepper(ABC):
```
IvarStefansson: If we were to think of anything other than implicit Euler, would the implementation localize to the model? I'm not convinced that's the optimal design. I don't suggest we do a major change now but want to avoid making stupid design choices in this iteration.

IvarStefansson (commented):
I like the design. I think we should leave what is left in the TimeManager for this PR as far as possible. 

Let's meet and discuss how to proceed, @Yuriyzabegaev.

src/porepy/time/time_stepper.py
IvarStefansson: before nonlinear loop

src/porepy/time/time_stepper.py
```diff
@@ -0,0 +1,423 @@
...
+        """
+        super().__init__(time_manager, params)
+        self.max_attempts = time_manager.recomp_max
+        """Maximum number of retry attempts."""
+
+        # Sanity checks.
+        assert not self.time_manager.is_constant, (
+            "AdaptiveTimeStepper should not be used with constant time step size."
+        )
+        assert self.max_attempts > 0, "max_attempts must be greater than 0."
```
IvarStefansson, line 69: If we permit max = 0, we can get rid of the Direct/Constant Time stepper.

Yuriyzabegaev: did this

src/porepy/models/model_runner.py
Yuriyzabegaev: So far there are no particular parameters related to the time stepper. All information is passed through the time manager, so I removed params argument from the time stepper.

src/porepy/time/time_step_acceptance.py
Yuriyzabegaev: All time step acceptance logic is removed from the time stepper. (You accept a time step if the nonlinear solver is successful).

Yuriyzabegaev: Due to this I resolve all the comments below related to time step acceptance.

src/porepy/models/model_runner.py
```diff
@@ -4,57 +4,57 @@
...
+]
 
 # Module-wide logger
 logger = logging.getLogger(__name__)
 
 
-class SimulationStatus(StrEnum):
-    """Enumeration of potential simulation statuses."""
+@dataclass
+class ModelRunnerStatus(ABC):
```
Yuriyzabegaev, line 38: The ModelRunnerStatus is a replacement for the SimulationStatus. It has only two states: Success (simulation completed successfully) and Failure (simulation did not complete successfuly).

I use dataclasses for enums to associate each state with data. This potentially allows for extension in the data you return, and I also see it as a step towards more explicit statistics (where all statistics is explicitly returned in such object). This is not introduced in this pr though.

src/porepy/models/model_runner.py
```diff
@@ -246,177 +256,73 @@ def init_time_progressbar(self) -> None:
...
-                    simulation_status = SimulationStatus.STOPPED
-                    self.logging(simulation_status)
-
-        else:
-            raise ValueError(f"Unrecognized solver stats {solver_status}.")
-
-        if simulation_status.is_stopped():
-            logger.warning("Simulation stopped.")
-            raise RuntimeError("Simulation stopped.")
+        if simulation_status.is_failure():
```
Yuriyzabegaev, line 284: Recent issue #1657 asked to raise on failure, so I obey.

src/porepy/models/model_runner.py
```diff
@@ -246,177 +256,73 @@ def init_time_progressbar(self) -> None:
...
+            f"Time step size {self.model.time_manager.dt:.2e}"
         )
-        if self._is_time_dependent:
-            assert isinstance(self.model.nonlinear_solver_statistics, pp.TimeStatistics)
-            self.model.nonlinear_solver_statistics.log_time_information(
-                self.model.time_manager.time_index,
-                self.model.time_manager.time,
-                self.model.time_manager.dt,
-                self.model.time_manager.final_time_reached(),
+        self.time_progressbar.update(
```
Yuriyzabegaev: Comment to myself: ChatGPT screams that this is a mistake since tqdm.update must accept an increment. Need to double-check how it was before this PR.

src/porepy/models/protocol.py
```diff
@@ -666,6 +666,9 @@ def _is_reference_component_eliminated(self) -> bool:
             """Returns True if ``params['eliminate_reference_component'] == True`.
             Defaults to True."""
 
+        def before_time_step(self) -> None:
```
Yuriyzabegaev, line 669: Updated the protocol with recently introduced methods.

src/porepy/numerics/nonlinear/convergence_check.py
```diff
@@ -22,47 +22,18 @@
 # Status and info classes
 
 
-class SolverStatus(StrEnum):
```
Yuriyzabegaev, line 25: SolverStatus is gone. It is replaced with separate TimeStepperStatus and NonlinearSolverStatus

src/porepy/numerics/nonlinear/convergence_check.py
```diff
@@ -22,47 +22,18 @@
...
     """Enumeration of potential convergence statuses."""
 
     CONVERGED = "converged"
-    """Convergence criterion is satisfied / Divergence criterion is not satisfied."""
-    NOT_CONVERGED = "not_converged"
-    """Convergence criterion is not satisfied."""
-    DIVERGED = "diverged"
-    """Divergence criterion is satisfied."""
+    """Convergence criterion is satisfied."""
+    CONTINUE_ITERATING = "continue_iterating"
```
Yuriyzabegaev, line 30: Renamed NOT_CONVERGED to CONTINUE_ITERATING and DIVERGED to FAILED after discussion with @IvarStefansson. 

Also changed the behavior of divergence criteria.
Previously:
* if not diverged -> return CONVERGED
Now:
* if not diverged -> return CONTINUE_ITERATING

I think this is less confusing. Let's discuss.

src/porepy/numerics/nonlinear/nonlinear_solver_status.py
```diff
@@ -0,0 +1,43 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+
+from porepy.numerics.nonlinear.convergence_check import ConvergenceStatusCollection
+
+
+@dataclass
+class NonlinearSolverStatus(ABC):
```
Yuriyzabegaev, line 8: Introduced NonlinearSolverStatus instead of SolverStatus. It now has associated data: convergence_statuses and divergence_statuses, so when you pass it around, you can explicitly see, why a verdict was rendered. Note that I don't put these data fields to the base class on purpose, it is purely abstract.

After our discussion with @keileg and @IvarStefansson we agreed that now NonlinearSolverStatus and TimeStepperStatus serve two purposes: (i) return codes and (ii) statistics logging. For the latter, the method "serialize" provides identical behavior to what was logged before this PR. I didn't want to make it `__str__`, because it does not include the fields data.

src/porepy/numerics/linear_solvers.py
```diff
@@ -87,11 +93,58 @@ def solve(self, model: SolutionStrategy) -> SolverStatus:
...
+
+        Parameters:
+            convergence_status: Multiple convergence statuses from different criteria.
+            divergence_status: Multiple divergence statuses from variaous criteria.
+
+        Returns:
+            NonlinearSolverStatus: Either Converged or Failed.
+
+        """
+        # YZ: This is a duplicated method of NewtonSolver, but I find it acceptible for
```
Yuriyzabegaev: Note the comment. Is it ok?

src/porepy/numerics/linear_solvers.py
```diff
@@ -223,25 +240,35 @@ def check_convergence(
         """
         # Fetch the residual.
         residual = model.equation_system.assemble(evaluate_jacobian=False)
+        iterate = model.equation_system.get_variable_values(iterate_index=0)
 
         # Check convergence status based on current iteration.
         convergence_status, convergence_info = self.convergence_criteria.check(
             increment=nonlinear_increment,
             residual=residual,
+            # Reference values don't make sense for a linear solver, but passed to
```
Yuriyzabegaev, line 209: Added these for completeness, because I had to use the MaxIterationDivergenceCriterion in combination with the LinearSolver.

src/porepy/time_stepper/time_step_control.py
```diff
@@ -563,64 +563,58 @@ def _adaptation_based_on_recomputation(self) -> None:
...
+        # by self.correction_based_on_dt_min() in a subsequent correction step).
+        # Thus, to avoid pointless iterations, we raise an error.
+        if self.dt == self.dt_min_max[0]:
+            msg = (
+                "Recomputation will not have any effect since the time step "
+                f"achieved its minimum admissible value -> dt = dt_min = {self.dt}."
+            )
+            raise ValueError(msg)
 
-        if self._recomp_num < self.recomp_max:
```
Yuriyzabegaev, line 570: I removed this condition, because now it's always true. TimeManager does not track `_recomp_num` no longer.

src/porepy/time_stepper/time_step_control.py
```diff
@@ -644,6 +638,12 @@ def _correction_based_on_dt_max(self) -> None:
 
     def _correction_based_on_schedule(self) -> None:
         """Correct time step if time + dt > scheduled_time."""
+        # When moving this to the TimeStepper, we should make this more efficient and
```
Yuriyzabegaev, line 625: Some new comments left here by @jwboth and me to keep in mind when refactoring this.

src/porepy/time_stepper/time_step_status.py
```diff
@@ -0,0 +1,82 @@
+"""Collection of objects related to time-step acceptance checking.
+
+Note: Statuses (TimeStepperStatus, NonlinearSolverStatus etc.) serve two purposes: (i)
```
Yuriyzabegaev, line 3: Note this discussion

src/porepy/viz/solver_statistics.py
```diff
@@ -63,14 +69,27 @@ class SolverStatistics:
...
     """Number of domains in each dimension."""
-    simulation_status: SimulationStatus = field(default=SimulationStatus.IN_PROGRESS)
-    """Last logged simulation status."""
-    simulation_status_history: list[SimulationStatus] = field(default_factory=list)
-    """Simulation status history."""
-    solver_status: SolverStatus = field(default=SolverStatus.IN_PROGRESS)
-    """Last logged solver status."""
-    solver_status_history: list[SolverStatus] = field(default_factory=list)
-    """Solver status history."""
+    solver_status: NonlinearSolverStatus = field(
```
Yuriyzabegaev, line 72: Left the field names the same as before for compatability (solver_status and simulation_status), but now they are mapped to NonlinearSolverStatus and TimeStepperStatus, respectively

tests/numerics/nonlinear/test_convergence_check.py
```diff
@@ -53,43 +52,6 @@ def multiphysics_check(info, expected_value):
     )
 
 
-# ! ---- SIMULATION STATUS TESTS ---- !
-
-
-def test_simulation_status_methods():
```
Yuriyzabegaev, line 59: removed these tests since removed the corresponding enums.

tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -363,168 +431,18 @@ def test_solve_convergence_time_dependent(default_newton_solver):
...
     # Second time step.
     model.time_manager.increase_time()
     model.time_manager.increase_time_index()
     solver_status = solver.solve(model)
 
     # Check simulation status.
-    assert solver_status == SolverStatus.SUCCESSFUL
-
-
-def test_solve_convergence_time_dependent_statistics(default_newton_solver):
```
Yuriyzabegaev, line 377: Moved to test_time_stepper, because the responsibility on statistics writing is on it

tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -348,7 +416,7 @@ def test_solve_convergence_statistics(default_newton_solver):
     Path("solver_statistics.json").unlink()
 
 
-def test_solve_convergence_time_dependent(default_newton_solver):
```
Yuriyzabegaev, line 351: Did not move this test to test_time_stepper. Despite it's about a time-dependent model, it covers the nonlinear solver behavior, abstracted from the TimeStepper.

tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -579,13 +494,13 @@ def test_solve_failure_statistics(default_newton_solver):
...
-                        "inc_inf": "diverged",
-                        "res_inf": "diverged",
-                        "inc_nan": "converged",
-                        "res_nan": "diverged",
+                        "inc_abs": "continue_iterating",
+                        "res_abs": "continue_iterating",
+                        "max_iter": "continue_iterating",
+                        "inc_inf": "failed",
+                        "res_inf": "failed",
+                        "inc_nan": "continue_iterating",
```
Yuriyzabegaev: This is where we see the changed behavior of divergence criteria (see my comment above). If it did not diverge, it previously returned "converge", now it is "continue_iterating".

tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -649,234 +564,10 @@ def test_solve_failure_time_dependent(default_newton_solver):
 
     # Check simulation status.
     assert model.time_manager.final_time_reached()
-    assert solver_status == SolverStatus.SUCCESSFUL
-
-
-def test_solve_failure_time_dependent_statistics(default_newton_solver):
```
Yuriyzabegaev, line 655: Moved to test_time_stepper

tests/time_stepper/test_time_step_control.py
```diff
@@ -464,33 +468,6 @@ def test_raise_warning_recompute_solution_true_for_constant_time_step(self):
             time_manager.compute_time_step(recompute_solution=True)
         assert str(record[0].message) == msg
 
-    def test_non_recomputed_solution_conditions(self):
```
Yuriyzabegaev, line 467: Removed this test since the TimeManager is not responsible for counting the retries anymore. Equivalent test is introduced in test_time_stepper.

tests/time_stepper/test_time_step_control.py
```diff
@@ -768,189 +750,3 @@ def test_io_time_information(self):
...
-        # Case 4: The time step fails right before the schedule point. Expected to
-        # decrease dt and meet the schedule regardless.
-        {
-            "num_nonlinear_iterations": [1, 1, 1, 1, 1],
-            "time_step_converged": [True, False, True, True, True],
-            "exported_dt_expected": [1, 0.35, 0.105, 0.21, 0.035],
-        },
-    ],
-)
-def test_model_time_step_control(params: dict):
```
Yuriyzabegaev, line 879: moved to test_time_stepper

tests/time/test_time_stepper.py
```diff
@@ -0,0 +1,807 @@
...
+    time_manager = TimeManager(
+        schedule=[0, 1], dt_init=1, constant_dt=False, dt_min_max=(0.5, 1)
+    )
+    time_stepper = TimeStepper(time_manager=time_manager)
+
+    # It will attempt to make a time step twice here, with dt=1 and dt=0.5. Both will
+    # fail after two unsuccessful nonlinear iterations.
+    status = time_stepper.perform_time_step(model=model, solver=solver)
+    assert status.is_failure()
+    # Note that the first attempt (with dt=1) will not be saved in the file. The second
```
Yuriyzabegaev, line 702: Note this comment

Yuriyzabegaev (commented):
I highlighted some important changes, hope it helps reviewing it.

src/porepy/models/model_runner.py
Yuriyzabegaev: At some point, possibly during this PR and likely by me, the progressbar was broken. I fixed it and made an attempt to improve it here, having the change approved by @keileg.

src/porepy/numerics/nonlinear/convergence_check.py
```diff
@@ -627,8 +602,8 @@ def check(self, *args, **kwargs) -> ConvergenceStatus:
         status, _ = CombinedConvergenceCriterion.check(
             cast(CombinedConvergenceCriterion, self), *args, **kwargs
         )
-        if status.is_not_converged():
-            status = ConvergenceStatus.DIVERGED
+        if status.is_iterating():
+            status = ConvergenceStatus.FAILED
```
keileg, line 604: This looks strange. Please explain.

src/porepy/numerics/nonlinear/convergence_check.py
```diff
@@ -917,6 +892,6 @@ def check(self, num_iterations: int, **kwargs) -> ConvergenceStatus:
         """
         if num_iterations >= self.max_iterations:
             logger.info(self.divergence_msg())
-            return ConvergenceStatus.DIVERGED
+            return ConvergenceStatus.FAILED
         else:
-            return ConvergenceStatus.CONVERGED
+            return ConvergenceStatus.CONTINUE_ITERATING
```
keileg: This is a change in logic: If a criteria intersecter requires that all criteria are converged, this will never pass. Please check this.

src/porepy/numerics/nonlinear/convergence_check.py
```diff
@@ -515,9 +488,11 @@ def check(self, *args, **kwargs) -> ConvergenceStatus:
         status, _ = RelativeConvergenceCriterion.check(
             cast(RelativeConvergenceCriterion, self), *args, **kwargs
         )
-        if status.is_not_converged():
-            status = ConvergenceStatus.DIVERGED
+        if status.is_iterating():
+            status = ConvergenceStatus.FAILED
             logger.info(self.divergence_msg())
+        elif status.is_converged():
```
keileg: Will this not doom the check never to return 'converged'?

keileg: The same comment applies to multiple cases in this file.

src/porepy/numerics/nonlinear/nonlinear_solver_status.py
```diff
@@ -0,0 +1,43 @@
...
+    """A status object used to indicate the NewtonSolver (or LinearSolver) state."""
+
+    @abstractmethod
+    def serialize(self) -> str:
+        """Return the stable string representation used in stored statistics."""
+
+    def is_converged(self) -> bool:
+        """Whether the nonlinear system is solved successfully."""
+        return isinstance(self, NonlinearSolverStatusConverged)
```
keileg, line 29: This must break with some OOP principle. A superclass should not be aware of its subclasses. Granted, these are minimal classes, but still.

src/porepy/numerics/linear_solvers.py
keileg: I have come to regret similar decisions so many times (the word 'planning' is dangerous)that I would prefer if we can find a joint home for the method, even if it is temporary. I think it will not be wrong to put it in a new module - perhaps we will move it later, but at least it is not duplicated.

The same applies to update_solver_statistics in this class, which I noted was called from the nonlinear solver.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -250,16 +253,68 @@ def solve(self, model: SolutionStrategy) -> SolverStatus:
...
 
-        # Conclude on the solver status.
+        # Summarizing the convergence message from multiple criteria into an overall
+        # status.
         solver_status = self.summarize_solver_status(
-            model, convergence_status, divergence_status
+            convergence_status, divergence_status
+        )
+
+        # YZ: This hack is to be addressed when we find a better way to organize
```
keileg: See comment in LinearSolver regarding the duplicated summarize_solver_status.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -250,16 +253,68 @@ def solve(self, model: SolutionStrategy) -> SolverStatus:
...
         return solver_status
 
+    def summarize_solver_status(
+        self,
+        convergence_status: ConvergenceStatusCollection,
+        divergence_status: ConvergenceStatusCollection,
+    ) -> NonlinearSolverStatus:
+        """Consider a collection of convergence and divergence statuses from multiple
+        criteria and make an overall verdict on whether we accept the sollution or not.
+
```
keileg: Perhaps document that the method is meant to be called after the nonlinear solver is done.

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -311,59 +366,6 @@ def nonlinear_loop(
...
 
-    def summarize_solver_status(
-        self,
-        model: SolutionStrategy,
-        convergence_status: ConvergenceStatusCollection,
-        divergence_status: ConvergenceStatusCollection,
-    ) -> SolverStatus:
-        """Conclude on the overall solver status.
-
-        NOTE: Convergence status takes precedence over divergence status.
```
keileg, line 322: This comment and its implication seem to have disappeared in the revision. It may be fine - it should be a responsibility of the convergence status collection, not the solver, but please confirm.

src/porepy/models/model_runner.py
```diff
@@ -4,57 +4,60 @@
...
 
-class SimulationStatus(StrEnum):
-    """Enumeration of potential simulation statuses."""
+@dataclass
+class ModelRunnerStatus(ABC):
+    """A status object used to indicate the ModelRunner state."""
+
+    def is_success(self) -> bool:
+        """Whether the simulation finished successfully."""
+        return isinstance(self, ModelRunnerStatusSuccess)
```
keileg, line 54: Same comment as for the nonlinear solver status: A superclass should not be aware of its subclasses. I am not sure if an exception is okay here.

src/porepy/models/model_runner.py
```diff
@@ -146,7 +149,12 @@ class ModelRunner:
 
     """
 
-    def __init__(self, model: pp.SolutionStrategy, params: dict | None = None) -> None:
+    def __init__(
+        self,
+        model: SolutionStrategy,
+        params: Optional[dict] = None,
+        time_stepper: Optional[TimeStepper] = None,
```
keileg, line 169: Please update docstring.

src/porepy/models/model_runner.py
keileg: But you don't like it?

src/porepy/models/model_runner.py
```diff
@@ -212,211 +228,95 @@ def init_time_progressbar(self) -> None:
...
+                time_step_status = self.time_stepper.perform_time_step(
+                    self.model, self.solver
+                )
+
+                # Update the progressbar after the time step.
+                if isinstance(time_step_status, TimeStepperStatusSuccess):
+                    self.time_progressbar.update(n=time_step_status.dt)
+
+                # Abort simulation if time step was stopped.
+                match time_step_status:
```
keileg: Why match here?

src/porepy/numerics/nonlinear/nonlinear_solvers.py
```diff
@@ -250,16 +253,68 @@ def solve(self, model: SolutionStrategy) -> SolverStatus:
...
+        # status.
         solver_status = self.summarize_solver_status(
-            model, convergence_status, divergence_status
+            convergence_status, divergence_status
+        )
+
+        # YZ: This hack is to be addressed when we find a better way to organize
+        # statistics collection.
+        # Logging basic discretization-related information and overall simulation status
+        pp.LinearSolver.update_solver_statistics(
```
keileg: Why not the similarly named method in this class?

src/porepy/numerics/time_step_control.py
```diff
@@ -563,64 +563,58 @@ def _adaptation_based_on_recomputation(self) -> None:
...
+
+        # Note that iterations is not really used here. So, as long as
+        # recompute_solution = True and recomputation_attempts <
+        # max_recomp_attempts, the method is entirely agnostic to the number of
+        # iterations passed. This design choice was made to give more flexibility,
+        # in the sense that we are not limiting the recomputation criteria to _only_
+        # reaching the maximum number of iterations, even though that is the primary
+        # intended usage.
+
+        # S1, S2 and S4 are removed as the TimeStepper is responsible for this now.
```
keileg: Please update the above documentation accordingly.

src/porepy/time_stepper/time_step_control.py
```diff
@@ -158,8 +158,7 @@ class TimeManager:
             Factor by which the time step will be multiplied in case the solution must
```
keileg, line 158: Should the file be moved to the folder 'time'?

src/porepy/time_stepper/time_step_status.py
keileg: Looks fine, thanks for leaving the reminder here.

src/porepy/time/time_stepper.py
```diff
@@ -0,0 +1,239 @@
...
+    TimeStepperStatusFailure,
+    TimeStepperStatusSuccess,
+)
+from porepy.viz.solver_statistics import NonlinearSolverStatistics
+
+logger = logging.getLogger(__name__)
+
+
+class TimeStepper:
+    """Base class for time-stepping strategies.
```
keileg: I will need a refresher on the planned responsibilities for this and the time manager class. Keywords: Is this truly a base class, or does it also define 1-2 different strategies for dt computation?

src/porepy/time/time_stepper.py
```diff
@@ -0,0 +1,239 @@
...
+            time_manager.recomp_max = 0
+
+        assert self.max_attempts > 0, "max_attempts must be greater than 0."
+
+    def perform_time_step(
+        self,
+        model: pp.PorePyModel,
+        solver: pp.NewtonSolver | pp.LinearSolver,
+    ) -> TimeStepperStatusSuccess | TimeStepperStatusFailure:
+        """Perform a time step. If the nonlinear solver fails, cut the time step and
```
keileg: Cut -> alter? As @IvarStefansson has pointed out repeatedly to me, some problems (e.g., initialization) becomes easier with increasing time step size.

tests/models/test_contact_mechanics.py
```diff
@@ -173,9 +173,7 @@ def test_contact_mechanics_convergence(nd, formulation):
     model_class = add_mixin(grid_classes[nd], tester_classes[formulation])
     model = model_class(params)
 
-    # Run and capture solver info.
-    pp.ModelRunner(model).run()
+    # Run the simulation.
+    status = pp.ModelRunner(model).run()
 
-    # Fetch convergence status from statistics.
-    status = model.nonlinear_solver_statistics.simulation_status
```
keileg, line 180: This looks much better now.

keileg (commented):
Partial review, I ran out of energy before having looked at the tests in earnest. I will try to return to these later today.

tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -253,7 +322,7 @@ def test_increase_iteration_index(default_newton_solver):
     assert solver.iteration_index == 2
 
 
-def test_solve_convergence(default_newton_solver):
+def test_solve_convergence(default_newton_solver: pp.NewtonSolver):
```
keileg, line 326: Any reason to introduce typing here specifically?

tests/numerics/nonlinear/test_nonlinear_solvers.py
keileg: Do we have a better name for the test then?

tests/numerics/nonlinear/test_nonlinear_solvers.py
keileg: Let's discuss the implications of this.

tests/time_stepper/test_time_step_control.py
keileg: Does this also apply to the model class just above?

tests/time/test_time_stepper.py
keileg: I think we should make a GH issue with stuff we need to remember regarding the solver statistics. Likely this applies to other similar issues.

src/porepy/models/model_runner.py
Yuriyzabegaev: I'm not sure if both it and returning a status is meaningful. It could return a failure status, and if you want to check it - you do, and also check for a reason. I now wrap a status inside a runtime error, so you can unwrap the error to check the failure status.

src/porepy/numerics/nonlinear/convergence_check.py
Yuriyzabegaev: It's a divergence criterion, why should it ever return "converged"?

src/porepy/numerics/nonlinear/convergence_check.py
Yuriyzabegaev: The class we are in is a divergence criterion. It is implemented in way that it reuses the convergence criterion code with a divergence tolerance. If we are below threshold, the inner convergence criterion returns "converged". For us, it means we are below divergence threshold, so we continue iterating. If we are above the threshold, the inner convergence criterion returns "continue iterating", which means "diverged" for us. I see the error here and I will fix it, but this logic applies to all the other divergence criteria in this file.

src/porepy/numerics/nonlinear/convergence_check.py
Yuriyzabegaev: It never checks on the intersection of convergence and divergence criteria. It asks convergence criteria "did we converge" and divergence criteria "did we diverge"

src/porepy/numerics/nonlinear/nonlinear_solvers.py
Yuriyzabegaev: Because they must be called in different points of time (NonlinearSolver.update_solver_statistics at each iteration and LinearSolver.update_solver_statistics once per nonlinear solve). I didn't change it from how it was before. Do you think we should improve this logic in this pr?

src/porepy/models/model_runner.py
Yuriyzabegaev: added docstring and comment after discussion

src/porepy/models/model_runner.py
Yuriyzabegaev: updated

src/porepy/models/model_runner.py
Yuriyzabegaev: replaced with isinstance

src/porepy/numerics/nonlinear/convergence_check.py
Yuriyzabegaev: the logic of divergence criteria is reverted back to what it was

src/porepy/numerics/nonlinear/convergence_check.py
Yuriyzabegaev: the logic is reverted to what it was

src/porepy/numerics/nonlinear/convergence_check.py
Yuriyzabegaev: nevermind, reverted it

src/porepy/numerics/nonlinear/nonlinear_solver_status.py
Yuriyzabegaev: added a comment and a docstring

src/porepy/numerics/nonlinear/nonlinear_solvers.py
Yuriyzabegaev: moved common parts into a separate file

src/porepy/numerics/nonlinear/nonlinear_solvers.py
Yuriyzabegaev: documented

src/porepy/numerics/nonlinear/nonlinear_solvers.py
Yuriyzabegaev: Updated the comment

src/porepy/numerics/linear_solvers.py
Yuriyzabegaev: moved into a separate file

src/porepy/time_stepper/time_step_control.py
Yuriyzabegaev: moved

src/porepy/numerics/time_step_control.py
Yuriyzabegaev: updated

src/porepy/time/time_stepper.py
Yuriyzabegaev: This class is not polymorphic. Removing the word "base" to avoid confusion. 

Now, all time step adjustments are done by a time stepper. Defining strategies for dt computations is a future change (not this PR), as we discussed with @IvarStefansson.

src/porepy/time/time_stepper.py
Yuriyzabegaev: edited

tests/numerics/nonlinear/test_nonlinear_solvers.py
Yuriyzabegaev: No reason, just for vscode to give me type hints. Should I revert it?

tests/time_stepper/test_time_step_control.py
Yuriyzabegaev: yes

tests/numerics/nonlinear/test_nonlinear_solvers.py
Yuriyzabegaev: I reverted it. Now it's the same way logic that was before

src/porepy/numerics/nonlinear/nonlinear_solvers.py
keileg: Let's leave it as is for now.

tests/numerics/nonlinear/test_nonlinear_solvers.py
keileg: Let's keep it.

src/porepy/time/time_stepper.py
Yuriyzabegaev: After discussion with @keileg we decided that the distinction between before_time_step and before_nonlinear_loop can be important and we keep it for now

Yuriyzabegaev: @keileg decided to keep the name `model.before_time_step`, because the model should not have concept of trial time steps.

src/porepy/time/time_stepper.py
Yuriyzabegaev: After the changes in the code, we restructured this loop and decided to not use both before_ and after_ methods in this case, because this class is not meant for subclassing anymore.

src/porepy/time/time_stepper.py
keileg: Fair question. @Yuriyzabegaev and I discussed the matter, but could not identify a likely scenario where the current design would go directly against an abstraction needed in a more general setting. That is not a full answer, but I think it is the best we can do for now. Resolving.

keileg (approved):

src/porepy/viz/solver_statistics.py
```diff
@@ -63,14 +69,27 @@ class SolverStatistics:
...
+        default_factory=lambda: NonlinearSolverStatusConverged(
+            convergence_statuses=ConvergenceStatusCollection(),
+            divergence_statuses=ConvergenceStatusCollection(),
+        )
+    )
+    """Nonlinear solver status history."""
```
IvarStefansson: Confused docstrings for status and history.

src/porepy/viz/solver_statistics.py
```diff
@@ -63,14 +69,27 @@ class SolverStatistics:
...
+            attempt=-1,
+            nonlinear_solver_status=NonlinearSolverStatusConverged(
+                convergence_statuses=ConvergenceStatusCollection(),
+                divergence_statuses=ConvergenceStatusCollection(),
+            ),
+        )
+    )
+    """Simulation time step status."""
+    simulation_status_history: list[TimeStepperStatus] = field(default_factory=list)
+    """Simulation time step status history."""
```
IvarStefansson, line 92: Please verify that this is indeed time step.

src/porepy/time_stepper/time_step_control.py
```diff
@@ -563,64 +563,42 @@ def _adaptation_based_on_recomputation(self) -> None:
...
-            # the next iteration (any decrease in time step will be corrected to dt_min
-            # by self.correction_based_on_dt_min() in a subsequent correction step).
-            # Thus, to avoid pointless iterations, we raise an error.
-            if self.dt == self.dt_min_max[0]:
-                msg = (
-                    "Recomputation will not have any effect since the time step "
-                    f"achieved its minimum admissible value -> dt = dt_min = {self.dt}."
-                )
-                raise ValueError(msg)
+        # Raise a warning if iterations is not None
```
IvarStefansson: Why has iterations have to be None?

src/porepy/time_stepper/time_step_control.py
```diff
@@ -661,6 +645,13 @@ def _correction_based_on_schedule(self) -> None:
                     )
                 return
 
+            # Consider dt=1, t=0.999999, schedule_time=1. We'll decrease the time step
+            # to dt=1e-6 and have to slowly increase it. When refactored, we should
+            # track of previous dt, and return to that value if we expected to hit the
+            # schedule. There is no reason to start increasing the dt from scratch.
```
IvarStefansson, line 651: Note: I think we might want the option of passing user-defined values (min, max, initial, optimal range etc.) for each schedule interval. For this particular case (initial value of dt at new interval), I agree that if not passed, fallback should be final non-chopped value of previous interval rather than the chopped one. Perhaps it is useful to introduce the concept of "target dt" or similar, which is what the iteration-based TimeStepSuggestor (or whatever we end up calling it) will provide. Then each suggestor could store the previous value if required for its algorithm. However, not obvious this is clean, since it needs to be adapted if recomputation is triggered.

Sorry about the rambling. Meta note: I have some thoughts here, which it may be useful to discuss before next iteration/PR of the project. Even if it turns out my ideas are misguided, the discussion may save you from me having opinions down the road.

src/porepy/time_stepper/time_stepper.py
```diff
@@ -0,0 +1,239 @@
...
+
+    Responsibilities:
+    - Orchestrate the single time-step workflow to be called from the model runner.
+    - Execute trials (delegating nonlinear solves)
+    - Adjust dt (currently delegated to TimeManager)
+
+    Workflow:
+    1. For each retry (up to max_retries):
+       a. Execute trial with current dt
+       b. If success: update time, adapt dt for next step, return
```
IvarStefansson: Isn't this misleading? Time is updated before trial. However, solution values are updated after. So I'd say
       a. Update trial time
       b. Execute trial with current dt
       c. If success: update solution values, adapt dt for next step, return
       d. If rejected: reduce dt, loop, revert trial time
Also: Is reduce dt in the contract? Or is it one of in theory several actions that could be taken to retry? Perhaps those other options would be on the nonlinear solver side.

IvarStefansson: That being said: I love how this class is documented with responsibilities and workflow.

src/porepy/time_stepper/time_stepper.py
```diff
@@ -0,0 +1,239 @@
...
+    ) -> NonlinearSolverStatus:
+        """Perform a nonlinear solve to make the time step.
+
+        Returns:
+            The nonlinear solver status (converged/failed).
+
+        """
+
+        # Execute trial time step.
+        model.before_time_step()
```
IvarStefansson, line 204: Is it obvious this should happen at each trial? Please make sure that contract is clearly documented.

src/porepy/time_stepper/time_stepper.py
```diff
@@ -0,0 +1,239 @@
...
+            self.time_manager.dt,
+            self.time_manager.final_time_reached(),
+        )
+
+
+def _log_time_step(time_manager: pp.TimeManager) -> None:
+    """Log the current state of the time step."""
+    logger.info(
+        f"Time step #{time_manager.time_index}: "
+        f"dt={time_manager.dt:.2e}, "
```
IvarStefansson, line 235: I'm tempted to ask for higher precision here. Opinions based on your experience with "realistic" simulations, @Yuriyzabegaev?

tests/time_stepper/test_time_stepper.py
```diff
@@ -0,0 +1,807 @@
...
+)
+from porepy.numerics.nonlinear.nonlinear_solvers import NewtonSolver
+from porepy.time_stepper.time_step_control import TimeManager
+from porepy.time_stepper.time_stepper import TimeStepper
+from porepy.viz.solver_statistics import (
+    NonlinearSolverAndTimeStatistics,
+    SolverStatisticsFactory,
+)
+
+# MARK: test_model_delegate_methods_called
```
IvarStefansson, line 36: I think we usually don't use these. Do we want to introduce them?

tests/time_stepper/test_time_stepper.py
```diff
@@ -0,0 +1,807 @@
...
+        {
+            "constant_dt": True,
+            "num_nonlinear_iterations": [1, 8, 1],
+            "time_step_converged": [True, True, True],
+            "exported_dt_expected": [1, 1, 1],
+            "schedule_end": 3,
+        },
+    ],
+)
+def test_model_time_step_control(params: dict):
```
IvarStefansson, line 470: Shouldn't this live in test_time_step_control.py?

IvarStefansson (commented):
Some more or less fair comments. Note that I had to take some of the tests on faith due to max_PR_review_time being reached (the quota is reset on July 26. To increase your review budget, please see our upgrade options. Or make sure the weather is terrible next week, in which case yours truly will postpone some of his vacation). However, I realise they are important, as they define a truth. I trust you have reviewed thoroughly, @keileg.

src/porepy/time_stepper/time_stepper.py
Yuriyzabegaev: It's the same as it was before. In my opinion it's fine, a single digit after floating point would be enough for me. E.g., when I see something like: "t=1.5e+4 of 1.0e+5, dt=1.7e+3", I realize that we're 15% done, and each time step does 0.1% of the simulation. I don't need more precision here.

src/porepy/viz/solver_statistics.py
Yuriyzabegaev: I agree they are confusing! And I'm afraid improving the docstrings won't help here, the logic should be reconsidered instead.  But I deliberately avoided doing anything related to it in this PR. The solver_status and simulation_status had docstrings "Last logged solver/simulation status.". Should I revert?

src/porepy/viz/solver_statistics.py
Yuriyzabegaev: Yes. Each entry is for a single time step. This is the same as it was before this PR, just specified it in the docstring.

src/porepy/time_stepper/time_step_control.py
Yuriyzabegaev: This comment is left long before this PR. The logic is the following. You can in principle pass "iterations" to this function, but it will cause no effect. Thus raise a warning.

tests/time_stepper/test_time_stepper.py
keileg: I have started introducing similar comments lately and find them useful, in particular when trying to write shorter functions. So I would vote for allowing them in general and leave some discretion on when to use them.

Yuriyzabegaev: I see the benefit in this particular file because it consists of only 3 tests, but each involves quite a lot of setup with mockup classes. This plus how vscode highlights these "mark" sections makes it easier for me to navigate in it.

tests/time_stepper/test_time_stepper.py
Yuriyzabegaev: test_time_step_control.py mirrors time_step_control.py, which only defines the TimeManager. Since we will be moving time control responsibility away from the TimeManager, I though that this is a better place for such high-level integration test.

src/porepy/time_stepper/time_stepper.py
Yuriyzabegaev: Added documentation

tests/time_stepper/test_time_stepper.py
Yuriyzabegaev: Improved documentation

Issue #1674, opened 2026-05-07
BC value units


IvarStefansson:
There is confusion and inconsistent/lacking documentation of the physical (SI) units of bc values as returned by the `bc_values_*` methods of the models, particularly those of secondary variables. We should
1. Deduce the correct units in each case.
2. Create tests checking programmatically the deduction is correct by comparing model runs with nontrivial bc values with default unitary Units to runs with non-unitary Units.
3. Update and improve documentation throughout code, including tutorials and examples.
4. Check for any bugs in usage, particularly tutorials, examples and applications.

PR #1683, opened 2026-05-27
Issue #1674

mikeljordan:
This PR adds unit-invariance tests for boundary-condition value methods, to verify their SI units using porepy.Units rescaling. Each test applies non-unitary internal unit systems, solves a small probe model, recovers the relevant observable to SI units, and checks that the recovered result is invariant under the rescaling. A mismatch between the declared and true SI dimensions of the BC produces a clean monomial discrepancy in the failure ratio, which makes the diagnostic actionable: the empirical exponent identifies the correct unit.

The covered BC methods are:
- `bc_values_darcy_flux`
- `bc_values_fluid_flux`
- `bc_values_stress`
- `bc_values_fourier_flux`
- `bc_values_enthalpy_flux`
- `bc_values_component_flux`

Docstrings of the affected methods, along with examples and tutorials referencing them, have been updated to reflect the verified SI unit definitions.
Co-authored-by: @RonghaoCui 

## Types of changes

- [ ] Other:

## Checklist


src/porepy/applications/test_utils/well_models.py
IvarStefansson: Please update docstring.

tests/models/test_boundary_condition.py
```diff
@@ -10,6 +15,9 @@
     SquareDomainOrthogonalFractures,
 )
 from porepy.applications.test_utils.models import MassBalance as MassBalance_
+
+# from tests.functional.setups.linear_tracer import TracerFlowModel_1p
```
IvarStefansson: Should we remove this line?

tests/models/test_boundary_condition.py
```diff
@@ -451,3 +459,483 @@ class LocalReferenceModel(reference_bc_class, reference_model_class):
...
+    declared_bc_unit: str
+    extraction_subdomain_dim: int = 2
+
+
+#: Unit scalings probing each axis of the BC's SI dimensions.
+_DEFAULT_UNIT_SCALINGS: list[pp.Units] = [
+    pp.Units(m=10.0),
+    pp.Units(kg=0.01),
+    pp.Units(m=100.0),
+    pp.Units(m=10.0, kg=0.01, K=10.0),
```
IvarStefansson: I suggest a more orthogonal set of values to avoid spurious cancellation effects. E.g. m=3, kg=7, K=0.23 or whatever.

tests/models/test_boundary_condition.py
```diff
@@ -451,3 +459,483 @@ class LocalReferenceModel(reference_bc_class, reference_model_class):
...
+        atol=1e-10,
+    )
+
+
+@pytest.mark.parametrize(
+    "spec",
+    _BC_UNIT_INVARIANCE_SPECS,
+    ids=_bc_spec_label,
+)
+def test_bc_values_baseline_well_posed(spec: _BCUnitInvarianceSpec) -> None:
```
IvarStefansson: This test should run on exactly the same models as the other one you introduced, right? If that is the case, we should avoid rerunning the simulations to save test time. I have two suggestions for how we could achieve that (but am open to other approaches): 1 insert the check into the other test or 2 convert the construction (including the running of the simulations) into a parametrized fixture?

tests/models/test_boundary_condition.py
```diff
@@ -451,3 +459,483 @@ class LocalReferenceModel(reference_bc_class, reference_model_class):
...
+    "spec",
+    _BC_UNIT_INVARIANCE_SPECS,
+    ids=_bc_spec_label,
+)
+@pytest.mark.parametrize(
+    "units",
+    _DEFAULT_UNIT_SCALINGS,
+    ids=_unit_scaling_label,
+)
+def test_bc_values_unit_invariance(
```
IvarStefansson, line 912: Do you agree we could mark this test as slow (weekly run instead of each PR)?

IvarStefansson (commented):
Nice work. Some relatively minor comments only. 
One is included here to circumvent notebook reviewing, which is a pain: Please add units of flux BCs in docstrings in boundary_conditions.ipynb.

tests/models/test_boundary_condition.py
mikeljordan: Yes, this won't be a problem, the test should hold for any values of m, kg, K. Will do that now.

tests/models/test_boundary_condition.py
mikeljordan: Good point. I’ll fold the check into the unit-invariance assertion, following your first suggestion and remove the redundant test.

tests/models/test_boundary_condition.py
mikeljordan: Yes, I agree. I’ll add the slow marker to the test.

tests/models/test_boundary_condition.py
RonghaoCui: Yes, this line has been removed.

src/porepy/applications/test_utils/well_models.py
RonghaoCui: The docstrings on units of flux BCs in the tutorial have been added. 
The boundary condition docstrings in well_models.py have also been updated. A new review for well_models.py is requested.

tests/models/test_boundary_condition.py
IvarStefansson: Thanks. Just to clarify: I was fine with not changing all units for all types. I was only referring to the last case, the one which had non-default values for all three units.  Feel free to resolve again if you want to keep as is or ask if I'm being confusing.

tutorials/boundary_conditions.ipynb
```diff
@@ -370,7 +380,13 @@
                 "\n",
                 "    # Note that now this is bc_values_darcy_flux, not the bc_values_fluid_flux.\n",
                 "    def bc_values_darcy_flux(self, bg: pp.BoundaryGrid) -> np.ndarray:\n",
-                "        \"\"\"Setting the Darcy flux values on the west boundary.\"\"\"\n",
+                "        \"\"\"Setting the mobility-free Darcy flux values on the west boundary.\n",
+                "        \n",
+                "        The values correspond to the integrated quantity of mobility-free Darcy \n",
+                "        flux over the boundary face on Neumann boundary. In two dimensions, the \n",
+                "        SI unit of integrated mobility-free Darcy flux is m^2*Pa.\n",
```
IvarStefansson: Great, this will be much appreciated! Optional suggestion: Since this is a tutorial, should we be nice to the user and spell out the 3d case as well?

IvarStefansson (approved):
Thanks. This is going to be useful for users new and old!

Please have a look at the new comments, then merge.

tutorials/boundary_conditions.ipynb
mikeljordan: Yes I agree, we will include that

PR #1684, opened 2026-05-28
Distinction SolverStatus and SimulationStatus

### LINKED ISSUE #1657: Failed simulations should throw error

### LINKED ISSUE #1672: Distinction between SolverStatus and SimulationStatus
Currently, `SimulationStatus` is used for both defining a `solver_status` (return object from a nonlinear solver) and `simulation_status`. A clean up is required introducing a clearer distinction. In addition, a mild refactoring of the respective organization and logic during assignment and reaction is foreseen. The following procedure is suggested:

- [x] Introduce separate `EnumStr` for `SimulationStatus` and `SolverStatus` with possible states `SUCCESS`, `IN_PROGRESS`, `FAILURE`, `FATAL`. While `FAILURE` shall indicate potentially recoverable errors (nonlinear solver did not converge), `FATAL` defines non-recoverable errors. This distinction allows for simply passing a fatal status to break all loops and abort the simulation without omitting potential export or statistics saving.
- [x] Clean up the current mixed use of `SimulationStatus` at the interface between `pp.ModelRunner` and the solvers `pp.NewtonSolver` and `pp.LinearSolver`.
- [x] (Midly!) refactor `pp.ModelRunner.after_time_step` and let `solver_status.is_stopped()` (/fatal) explicitly result in the assignment `simulation_status = fatal`. Reminder: Fatal errors are not recoverable and therefore are sort of passed through and shall result in abortion of the simulation.
- [x] If `simulation_status.is_fatal` close the simulation with throwing an error just in the end, making sure statistics etc are printed to file (otherwise only warnings will be thrown) (directly addresses #1657)
- [x] `pp.SolverStatistics` will need an update as well. Keep that update as simple as possible.

Heads-up: This is an intermediate step. It is foreseen that a further layer will be introduced distinguishing between stepping through time (outer layer) and attempting a single time step. The latter currently has two modes: one-shot (constant dt) or iterative (adaptive dt reacting e.g. to solver performance). There is a strong analogy to the nonlinear solver (Newton and Linear). Thus, with additional success criteria for controlling a time_step_status object (can even fail if the newton converged, but with e.g. to many time steps, too large increment in time a la Appleyard, fancy aposteriori estimates) the same logic will be introduced for computing a time step as computing an solution to a (non)linear problem. The outcome of a solution to this issue, will still mix up these two layers through the `ModelRunner` and associated `simulation_status`. The split will be a separate follow-up.

vlipovac:
Closes #1672 
Closes #1657 

## Proposed changes

Changes as per issue #1672 

## Types of changes

- [ ] Other:

## Checklist


src/porepy/viz/solver_statistics.py
```diff
@@ -355,7 +374,7 @@ def append_global_data(self, data: dict[str, dict]) -> dict[str, dict]:
         for simulation_status, num_iterations in zip(
             self.simulation_status_history, self.num_iterations_history
         ):
-            if simulation_status != SimulationStatus.SUCCESSFUL:
+            if simulation_status != SolverStatus.SUCCESSFUL:
```
jwboth: Shouldn't this all be solver_status objects and histories?

src/porepy/viz/solver_statistics.py
```diff
@@ -451,7 +479,7 @@ def append_global_data(self, data: dict[str, dict]) -> dict[str, dict]:
         total_num_failed_time_steps = 0
         for simulation_status in self.simulation_status_history:
             total_num_time_steps += 1
-            if simulation_status != SimulationStatus.SUCCESSFUL:
+            if simulation_status != SolverStatus.SUCCESSFUL:
```
jwboth: Same here. Use `solver_status` and `solver_status_history`?

tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -387,11 +389,11 @@ def test_solve_convergence_time_dependent_statistics(default_newton_solver):
...
             "num_cells": {},
             "num_domains": {},
-            "simulation_status_history": ["successful"],
-            "final_simulation_status": "successful",
+            "simulation_status_history": [],
+            "final_simulation_status": None,
             "num_entries": 1,
             "final_time_reached": 0,
-            "total_num_time_steps": 1,
+            "total_num_time_steps": 0,
```
jwboth: Reminder for myself. Why was this change needed?

tests/numerics/nonlinear/test_nonlinear_solvers.py
```diff
@@ -542,12 +548,12 @@ def test_solve_failure_statistics(default_newton_solver):
...
                     "num_domains": {},
-                    "simulation_status_history": ["failed"],
-                    "final_simulation_status": "failed",
+                    "simulation_status_history": [],
+                    "final_simulation_status": None,
                     "num_entries": 1,
                     "num_iterations_history": [2],
                     "total_num_iterations": 2,
-                    "total_num_waisted_iterations": 2,
+                    "total_num_waisted_iterations": 0,
```
jwboth: Reminder for myself: Why are not iterations waisted after these changes?

src/porepy/viz/solver_statistics.py
```diff
@@ -381,6 +407,8 @@ def append_iterative_data(self, data: dict[str, dict]) -> dict[str, dict]:
                     if len(self.simulation_status_history) == 0
                     else self.simulation_status_history[-1]
                 ),
+                "solver_status": str_solver_status_history,
```
jwboth: Better call the key `solver_status_history`?

jwboth (commented):
I am starting with some minor questions. While I am thinking about `test_solver_statistics` and the failing tests, feel free to reply.

src/porepy/viz/solver_statistics.py
jwboth: Resolved.

src/porepy/viz/solver_statistics.py
jwboth: Resolved.

tests/numerics/nonlinear/test_nonlinear_solvers.py
jwboth: Reverted by explicitly logging simulation status.

tests/numerics/nonlinear/test_nonlinear_solvers.py
jwboth: Reverted by explicitly logging simulation status.

src/porepy/viz/solver_statistics.py
jwboth: Updated.

tests/models/test_thermoporomechanics.py
```diff
@@ -466,12 +466,12 @@ def test_unit_conversion(units: dict, model_class: type):
 
     # Create model and run simulation
     reference_model = model_class(model_params_ref)
-    pp.ModelRunner(reference_model).run()
```
jwboth, line 469: @IvarStefansson This PR fixes https://github.com/pmgbergen/porepy/issues/1657

Several tests actually should have failed due to strict tolerances. Are you for now OK with fixing those tests by adapting the tolerances locally? It is not many tests. I assume at later stage, we need some mild overhaul when discussing default tolerances and criteria.

tests/models/test_solution_strategy.py
```diff
@@ -283,14 +283,24 @@ def test_targeted_rediscretization(model_class):
...
     # A model object with full rediscretization.
     full_model: pp.PorePyModel = rediscretization_model_class(model_params)
-    pp.ModelRunner(full_model, solver_params).run()
+    try:
+        pp.ModelRunner(full_model, solver_params).run()
+    except RuntimeError:
+        # RuntimeError expected due to unattainable convergence criteria
+        # with only two iterations and zero tolerances. Allow RuntimeError without
+        # stopping the test.
+        pass
```
jwboth, line 292: @IvarStefansson Are you also OK with this?

Here, the simulation also technically does not run through. However, the simulation is only run to utilize the setup phase and a simulation is not meaningful anyways. So should be OK here, I think.

jwboth (commented):
I have updated the code. Main changes:

- reverted "nulling" several reported statistics in the tests, and instead changed the code such that the statistics are updated and reported to file
- relaxed tolerances to allow simulations to run through and not be stopped
- minor changes to the solver statistics

@vlipovac @aband Could you double check whether my changes are in line with your intentions?

Other than that, your looked good to me! Thanks a lot.

jwboth (approved):

tests/models/test_solution_strategy.py
IvarStefansson: Is it an option to simply call prepare_simulation instead of run?

tests/models/test_thermoporomechanics.py
IvarStefansson: This seems totally fair here. If you feel like you are kicking cans, as your later stage comment may suggest, please keep track of them as issue(s).

jwboth: Done. See #1690

tests/models/test_solution_strategy.py
jwboth: Not directly. Your suggestion stands in conflict with the (I assume intentionally designed) choice to make use of `RediscretizationTest` which provides a mechanism to store the linear systems for later comparison.

PR #1686, opened 2026-06-01
Initialization

jwboth:
Three-fold extension:
- Distinction between reference, relative and absolute variables (explicitly defined for each physics - only considering HM cases so far). If reference states are not set/updated the simulations should run as initially.
- Constitutive laws in terms of relative variables.
- Standardized initialization strategy as part of the simulation preparation - seamless transition from initialization to simulation.

The latter is defined for initialization of quasi-static initial conditions as well as reference states.

## Types of changes

- [ ] Other:

## Checklist


src/porepy/models/momentum_balance.py
```diff
@@ -417,6 +417,8 @@ class VariablesMomentumBalance(VariableMixin):
 
     displacement_variable: str
     interface_displacement_variable: str
+    has_momentum_balance_reference_state: bool = False
+    """Flag whether reference states have been defined."""
```
jwboth, line 421: The changes in this class are representative for changes in other core model classes. Be critical about:
- [ ] explicit introduction of essentially three versions of a variable (the absolute one derives from the others by addition) 
- [ ] the concept behind updating and shifting values including the flag above - it feels like one should have maybe the possibility to do this in a more global way and not model by model. Is it possible to utilize some decorators or collect variables in respective containers? Bit like with `previous_timestep()`. Argument against a global approach would possibly be memory. One could also argue for (like @IvarStefansson did yesterday) that we could ignore reference states in the remaining physics (pressure, temperature etc) since displacements are the only singular object and require correct treatment in terms of modelling, while the remaining once may instead only impact the numerical properties like conditioning.
- [ ] the imposed DataSavingMixin that always exports all three variants for the variables (maybe too much?). Reference variables should for instance also not change over time and could be handled as constants (see the not so much used functionality of the Exporter)

src/porepy/models/constitutive_laws.py
```diff
@@ -33,8 +33,67 @@
...
 
 
+class MechanicalAperture:
+    """Mechanical aperture of fractures wrt reference configuration."""
+
+    normal_component: Callable[[list[pp.Grid]], pp.ad.Operator]
+    """Operator giving the normal component of a vector on fractures."""
+
+    displacement_jump: Callable[[list[pp.Grid]], pp.ad.Operator]
+    """Displacement jump on fractures wrt reference ."""
```
jwboth, line 43: Be critical about the handling of the difference apertures and the now explicit use of relative displacements.

src/porepy/models/constitutive_laws.py
```diff
@@ -2938,6 +2990,13 @@ class LinearElasticMechanicalStress(pp.PorePyModel):
 
     To be used in mechanical problems, e.g. force balance.
 
+    NOTE: The implementation of the mechanical stress below is in absolute terms.
+    The stress could be formulated wrt a reference configuration, in which case
+    a reference (background) stress would be required, only depending on the reference
+    displacement. Under the assumption of linearity, the summation of the reference and
+    the linear mechanical stress below cancels the reference contributions. For
+    simplicity, background stress is not supported in the current implementation.
+
```
jwboth, line 2999: Should one think more about including explicit background stress measurements or possibly a homogeneous background stress and then formulate the mechanical stress in terms of relative quantities?

src/porepy/models/constitutive_laws.py
```diff
@@ -3562,7 +3625,8 @@ def pressure_stress(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
         # sd_primary, thus there is no need for a subdomain projection.
         stress: pp.ad.Operator = discr.scalar_gradient(
             self.darcy_keyword
-        ) @ self.perturbation_from_reference("pressure", subdomains)
+        ) @ self.pressure(subdomains)
+        # TODO: Safe to remove? self.perturbation_from_reference("pressure", subdomains)
```
jwboth: Note this here. I am quite confident but better to be critical. On fracture interfaces we want to be able to consider the balance of absolute stresses and not relative ones. In the bulk, perturbation by scalars should not make any difference when taking the divergence of a stress.

src/porepy/models/constitutive_laws.py
```diff
@@ -4549,7 +4619,7 @@ def porosity(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
 
 
 class PoroMechanicsPorosity(pp.PorePyModel):
```
jwboth, line 4618: Note that I did not cover all constitutive laws, e.g. code related to BartonBandis , damage etc.

What should be the right strategy? Coverage or sanity checks throwing an error/warning if reference states are used e.g. for unsupported constitutive laws or models.

src/porepy/models/contact_mechanics.py
```diff
@@ -364,8 +369,18 @@ def create_variables(self) -> None:
...
+    ) -> pp.ad.Operator:
+        """Reference displacement in the matrix."""
+        self.init_reference_contact_mechanics()
+        return pp.ad.TimeDependentDenseArray(
+            name=f"reference_" + self.contact_traction_variable,
+            domains=domains,
+        )
+
+    def relative_contact_traction(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
+        """Fracture contact traction increment [-].
```
jwboth, line 383: Be critical here: Do we really need reference and relative versions of all variables? There is no constitutive law that will benefit from contact traction changes, or? Same would apply for saturation-type variables in multiphase flow settings. If there should really be an impact on the linear/nonlinear conditioning of the problem by solving for increments, one should maybe consider solving for increments wrt the previous time step?

jwboth (commented):
@IvarStefansson @keileg I tried highlighting the pieces in the code that I suspect to require a critical look. Hope it helps for reviewing the suggested concepts.

src/porepy/models/constitutive_laws.py
```diff
@@ -33,8 +33,67 @@
...
+            Operator for the reference mechanical aperture.
+
+        Raises:
+             AssertionError: If the subdomains are not fractures, i.e. have dimension
+                `nd - 1`.
+
+        """
+        return Scalar(0.0, "reference_mechanical_aperture")
+
+    def mechanical_aperture(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
```
IvarStefansson, line 64: Consider if this makes reference fracture gap redundant/is already covered by that.

src/porepy/models/constitutive_laws.py
```diff
@@ -303,7 +364,7 @@ def specific_volume(self, grids: pp.GridLikeSequence) -> pp.ad.Operator:
         return specific_volume
 
 
-class DisplacementJumpAperture(DimensionReduction):
+class DisplacementJumpAperture(DimensionReduction, MechanicalAperture):
```
IvarStefansson, line 367: We need to be very careful about what this does to the interpretation of aperture (hydraulic vs mechanical)

src/porepy/models/constitutive_laws.py
IvarStefansson: I think that normatively the answer is yes. If we can't remove it, I have misunderstood what the purpose of the update is.

src/porepy/models/contact_mechanics.py
```diff
@@ -384,11 +399,87 @@ def contact_traction(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
...
+            for index in [{"iterate_index": 0}, {"time_step_index": 0}]:
+                pp.set_solution_values(
+                    name="reference_" + self.contact_traction_variable,
+                    values=np.zeros(self.nd * sd.num_cells),
+                    data=data,
+                    **index,
+                )
+
+        # Write logging info in yellow color
+        logger.info("\033[93m" + "Initialized reference state." + "\033[0m")
```
IvarStefansson: Why yellow?

src/porepy/models/fluid_mass_balance.py
```diff
@@ -543,15 +543,18 @@ def update_boundary_values_primary_variables(self) -> None:
...
 
-    pressure: Callable[[pp.SubdomainsOrBoundaries], pp.ad.Operator]
+    # TODO: OK to misuse the naming here and use ic_value_* for
+    # initial conditions of the primary variable which is relative?
+
+    relative_pressure: Callable[[pp.SubdomainsOrBoundaries], pp.ad.Operator]
     """See :class:`VariablesSinglePhaseFlow`."""
 
-    interface_darcy_flux: Callable[
+    relative_interface_darcy_flux: Callable[
```
IvarStefansson, line 552: Do we need relative fluxes?

src/porepy/models/momentum_balance.py
IvarStefansson: I agree that there seems to be unnecessary amount of boilerplate.

src/porepy/models/constitutive_laws.py
jwboth: Removed.

src/porepy/models/contact_mechanics.py
jwboth: Switched back to default font color.

keileg:
@jwboth Have I understood correctly that this PR is being replaced by the series of newer PRs on the same topic?

jwboth:
@keileg yes, it was kept for reference, but it was hopeless to base a solution for the identified weak points based on the herein presented implementation. 

> @jwboth Have I understood correctly that this PR is being replaced by the series of newer PRs on the same topic?

keileg:
Superseded by #1696, #1707 #1708 etc.

PR #1696, opened 2026-06-19
Reference states

jwboth:
## Proposed changes

This PR introduces an additional state of `Variable` and `TimeDependentDenseArray` following the style of `previous_iteration` and `previous_time_step`. These now also are able to hold reference values which can be obtained in AD operator form through call of the method `op.reference()` for any AD operator `op`, e.g., `model.pressure(subdomains).reference()`. In addition, perturbations to reference states can be conventiently fetched by `op.perturbation_from_reference()`. If not set, reference values are 0. No optimization in terms of emtpy operators has been implemented but could be foreseen in near future.

The API for setting values has been extended aiming at canonical extension of current workflows (`EquationSystem.set/get_variable_values` and `pp.set/get_solution_values`). An additional keyword `reference` needs to be set true.

As an example (hopefully of actual value) is added utilizing perturbations from reference states: Porosity with mere rock compressibility simplifying standard two-way coupled poromechanics. It currently mainly serves for testing the potential integration of reference states in models.

NOTE: This PR does not integrate reference states in models and variables like currently proposed in #1686 (initialization). The aim is to base further work on initialization on the current PR.

## Types of changes

- [ ] Other:

## Checklist


src/porepy/compositional/materials.py
```diff
@@ -363,6 +363,7 @@ class SolidConstants(Constants):
             "permeability": "m^2",
             "porosity": "-",
             "residual_aperture": "m",
+            "rock_compressibility": "Pa^-1",
```
IvarStefansson, line 366: The number of constants in the base concstants keeps growing. Some of them may be confusing/incompatible. Compressibility is already accounted for in the stiffness. I think we need to assess at some point whether to split into smaller units of constants tailored to different models. If you agree, let's create an issue.

src/porepy/numerics/ad/ad_utils.py
```diff
@@ -529,6 +546,96 @@ def shift_solution_values(
...
+        data: The data dictionary.
+
+    Returns:
+        A copy of the values stored at the passed category and name.
+        0 if no values are stored for the passed category and name.
+
+    """
+    try:
+        value = data[pp.REFERENCE_SOLUTIONS][name].copy()
+    except KeyError as err:
```
IvarStefansson: Are we sure we want to have fallback?

src/porepy/numerics/ad/ad_utils.py
```diff
@@ -445,6 +451,13 @@ def get_solution_values(
         A copy of the values stored at the passed index.
 
     """
+    if reference:
```
IvarStefansson: I think this precedence makes sense. But we need to document the rules for previous time steps and iterates and references somewhere.

src/porepy/numerics/ad/operators.py
```diff
@@ -107,6 +111,22 @@ def _get_previous_time_or_iterate(
         return new_op
 
 
+def _get_reference(op: Operator) -> Operator:
+    """Helper function for providing correct AD structure for reference operators."""
```
IvarStefansson: Needs documentation. Also suspect this precedence logic is different from get_solution_values. The example (op - op.previous_time_step).reference may be instructive. I think I'd expect zero rather than reference - previous time.

src/porepy/numerics/ad/operators.py
```diff
@@ -984,6 +1017,10 @@ def time_step_index(self) -> int | None:
         - ...
```
IvarStefansson: Update docstring

src/porepy/numerics/ad/operators.py
```diff
@@ -1165,6 +1206,84 @@ def previous_iteration(
...
+        domains: Optional[pp.GridLikeSequence] = None,
+        operation: Optional[Operations] = None,
+        children: Optional[Sequence[Operator]] = None,
+    ) -> None:
+        super().__init__(
+            name=name, domains=domains, operation=operation, children=children
+        )
+
+        self.original_operator: Operator
+        """Reference to the operator representing this operator at the current time and
```
IvarStefansson: Unfortunate use of "Reference".

src/porepy/numerics/ad/operators.py
```diff
@@ -1165,6 +1206,84 @@ def previous_iteration(
...
+        iterate.
+
+        This attribute is only available in operators representing previous time steps.
```
IvarStefansson: Fix comment.

src/porepy/numerics/ad/operators.py
```diff
@@ -1165,6 +1206,84 @@ def previous_iteration(
...
+
+    def reference(self: _ReferenceOperator) -> _ReferenceOperator:
+        """Returns a copy of the reference operator with an advanced time-step
+        index.
+
+        Reference operators do not invoke the recursion (like the base class),
+        but represent a leaf in the recursion tree.
+
+        Note:
+            You cannot create operators at the previous time step from operators which
```
IvarStefansson: Misplaced note

src/porepy/numerics/ad/operators.py
```diff
@@ -1165,6 +1206,84 @@ def previous_iteration(
...
+            are at some previous iterate. Use the :attr:`original_operator` instead.
+
+        Raises:
+            ValueError: If this instance represents an operator at a previous iterate.
+            AssertionError: If ``steps`` is not strictly positive.
```
IvarStefansson: Delete

src/porepy/numerics/ad/operators.py
```diff
@@ -1165,6 +1206,84 @@ def previous_iteration(
...
+            return self
+
+        # TODO copy or deepcopy? Is this enough for every operator class?
+        op = copy.copy(self)
+        # Delete the cached key, so that this must be regenerated for the new operator,
+        # which is different from the original one.
+        op._cached_key = None
+        op._is_reference = True
+
+        if isinstance(self, TimeDependentOperator) and self.is_previous_time:
```
IvarStefansson: Can we explain what this means practically and conceptually?

src/porepy/numerics/ad/operators.py
```diff
@@ -1443,6 +1562,7 @@ def _key(self) -> str:
                 f"(time_dependent_dense_array, name={self.name}, domains={domain_ids} "
                 f"previous_time={self.is_previous_time},"
                 f"time_step_index={self.time_step_index})"
+                f"reference={self.is_reference})"
```
IvarStefansson, line 1186: How does this render?

src/porepy/numerics/ad/operators.py
```diff
@@ -1394,7 +1513,7 @@ def parse(self, mdg: pp.MixedDimensionalGrid) -> np.ndarray:
         return self._values
 
 
-class TimeDependentDenseArray(TimeDependentOperator):
+class TimeDependentDenseArray(TimeDependentOperator, ReferenceOperator):
```
IvarStefansson: Have you thought through inheritance order? Are there common methods?

src/porepy/numerics/ad/operators.py
```diff
@@ -1578,7 +1705,7 @@ def set_value(self, value: float) -> None:
         self._value = value
 
 
-class Variable(TimeDependentOperator, IterativeOperator):
+class Variable(TimeDependentOperator, IterativeOperator, ReferenceOperator):
```
IvarStefansson: Consider MRO

IvarStefansson (commented):
Main comment: Think about precendence for reference - time step - iteration

src/porepy/compositional/materials.py
keileg: https://github.com/pmgbergen/porepy/issues/1697

src/porepy/numerics/ad/ad_utils.py
keileg: I would say no, I prefer to have the code crash on me. @jwboth what do you think?

keileg: Update after a chat with @jwboth: A default value will make it possible to take the perturbations of general operators, without having to explicitly set a zero reference on all leaves (in the operator tree sense). This seems like a nice feature to have and will be useful for the initialization. The original implementation will not work, since under parsing this may lead to arithmetic operations with incompatible operands. The solution we ended up with for now is implemented below; this is not elegant and could have been less computationally expensive, but it should do the job for now.

src/porepy/numerics/ad/operators.py
jwboth: Only __init__ which I stumbled upon. Why do `TimeDependentOperator` and `IterativeOperator`, `ReferenceOperator` functions of `__init_`? I did not double check, but arent they never called?

src/porepy/numerics/ad/ad_utils.py
IvarStefansson: Which types of leaves will this be invoked for? I'm not certain I grasp all possible consequences of this.

src/porepy/numerics/ad/operators.py
keileg: This is in TimeDependentOperator and should be fine?

src/porepy/numerics/ad/operators.py
keileg: Should be better now.

src/porepy/numerics/ad/operators.py
keileg: This and the below if-statement can never be true. Will delete.

keileg: .. because they will have hit identical ifs with return statements above, that is.

src/porepy/numerics/ad/operators.py
keileg: I made the time dependent and reference operators into mixins since they are pure extensions of a 'core' Operator. That should take care of any issues with inheritance order etc.

src/porepy/numerics/ad/operators.py
keileg: See above comment regarding the use of mixins

src/porepy/numerics/ad/ad_utils.py
keileg: Moved all get and set methods to a new module, will document at the module level there.

src/porepy/numerics/ad/ad_utils.py
keileg: @jwboth could you please comment on this one.

src/porepy/numerics/ad/_operator_states.py
```diff
@@ -0,0 +1,422 @@
...
+The mixins will modify how an operator is evaluated:
+  - TimeDependentOperator: The operator represents a previous time step, and its value
+    will be fetched from the time step which it represents, see class documentation for
+    details.
+  - IterativeOperator: The operator represents a previous iterate, and its value will be
+    fetched from the iterate which it represents, see class documentation for details.
+  - ReferenceOperator: The operator represents a reference value and will evaluate this
+    reference.
+
+When combining the mixins, their behavior is prioritized as follows:
```
keileg, line 12: This list, which hopefully is complete, defines the prioritization logic of various operator states. Please review carefully and keep in mind when assessing the rest of the code.

src/porepy/models/constitutive_laws.py
```diff
@@ -4741,6 +4741,68 @@ def porosity(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
         return Scalar(self.solid.porosity, "porosity")
 
 
+class CompressibleRockPorosity(pp.PorePyModel):
```
keileg: @IvarStefansson I have not taken a stand on what to do with this class. My initial thought is, do we really need it, or is it 'nice to have'? If its main purpose is as a testing ground for initialization routines, there is information in that.

src/porepy/numerics/ad/get_set_values.py
```diff
@@ -0,0 +1,353 @@
+"""This module contains helper functions for setting, getting and shifting values in the
```
keileg, line 1: This module is the new home for get and set methods (moved from ad_utils). I think that making a move was the right decision, whether this is the right placement is less clear.

keileg:
I have done a partial re-implementation of this PR. Main points:
1. Iterate, TimeDependent and Reference operators are now mixin classes. They have been moved to a separate module, operator_states.
2. The test suite for these mixin classes has been reworked. The test coverage should be better, as should test documentation.
3. The get- and set-methods for variable states were moved out of ad_utils to a separate file.
4. The rules of precedence for taking reference, previous iterate and previous time step have been revised and documented at the module level in operator states. I believe the rules are implemented consistently throughout the code base, but this should be carefully examined in the review.

src/porepy/numerics/ad/_ad_parser.py
```diff
@@ -197,7 +197,7 @@ def _evaluate_single(
         #    them according to the operator.
         if op.is_leaf():
             if isinstance(op, pp.ad.MixedDimensionalVariable):
-                if op.is_previous_iterate or op.is_previous_time:
+                if op.is_previous_iterate or op.is_previous_time or op.is_reference:
```
IvarStefansson, line 200: Should we consider introducing op.is_leaf()?

src/porepy/numerics/ad/get_set_values.py
IvarStefansson: Could you explain why these public functions are in a private module?

src/porepy/numerics/ad/_operator_states.py
```diff
@@ -0,0 +1,423 @@
...
+  - Calling the method reference() on an operator which is a ReferenceOperator will
+    return the operator in a reference state, independent of the original state (default
+    state, previous iterate, previous timestep) of the operator.
+  - Taking the previous time step or previous iterate of an operator which is a
+    ReferenceOperator will return the operator itself. Taking the reference is a one-way
```
IvarStefansson:
```suggestion
    ReferenceOperator will return the reference operator itself. Taking the reference is a one-way
```

src/porepy/numerics/ad/operators.py
```diff
@@ -1805,13 +1556,15 @@ def __init__(self, variables: list[Variable]) -> None:
...
         current_iter = []
+        reference = []
         names = []
         domains = []
 
         for var in variables:
             time_indices.append(var.time_step_index)
             iter_indices.append(var.iterate_index)
             current_iter.append(var.is_current_iterate)
+            reference.append(var.is_reference)
```
IvarStefansson, line 1567: Should we check that len set reference == 1 below?

src/porepy/models/abstract_equations.py
```diff
@@ -506,20 +506,30 @@ def create_variables(self) -> None:
         """
         pass
 
-    def perturbation_from_reference(self, name: str, grids: list[pp.Grid]):
-        """Perturbation of some quantity ``name`` from its reference value.
+    def perturbation_from_thermodynamic_state(self, name: str, grids: list[pp.Grid]):
```
IvarStefansson, line 509: Note on naming: My changes introduced the distinction model.perturbation_from_thermodynamic_state and operator.perturbation_from_reference. The latter is currently semi-explicitly understood as the reference configuration for mechanics-related linearization. We could consider making that explicit. I don't like the thought of doing that in the operator method. But we could wrap in a helper (function or model method?)   perturbation_from_mechanics_linearization_pleasefindbettername() returning the operator perturbation. No idea if this helps, but could facilitate future "other" reference states.

tests/numerics/ad/test_operator_states.py
```diff
@@ -0,0 +1,386 @@
...
+    def test_time_and_iterative_repeated_calls(
+        self, time_dependent: bool, operator_type: str
+    ):
+        """Verify the behavior of repeated calls to previous_timestep and
+        previous_iteration."""
+        operator = self._get_operator(operator_type)
+
+        index_key, _, prev_method, _ = self._get_index_keys(time_dependent)
+
+        for sd, data in self.mdg.subdomains(return_data=True):
```
IvarStefansson: The heck happened to the values we iterate over?

tests/numerics/ad/test_operator_states.py
```diff
@@ -0,0 +1,386 @@
...
+        if state == "current":
+            val = self.iter_value
+        elif state == "previous_timestep":
+            val = self.time_value
+        elif state == "previous_iteration":
+            val = self.iter_value
+        else:
+            raise NotImplementedError(f"State {state} not implemented.")
+
+        val = self._adjust_combined_expectence(val, op_type)
```
IvarStefansson: The logic of these expectations is a bit heavy to my taste. I can live with it, but would prefer a more straightforward implementation and/or logic.

tests/numerics/ad/test_operator_states.py
```diff
@@ -0,0 +1,386 @@
...
+        return val, jac_val
+
+    def _change_reference_values(self, op_type, change_type):
+        """Helper method to administer changes to the reference values and return the
+        expected value according to the change type.
+        """
+        vec = np.ones(self.num_cells)
+
+        if change_type is "delete":
+            # TODO: Should we rather erase here, so that the reference will return
```
IvarStefansson: Please resolve.

IvarStefansson (commented):
Mainly very good. Not a huge fan of the new tests, though. Please have a look, then we can discuss if needed.

src/porepy/models/abstract_equations.py
keileg: You may have a point, I'm not sure. What if we keep the code as is for now and introduce the helper if the need arise?

src/porepy/numerics/ad/_ad_parser.py
keileg: I guess you mean 'is_current' or similar? We use this combination twice in this module and, from what I can tell, that is it. So introduce if we use it more later?

src/porepy/numerics/ad/get_set_values.py
keileg: Not really. Will make public.

tests/numerics/ad/test_operator_states.py
keileg: Thanks.

tests/numerics/ad/test_operator_states.py
keileg: Had already implemented the erasure. Delete comment.

tests/numerics/ad/test_operator_states.py
keileg: This is a fair point, but it is also the most straightforward way I found to make meaningful tests that 
i) probe the nuances between the various states
ii) cover different types of operators
iii) exploit synergies between the setups and also parametrizations.
I am afraid that a more straightforward setup would require much more code and, in the main, a heavier maintenance cost. Still, I will see if the logic can be marginally simplified and certainly add more comments.

IvarStefansson (approved):

PR #1724, opened 2026-07-21
Cleanup equation system and introduce indexers

Yuriyzabegaev:
The goal of this PR is to move all DoF management away from the equation system and introduce `EquationIndexer` and `VariableIndexer` objects, which now control it. This is done as a preparation step for introducing nonlinear solvers that work on a subset of equations and variables, and need their own indexers.

Breaking API changes:
- removed `equation_system.assembled_equation_indices`. Introduced a method `equation_system.construct_assembled_matrix_indexers(equations, variables)`, which constructs indices of a subsystem produced by `assemble(equations, variables)`. 
- `equation_system.equation_image_space_composition` prints a deprecation warning but it still available. The only difference - it does not store equations defined on none domains. This might be a breaking change for a user code like ```for eq in equation_system.equations: ... = equation_image_space_composition[eq]```. 

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/fracture_damage.py
+++ b/src/porepy/models/fracture_damage.py
@@ -293,8 +293,9 @@ class DilationDamageEquation(FractureDamageEquations):
         """Update the dilation damage equation to include new term."""
         super().before_nonlinear_loop()
         fractures = self.mdg.subdomains(dim=self.nd - 1)
-        self.equation_system._equations[self.dilation_damage_equation_name] = (
-            self.dilation_damage_equation(fractures)
+        self.equation_system.update_equation(
+            equation_name=self.dilation_damage_equation_name,
+            new_equation=self.dilation_damage_equation(fractures),
         )
 
     def dilation_damage_equation(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
--- a/src/porepy/models/metric.py
+++ b/src/porepy/models/metric.py
@@ -183,8 +183,7 @@ class VariableBasedLebesgueMetric(LebesgueMetric):
             dict[str, float]: measure of values for each variable block.
 
         """
-        equation_system = self.model.equation_system
-        variable_indexer = equation_system.variable_indexer
+        variable_indexer = self.variable_indexer
 
         # Sanity check: Ensure that variables are defined on cells.
         for variable in variable_indexer.variable_dofs:
@@ -194,7 +193,7 @@ class VariableBasedLebesgueMetric(LebesgueMetric):
                     """variables defined on cells."""
                 )
 
-        norms = {v.name: 0.0 for v in self.model.equation_system.variables}
+        norms = {v.name: 0.0 for v in variable_indexer.variable_dofs}
 
         for variable, indices in variable_indexer.variable_dofs.items():
             variable_values = pp.ad.DenseArray(values[indices])
--- a/src/porepy/models/metric.py
+++ b/src/porepy/models/metric.py
@@ -96,6 +96,8 @@ class EquationBasedEuclideanMetric(EuclideanMetric):
         equation_indexer: Optional[pp.ad.EquationIndexer] = None,
     ) -> None:
         self.model = model
+        # TODO YZ: Norms should accept equation tags instead of the indexer. This will
+        # be addressed in the PR downstream.
         if equation_indexer is None:
             equation_indexer = model.equation_system.equation_indexer
         self.equation_indexer = equation_indexer
@@ -186,18 +188,18 @@ class VariableBasedLebesgueMetric(LebesgueMetric):
         variable_indexer = self.variable_indexer
 
         # Sanity check: Ensure that variables are defined on cells.
-        for variable in variable_indexer.variable_dofs:
+        for variable in variable_indexer.indices:
             if not variable._faces == 0 and variable._nodes == 0:
                 raise NotImplementedError(
                     """VariableBasedLebesgueMetric currently only supports """
                     """variables defined on cells."""
                 )
 
-        norms = {v.name: 0.0 for v in variable_indexer.variable_dofs}
+        norms = {v.name: 0.0 for v in variable_indexer.indices}
 
-        for variable, indices in variable_indexer.variable_dofs.items():
+        for variable, indices in variable_indexer.indices.items():
             variable_values = pp.ad.DenseArray(values[indices])
-            dim = variable.dof_info["cells"]  # + variable._faces + variable._nodes,
+            dim = variable.dof_info["cells"]
             domains: pp.GridLikeSequence = [variable.domain]  # type: ignore[assignment]
             norms[variable.name] += (
                 self._lebesgue2_norm(variable_values, dim, domains) ** 2
@@ -226,6 +228,8 @@ class EquationBasedLebesgueMetric(LebesgueMetric):
         equation_indexer: Optional[pp.ad.EquationIndexer] = None,
     ) -> None:
         super().__init__(model)
+        # TODO YZ: Norms should accept equation tags instead of the indexer. This will
+        # be addressed in the PR downstream.
         if equation_indexer is None:
             equation_indexer = model.equation_system.equation_indexer
         self.equation_indexer = equation_indexer
--- a/src/porepy/models/metric.py
+++ b/src/porepy/models/metric.py
@@ -5,6 +5,7 @@ From plain Euclidean norms to model-specific L2 norms of states and equations.
 """
 
 from functools import partial
+from typing import cast
 
 import numpy as np
 
@@ -89,8 +90,9 @@ class EquationBasedEuclideanMetric(EuclideanMetric):
 
     """
 
-    def __init__(self, model) -> None:
+    def __init__(self, model, equation_indexer: pp.ad.EquationIndexer) -> None:
         self.model = model
+        self.equation_indexer = equation_indexer
 
     def __call__(self, values: np.ndarray) -> dict[str, float]:  # type: ignore[override]
         """Compute the Euclidean norm of each separate equation.
@@ -103,8 +105,13 @@ class EquationBasedEuclideanMetric(EuclideanMetric):
 
         """
         norms = {}
-        equation_blocks = self.model.equation_system.assembled_equation_indices
-        for name, indices in equation_blocks.items():
+        equation_image_space_composition = (
+            self.equation_indexer.equation_image_space_composition
+        )
+        offset = 0
+        for name, blocks in equation_image_space_composition.items():
+            indices = np.concatenate(list(blocks.values())) + offset
+            offset += len(indices)
             norms[name] = self._euclidean_norm(values[indices])
         return norms
 
@@ -192,6 +199,10 @@ class EquationBasedLebesgueMetric(LebesgueMetric):
 
     """
 
+    def __init__(self, model, equation_indexer: pp.ad.EquationIndexer) -> None:
+        super().__init__(model)
+        self.equation_indexer = equation_indexer
+
     def __call__(self, values: np.ndarray) -> dict[str, float]:
         """Compute the Lebesgue L2 norm of each separate equation.
 
@@ -202,33 +213,24 @@ class EquationBasedLebesgueMetric(LebesgueMetric):
             dict[str, float]: measure of values for each equation block
 
         """
-        norms = {name: 0.0 for name in self.model.equation_system.equations}
-
-        equation_blocks = {}
-        for name in self.model.equation_system.equations:
-            if name not in self.model.equation_system.assembled_equation_indices:
+        equation_system = self.model.equation_system
+        equation_image_space_composition = (
+            self.equation_indexer.equation_image_space_composition
+        )
+        norms = {name: 0.0 for name in equation_image_space_composition}
+        offset = 0
+        for name, blocks in equation_image_space_composition.items():
+            indices = np.concatenate(list(blocks.values())) + offset
+            offset += len(indices)
+            if indices.size == 0:
                 continue
-            indices = self.model.equation_system.assembled_equation_indices[name]
-            if len(indices) == 0:
-                continue
-            equation_blocks[name] = (
-                indices,
-                list(
-                    self.model.equation_system.equation_image_space_composition[
-                        name
-                    ].keys()
-                ),
-                self.model.equation_system.equation_image_size_info[name]["cells"],
+            domains = cast(pp.GridLikeSequence, list(blocks.keys()))
+            equation_dim = equation_system.equation_image_size_info[name]["cells"]
+            equation_values = values[indices].reshape((equation_dim, -1), order="F")
+            cell_weights = np.hstack([domain.cell_volumes for domain in domains])
+            intensive_equation_values = pp.ad.DenseArray(
+                np.linalg.norm(equation_values, ord=2, axis=0) / cell_weights
             )
-        for name, (indices, sd, eq_dim) in equation_blocks.items():
-            if len(sd) == 0:
-                norms[name] = 0.0
-            else:
-                equation_values = values[indices].reshape((eq_dim, -1), order="F")
-                cell_weights = np.hstack([_sd.cell_volumes for _sd in sd])
-                intensive_equation_values = pp.ad.DenseArray(
-                    np.linalg.norm(equation_values, ord=2, axis=0) / cell_weights
-                )
-                norms[name] = self._lebesgue2_norm(intensive_equation_values, 1, sd)
+            norms[name] = self._lebesgue2_norm(intensive_equation_values, 1, domains)
 
         return norms
--- a/src/porepy/models/metric.py
+++ b/src/porepy/models/metric.py
@@ -5,7 +5,7 @@ From plain Euclidean norms to model-specific L2 norms of states and equations.
 """
 
 from functools import partial
-from typing import cast
+from typing import Optional, cast
 
 import numpy as np
 
@@ -42,13 +42,24 @@ class EuclideanMetric:
 
 
 class VariableBasedEuclideanMetric(EuclideanMetric):
-    """Plain Euclidean norm for variables and equations, computed per variable and
-    equation block.
+    """Plain Euclidean norm for variables, computed per variable block.
+
+    Parameters:
+        model: The model used to compute the metric.
+        variable_indexer: Define a subset of variables to evaluate the metric on. If
+            None (default), uses all variables.
 
     """
 
-    def __init__(self, model) -> None:
+    def __init__(
+        self,
+        model: pp.PorePyModel,
+        variable_indexer: Optional[pp.ad.VariableIndexer] = None,
+    ) -> None:
         self.model = model
+        if variable_indexer is None:
+            variable_indexer = model.equation_system.variable_indexer
+        self.variable_indexer = variable_indexer
 
     def __call__(self, values: np.ndarray) -> dict[str, float]:  # type: ignore[override]
         """Compute the Euclidean norm of each separate variable.
@@ -60,38 +71,33 @@ class VariableBasedEuclideanMetric(EuclideanMetric):
             dict[str, float]: measure of values for each variable block
 
         """
-        # Collect variable blocks based on variable names
-        variable_names = [
-            variable.name for variable in self.model.equation_system.variables
-        ]
-        variable_blocks = {
-            (variable.name, variable.domain): (
-                self.model.equation_system.dofs_of([variable])
-            )
-            for variable in self.model.equation_system.variables
-        }
-        concatenated_variable_blocks: dict[str, list[int]] = {
-            name: [] for name in set(variable_names)
-        }
-        for (name, _), indices in variable_blocks.items():
-            concatenated_variable_blocks[name].extend(indices)
-
         # Compute norms for each variable block
-        norms = {name: 0.0 for name in set(variable_names)}
-        for name, indices in concatenated_variable_blocks.items():
+        norms: dict[str, float] = {}
+        for name, domains_dofs in self.variable_indexer.group_by_name().items():
+            indices = np.concatenate(list(domains_dofs.values()))
             norms[name] = self._euclidean_norm(values[indices])
 
         return norms
 
 
 class EquationBasedEuclideanMetric(EuclideanMetric):
-    """Plain Euclidean norm for variables and equations, computed per variable and
-    equation block.
+    """Plain Euclidean norm for equations, computed per equation block.
+
+    Parameters:
+        model: The model used to compute the metric.
+        equation_indexer: Define a subset of equations to evaluate the metric on. If
+            None (default), uses all equations.
 
     """
 
-    def __init__(self, model, equation_indexer: pp.ad.EquationIndexer) -> None:
+    def __init__(
+        self,
+        model: pp.PorePyModel,
+        equation_indexer: Optional[pp.ad.EquationIndexer] = None,
+    ) -> None:
         self.model = model
+        if equation_indexer is None:
+            equation_indexer = model.equation_system.equation_indexer
         self.equation_indexer = equation_indexer
 
     def __call__(self, values: np.ndarray) -> dict[str, float]:  # type: ignore[override]
@@ -105,19 +111,14 @@ class EquationBasedEuclideanMetric(EuclideanMetric):
 
         """
         norms = {}
-        equation_image_space_composition = (
-            self.equation_indexer.equation_image_space_composition
-        )
-        offset = 0
-        for name, blocks in equation_image_space_composition.items():
-            indices = np.concatenate(list(blocks.values())) + offset
-            offset += len(indices)
+        for name, domains_dofs in self.equation_indexer.group_by_name().items():
+            indices = np.concatenate(list(domains_dofs.values()))
             norms[name] = self._euclidean_norm(values[indices])
         return norms
 
 
 class LebesgueMetric:
-    def __init__(self, model) -> None:
+    def __init__(self, model: pp.PorePyModel) -> None:
         self.model = model
 
     def _lebesgue2_norm(
@@ -153,7 +154,24 @@ class LebesgueMetric:
 
 
 class VariableBasedLebesgueMetric(LebesgueMetric):
-    """Lebesgue L2 norm for variables and equations, computed per variable block."""
+    """Lebesgue L2 norm for variables, computed per variable block.
+
+    Parameters:
+        model: The model used to compute the metric.
+        variable_indexer: Define a subset of variables to evaluate the metric on. If
+            None (default), uses all variables.
+
+    """
+
+    def __init__(
+        self,
+        model: pp.PorePyModel,
+        variable_indexer: Optional[pp.ad.VariableIndexer] = None,
+    ) -> None:
+        super().__init__(model)
+        if variable_indexer is None:
+            variable_indexer = model.equation_system.variable_indexer
+        self.variable_indexer = variable_indexer
 
     def __call__(self, values: np.ndarray) -> dict[str, float]:
         """Compute the Lebesgue L2 norm of each separate variable.
@@ -165,8 +183,11 @@ class VariableBasedLebesgueMetric(LebesgueMetric):
             dict[str, float]: measure of values for each variable block.
 
         """
+        equation_system = self.model.equation_system
+        variable_indexer = equation_system.variable_indexer
+
         # Sanity check: Ensure that variables are defined on cells.
-        for variable in self.model.equation_system.variables:
+        for variable in variable_indexer.variable_dofs:
             if not variable._faces == 0 and variable._nodes == 0:
                 raise NotImplementedError(
                     """VariableBasedLebesgueMetric currently only supports """
@@ -174,17 +195,13 @@ class VariableBasedLebesgueMetric(LebesgueMetric):
                 )
 
         norms = {v.name: 0.0 for v in self.model.equation_system.variables}
-        variable_blocks = {
-            (variable.name, variable.domain): (
-                self.model.equation_system.dofs_of([variable]),
-                variable._cells,  # + variable._faces + variable._nodes,
-            )
-            for variable in self.model.equation_system.variables
-        }
-        for (name, sd), (indices, variable_dim) in variable_blocks.items():
+
+        for variable, indices in variable_indexer.variable_dofs.items():
             variable_values = pp.ad.DenseArray(values[indices])
-            norms[name] += (
-                self._lebesgue2_norm(variable_values, variable_dim, [sd]) ** 2
+            dim = variable.dof_info["cells"]  # + variable._faces + variable._nodes,
+            domains: pp.GridLikeSequence = [variable.domain]  # type: ignore[assignment]
+            norms[variable.name] += (
+                self._lebesgue2_norm(variable_values, dim, domains) ** 2
             )
         for name in norms:
             norms[name] = np.sqrt(norms[name])
@@ -197,10 +214,21 @@ class EquationBasedLebesgueMetric(LebesgueMetric):
 
     NOTE: Assumes equations are intensive quantities and defined only on cells.
 
+    Parameters:
+        model: The model used to compute the metric.
+        equation_indexer: Define a subset of equations to evaluate the metric on. If
+            None (default), uses all equations.
+
     """
 
-    def __init__(self, model, equation_indexer: pp.ad.EquationIndexer) -> None:
+    def __init__(
+        self,
+        model: pp.PorePyModel,
+        equation_indexer: Optional[pp.ad.EquationIndexer] = None,
+    ) -> None:
         super().__init__(model)
+        if equation_indexer is None:
+            equation_indexer = model.equation_system.equation_indexer
         self.equation_indexer = equation_indexer
 
     def __call__(self, values: np.ndarray) -> dict[str, float]:
@@ -214,17 +242,10 @@ class EquationBasedLebesgueMetric(LebesgueMetric):
 
         """
         equation_system = self.model.equation_system
-        equation_image_space_composition = (
-            self.equation_indexer.equation_image_space_composition
-        )
-        norms = {name: 0.0 for name in equation_image_space_composition}
-        offset = 0
-        for name, blocks in equation_image_space_composition.items():
-            indices = np.concatenate(list(blocks.values())) + offset
-            offset += len(indices)
-            if indices.size == 0:
-                continue
-            domains = cast(pp.GridLikeSequence, list(blocks.keys()))
+        norms: dict[str, float] = {}
+        for name, grids_dofs in self.equation_indexer.group_by_name().items():
+            indices = np.concatenate(list(grids_dofs.values()))
+            domains = cast(pp.GridLikeSequence, list(grids_dofs.keys()))
             equation_dim = equation_system.equation_image_size_info[name]["cells"]
             equation_values = values[indices].reshape((equation_dim, -1), order="F")
             cell_weights = np.hstack([domain.cell_volumes for domain in domains])
--- a/src/porepy/numerics/ad/equation_system.py
+++ b/src/porepy/numerics/ad/equation_system.py
@@ -1268,8 +1268,7 @@ class EquationSystem:
                 defined.
 
         Returns:
-            A dictionary with the index set of the restricted equations (referring to
-            equation rows) as values. If no restriction is given, the value is None.
+            A list of atomic equation-domain identifiers.
 
         """
         equations_on_domains: list[pp.ad.EquationOnDomain] = []
@@ -1562,14 +1561,14 @@ class EquationSystem:
         if variables is not None:
             # Respect the ordering of the input list of variables.
             variable_indexer = self.variable_indexer.construct_restricted_indexer(
-                self._parse_variable_type(variables=variables, ordered=False)
+                self._parse_variable_type(variables=variables, ordered=True)
             )
         else:
             variable_indexer = self.variable_indexer
 
         if equations is not None:
             equation_indexer = self.equation_indexer.construct_restricted_indexer(
-                self._parse_equations(equations=equations, ordered=False)
+                self._parse_equations(equations=equations, ordered=True)
             )
         else:
             equation_indexer = self.equation_indexer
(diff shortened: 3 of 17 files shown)


src/porepy/models/fracture_damage.py
```diff
@@ -293,8 +293,9 @@ def before_nonlinear_loop(self):
         """Update the dilation damage equation to include new term."""
         super().before_nonlinear_loop()
         fractures = self.mdg.subdomains(dim=self.nd - 1)
-        self.equation_system._equations[self.dilation_damage_equation_name] = (
-            self.dilation_damage_equation(fractures)
+        self.equation_system.update_equation(
```
Yuriyzabegaev, line 296: Was it intentional that it updated this equation directly and not through `update_equation`? Had to change to make the tests pass.

src/porepy/models/metric.py
```diff
@@ -192,6 +199,10 @@ class EquationBasedLebesgueMetric(LebesgueMetric):
 
     """
 
+    def __init__(self, model, equation_indexer: pp.ad.EquationIndexer) -> None:
+        super().__init__(model)
+        self.equation_indexer = equation_indexer
```
Yuriyzabegaev, line 235: equation-based convergence metrics have to work for a subset of equations. They worked before by relying on `assembled_equation_indices`. Replaced it by explicitly passing them the indexer in the constructor. If the residual vector they will receive is restricted, the indexer they accept in initialization also must be restricted.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -184,35 +159,21 @@ def __init__(self, mdg: pp.MixedDimensionalGrid) -> None:
...
-
-        This is an optimized structure, meaning the order of entries is created in
-        :meth:`_cluster_dofs_gridwise`.
+        self._variable_indexer: pp.ad.VariableIndexer | None = None
+        """Indexer defining the ordering of the variables (columns) when multiple
+        variable values are packed in a single vector.
 
         """
-
-        self._variable_num_dofs: np.ndarray = np.array([], dtype=int)
```
Yuriyzabegaev, line 200: Removed all attributes related to dofs bookkeeping. They are now indexers' responsibility.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -275,33 +236,18 @@ def SubSystem(
...
-
-                # Update variable numbers.
-                new_equation_system._variable_dof_type[variable.id] = (
-                    self._variable_dof_type[variable.id]
-                )
-
-                # Create DOFs.
-                new_equation_system._append_dofs(variable)
+        # Loop over new system's variables, old DOF ordering is preserved.
+        for variable in variables:
```
Yuriyzabegaev, line 231: Here and in many places below, the original pattern was to loop over all registered variables and do something for a requested subset of variables. This is so to preserve the internal order, and neglect the input order. I centralized this logic in `self._parse_variable_type(..., ordered=True)`, which now returns correctly ordered subset of variables, so we can iterate over them instead.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -321,8 +267,17 @@ def equation_image_space_composition(
         and their block indices in the global system, for every equation
         set in this EquationSystem.
 
+        Note: It does not include equations with empty domains.
+
         """
-        return self._equation_image_space_composition
+        warn(
```
Yuriyzabegaev, line 264: Note: equation_image_space_composition is deprecated

src/porepy/numerics/ad/equation_system.py
```diff
@@ -348,6 +303,159 @@ def variable_domains(self) -> list[pp.GridLike]:
...
+                    dofs_per_grid = (
+                        domain.num_cells * dofs_info.get("cells", 0)  # cells
+                        + domain.num_faces * dofs_info.get("faces", 0)  # faces
+                        + domain.num_nodes * dofs_info.get("nodes", 0)  # nodes
+                    )
+                elif isinstance(domain, pp.MortarGrid):
+                    # Mortar grid has no faces.
+                    dofs_per_grid = (
+                        domain.num_cells * dofs_info.get("cells", 0)  # cells
+                        + domain.num_nodes * dofs_info.get("nodes", 0)  # nodes
```
Yuriyzabegaev, line 385: Before, the interface equation dofs were computed as `domain.num_cells * dofs_info.get("cells", 0)` only (no nodal dofs). I realize there are no such practical cases, but I think it was a bug.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1053,68 +1042,18 @@ def dofs_of(self, variables: VariableList) -> np.ndarray:
...
-
-        # Concatenate indices, if any
-        if len(indices) > 0:
-            all_indices = np.concatenate(indices, dtype=int)
-        else:
-            all_indices = np.array([], dtype=int)
-
-        return all_indices
-
-    def identify_dof(self, dof: int) -> Variable:
```
Yuriyzabegaev, line 1084: I removed this method since it is not used anywhere and not tested. Should I provide a similar method for `VariableIndexer`?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1203,55 +1133,35 @@ def set_equation(
...
-                image_info.update({sd: block_idx})
-                # Remove the subdomain from the domain list.
-                # Ignore mypy error here, since we know that sd is in grids.
-                grids.remove(sd)  # type: ignore
-
-        for intf in self.mdg.interfaces():
-            if intf in grids:
-                # Equations on interfaces can only be defined on cells.
-                num_equ_per_grid = int(
-                    intf.num_cells * equations_per_grid_entity.get("cells", 0)
```
Yuriyzabegaev, line 1231: See my comment above about counting equation DoFs on interface grids.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -2212,3 +2094,41 @@ def __str__(self) -> str:
             s += "\n\t".join(eq_names) + "\n"
 
         return s
+
+
+def cluster_dofs_gridwise(variables: list[pp.ad.Variable]) -> list[pp.ad.Variable]:
```
Yuriyzabegaev, line 2028: Moved it here as a unitity function for now. Ideally, want to move it to the linear solver module.

src/porepy/numerics/ad/operators.py
```diff
@@ -1457,6 +1457,15 @@ def size(self) -> int:
             )
         raise ValueError()
 
+    @property
+    def dof_info(self) -> dict[Literal["cells", "faces", "nodes"], int]:
```
Yuriyzabegaev, line 1461: added this method to `Variable`, to make read access to this data public.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1203,55 +1133,35 @@ def set_equation(
...
         # Assert the equation is not defined on an unknown domain.
-        assert len(grids) == 0
+        known_domains = set(self.mdg.subdomains()) | set(self.mdg.interfaces())
+        unknown_domains = set(grids).difference(known_domains)
+        assert not unknown_domains, (
+            f"Equation defined on unknown domains: {unknown_domains}"
+        )
+
+        # Store the ordered equation domains. The equation may
+        # already contain domains, in which case make sure they are equal.
```
Yuriyzabegaev, line 1076: Important change. Now, it is the equation operator that is the only source of truth about what domains it is defined on. This is a preparation for the change where operators deducing their definition domains based on their children.

Yuriyzabegaev (commented):
A few details worth highlighting.

src/porepy/models/metric.py
```diff
@@ -60,38 +71,33 @@ def __call__(self, values: np.ndarray) -> dict[str, float]:  # type: ignore[over
...
-        concatenated_variable_blocks: dict[str, list[int]] = {
-            name: [] for name in set(variable_names)
-        }
-        for (name, _), indices in variable_blocks.items():
-            concatenated_variable_blocks[name].extend(indices)
-
         # Compute norms for each variable block
-        norms = {name: 0.0 for name in set(variable_names)}
-        for name, indices in concatenated_variable_blocks.items():
+        norms: dict[str, float] = {}
```
Yuriyzabegaev, line 75: I want to believe that the the operations with dof indices become less cumbersome with this update. Changes in this file is the illustration for this.

Yuriyzabegaev:
I also think that indexers require thought-through test cases, which I haven't written yet. I'm postponing this to after I get an approval on this change, to avoid doing useless work, if significant changes are requested.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -265,7 +265,7 @@ def equation_image_space_composition(
             "equation_system.equation_image_space_composition is deprecated and will be"
             " removed. See equation_system.equation_indexer, or "
             "equation_system.construct_assembled_matrix_indexers.",
-            category=DeprecationWarning,
+            category=FutureWarning,
```
Yuriyzabegaev, line 268: I discovered now that `DeprecationWarning` is not printed in stdout by default, if you don't configure logging, while `FutureWarning` does. Does this mean that nobody will ever see all the deprecation warnings we add in code? Or am I the only one that silly that I don't configure my logger to print deprecation warning in every file I run?

Yuriyzabegaev: To be precice, you need this in your every runscript:

`warnings.simplefilter("default", DeprecationWarning)`

src/porepy/models/fracture_damage.py
IvarStefansson: I can't see any reason why it shouldn't use the method. Do you know why you had to change to pass tests, though?

src/porepy/models/metric.py
IvarStefansson: Nice!

src/porepy/models/metric.py
```diff
@@ -158,26 +183,24 @@ def __call__(self, values: np.ndarray) -> dict[str, float]:
...
-            for variable in self.model.equation_system.variables
-        }
-        for (name, sd), (indices, variable_dim) in variable_blocks.items():
+        norms = {v.name: 0.0 for v in variable_indexer.variable_dofs}
+
+        for variable, indices in variable_indexer.variable_dofs.items():
             variable_values = pp.ad.DenseArray(values[indices])
-            norms[name] += (
-                self._lebesgue2_norm(variable_values, variable_dim, [sd]) ** 2
+            dim = variable.dof_info["cells"]  # + variable._faces + variable._nodes,
```
IvarStefansson: Clean?

src/porepy/viz/diagnostics_mixin.py
```diff
@@ -170,6 +180,22 @@ def get_max(mat: spmatrix, equation_name: str, variable_name: str) -> float:
             "Cannot diagnose a linear system whose matrix was freed."
         )
```
IvarStefansson, line 178: Noticed the grouping_ typing assignment above. What is your (and @keileg's) take on this pattern vs. using cast?

tests/functional/test_buoyancy_flow.py
```diff
@@ -96,12 +96,17 @@ def _run_buoyancy_model(
     geometry_class = geometry2d if dim == 2 else geometry3d
     model_class = add_mixin(geometry_class, model_class)
     model = model_class(model_params)
+    model.prepare_simulation()
```
IvarStefansson, line 100: Why is this split necessary? That looks like a design failure to me.

tests/models/test_metric.py
```diff
@@ -428,8 +453,8 @@ def test_equation_based_lebesgue_metric_with_model(random_polynomial_setup):
             "exp_y": random_polynomial_setup["exponents_y"],
         }
     )
-    m_eq = pp.EquationBasedLebesgueMetric(model)
     model.prepare_simulation()
+    m_eq = pp.EquationBasedLebesgueMetric(model, model.equation_system.equation_indexer)
```
IvarStefansson: Is it necessary to actively pass indexer here or would the default None work?

tests/numerics/ad/test_equation_system.py
```diff
@@ -976,6 +974,9 @@ def test_set_remove_equations(model: EquationSystemMockModel):
         grids=model.subdomains,
         equations_per_grid_entity=dof_info_subdomain,
     )
+    equation_subdomain_blocks = (
+        equation_system.equation_indexer.equation_image_space_composition
```
IvarStefansson, line 978: General style comment: I think having to split multiple lines like this would have warranted e.g.
`indexer = equation_system.equation_indexer`.
Not important and completely fine with keeping as is.

tests/numerics/ad/test_equation_system.py
```diff
@@ -1117,6 +1139,19 @@ def test_parse_variable_like(model: EquationSystemMockModel):
     )
 
 
+@pytest.mark.parametrize("ordered", [True, False])
+def test_parse_variable_type_rejects_unknown_variable(
```
IvarStefansson, line 1143: The test logic here relies on scope fixture, right? Since this is not quite trivial, please document.

src/porepy/numerics/ad/indexers.py
```diff
@@ -0,0 +1,246 @@
...
+        variables: dict[str, dict[pp.GridLike, np.ndarray]] = {}
+        for variable, dofs in self.variable_dofs.items():
+            if len(dofs) == 0:
+                continue
+            # Get by key variable.name, if not found, initialize it with an empty dict.
+            variables.setdefault(variable.name, {})[variable.domain] = dofs
+        return variables
+
+
+class EquationIndexer:
```
IvarStefansson, line 165: Throughout: I'd prefer relating the DoF concept directly to variables and not use it for equations and their indices.

src/porepy/numerics/ad/equation_system.py
IvarStefansson: I think we can safely say you're not the only one. I don't get why FutureWarning should be higher level than DepracationWarning. Anyway, let's discuss with @keileg.

src/porepy/numerics/ad/equation_system.py
IvarStefansson: What about faces?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -651,37 +693,27 @@ def get_variable_values(
...
-                val = pp.get_solution_values(
-                    variable.name,
-                    data_cache[domain],
-                    time_step_index=time_step_index,
-                    iterate_index=iterate_index,
-                    reference=reference,
-                )
-                # NOTE get_solution_values already returns a copy
-                values.append(val)
+        # Indexer determines ordering.
```
IvarStefansson: Am I being stupid or is (the placement of) this comment a bit strange?

src/porepy/numerics/ad/equation_system.py
IvarStefansson: Yes, I think so, unless there is a different convenient way of knowing which variable a dof corresponds to. Can be very useful when debugging.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1320,34 +1169,12 @@ def update_equation(
...
-        The method loops through the variables and updates the number of fine-scale
-        degree of freedom. The system size will be updated if the grid has changed or
-        (perhaps less realistically) a variable has had its number of dofs per grid
-        quantity changed.
-
-        NOTE: This method is experimental and should be used with caution. After this
-        method has been called, other attributes of the class that depend on the number
-        of dofs (such as _equation_image_space_composition) will be outdated and should
-        be used with care.
+        Used if the domain of definition of variables has been changed externally.
```
IvarStefansson: Is the name of this method intuitive to you?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1763,8 +1529,51 @@ def assemble(
...
+            variable_indexer = self.variable_indexer
+            # Respect the ordering of the input list of variables.
+            variables_ = self._parse_variable_type(variables=variables, ordered=True)
+            col_proj = [variable_indexer.variable_dofs[var] for var in variables_]
+            column_projection = (
+                np.concatenate(col_proj)
+                if len(col_proj) > 0
+                else np.empty(0, dtype=int)
+            )
+            A = A[:, column_projection]
```
IvarStefansson, line 1536: What is the efficiency of this slicing? Candidates: 
1. short-circuit if passed variables is in indexer's ordering. 
2. Convert tocsc, slice, tocsr. Don't know if/when that is more efficient.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1792,7 +1601,7 @@ def assemble_schur_complement_system(
         .. math::
 
             \left( A_{pp} - A_{ps} * A_{ss}^{-1} * A_{sp}\right) * x_p
-            = b_p - A_{ps} * A_{ss} * b_s
+            = b_p - A_{ps} * A_{ss}^{-1} * b_s
```
IvarStefansson, line 1599: Indeed!

IvarStefansson (commented):
Looks good to me. Some relatively minor comments. Deferring to @keileg for approval.

src/porepy/models/fracture_damage.py
Yuriyzabegaev: Because modifying `_equations` directly does not trigger recomputation of indexers

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: Interfaces have no `num_faces` property

tests/numerics/ad/test_equation_system.py
Yuriyzabegaev: Absolutely! I always prefer to split them, missed this one

src/porepy/models/fracture_damage.py
IvarStefansson: Got it!

src/porepy/numerics/ad/indexers.py
```diff
@@ -0,0 +1,247 @@
...
+    variables on multiple grids in a contiguous array.
+
+    For a data array with a different arrangement (e.g., produced by taking a subset of
+    variables), a new indexer needs to be constructed.
+
+    """
+
+    def __init__(self, variable_dofs: dict[pp.ad.Variable, np.ndarray]) -> None:
+        self.variable_dofs: dict[pp.ad.Variable, np.ndarray] = variable_dofs
+        """An ordered mapping of atomic variables to their DoF indices. The keys are
```
keileg: Are ordered or should be ordered?

src/porepy/numerics/ad/indexers.py
```diff
@@ -0,0 +1,247 @@
...
+
+        Returns:
+            an index array of `shape=(M,)`, where `0 <= M <= num_dofs`.
+
+        """
+        projections = []
+        for variable in variables:
+            dofs = self.variable_dofs.get(variable, None)
+            if dofs is None:
+                raise ValueError(
```
keileg, line 78: Please mention in docstring.

src/porepy/numerics/ad/indexers.py
```diff
@@ -0,0 +1,247 @@
...
+            projections.append(dofs)
+        return (
+            np.concatenate(projections)
+            if len(projections) > 0
+            else np.empty(0, dtype=int)
+        )
+
+    def construct_restricted_indexer(
+        self, variables: list[pp.ad.Variable]
+    ) -> "VariableIndexer":
```
keileg: Can't the string representation be avoided by the import of annotations?

src/porepy/numerics/ad/indexers.py
keileg: If you mean that 'DoF' means something different than what is needed for ordering and bookkeeping, I agree.

src/porepy/numerics/ad/indexers.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+class EquationIndexer:
+    """An equation indexer determines the arrangement of DoFs corresponding to multiple
+    equations on multiple grids in a contiguous array.
+
+    For a data array with a different arrangement (e.g., produced by taking a subset of
+    equations), a new indexer needs to be constructed.
+
+    Implementation note: There is an unfortunate asymmetry between this and
```
keileg, line 172: Is this asymmetry something we should address at some later point? I can never decide how far to pursue the resemblance between equations and variables.

src/porepy/numerics/ad/indexers.py
```diff
@@ -0,0 +1,247 @@
...
+                offset_per_equation += dofs.size
+            global_offset += offset_per_equation
+
+        self.equation_dofs: dict[EquationOnDomain, np.ndarray] = equation_dofs
+        """An ordered mapping of atomic equations to their DoF indices. The keys are
+        ordered, in the sense that if key A goes before key B, DoFs of key A are located
+        before DoFs of key B.
+
+        """
+        self.equation_image_space_composition: dict[
```
keileg, line 202: Is this the time to simplify this name, or should it happen later?

src/porepy/numerics/ad/indexers.py
```diff
@@ -0,0 +1,247 @@
...
+        Note: It does not include equations with empty domains.
+
+        """
+
+    def group_by_name(self) -> dict[str, dict[pp.GridLike, np.ndarray]]:
+        """Group :attr:`equation_dofs` by equation names.
+
+        Domains with no dofs are ignored.
+
+        Offset between equations is assumed.
```
keileg, line 221: What is meant by this?

src/porepy/numerics/ad/indexers.py
keileg: Even better, I see it is about to die. Hurrah!

src/porepy/numerics/ad/equation_system.py
```diff
@@ -348,6 +294,108 @@ def variable_domains(self) -> list[pp.GridLike]:
...
+        EquationSystem. If an equation is defined on multiple grids, the respective
+        row-block is internally ordered as given by the mixed-dimensional grid
+        (for sd in subdomains, for intf in interfaces).
+
+        """
+        if self._equation_indexer is None:
+            self._equation_indexer = self.construct_equation_indexer()
+        return self._equation_indexer
+
+    def construct_variable_indexer(self) -> pp.ad.VariableIndexer:
```
keileg: Should this be private? Just asking, no strong opinions.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -348,6 +294,108 @@ def variable_domains(self) -> list[pp.GridLike]:
...
+                        domain.num_cells * dofs_info.get("cells", 0)  # cells
+                        + domain.num_nodes * dofs_info.get("nodes", 0)  # nodes
+                    )
+                else:
+                    raise ValueError(f"Unknown domain type: {domain}")
+                dofs = np.arange(dofs_per_grid) + offset
+                dofs_on_domains[domain] = dofs
+                offset += len(dofs)
+
+            # Filter out equations not defined anywhere.
```
keileg, line 393: Let's talk through this just to be sure.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -348,6 +294,108 @@ def variable_domains(self) -> list[pp.GridLike]:
...
+
+    @property
+    def equation_indexer(self) -> pp.ad.EquationIndexer:
+        """Indexer defining the ordering of the equations (rows) when multiple
+        equation values are packed in a single vector.
+
+        Initialized lazily, upon first request. Changes in equation system may
+        invalidate the indexer, leading to its recomputation.
+
+        By default, the row blocks are in the same order as equations were added to this
```
keileg: Is it possible to change this default?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -726,35 +758,33 @@ def set_variable_values(
...
-                # 3. Move dissection forward.
-                dof_start = dof_end
+        # Normalize the variable input. Using equation_system's ordering of variables.
+        variables = self._parse_variable_type(variables, ordered=True)
+
+        for variable in variables:
+            # 1. Slice the vector to local size
+            num_dofs = variable.size
+            # Extract local vector.
+            # This will raise errors if indexation is out of range.
```
keileg: This is not true, both start and end indices may be out of range and numpy will just return whatever is meaningful based on those indices.

src/porepy/numerics/ad/equation_system.py
keileg: Something like 'reset_variable_indices'? Or also mention equations?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1028,17 +965,6 @@ def projection_to(self, variables: Optional[VariableList] = None) -> sps.csr_mat
         else:
             return sps.csr_matrix((0, num_dofs))
 
-    @lru_cache
```
keileg, line 1031: The point of caching here was that this construction actually takes non-negligible time in practical simulations. Do you have a feeling for whether the need for caching is no longer there?

keileg: I suspect lines 992-3 will give the same issue in the new code

src/porepy/numerics/ad/equation_system.py
keileg: Please add, this is very useful. It could be useful to have something similar for the EquationIndexer as well if that is meaningful.

src/porepy/numerics/ad/equation_system.py
keileg: Good!

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1763,8 +1529,51 @@ def assemble(
         # grid-related column blocks by using the transposed projection to respective
         # subspace.
         # Multiply rhs by -1 to move to the rhs.
```
keileg: Move this line to return statement.

src/porepy/viz/diagnostics_mixin.py
keileg: I don't like any of them (but realize we need to do something). I think I prefer cast, since it localizes the mental effort needed for parsing the meaning, compared to grouping_, which I may hit upon any place it is used.

src/porepy/numerics/ad/indexers.py
```diff
@@ -0,0 +1,247 @@
...
+    :class:`EquationOnDomain` to query DoFs and not
+    `dict[str, dict[pp.GridLike, np.ndarray]]`. See also the docstring of
+    :attr:`equation_image_space_composition`.
+
+    """
+
+    def __init__(
+        self, equation_image_composition: dict[str, dict[pp.GridLike, np.ndarray]]
+    ) -> None:
+        equation_dofs: dict[EquationOnDomain, np.ndarray] = {}
```
keileg: Can this be simplified to dofs (or perhaps better indices, cf Ivar's comment elsewhere).

keileg: Even better, is it meaningful to enable iterations over this attribute directly through the class?

keileg: Same with VariableIndexer, naturally.

src/porepy/models/metric.py
keileg: This requires the EquationIndexer to be avaialble at initialization of the metric, which will often be during the simulation setup. Do you see any troubles from this?

tests/numerics/ad/test_equation_system.py
```diff
@@ -1128,39 +1163,51 @@ def test_parse_single_equation(model: EquationSystemMockModel):
...
-        name = eq if isinstance(eq, str) else eq.name
+        name = eq_or_name if isinstance(eq_or_name, str) else eq_or_name.name
+        eq = equation_system.equations[name]
 
         # First parse the equation as it is, without any restriction.
         # This should give back the full equation with no restriction.
-        restriction_1 = equation_system._parse_single_equation(eq)
-        assert len(restriction_1) == 1
+        restriction_1 = equation_system._parse_equations([eq_or_name])
+        assert len(restriction_1) == 4
```
keileg, line 1194: What changed?

tests/numerics/ad/test_indexers.py
```diff
@@ -0,0 +1,195 @@
...
+    equation_b_second = EquationOnDomain("equation_b", second)
+    equation_b_third = EquationOnDomain("equation_b", third)
+    indexer = EquationIndexer(
+        equation_image_composition={
+            "equation_a": {first: np.arange(2), second: np.arange(2, 5)},
+            "equation_b": {second: np.arange(1), third: np.arange(1, 3)},
+        }
+    )
+
+    restricted = indexer.construct_restricted_indexer(
```
keileg, line 152: Please add one sentence explaining the intended logic here.

keileg (commented):
Mainly minor comments from my side.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: I suggest renaming it to `reset_variable_equation_indices`. 

Previously, it only mentioned variables and not equations, since equation_image_space_composition was constructed ad-hoc each time we assemble the jacobian.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: ok

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: I don't know how efficient columns slicing is compared to multiplying by a matrix to achieve the same thing. Before, it always was `return A * column_projection, -rhs_cat`, even if variables is None.

Now it only slices if variables are not None, which implies some restriction. So it already does the short-circuit you mention in 1.

I don't think it's worth optimizing more, since some PRs ahead, it computes the Jacobian with a sub-set of variables without a need for slicing at all.

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: Should be ordered. But I believe this assumption is not used anywhere.

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: done, thanks

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: seems like yes! thanks

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: A few PRs ahead, `EquationIndexer` and `VariableIndexer` will be united in a single class. `equation_dofs` and `variable_dofs` will become `dofs_of_operators`. By this logic, I suggest replacing:
- `equation_dofs` -> `equation_indices`
- `variable_dofs` -> `variable_indices`
- `dofs_of_operators` -> `indices` (I don't like using the word operator in this context)

Sounds good?

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: Yes, this asymmetry as addressed 1-2 PRs ahead!

Regarding the second statement, I convinced myself that what the indexers turned out to be after these PRs became quite quite convenient and concise. Let's see if you agree or not!

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: Possibly yes! Let's come back to "iterate over this attribute directly through the class" a few PRs later, when you see what the indexer API converged to?

src/porepy/viz/diagnostics_mixin.py
Yuriyzabegaev: Ok, will replace it with cast and follow your lead.

tests/functional/test_buoyancy_flow.py
Yuriyzabegaev: Some PRs ahead I introduce `pp.solvers.EquationTag`, which solves exactly this problem: to define a set of equations you operate on before initializing the model.

What I suggest to do here is to add a TODO and address it in the PR that introduces tags. Ok?

tests/models/test_metric.py
Yuriyzabegaev: You're right, default would work. Will revert it.

tests/numerics/ad/test_equation_system.py
Yuriyzabegaev: done

src/porepy/models/metric.py
Yuriyzabegaev: Ivar answered a similar question about test_metrics.py. I will copy my answer here:


Some PRs ahead I introduce pp.solvers.EquationTag, which solves exactly this problem: to define a set of equations you operate on before initializing the model.

What I suggest to do here is to add a TODO and address it in the PR that introduces tags. Ok?

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: By now it would be possible only by subclassing EquationSystem, which I don't like. If you feel like it, I can add a parameter callable, that rearranges the equations. But we are quite restricted in how you can rearrange the equations due to the AD machinery.

I think the right solution is to remove "by default" from the docstring.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: Yes, changing it to private.

src/porepy/models/metric.py
Yuriyzabegaev: I left it here (it was the same before the PR) to draw you attention to this. Is this intended?

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: Probably I can just remove this comment, since there's a similar one a few lines above.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: You're right. It's an old comment which I didn't touch. I will correct it.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: But old and new `dofs_of` still call `np.concatenate`. 

The method `_global_variable_dofs` is replaced by the variable indexer, and it's construction is cached. Let's discuss in person?

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: Added to VariableIndexer. Some PRs later, VariableIndexer and EquationIndexer will become the same thing, so it will automatically propagate on it.

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: Ok, it's actually used everywhere. Let's discuss

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: Let's discuss in person. I need help expressing what I mean...

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: Resolution: make an reminder issue.

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: "indices are unique for this vector"

keileg (approved):
I'll formally approve this, trusting that you will take care of the rest of the changes. Let me know if you want to discuss anything.

src/porepy/models/metric.py
Yuriyzabegaev: Decided to clean up. There already is a check for it above, so not adding an additional docstring,

src/porepy/models/metric.py
Yuriyzabegaev: After discussion with Eirik, agreed on doing so.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: Agreed on doing so after the discussion.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: #1734

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: Looked at it with @keileg, decided it's fine.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: Renamed to it.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: After discussion with @keileg, decided to keep it as it is, considering the things above.

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: Added in the class docstring

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: Renamed both to just `indices`.

src/porepy/numerics/ad/indexers.py
Yuriyzabegaev: added a TODO about it.

tests/functional/test_buoyancy_flow.py
Yuriyzabegaev: Added a TODO. See a similar discussion in `metric.py`.

tests/numerics/ad/test_equation_system.py
Yuriyzabegaev: _parse_equations returned 4 atomic equation operators, corresponding to a single equation name defined on 4 domains.

Previously, _parse_single_equation returned a dict with a single equation_name key.

This is private, so I decided to change this contract. Is it ok?

tests/numerics/ad/test_indexers.py
Yuriyzabegaev: added

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: #1735

PR #1726, opened 2026-07-24
Schur complement reduction

Yuriyzabegaev:
## Proposed changes

Making Schur complement reduction a standalone linear solver, instead of spreading its logic throughout the equation system and the model.

This PR follows after #1724.

It is a breaking change for those who rely on Schur complement reduction. The new usage example is provided.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
+++ b/src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
@@ -1,7 +1,7 @@
 """This example shows how to apply the Schur complement reduction-based linear solver.
 
-The considered model a simplified model of 3-phase 2-component flow, which do not
-represent any real physics and should not be used in modeling.
+The considered model is a simplified model of three-phase, two-component flow. It does
+not represent any real physics and should not be used in modeling.
 
 The setup is heavily inspired by `tests/functional/test_linear_tracer.py`.
 
@@ -30,7 +30,7 @@ def run_example():
 
     model = TracerFlowModel_3p(model_params)
 
-    # Setting dt and end time schedule according to cfl condition and approximate
+    # Setting dt and end time schedule according to the CFL condition and approximate
     # flow velocity.
     model.prepare_simulation()
     sd = model.mdg.subdomains()[0]
@@ -47,8 +47,8 @@ def run_example():
     # Initializing a nonlinear solver with a custom linear solver that applies a Schur
     # complement reduction. It requires us to list the primary variables and equations.
     # The default tags are available for the standard equations and variables in PorePy.
-    # This model uses a non-standard equation and variable "z_tracer". Creating custom
-    # tags for them.
+    # This model uses a non-standard equation and variable, "z_tracer", so we create
+    # custom tags for them.
     nonlinear_solver = pp.solvers.NewtonSolver(
         linear_solver=pp.solvers.SchurComplementReductionLinearSolver(
             primary_equation_tags=[
--- a/src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
+++ b/src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
@@ -19,7 +19,7 @@ def run_example():
     model_params = {
         "material_constants": {
             "solid": pp.SolidConstants(
-                porosity=1.0, permeability=1.0, residual_aperture=1
+                porosity=1.0, residual_aperture=1
             ),
         },
         "meshing_arguments": {"cell_size": SimplePipe2D.pipe_length / 10},
--- a/src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
+++ b/src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
@@ -1,49 +1,58 @@
 """This example shows how to apply the Schur complement reduction-based linear solver.
 
-The considered model is a simplified model of three-phase, two-component flow. It does
+The considered model is a simplified model of single-phase, two-component flow. It does
 not represent any real physics and should not be used in modeling.
 
-The setup is heavily inspired by `tests/functional/test_linear_tracer.py`.
+The setup is heavily inspired by `examples/tracer_flow.py`.
 
 """
 
 import logging
 
 import porepy as pp
-from tests.functional.setups.linear_tracer import SimplePipe2D, TracerFlowModel_3p
+from porepy.examples.tracer_flow import TracerFlowModel
 
 
 def run_example():
-    logging.basicConfig(level=logging.INFO)
+    # The same initialization as in examples/tracer_flow.py.
 
-    model_params = {
-        "material_constants": {
-            "solid": pp.SolidConstants(
-                porosity=1.0, residual_aperture=1
-            ),
-        },
-        "meshing_arguments": {"cell_size": SimplePipe2D.pipe_length / 10},
-        "prepare_simulation": False,  # Need to do it manually to determine CFL.
-        "times_to_export": [],
-        "equilibrium_condition": "dummy",
-    }
-
-    model = TracerFlowModel_3p(model_params)
-
-    # Setting dt and end time schedule according to the CFL condition and approximate
-    # flow velocity.
-    model.prepare_simulation()
-    sd = model.mdg.subdomains()[0]
-    dt = model.exact_sol.dt_from_cfl(sd)
+    # Initial time step 60 seconds.
+    dt_init = pp.MINUTE
+    # Simulation time 20 minutes.
+    T_end = 20 * pp.MINUTE
+    # min max time step size is 6 seconds and 10 minutes respectively
+    dt_min_max = (0.1 * dt_init, 10 * pp.MINUTE)
+    # parameters for Newton solver
+    max_iterations = 80
+    newton_tol = 1e-6
+    newton_tol_increment = newton_tol
 
     time_manager = pp.TimeManager(
-        schedule=[0, 3 * dt, 6 * dt, 9 * dt, 10 * dt],
-        dt_init=dt,
-        constant_dt=True,
+        schedule=[0, T_end],
+        dt_init=dt_init,
+        dt_min_max=dt_min_max,
+        iter_max=max_iterations,
+        iter_optimal_range=(2, 10),
+        iter_relax_factors=(0.8, 1.2),
+        recomp_factor=0.8,
+        recomp_max=5,
     )
-    model.ad_time_step.set_value(dt)
-    model.time_manager = time_manager
 
+    params = {
+        "material_constants": {
+            # Solid with impermeable fractures.
+            "solid": pp.SolidConstants(
+                porosity=0.1, permeability=1e-7, normal_permeability=1e-19
+            ),
+        },
+        "fracture_indices": [0, 1],
+        "time_manager": time_manager,
+        "meshing_arguments": {"cell_size": 0.05},
+        "grid_type": "simplex",
+    }
+    model = TracerFlowModel(params)  # type: ignore[abstract]
+
+    # This part is different from examples/tracer_flow.py:
     # Initializing a nonlinear solver with a custom linear solver that applies a Schur
     # complement reduction. It requires us to list the primary variables and equations.
     # The default tags are available for the standard equations and variables in PorePy.
@@ -53,21 +62,31 @@ def run_example():
         linear_solver=pp.solvers.SchurComplementReductionLinearSolver(
             primary_equation_tags=[
                 pp.solvers.DefaultEquationTags.mass_balance,
-                pp.solvers.DefaultEquationTags.energy_balance,
+                pp.solvers.DefaultEquationTags.interface_darcy_flux,
+                pp.solvers.DefaultEquationTags.well_flux,
                 pp.solvers.EquationTag(name="component_mass_balance_equation_tracer"),
             ],
             primary_variable_tags=[
                 pp.solvers.DefaultVariableTags.pressure,
-                pp.solvers.DefaultVariableTags.enthalpy,
+                pp.solvers.DefaultVariableTags.interface_darcy_flux,
+                pp.solvers.DefaultVariableTags.well_flux,
                 pp.solvers.VariableTag(name="z_tracer"),
             ],
             # Using a direct linear solver as the inner algorithm. It can be replaced
             # with an iterative solver.
             primary_linear_solver=pp.solvers.LinearSolverDirect(),
-        )
+        ),
+        params={
+            "nl_max_iterations": max_iterations,
+            "nl_convergence_inc_atol": newton_tol_increment,
+            "nl_convergence_res_atol": newton_tol,
+        },
     )
-    pp.ModelRunner(model, model_params, nonlinear_solver=nonlinear_solver).run()
+    runner = pp.ModelRunner(model, nonlinear_solver=nonlinear_solver)
+    runner.run()
+    return [model]
 
 
 if __name__ == "__main__":
+    logging.basicConfig(level=logging.INFO)
     run_example()
--- a/src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
+++ b/src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
@@ -1,0 +1,73 @@
+"""This example shows how to apply the Schur complement reduction-based linear solver.
+
+The considered model a simplified model of 3-phase 2-component flow, which do not
+represent any real physics and should not be used in modeling.
+
+The setup is heavily inspired by `tests/functional/test_linear_tracer.py`.
+
+"""
+
+import logging
+
+import porepy as pp
+from tests.functional.setups.linear_tracer import SimplePipe2D, TracerFlowModel_3p
+
+
+def run_example():
+    logging.basicConfig(level=logging.INFO)
+
+    model_params = {
+        "material_constants": {
+            "solid": pp.SolidConstants(
+                porosity=1.0, permeability=1.0, residual_aperture=1
+            ),
+        },
+        "meshing_arguments": {"cell_size": SimplePipe2D.pipe_length / 10},
+        "prepare_simulation": False,  # Need to do it manually to determine CFL.
+        "times_to_export": [],
+        "equilibrium_condition": "dummy",
+    }
+
+    model = TracerFlowModel_3p(model_params)
+
+    # Setting dt and end time schedule according to cfl condition and approximate
+    # flow velocity.
+    model.prepare_simulation()
+    sd = model.mdg.subdomains()[0]
+    dt = model.exact_sol.dt_from_cfl(sd)
+
+    time_manager = pp.TimeManager(
+        schedule=[0, 3 * dt, 6 * dt, 9 * dt, 10 * dt],
+        dt_init=dt,
+        constant_dt=True,
+    )
+    model.ad_time_step.set_value(dt)
+    model.time_manager = time_manager
+
+    # Initializing a nonlinear solver with a custom linear solver that applies a Schur
+    # complement reduction. It requires us to list the primary variables and equations.
+    # The default tags are available for the standard equations and variables in PorePy.
+    # This model uses a non-standard equation and variable "z_tracer". Creating custom
+    # tags for them.
+    nonlinear_solver = pp.solvers.NewtonSolver(
+        linear_solver=pp.solvers.SchurComplementReductionLinearSolver(
+            primary_equation_tags=[
+                pp.solvers.DefaultEquationTags.mass_balance,
+                pp.solvers.DefaultEquationTags.energy_balance,
+                pp.solvers.EquationTag(name="component_mass_balance_equation_tracer"),
+            ],
+            primary_variable_tags=[
+                pp.solvers.DefaultVariableTags.pressure,
+                pp.solvers.DefaultVariableTags.enthalpy,
+                pp.solvers.VariableTag(name="z_tracer"),
+            ],
+            # Using a direct linear solver as the inner algorithm. It can be replaced
+            # with an iterative solver.
+            primary_linear_solver=pp.solvers.LinearSolverDirect(),
+        )
+    )
+    pp.ModelRunner(model, model_params, nonlinear_solver=nonlinear_solver).run()
+
+
+if __name__ == "__main__":
+    run_example()
--- a/src/porepy/models/solution_strategy.py
+++ b/src/porepy/models/solution_strategy.py
@@ -42,12 +42,14 @@ class SolutionStrategy(pp.PorePyModel):
 
         # Print deprecation warning for an old parameter.
         if "linear_solver" in params:
-            logger.warning(
+            warn(
                 "Linear solver was moved outside the PorePy model. If you previously "
                 "passed 'linear_solver' backend string (e.g. 'pypardiso') in model "
                 "params, replace it with "
                 "pp.solvers.NewtonSolver(linear_solver=pp.LinearSolverDirect(backend="
-                "'pypardiso')). The current passed value is ignored."
+                "'pypardiso')). The current passed value is ignored.",
+                category=DeprecationWarning,
+                stacklevel=2,
             )
 
         # Set default parameters, these will be overwritten by any parameters passed.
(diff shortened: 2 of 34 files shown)


src/porepy/models/solution_strategy.py
```diff
@@ -42,12 +42,14 @@ def __init__(self, params: Optional[dict] = None):
 
         # Print deprecation warning for an old parameter.
         if "linear_solver" in params:
-            logger.warning(
+            warn(
```
Yuriyzabegaev, line 45: warn prints a signle time, does not spam every time the function is called.

src/porepy/numerics/ad/indexers.py
```diff
@@ -231,16 +200,81 @@ def construct_restricted_indexer(
...
+                )
+            projections.append(dofs)
+        return (
+            np.concatenate(projections)
+            if len(projections) > 0
+            else np.empty(0, dtype=int)
+        )
+
+
+class EquationSystemIndexer(EquationIndexer):
```
Yuriyzabegaev, line 269: This introduces a subclass of `EquationIndexer`, `EquationSystemIndexer` (maybe should find a better name, but I think this name is fine, you'll see why).

The subclass contains a single additional field `equation_image_space_composition`. This is what was before the attribute of the equation system. It has indices local to each equation. It is needed only to do slicing within `equation_system.asseble`, because it slices each equation before they are glued in a global system. It is not needed anywhere else.

You can pass this indexer outside the linear system, and it will behave as a normal EquationIndexer, so I think inheritance is justified and very natural in this scenario. Hope it's not confusing.

src/porepy/numerics/solvers/linear_solvers/linear_solver.py
```diff
@@ -105,6 +105,16 @@ class LinearSystem:
...
+    equation_indexer: pp.ad.EquationIndexer
+    """Indexer to map equations and their definition domains to the DoFs of the `matrix`
+    and the `rhs`.
+    
+    """
+    variable_indexer: pp.ad.VariableIndexer
+    """Indexer to map variables and their definition domains to the DoFs of the
+    `matrix`.
+    
+    """
```
Yuriyzabegaev, line 117: Equation and variable indexers are now first-class passengers in the linear system!

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
```diff
@@ -0,0 +1,376 @@
...
+        """The indices arrays needed for the algorithm to operate. Initialized lazily in
+        :meth:`_initialize_data` the first time :meth:`solve_linear_system` is invoked.
+
+        """
+
+    def _initialize_data(
+        self, linear_system: LinearSystem
+    ) -> _SchurComplementReductionData:
+        """Lazy initialization of arrays. Applied when"""
+        primary_eqs, secondary_eqs = _filter_by_tags(
```
Yuriyzabegaev, line 125: this function is a strong candidate on becoming the indexer's method. But I'm not rushing to add it there until some other code needs it.

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
```diff
@@ -0,0 +1,376 @@
...
+
+        # Concatenate the primary and the secondary solution blocks.
+        solution = np.empty(len(sol_s) + len(sol_p), dtype=sol_s.dtype)
+        solution[data.primary_var_dofs] = sol_p
+        solution[data.secondary_var_dofs] = sol_s
+
+        return solution, _wrap_primary_solver_status(status, solve_time=time() - t0)
+
+
+def rearrange_matrix_as_array_of_structures(
```
Yuriyzabegaev, line 252: this function replaces `generate_permutation_to_block_diag_matrix`. It is less general, see docstring. But I really don't see a practical case that it does not cover.

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
```diff
@@ -0,0 +1,376 @@
...
+
+    var_dofs_by_grid: dict[pp.GridLike, list[np.ndarray]] = {
+        grid: [] for grid in unique_grids_variables
+    }
+    for var, dofs in var_indexer.variable_dofs.items():
+        var_dofs_by_grid[var.domain].append(dofs)
+
+    # Sanity check that assumes that the number of equations and variables is the same
+    # on every grid. If it turns out that we define some equations only on certain
+    # grids, this funciton can be extended. So far, no such case is known to YZ.
```
Yuriyzabegaev: This sanity check is needed to ensure all blocks are of the same size. If not, this function can be generalized. Will there every be such a case?

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
```diff
@@ -0,0 +1,376 @@
...
+    secondary_eq_perm: np.ndarray
+    secondary_var_perm: np.ndarray
+    secondary_block_sizes: np.ndarray
+
+
+def _concatenate_safe(arrays: list[np.ndarray]) -> np.ndarray:
+    """Concatenate a list of arrays and return an empty array if the list is empty."""
+    if len(arrays) > 0:
+        return np.concatenate(arrays)
+    return np.empty(0)
```
Yuriyzabegaev, line 379: I really wish this becomes a public utility function that lives nearby indexers.

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
```diff
@@ -0,0 +1,376 @@
...
+            primary_solver_status=primary_solver_status,
+        )
+    else:
+        return SchurComplementReductionStatusFailure(
+            primary_solver_status=primary_solver_status,
+            reason="primary linear solver failed",
+        )
+
+
+def _filter_by_tags[T: (pp.ad.EquationOnDomain, pp.ad.Variable)](
```
Yuriyzabegaev, line 398: `pp.ad.Variable` and `EquationOnDomain` both have fields `name` and `domain`. So this is a generic function, that either of them, and returns the same type.

src/porepy/numerics/solvers/equation_variable_tags.py
```diff
@@ -0,0 +1,138 @@
+"""This module defines `EquationTag` and `VariableTag`, which identify equations
```
Yuriyzabegaev, line 1: tags are defined in the solvers namespace, e.g. `pp.solvers.EquationTag`. 

I don't want this to become a disruptive change "now we use these tags everywhere". Let it be scoped as something to define solvers. If they turn out to be convenient, we can consider extending their application. Fair?

tests/functional/test_linear_tracer.py
```diff
@@ -57,7 +57,22 @@ def results(request: pytest.FixtureRequest) -> list[LinearTracerSaveData]:
     )
     model.ad_time_step.set_value(dt)
     model.time_manager = time_manager
-    pp.ModelRunner(model, model_params).run()
+    nonlinear_solver = pp.solvers.NewtonSolver(
```
Yuriyzabegaev, line 64: Previously, this test did not use Schur complement reduction. Changing it, because it's an integration test. The solved problem is so slow, so it does not affect the test run time anyhow.

tests/models/test_model_runner.py
```diff
@@ -27,17 +27,17 @@ def initial_condition(self):
             values = np.full(self.equation_system.num_dofs(), STATE_VALUE)
             self.equation_system.set_variable_values(values, iterate_index=0)
 
-        def assemble_linear_system(self) -> pp.solvers.LinearSystem:
+        def before_nonlinear_iteration(self) -> None:
```
Yuriyzabegaev, line 30: changed because `PorePyModel.assemble_linear_system` became deprecated. Test did not become weaker.

tests/models/test_solution_strategy.py
```diff
@@ -513,161 +510,3 @@ def test_linear_or_nonlinear_model(params: dict):
 
     model = models.model(model_type=model_name, dim=2, num_fracs=num_fracs)
     assert model._is_nonlinear_problem() == is_nonlinear
-
-
-# ---------------------------------------------------------------------
-# Tests for block‐permutation and inversion in SolutionStrategy
-# ---------------------------------------------------------------------
```
Yuriyzabegaev, line 520: Moved to `test_schur_complement_reduction.py`

src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
```diff
@@ -0,0 +1,73 @@
+"""This example shows how to apply the Schur complement reduction-based linear solver.
+
+The considered model is a simplified model of three-phase, two-component flow. It does
+not represent any real physics and should not be used in modeling.
+
+The setup is heavily inspired by `tests/functional/test_linear_tracer.py`.
```
keileg: This may indicate that said file does not belong to the tests. Or perhaps it just reflects that something is wrong with the organization of such files - that would not be a surprise. @IvarStefansson what do you think?

src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
```diff
@@ -0,0 +1,73 @@
...
+from tests.functional.setups.linear_tracer import SimplePipe2D, TracerFlowModel_3p
+
+
+def run_example():
+    logging.basicConfig(level=logging.INFO)
+
+    model_params = {
+        "material_constants": {
+            "solid": pp.SolidConstants(
+                porosity=1.0, permeability=1.0, residual_aperture=1
```
keileg:
```suggestion
                porosity=1.0, residual_aperture=1
```
Permeability is one by default.

src/porepy/numerics/solvers/equation_variable_tags.py
keileg: I believe so. I guess we in principle could have used the tags directly in the model instead of hard-coded strings for equation and variable names? I can see the current structure, where we rely on typing the same name in the model and here, is error-prone. On the other hand, nobody ever change the names, so perhaps it is safe.

src/porepy/numerics/solvers/equation_variable_tags.py
```diff
@@ -0,0 +1,140 @@
...
+
+@dataclass(frozen=True)
+class DomainFilter(ABC):
+    """A filter to restrict domains of a variable or an equation.
+
+    Used in :class:`EquationTag` and :class:`VariableTag`."""
+
+    @abstractmethod
+    def filter(self, domain: pp.GridLike) -> bool:
+        """Whether this `domain` is included in the domains where the equation /
```
keileg, line 37: Question for future extensions: Should the filter raise if it removes a domain which the operator is not defined on?

src/porepy/numerics/solvers/equation_variable_tags.py
```diff
@@ -0,0 +1,140 @@
...
+    """Variable name."""
+    defined_on: DomainFilter = Anywhere()
+    """Operational domains of the variable identified by this tag. Possibly a subset of
+    all domains where this variable is defined by a PorePy model.
+
+    """
+
+
+class DefaultEquationTags:
+    """A namespace for all known equations defined by standard PorePy models."""
```
keileg, line 93: How do we make sure the list is updated if new vraiables are introduced? Can this be tested? Also, if I add an equation in a runscript, should I also update this class somehow?

src/porepy/numerics/solvers/equation_variable_tags.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # Mass balance
```
keileg, line 95: I believe that everything compositional is missing. The same applies to the variables.

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
```diff
@@ -0,0 +1,400 @@
...
+        primary_eqs, secondary_eqs = _filter_by_tags(
+            all_operators=linear_system.equation_indexer.equation_dofs,
+            tags=self.primary_equation_tags,
+        )
+        primary_vars, secondary_vars = _filter_by_tags(
+            all_operators=linear_system.variable_indexer.variable_dofs,
+            tags=self.primary_variable_tags,
+        )
+
+        eq_indexer = linear_system.equation_indexer
```
keileg: Move above L104 and use also in definition of primary/secondary?

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
keileg: No, it should be placed in array_operations. We must have similar code in hundreds of places throughout PorePy.

keileg: But not this PR.

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
```diff
@@ -0,0 +1,400 @@
...
+    ```
+    | A_ss A_sp |   | x_s |   | b_s |
+    | A_ps A_pp | * | x_p | = | b_p |,
+    ```
+    where subscripts 'p' and 's' denote primary and secondary blocks, respectively. The
+    primary block is defined by parameters `primary_equation_tags` and
+    `primary_variable_tags`. The secondary block is a complement to it.
+
+    The submatrix `A_ss` must be invertible, and computing its inverse must be cheap
+    computationally, e.g., it should be block diagonal. Then, the Schur complement
```
keileg: I think we can be clearer: The present implementation assumes A_ss is block diagonal, as this is the most relevant case for the intended usage.

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
```diff
@@ -0,0 +1,400 @@
...
+    for list_of_dofs in var_dofs_by_grid.values():
+        if bs != len(list_of_dofs):
+            raise ValueError(
+                "All variables must be defined on the same set of domains."
+            )
+
+    if bs == 0:
+        return np.empty(0, dtype=int), np.empty(0, dtype=int), np.empty(0, dtype=int)
+
+    assert var_indexer.num_dofs % bs == 0
```
keileg: Will this also work if the system contains variables defined on only some domains, say displacements (which also introduce a heterogeneous number of dofs) that are not included in the secondary block?

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1525,22 +1466,31 @@ def assemble(
...
-        return A, -rhs_cat
+
+        return pp.solvers.LinearSystem(
+            matrix=A,
+            rhs=-rhs_cat,
+            equation_indexer=equation_indexer,
+            variable_indexer=variable_indexer,
+        )
 
     def construct_assembled_matrix_indexers(
```
keileg: Should this method be private? Just checking.

src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
```diff
@@ -0,0 +1,73 @@
...
+not represent any real physics and should not be used in modeling.
+
+The setup is heavily inspired by `tests/functional/test_linear_tracer.py`.
+
+"""
+
+import logging
+
+import porepy as pp
+from tests.functional.setups.linear_tracer import SimplePipe2D, TracerFlowModel_3p
```
Yuriyzabegaev: This (import from tests) should not be like this. Let's discuss the right way to present examples. My intent was not to make a tutorial, since this is quite advanced. But if someone needs it, they can find the working example of this functionality here.

tests/numerics/ad/test_equation_system.py
```diff
@@ -1286,6 +1286,38 @@ def test_parse_equations(model: EquationSystemMockModel):
     }
 
 
+def test_domain_restricted_assembly_has_local_equation_indices(
```
keileg: I have the same comment here as to the previous PR, that the tests need some more comments. I guess the answer is, this is a draft, tests to be improved?

keileg (commented):
Mainly minor comments. I guess the point about tests not being finalized (commented) carries over from the previous PR?

src/porepy/numerics/solvers/equation_variable_tags.py
Yuriyzabegaev: I have had precisely a case when it's not safe. Say you want to add a constant pressure constraint for a single cell on top of the mass balance equation. You add an operator that takes the "mass_balance_equation" operator as a child and does some conditional logic. You naturally name the new operator "constrained_mass_balance_equation" and register it in the equation system. It breaks everything that expected "mass_balance_equation" (true story).

My opinion about it is that operator_name != equation_name (registered in the equation system). Operator names are useful for debugging and finding a certain operator in a tree. Equation name is useful to fetch an equation from the equation system and do something with it (assemble subsystem, linear/nonlinear solver, partial residual, etc).

Whether tags is the way to go or not - I'm unsure.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: I don't have a strong opinion. Let's make it private.

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
Yuriyzabegaev: Maybe
"the present implementation assumes A_ss can be permuted in a block diagonal form"?

src/porepy/numerics/solvers/equation_variable_tags.py
Yuriyzabegaev: I think I didn't get it

src/porepy/numerics/solvers/equation_variable_tags.py
Yuriyzabegaev: see how it's done in `schur_complement_reduction_linear_tracer_3p.py`:

```
            primary_equation_tags=[
                pp.solvers.DefaultEquationTags.mass_balance,
                pp.solvers.DefaultEquationTags.energy_balance,
                pp.solvers.EquationTag(name="component_mass_balance_equation_tracer"),
            ],
```

This is a list of defaults. If you add a non-default equation/variable, you're free to declare a tag for it.

Regarding testing, unclear to me. We can make some regex magic, similarly to `test_example_parameters`. But not anything smarter, while the equation names are hard-coded literals.

src/porepy/numerics/solvers/equation_variable_tags.py
Yuriyzabegaev: This is because compositional names are constructed dynamically, depending on the composition. See my comment above about custom tags.

tests/functional/test_linear_tracer.py
Yuriyzabegaev: *fast, not slow (typo)

tests/numerics/ad/test_equation_system.py
Yuriyzabegaev: yes, but these tests must be improved before this PR is green lighted.

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
Yuriyzabegaev: I left an assertion above that raises if that is the case. And left a comment about it, that this function can be extended to cover this case if needed. If you want, I can do it in this PR

src/porepy/numerics/solvers/linear_solvers/schur_complement_reduction.py
Yuriyzabegaev: done

src/porepy/examples/solvers/schur_complement_reduction_linear_tracer_3p.py
Yuriyzabegaev: Fixed by importing tracer flow from src/examples

Yuriyzabegaev:
@keileg I have cleaned up tests. Please have one final look.

tests/numerics/ad/test_equation_system.py
```diff
@@ -1050,18 +1041,6 @@ def test_set_remove_equations(model: EquationSystemMockModel):
     )
 
 
-def test_update_equation_on_empty_domain(model: EquationSystemMockModel) -> None:
```
Yuriyzabegaev, line 1053: removed this test that I introduced myself. It's overlapping with test_set_remove_equations.

tests/numerics/ad/test_equation_system.py
```diff
@@ -1287,91 +1267,37 @@ def test_parse_equations(model: EquationSystemMockModel):
...
-        ["z", "w"],  # Combination of simple and merged.
+        None,  # None gives the full system.
+        [],  # An empty list will give a system with zero rows.
+        ["eq_single_subdomain"],  # A single equation.
+        ["eq_single_interface", "eq_all_subdomains"],  # Combination of two equations.
+        # Combination of two equations, reversed order.
+        ["eq_all_subdomains", "eq_single_interface"],
     ],
 )
-def test_secondary_variable_assembly(model: EquationSystemMockModel, var_names):
```
Yuriyzabegaev, line 1299: Removed this test. It fully overlaps with test_assemble. I did not introduce this test myself. I think it's left from the times we had separate functions assemble and assemble_subsystem.

tests/numerics/ad/test_equation_system.py
```diff
@@ -1287,91 +1267,37 @@ def test_parse_equations(model: EquationSystemMockModel):
...
-        ],
+        None,  # None gives the full system.
+        [],  # An empty list will give a system with zero columns.
+        ["x"],  # A single variable.
+        ["x", "w"],  # Combination of two variables.
+        ["w", "x"],  # Combination of two variables, reversed order.
     ],
 )
-def test_assemble(model: EquationSystemMockModel, equation_variables):
+def test_assemble(
```
Yuriyzabegaev, line 1290: Changed parametrization to test a cartesian product of possible input parameters.

tests/numerics/ad/test_equation_system.py
```diff
@@ -1434,112 +1362,66 @@ def test_assemble(model: EquationSystemMockModel, equation_variables):
...
-        )
-        assert num_dofs == model.block_size(name)
-
-    variable_dofs = variable_indexer.indices
-    for name in var_names:
-        num_dofs = sum(
-            [dofs.size for var, dofs in variable_dofs.items() if var.name == name]
-        )
-        assert num_dofs == model.dof_ind(name).size
+    # The restricted linear system comes with indexers local to it. Check that they
```
Yuriyzabegaev, line 1359: Added all checks for indexers here.

tests/numerics/ad/test_equation_system.py
```diff
@@ -1434,112 +1362,66 @@ def test_assemble(model: EquationSystemMockModel, equation_variables):
...
-        ["x", "w"],
+        None,  # None gives the full system.
+        [],  # An empty list will give a system with zero columns.
+        ["x"],  # A single variable.
+        ["x", "w"],  # Combination of two variables.
+        ["w", "x"],  # Combination of two variables, reversed order.
     ],
 )
-def test_assembled_matrix_indexers_match_assembly_order(
+def test_extract_subsystem(
```
Yuriyzabegaev, line 1411: Changed parametrization to test a cartesian product of possible input parameters, since the parametrization here is the same as in test_assembly, which I updated.

tests/numerics/ad/test_equation_system.py
```diff
@@ -1434,112 +1362,66 @@ def test_assemble(model: EquationSystemMockModel, equation_variables):
...
-        ["w", "x"],
-        ["x", "w"],
+        None,  # None gives the full system.
+        [],  # An empty list will give a system with zero columns.
+        ["x"],  # A single variable.
+        ["x", "w"],  # Combination of two variables.
+        ["w", "x"],  # Combination of two variables, reversed order.
     ],
 )
-def test_assembled_matrix_indexers_match_assembly_order(
```
Yuriyzabegaev, line 1475: Removed the test that I myself introduced. It fully overlaps with test_assemble.

keileg:
> @keileg I have cleaned up tests. Please have one final look.

@Yuriyzabegaev  looks good to me - I cleaned up a minor comment.

keileg (approved):

PR #1728, opened 2026-07-26
Let numpy defer to the Ad operator overloads instead of broadcasting

### LINKED ISSUE #819: Anomaly in overload of  `Ad_array.__rmul__`
Hei,

I have tried to wrap my mind around what is happening here, but I have absolutely no clue

```
import numpy as np
import porepy as pp
import scipy.sparse as sps


a = pp.ad.Ad_array(np.array([1]), sps.csr_matrix((1,10)))
b = np.array([0.9])
c = b * a
print(type(c))  # c is of type numpy.ndarray
print(type(c[0]))  # but it is a numpy array containing Ad_arrays!
```

gaoflow:
#819 is tagged `wontfix` — "I am afraid it will never be solved. It is just what it is." — but numpy has a documented opt-out for exactly this case: with `__array_ufunc__ = None` on the class, `ndarray.__mul__` returns `NotImplemented` and Python calls the `__rmul__` that `AdArray`, `Operator` and `ArraySlicer` already implement. Without it, `numpy_array * ad_array` silently returns an object array holding one full `AdArray`, Jacobian included, per element: 4,000,000 stored Jacobian nonzeros instead of 2,000 at n=2000. The comparison operators are ufuncs too, so `numpy_array < ad_array` was captured the same way.

Auditing the rest of the surface turned up a second instance of the same "numpy compares elementwise" problem: the `if other == 0` guard in `Operator.__add__`, which is there to catch the `0` that `sum()` starts from, means `DenseArray + numpy_array` and `DenseArray + sparse_matrix` raise "truth value of an array is ambiguous" today — in the *supported* operand order. It now checks for a scalar before comparing.

Tested: the `dense`/`ad` case that `test_arithmetic_operations_on_ad_objects` skipped with a reference to this issue now passes, as does `other_mode="dense"` in `test_matrix_slicer_delayed_evaluation_ad_dense_float`; plus new tests for the arithmetic, comparison and addition cases. Full suite 3566 passed vs 3524 before, identical set of 45 pre-existing failures (tutorials, `warnings.deprecated` on py3.11); ruff, isort and mypy unchanged. One behaviour change worth flagging: a 0-d numpy array on the left now raises, consistently with `AdArray + np.array(2.0)`, which already did.

Closes #819.

### DIFF
--- a/src/porepy/numerics/ad/forward_mode.py
+++ b/src/porepy/numerics/ad/forward_mode.py
@@ -73,9 +73,11 @@ class AdArray:
 
     """
 
-    # Numpy defers binary operations to the __r*__ methods below only if this is set
-    # to None; without it, `numpy_array * ad_array` broadcasts into an object array
-    # holding one full AdArray, Jacobian included, per element (#819).
+    # Turn off numpy's ufuncs for this class, to avoid unexpected behavior when
+    # combining with numpy arrays. See
+    # https://numpy.org/neps/nep-0013-ufunc-overrides.html#turning-ufuncs-off for
+    # technical information and GH issue #819 for a description of the problem that can
+    # arise if this is not done.
     __array_ufunc__ = None
 
     def __init__(self, val: np.ndarray, jac: sps.spmatrix) -> None:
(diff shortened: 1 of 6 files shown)


keileg:
Thanks for the contribution. I made some changes to the documentation  to adapt it better to the existing style. Also, I removed the ufunc fix related to Ad Operators, since these objects are part of a computational graph that will never be mixed with numpy arrays (though that is not easy to see from the source code, sorry about that).

I will approve and merge once the tests pass, but I will take responsibility for the rest of the process. Again, thanks a lot.

keileg (approved):

PR #1732, opened 2026-07-31
Evaluate Operator on subset of variables

Yuriyzabegaev:
## Proposed changes

This allows to evaluate an operator on a subset of variables, without the need to slice the columns of the resulting Jacobian. The excluded variables are taken from the the current iterate. 

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/numerics/ad/_ad_parser.py
+++ b/src/porepy/numerics/ad/_ad_parser.py
@@ -67,6 +67,7 @@ class AdParser:
         equation_system: pp.ad.EquationSystem,
         derivative: Literal[False],
         state: np.ndarray | None,
+        variable_indexer: pp.ad.VariableIndexer,
     ) -> float | np.ndarray | sps.spmatrix: ...
 
     @overload
@@ -76,6 +77,7 @@ class AdParser:
         equation_system: pp.ad.EquationSystem,
         derivative: Literal[True],
         state: np.ndarray | None,
+        variable_indexer: pp.ad.VariableIndexer,
     ) -> pp.ad.AdArray: ...
 
     @overload
@@ -85,6 +87,7 @@ class AdParser:
         equation_system: pp.ad.EquationSystem,
         derivative: Literal[True],
         state: np.ndarray | None,
+        variable_indexer: pp.ad.VariableIndexer,
     ) -> list[pp.ad.AdArray]: ...
 
     @overload
@@ -94,6 +97,7 @@ class AdParser:
         equation_system: pp.ad.EquationSystem,
         derivative: Literal[False],
         state: np.ndarray | None,
+        variable_indexer: pp.ad.VariableIndexer,
     ) -> list[float | np.ndarray | sps.spmatrix]: ...
 
     def evaluate(
@@ -102,6 +106,7 @@ class AdParser:
         equation_system: pp.ad.EquationSystem,
         derivative: bool,
         state: np.ndarray | None,
+        variable_indexer: pp.ad.VariableIndexer,
     ) -> (
         float
         | np.ndarray
@@ -127,10 +132,11 @@ class AdParser:
                 or a list of values, or AdArrays if op is a list.
 
         """
-
         # Get the state of the system, if not provided.
         if state is None:
-            state = equation_system.get_variable_values(iterate_index=0)
+            state = equation_system.get_variable_values(
+                iterate_index=0, variables=list(variable_indexer.operators_to_dofs)
+            )
 
         # Create an AdArray representation of the state, if the derivative is requested.
         # If not, the state is used as is (as a numpy array).
@@ -140,10 +146,13 @@ class AdParser:
         # post-processing below.
         if isinstance(op, list):
             result_list = [
-                self._evaluate_single(o, ad_base, equation_system) for o in op
+                self._evaluate_single(o, ad_base, equation_system, variable_indexer)
+                for o in op
             ]
         else:
-            result_list = [self._evaluate_single(op, ad_base, equation_system)]
+            result_list = [
+                self._evaluate_single(op, ad_base, equation_system, variable_indexer)
+            ]
 
         # If the derivative is requested, the results should be AdArrays. Enforce this.
         if derivative:
@@ -155,7 +164,7 @@ class AdParser:
                 if isinstance(res, np.ndarray) and len(res.shape) == 1:
                     # Convert numpy arrays to AdArrays and update result_list.
                     result_list[index] = pp.ad.AdArray(
-                        res, sps.csr_matrix((res.shape[0], equation_system.num_dofs()))
+                        res, sps.csr_matrix((res.shape[0], variable_indexer.num_dofs))
                     )
                 elif isinstance(res, (sps.spmatrix, sps.sparray, np.ndarray)):
                     # This will cover numpy arrays of higher dimensions (> 1) and sparse
@@ -190,6 +199,7 @@ class AdParser:
         op: pp.ad.Operator,
         ad_base: np.ndarray | pp.ad.AdArray,
         equation_system: pp.EquationSystem,
+        variable_indexer: pp.ad.VariableIndexer,
     ) -> float | np.ndarray | sps.spmatrix | pp.ad.AdArray:
         """Evaluate a single operator.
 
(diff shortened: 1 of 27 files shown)


src/porepy/numerics/ad/_ad_parser.py
```diff
@@ -197,32 +207,47 @@ def _evaluate_single(
...
             if isinstance(op, pp.ad.MixedDimensionalVariable):
-                if op.is_previous_iterate or op.is_previous_time or op.is_reference:
-                    # Empty vector like the global vector of unknowns for prev time/iter
-                    # insert the values at the right dofs and slice.
-                    vals = np.empty_like(
-                        ad_base.val if isinstance(ad_base, pp.ad.AdArray) else ad_base
-                    )
-                    # List of indices for sub variables.
-                    dofs = []
-                    for sub_var in op.sub_vars:
```
Yuriyzabegaev, line 226: I don't fully understand why the slicing here is done the way it's done. Thus I might unintentionally broke some contract. Please check this change very carefully.

src/porepy/numerics/ad/_ad_parser.py
```diff
@@ -197,32 +207,45 @@ def _evaluate_single(
...
-                        sub_dofs = equation_system.dofs_of([sub_var])
-                        vals[sub_dofs] = sub_var.parse(equation_system.mdg)
-                        dofs.append(sub_dofs)
-
-                    return vals[np.hstack(dofs, dtype=int)] if dofs else np.array([])
+                not_active_variable = any(
+                    sub_var not in variable_indexer.indices for sub_var in op.sub_vars
+                )
+                if (
+                    op.is_previous_iterate
```
Yuriyzabegaev, line 258: The new logic is to treat inactive variables as whatever their `parse` method returns. It means, it will return the current `iterate` data for it.

Is it always the intended behavior? It is definitely backward-compatible with "assemble all columns and slice only columns we need". But can we at some point want something different?

src/porepy/numerics/ad/_ad_parser.py
Yuriyzabegaev, line 154: According to this, `state` must conform to the variable indexer.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1544,67 +1535,44 @@ def _construct_assembled_matrix_indexers(
...
+        variable_indexer: pp.ad.VariableIndexer | None = None,
     ) -> pp.number | np.ndarray | sps.spmatrix: ...
 
     @overload
     def evaluate(
         self,
         operator: list[pp.ad.Operator],
-    ) -> list[pp.number | np.ndarray | sps.spmatrix]: ...
-
-    @overload
```
Yuriyzabegaev, line 1555: Managed to remove superfluous overloads with help of an LLM. Mypy is happy,

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1620,7 +1588,9 @@ def evaluate(
             derivative: Whether to evaluate the derivative of the operator. Defaults to
                 False.
             state: State vector to evaluate the operator on. By default, the current
-                state is used.
+                state is used. TODO YZ: What if state if variable_indexer is used? The
```
Yuriyzabegaev: Note this!

src/porepy/numerics/ad/_ad_parser.py
IvarStefansson: Conform in the sense of having the same indexing/order, not just size, right? Perhaps add to docstring? Speaking of which, seems you haven't added parameter descriptions of `variable_indexer` to this or other methods.

src/porepy/numerics/ad/equation_system.py
IvarStefansson: This reveals there is something we've misunderstood about overload which I think is worth understanding. Let's parse through this in person with @keileg.

src/porepy/numerics/ad/equation_system.py
IvarStefansson: Very good question. Can we deduce a logic by considering what state is if not passed and variable_indexer is used? That would be the system as last assembled, right? Which is also what the indexer is supposed to correspond to in that case, no? If I'm right, my gut instinct is to be consistent with that. I'll gladly discuss this.

IvarStefansson (commented):
Posted three comments. Didn't feel confident about your other question(s). Let me know if you want to discuss, probably after Eirik has had a look.

Yuriyzabegaev:
The failing test is failing because its design is incompatible with the change of this PR. It calls `equation_system.evaluate`, but does not initialize data dictionaries (that the model typically initialize in prepare_simulation). It was not an issue previously, because the test passed `state` to `equation_system.evaluate`. Now it does not pass (since this variable is "disabled"), so the ad parser falls back to the data from the data dictionaries, and it is absent.

src/porepy/numerics/ad/_ad_parser.py
```diff
@@ -215,32 +225,45 @@ def _evaluate_single(
...
+                    or op.is_previous_time
+                    or op.is_reference
+                    or not_active_variable
+                ):
+                    vals = [
+                        sub_var.parse(equation_system.mdg) for sub_var in op.sub_vars
```
keileg, line 268: At a glance, something is missing here: The old use of sub_dofs made sure that the evaluated variable was put in the right place in the vector. This information is now in the variable indexer (I believe), hence by analogy, a call to that function should be necessary somehow.

src/porepy/numerics/ad/_ad_parser.py
keileg: I am afraid the contract is implicitly defined by the way the code works. Let's talk through this and see if we can make an explicit contract instead, and take the chance to roll this out.

src/porepy/numerics/ad/_ad_parser.py
keileg: I think the answer is yes, we may want more here. Example: Consider a function f(x, y) and assume that in an iterative scheme, I want to lag the evaluation of y to the previous time step (note: this is a quasi-Newton approach that we currently do not have implemented, but I hope to get there). I do not think this is possible with the current scheme. Let's talk.

src/porepy/numerics/ad/_ad_parser.py
```diff
@@ -215,32 +225,45 @@ def _evaluate_single(
...
+                        variable_indexer.indices[sub_var] for sub_var in op.sub_vars
+                    ]
+                    dofs = (
+                        np.concatenate(dofs_list)
+                        if len(dofs_list)
+                        else np.zeros(0, dtype=int)
+                    )
                     # Fetch the values from the state vector.
-                    return ad_base[equation_system.dofs_of([op])]
+                    return ad_base[dofs]
```
keileg, line 281: Again, this is a place where there may be an implicit contract on the variable ordering imposed by equation_system.dofs_of.

src/porepy/numerics/ad/equation_system.py
keileg: Sounds great. At least in part, the problem is me doing whatever it takes to make mypy happy and then moving rapidly on without trying for further consolidation (because I feel that, by that time, I have tried all possible configurations of method signatures).

src/porepy/numerics/ad/equation_system.py
```diff
@@ -1406,6 +1406,10 @@ def assemble(
         # Standardize and validate input, order it according to the equation_indexer.
         equations_on_domains = self._parse_equations(equations=equations, ordered=True)
 
+        equation_indexer, variable_indexer = self._construct_assembled_matrix_indexers(
```
keileg, line 1409: Should line 1480 be dropped?

src/porepy/numerics/ad/equation_system.py
keileg: Let's discuss. This is in part also related to the missing explicit contract for variable ordering etc. under parsing, see comments in AdParser.

keileg (commented):
Various small comments. I think we all agree on two things: 1) we want something like this; 2) the lack of a clear API for variable parsing is a real problem. I suggest we meet to hammer out an answer to number 2) to see if we can settle this.

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: Yes! Good catch

src/porepy/grids/mdg_generation.py
```diff
@@ -860,7 +860,7 @@ def create_mdg(
     # Keep a reference to the user-provided constraints before they are consumed
     # (and removed from kwargs) by the simplex preprocessing below. This is needed
     # for the fractures-outside-domain check further down.
-    constraints: np.ndarray = kwargs.get("constraints", np.array([]))
+    constraints: np.ndarray = np.array(kwargs.get("constraints", []))
```
Yuriyzabegaev: This fixes a slow test failure introduced by a commit a week ago, 4e0d7a4ca7c74f236569323a1f2a134ebfd6da38. 

```
ERROR tests/functional/test_manu_thermoporomech_nofrac.py::test_order_of_convergence[0-0-pressure] - AttributeError: 'list' object has no attribute 'size'
```

I did not investigate why exactly the test started to break after that change, by the logic looks sound.

src/porepy/numerics/ad/_ad_parser.py
```diff
@@ -215,32 +230,45 @@ def _evaluate_single(
...
-                    # List of indices for sub variables.
-                    dofs = []
-                    for sub_var in op.sub_vars:
-                        sub_dofs = equation_system.dofs_of([sub_var])
-                        vals[sub_dofs] = sub_var.parse(equation_system.mdg)
-                        dofs.append(sub_dofs)
-
-                    return vals[np.hstack(dofs, dtype=int)] if dofs else np.array([])
+                not_active_variable = any(
+                    sub_var not in variable_indexer.indices for sub_var in op.sub_vars
```
keileg, line 243: Change to not active if all variables are in active? If some are, some are not, raise.

keileg (approved):
Feel free to merge when you consider yourself done.

src/porepy/numerics/ad/_ad_parser.py
Yuriyzabegaev: left a comment

src/porepy/numerics/ad/_ad_parser.py
Yuriyzabegaev: added docstring

PR #1471, opened 2025-08-13
BUG: Mark failed time steps when exporting

pschultzendorff:
## Proposed changes
Attempt at fixing #1421.

The idea is to have `SolverStatistics` keep track of whether the current nonlinear loop has converged. When calling `DataSavingMixin.save_data_time_step`, `SolverStatistics.save` stores the convergence flag in the JSON file.  `DataSavingMixin.write_pvd_and_vtu`  then prefixes exported files with `failed_` in the case of non-convergence. Lastly, only the files corresponding to converged time steps are included in the `data.pvd` file for the full simulation.

The code works, except for the last part. To track failed time steps, I added `Exporter._exported_timesteps_convergence`, which masks non-converged steps before writing `data.pvd`. However, when `DataSavingMixin.write_pvd_and_vtu` calls `Exporter.write_pvd`, it passes `TimeManager.exported_times` for `times` and `None` for `file_extension`. The latter parameter is then replaced by `Exporter._exported_timesteps`. As far as I see, tracking exported time steps at two different places and coupling both in `Exporter.write_pvd` is prone to bugs.

Before I make this even more convoluted, we should discuss whether it is worth to clean this up and unify bookkeeping of exported time steps.

## Types of changes

- [ ] Minor change (e.g., dependency bumps, broken links).
- [x] Bugfix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] Testing (contribution related to testing of existing or new functionality).
- [ ] Documentation (contribution related to adding, improving, or fixing documentation).
- [ ] Maintenance (e.g., improve logic and performance, remove obsolete code).
- [ ] Other:

## Checklist

- [x] The documentation is up-to-date.
- [x] Static typing is included in the update.
- [x] This PR does not duplicate existing functionality.
- [ ] The update is covered by the test suite (including tests added in the PR).
- [ ] If new skipped tests have been introduced in this PR, `pytest` was run with the `--run-skipped` flag.

Issue #1421, opened 2025-04-30
`SolutionStrategy.after_nonlinear_failure` saves time step data causing weird visualization


vlipovac:
The base solution strategy at line 477 calls `self.save_data_time_step()` after a time step failes to converge.
If the time manager then goes back in time and reduces `dt`, and converges, it results in a weird visualization where the simulation jumps back and forth in time.

https://github.com/user-attachments/assets/304dd90a-8780-4b09-9295-494243dfb227

To me this is unexpected behavior, useful only for debugging maybe.
Should this be removed? Or at least made optional?

jwboth:
I do not think this is intended. I also realized a similar thing. The exporting counter is increased even though the time step has not passed convergence. In my view, this is a bug and can be declared as such.

pschultzendorff:
Just my two cents: ``save_data_time_step`` also calls ``solver_statistics.save`` in addition to exporting data. Removing ``self.save_data_time_step()`` from ``after_nonlinear_failure`` would eliminate any trace of failed time steps. Then one could no longer assess some common solver metrics such as total number of nonlinear iterations.

vlipovac:
Yes, that is important. The data extracted by `solver_statistics` could use a facelift in terms of flexibility how data is saved and stored, and when

PR #1737, opened 2026-08-06
Detach solver statistics save from data save, and always save

jwboth:
## Proposed changes

This PR suggests to detach saving solver statistics from solving data. While the latter only is performed after successful simulation (of a single time step for instationary simulations), the former shall be always performed, including after the simulation. This makes ensures that the solver statistics is up-to-date even in case of a failing simulation. In part related to #1421  and #1471.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/solution_strategy.py
+++ b/src/porepy/models/solution_strategy.py
@@ -467,7 +467,6 @@ class SolutionStrategy(pp.PorePyModel):
         Use this method for loop-specific post-processing.
 
         """
-        pass
 
     def after_nonlinear_failure(self) -> None:
         """Called after a nonlinear solver loop fails to converge.
@@ -478,7 +477,6 @@ class SolutionStrategy(pp.PorePyModel):
         nonlinear loops per time step and call this method after each failed loop.
 
         """
-        pass
 
     def after_time_step_convergence(self) -> None:
         """Called after a new time step solution has been achieved.
--- a/src/porepy/models/solution_strategy.py
+++ b/src/porepy/models/solution_strategy.py
@@ -467,6 +467,7 @@ class SolutionStrategy(pp.PorePyModel):
         Use this method for loop-specific post-processing.
 
         """
+        pass
 
     def after_nonlinear_failure(self) -> None:
         """Called after a nonlinear solver loop fails to converge.
@@ -477,6 +478,7 @@ class SolutionStrategy(pp.PorePyModel):
         nonlinear loops per time step and call this method after each failed loop.
 
         """
+        pass
 
     def after_time_step_convergence(self) -> None:
         """Called after a new time step solution has been achieved.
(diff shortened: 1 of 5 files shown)


src/porepy/models/solution_strategy.py
```diff
@@ -467,6 +467,7 @@ def after_nonlinear_convergence(self) -> None:
         Use this method for loop-specific post-processing.
 
         """
+        self.save_statistics()
```
Yuriyzabegaev: Why do we need to write statistics both in `nonlinear` and `time_step` callbacks?

tests/viz/test_solver_statistics_unit.py
Yuriyzabegaev, line 1: Is it github glitching, or is it an empty file? Same question about `test_solver_statistics_unit.py`

Yuriyzabegaev (commented):
Thumbs up for this change! I find it very logical to split data save and statistics save.

I have 2 concerns:

1. Can I still disable statistics writing, or is it always on? We might want to disable it for unrelated tests to avoid spamming with statistics files.

2. Soon on the way is the sequential nonlinear solver, that will call `before_nonlinear_*` and `after_nonlinear_*` multiple times per time step. Imagine poromechanics splitting, there will be 2 newton solvers inside. Each might want to have its own statistics, which will conflict with this logic. This is fine by now, but will require rethinking of who is responsible for writing solver statistics. Let's keep this in mind.

src/porepy/models/solution_strategy.py
jwboth: Fully agree. Removed.

tests/viz/test_solver_statistics_unit.py
jwboth: These files do not exist in my branch. I do not exactly know how they got there... these files were hanging around for some time, but I am not able to track them...

jwboth: Ah that is strange. Now after needing to perform a "git pull" they appeared. Removing the files. Sorry for the unexplainable chaos.

jwboth:
> Thumbs up for this change! I find it very logical to split data save and statistics save.
> 
> I have 2 concerns:
> 
> 1. Can I still disable statistics writing, or is it always on? We might want to disable it for unrelated tests to avoid spamming with statistics files.

As before, if not providing "solver_statistics_file_name" in the model parameters, no statistics file is writtin to file. So in most test, no solver statistics should be written, I presume only in the solver statistics tests and some workflow tests (I did not check the test suite).

> 2. Soon on the way is the sequential nonlinear solver, that will call `before_nonlinear_*` and `after_nonlinear_*` multiple times per time step. Imagine poromechanics splitting, there will be 2 newton solvers inside. Each might want to have its own statistics, which will conflict with this logic. This is fine by now, but will require rethinking of who is responsible for writing solver statistics. Let's keep this in mind.

Can't wait to use this :) I see how this can make things complicated. 

Comments:
- I do not attempt to find a unique solution here - I would see several design possibilities and believe there is easy solutions. It really depends on what is of interest to log and later retrieve. Is it the inner and/or outer nonlinear iterations? Type of solver used/stage of the solver? And how to deal with repeated solves due to fails as there is the potential for unsymmetrically many entrance into the subsolves if an earlier solve fails. 
- Currently one statistics "unit" is related to one time step. Adding a nested structure would possibly require a different logic.
- What is relevant for the saving. One can save as often one wants - after saving once, a second save does not mutate the object. Just omitting saving is not good. And the added saves after failure and simulation enable that both each time step will be logged, as well as the case of simulation failure is correctly reported to statistics.

src/porepy/models/solution_strategy.py
jwboth: Opened again. After quick thinking. One needs to write statistics after time_step and not after nonlinear, as the time step status needs to be logged after logging the `time_step_status` in the stepper. I am removing saving after `nonlinear`... sorry for the back and forth.

Yuriyzabegaev: I agree! This sounds reasonable.

Yuriyzabegaev (approved):
Looks good! I just removed extra `pass` to restore how it was before.

keileg (approved):

PR #1738, opened 2026-08-06
Split and reorder logging of time step/status during stepping.

jwboth:
## Proposed changes

The time stepper logs time information and status during stepping. This is done after computing the new dt (as this also return the time_step_status). Before that time is rolled back if the nonlinear solver did not converge. Thus, mutated time and dt are potentially logged. This PR suggests a split in the logging and places the logging of the time information at the start of the time step, such that always time and dt during the solver attempt is logged.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/time_stepper/time_stepper.py
+++ b/src/porepy/time_stepper/time_stepper.py
@@ -108,7 +108,7 @@ class TimeStepper:
             )
 
             # Log time step status for statistics.
-            self._update_statistics(model, time_step_status)
+            self._update_nonlinear_solver_statistics(model, time_step_status)
 
             # Update model (also saves logged statistics) based on trial results.
             self._update_model_after_trial(model, time_step_status)
@@ -227,7 +227,7 @@ class TimeStepper:
             self.time_manager.final_time_reached(),
         )
 
-    def _update_statistics(
+    def _update_nonlinear_solver_statistics(
         self, model: pp.PorePyModel, time_step_status: TimeStepperStatus
     ) -> None:
         """Update statistics from the time step."""
--- a/src/porepy/time_stepper/time_stepper.py
+++ b/src/porepy/time_stepper/time_stepper.py
@@ -110,6 +110,9 @@ class TimeStepper:
             # Log time step status for statistics.
             self._update_statistics(model, time_step_status)
 
+            # Update model (also saves logged statistics) based on trial results.
+            self._update_model_after_trial(model, time_step_status)
+
             # Return on success or error when there is no way to continue trying.
             if isinstance(
                 time_step_status, (TimeStepperStatusSuccess, TimeStepperStatusFailure)
@@ -201,13 +204,16 @@ class TimeStepper:
         model.before_time_step()
         nonlinear_solver_status = solver.solve(model)  # type: ignore
 
-        # Model update based on trial results.
-        if nonlinear_solver_status.is_converged():
-            model.after_time_step_convergence()
-        elif nonlinear_solver_status.is_failed():
-            model.after_time_step_failure()
-
         return nonlinear_solver_status
+
+    def _update_model_after_trial(
+        self, model: pp.PorePyModel, time_step_status: TimeStepperStatus
+    ) -> None:
+        """Model update based on trial results."""
+        if time_step_status.is_success():
+            model.after_time_step_convergence()
+        elif time_step_status.is_failure():
+            model.after_time_step_failure()
 
     def _update_time_statistics(self, model: pp.PorePyModel) -> None:
         """Update statistics from the time step."""
--- a/src/porepy/time_stepper/time_stepper.py
+++ b/src/porepy/time_stepper/time_stepper.py
@@ -91,6 +91,9 @@ class TimeStepper:
             self.time_manager.time = previous_time + self.time_manager.dt
             self.time_manager.time_index += 1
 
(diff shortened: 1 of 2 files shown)


jwboth:
@Yuriyzabegaev I suggest clarifying first #1737 . I see the failing test points towards an inconsistency that is due to lack of saving after failure. Instead of fixing the same test twice, let me do this after deciding whether the other PR is acceptable.

Yuriyzabegaev (commented):
The time stepper was left by me in a messy state, thanks for catching this! I have no comments about the code, just one general note.

We're currently working on a change that will take the rest of the time step control logic from the old TimeManager. This change will modify the code inside the time stepper quite a lot. It would be nice to have a test to avoid regression back to the incorrect behavior here.

I can do the test after we merge the previous two PRs.

jwboth:
@Yuriyzabegaev I added another change. I moved the communication to the model to the end of the routine. The reason is that otherwise the statistics has no chance to save the just logged `time_step_status`. I understand that this code is not finalized. In my view, these changes, just bring the current code in a state such that the integration at least works as intended. At the same time, I am expecting you will turn some of these bits fully around. Let me know what you think. I can also live with these changes only being local to continue using `SolverStatistics` for performance tracking.

Please have a careful look at the change how the model is reacting to time step status (and not solver_status). Feel free to find this ugly. I chose to make explicit what was going on.

jwboth:
> The time stepper was left by me in a messy state, thanks for catching this! I have no comments about the code, just one general note.
> 
> We're currently working on a change that will take the rest of the time step control logic from the old TimeManager. This change will modify the code inside the time stepper quite a lot. It would be nice to have a test to avoid regression back to the incorrect behavior here.
> 
> I can do the test after we merge the previous two PRs.

I had a look at the failing test. Should be fixed, incl. the previously missing entry for the first attempted time step.
I have not thought about the other test you mentioned. What do you think is missing?

tests/time_stepper/test_time_stepper.py
```diff
@@ -746,13 +740,31 @@ def test_solve_failure_time_dependent_statistics(statistics_path: Path):
                 "res_nan": "converged",
             },
         },
-        # Note that the key "0" is abscent. See the comment above for details.
+        "0": {
+            "simulation_status": "in_progress",
+            "solver_status": "failed",
+            "final_time_reached": 1,
+            "time_index": 1,
+            "time": 1,
```
Yuriyzabegaev, line 746: Maybe I misunderstood something, but is this not the original problem you created this issue with, @jwboth?

Mutated time is logged and nonlinear solver failed. Time at keys "0" and "1" should be 0, right?

tests/time_stepper/test_time_stepper.py
```diff
@@ -692,24 +694,16 @@ def test_solve_failure_time_dependent_statistics(statistics_path: Path):
     # fail after two unsuccessful nonlinear iterations.
     status = time_stepper.perform_time_step(model=model, solver=solver)
     assert status.is_failure()
-    # Note that the first attempt (with dt=1) will not be saved in the file. The second
```
Yuriyzabegaev, line 693: It is really nice that you fixed this! I kept the original (incorrect) behavior to avoid breaking something with you in mind, and now we can be consistent in this!

src/porepy/time_stepper/time_stepper.py
```diff
@@ -104,8 +107,11 @@ def perform_time_step(
                 nonlinear_solver_status, model, attempt
             )
 
-            # Save statistics.
-            self._update_time_statistics(model, time_step_status)
+            # Log time step status for statistics.
+            self._update_statistics(model, time_step_status)
```
Yuriyzabegaev: Would you agree on the name `_update_nonlinear_solver_statistics`?

Yuriyzabegaev:
> I have not thought about the other test you mentioned.

I looked at the test and it covers the issue. No need for another test. Thanks!

tests/time_stepper/test_time_stepper.py
jwboth: It depends, what one wants to be logged I presume. My sense was that we log the time that has been attempted to be solved during that time step, i.e., the trial time. In the first call of the solver (key "0"), starting from time 0, with dt=1, one aims at solving the problem at time=1. It failed, see solver_status. Then for the second call of a solver (still first time step, but second attempt, and here key "1"), the time step size had been reduced to 0.5, time is therefore 0.5, but again the solver fails. Finally, in the third solver-unit, the solver succeeds. In this mock test, the time step size was not cut, so also in key "2", dt=0.5 and time=0.5. In this sense, the logic is consistent across successful and failed time steps. And one can retell the solver behavior over time just based on reading the logged file. This was not the case before. Of course one can also design a different logic which differs between successful and failed time steps. Open for other suggestions.

src/porepy/time_stepper/time_stepper.py
jwboth: Agreed and changed.

jwboth:
Failing tests most certainly related with #1743.

tests/time_stepper/test_time_stepper.py
Yuriyzabegaev: Fair!

Yuriyzabegaev (approved):

keileg (approved):

PR #1742, opened 2026-08-12
Generalized lithostatic stress

jwboth:
This PR combines three extensions for the lithostatic stress boundary condition and closely related Neumann BC:
- The Neumann BC class can be customized to conveniently fix components of displacements, not only in single points (like mid points or corners as done currently) but on e.g. entire sides etc. The current change enables fixing a list/array of indices alongside single indices as before.
- The Lithostatic BC now is applicable also to 2D models.
- REMOVED from githistory: In view of initialization routines, in which consistent boundary conditions at time t=0 are required, a simple approach is suggested to control the activity of the lithostatic stress at time t=0. The original behavior remains to be the default behavior. If not liked, this is added as a single commit, which is easy to remove.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/applications/boundary_conditions/model_boundary_conditions.py
+++ b/src/porepy/applications/boundary_conditions/model_boundary_conditions.py
@@ -531,7 +531,7 @@ class LithostaticBoundaryStressValues(GravityMagnitude):
 
         """
         datum_lithostatic_stress = self.params.get(
-            "datum_lithostatic_stress", np.zeros(3)
+            "datum_lithostatic_stress", np.zeros(self.nd)
         )
         return self.units.convert_units(datum_lithostatic_stress, units="Pa")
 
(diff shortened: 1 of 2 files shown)


IvarStefansson (commented):
Thanks for the 2d generalisation and allowing picking other faces. As a side note, I actually don't think our current choice of faces is optimal in any sense. On the initial strength: I'm in two minds. On the one hand, I can see the logic. However, if the motivation is directly initialisation, I'm wondering if it is a bit premature, since we don't know exactly what that code is going to look like. Please share your own thoughts on this and/or elaborate on the motivation.

jwboth:
> Thanks for the 2d generalisation and allowing picking other faces. As a side note, I actually don't think our current choice of faces is optimal in any sense.

I have no more insight here than you, but I also created scenarios, which were asking for other choices than the default ones. 

> On the initial strength: I'm in two minds. On the one hand, I can see the logic. However, if the motivation is directly initialisation, I'm wondering if it is a bit premature, since we don't know exactly what that code is going to look like. Please share your own thoughts on this and/or elaborate on the motivation.

I see that! As mentioned in the description. It is easy to remove. And I have done so now. I will write my own copy of the LithostaticStress class which enables access to the boundary condition values also at time t=0 for consistent initialization.

IvarStefansson (approved):
OK. Let's do the trimmed-down version now. But fully open to accepting the other changes if and when we see their usefulness for initialisation.

PR #1745, opened 2026-08-13
Fixes #1743

### LINKED ISSUE #1743: np.cross fails in 2d resulting in failing tests
[In numpy 2.5 np.cross requires 3d.](https://numpy.org/doc/stable/reference/generated/numpy.cross.html)

Thus, two tests fail currently during build tests. (I believe this is the reason).

Note, that np.cross also is used across core modules.

keileg:
## Proposed changes
Takes care of two failing tests after update to numpy v2.5. Resolves #1743

As noted in the issue, we use `np.cross` extensively in the core of PorePy. I did a search in the full code base and came across one instance in an untested and (AFAIK) unused function - here I made a best-effort of a fix, deprecated the function and added a warning that the function should be used with caution. I did not verify that the fix actually worked, in my opinion this is not worth the effort. 

Beyond the fix, the remaining usages of `np.cross` should be fine from what I can tell.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/geometry/distances.py
+++ b/src/porepy/geometry/distances.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from typing import Union
+from warnings import warn
 
 import numpy as np
 from scipy.spatial import distance as scidist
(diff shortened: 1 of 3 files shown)


IvarStefansson (approved):
Thanks for fixing. Looks good to me.

PR #1747, opened 2026-08-14
Introduce source and target spaces for operator domains

keileg:
UPDATE: REVIEW GUIDANCE

An initial QC on the overall design was provided by @Yuriyzabegaev, leading to significant improvements in the code. All these comments should now have been taken care of. I suggest that the next review stage focus (at least) on the following points:
1. The syntax for operator spaces and when they must be declared. I have left some GH review comments in `constitutive_laws.py` and `test_constitutive_laws.py` to highlight what I think are the main points.
2. Higher level documentation: Should the `OperatorSpace` be documented  beyond usage in code, tutorials, and examples? The natural place to put it would be `equations.ipynb`, but that would be more natural with a slight reconfiguration of that file. Let me know what you think.
3. Also consider the other open GH comments. Most of these point to changes that are more detailed but still worth highlighting.

## Overview
This PR ensures that all Ad operators have a source and target (a better terminology would be 'domain' and 'range', but both these words are already overloaded in PorePy and Python in general). This streamlines the existing code, where some but far from all operators know where they are defined. Also, tests of operator size compatibility are added.

## Technical changes
This PR introduces the following changes (in increasing level of complexity - this is also a suggested order of review):
1. A `GridEntity` enum is now used to encode grid quantity. This replaces strings `"cell"` etc. 
2. A `DomainType` enum is encodes different types of domain. This replaces strings `"subdomain"` etc. It was necessary to construct some special domain types to cover unclear or flexible cases, see code documentation.
4. An `OperatorSpace` class, which wraps dof information, a domain (list of grids, mortar- og boundary grids) and domain type.
5. All operators must be constructed with a source and target `OperatorSpace`. Rolling this out required quite a few changes in the models folder, but these are mainly shallow edits.
6. The Ad `Operations` class (which now resides in its own module) infers its source and range from the source and range of its children operands. That is, if `A` and `B` are two operands and `x` is an operation, the resulting operator `A x B` will have its source and target automatically identified. If the triplet `{A, B, x}` are incompatible, an error will be raised at the time of construction (in contrast, this error will currently be raised during parsing, often with a less then intelligible error message). Also, for the classes `DenseAdArray` and `SparseAdArray` additional checks of array sizes are added to enhance the chances of catching errors.

Tests have been updated and new tests (in particular unit tests for the `OperatorSpace` class) have been added.

Resolves #1601 

## Types of changes

- [ ] Other:

## Checklist


src/porepy/models/constitutive_laws.py
```diff
@@ -199,7 +204,7 @@ def aperture(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
 
         for i, sd in enumerate(subdomains):
             # First make the local aperture array.
-            a_loc = pp.wrap_as_dense_ad_array(self.grid_aperture(sd))
+            a_loc = pp.wrap_as_dense_ad_array(self.grid_aperture(sd), grids=[sd])
```
Yuriyzabegaev, line 207: The syntax is getting a little bit verbose, but it's probably a price worth paying

src/porepy/numerics/ad/operator_space.py
```diff
@@ -0,0 +1,233 @@
...
+    @classmethod
+    def waived(cls) -> OperatorSpace:
+        """Return a sentinel space for operators whose domain/range check is waived."""
+        return cls(DomainType.waived, (), {})
+
+    @classmethod
+    def from_domains(
+        cls,
+        domains: Sequence[pp.Grid | pp.MortarGrid | pp.BoundaryGrid],
+        dof_info: dict[GridEntity, int],
```
Yuriyzabegaev: Would `dofs_info={"cells": 1}` (not correct syntax, but you get the idea) be a good default?

src/porepy/numerics/ad/operator_space.py
```diff
@@ -0,0 +1,233 @@
...
+    ) -> OperatorSpace:
+        """Construct an :class:`OperatorSpace` from a sequence of grids.
+
+        Parameters:
+            domains: Sequence of grid objects.  All grids must be of the same type
+                (all :class:`~porepy.Grid`, all :class:`~porepy.MortarGrid`, or all
+                :class:`~porepy.BoundaryGrid`).
+            dof_info: Mapping from :class:`~porepy.numerics.ad.GridEntity` to the number
+                of DOFs per entity.
+            domain_type: If given, the returned space is forced to have this
```
Yuriyzabegaev, line 194: And what does it mean if it's not given?

src/porepy/numerics/ad/grid_entity.py
```diff
@@ -0,0 +1,24 @@
+"""GridEntity enum for identifying grid entities (cells, faces, nodes)."""
+
+from __future__ import annotations
+
+import enum
+
+__all__ = ["GridEntity"]
+
+
+class GridEntity(enum.Enum):
```
Yuriyzabegaev, line 14: is there a reason why `dof_info` is `dict[GridEntity, int]` aside the historical one? Would this alternative be more convenient?
```
@dataclass
class GridEntities:  # the name is not ideal
    cells: int = 0,
    faces: int = 0,
    nodes: int = 0,
    # and post-init validation that at least one is non-zero
```

We pass it around everywhere and everywhere we access it as `dof_info.get('cells', 0)`. The current change improves the situation, but why stop when it's so close to perfect?

src/porepy/models/contact_mechanics.py
```diff
@@ -598,7 +640,7 @@ class RadialReturnTangentialContactMechanicsEquation:
     tangential_component: Callable[[list[pp.Grid]], pp.ad.Operator]
     """Mapping from a full vector to the tangential component."""
 
-    basis: Callable[[Sequence[pp.GridLike], int], list[pp.ad.Projection]]
+    basis: Callable[..., list[pp.ad.Projection]]
```
Yuriyzabegaev: why?

Yuriyzabegaev: (returning here after seeing the change in the function signature)

By the way, `basis` should be a part of `PorePyModel`, so including it here allows to remove this line completely.

src/porepy/models/geometry.py
```diff
@@ -846,14 +931,19 @@ def outwards_internal_boundary_normals(
         # Normalize by face area if requested.
         if unitary:
             # 1 over cell volumes on the interfaces
-            cell_volumes_inv = pp.ad.Scalar(1) / self.wrap_grid_attribute(
-                interfaces, "cell_volumes", dim=self.nd
-            )
+            cell_volumes_inv = pp.ad.Scalar(
+                1, domains=interfaces
```
Yuriyzabegaev, line 935: I don't understand why a scalar needs to know its domains. When is it practically useful?

src/porepy/models/metric.py
```diff
@@ -327,11 +342,18 @@ def __call__(self, values: np.ndarray) -> dict[str, float]:
         for name, grids_dofs in self.equation_indexer.group_by_name().items():
             indices = np.concatenate(list(grids_dofs.values()))
             domains = cast(pp.GridLikeSequence, list(grids_dofs.keys()))
-            equation_dim = equation_system.equation_image_size_info[name]["cells"]
+            equation_dim = equation_system.equation_image_size_info[name][
```
Yuriyzabegaev: Can this information now be fetched from the operator itself? (operator's OperatorSpace). If so, I vote for purging `equation_system.equation_image_size_info`.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -357,19 +354,27 @@ def construct_equation_indexer(self) -> pp.ad.EquationSystemIndexer:
...
-                        + domain.num_nodes * dofs_info.get("nodes", 0)  # nodes
+                        domain.num_cells
+                        * dofs_info.get(pp.ad.GridEntity.cells, 0)  # cells
+                        + domain.num_nodes
+                        * dofs_info.get(pp.ad.GridEntity.nodes, 0)  # nodes
                     )
                 else:
                     raise ValueError(f"Unknown domain type: {domain}")
 
+                assert dofs_per_grid > 0, (
```
Yuriyzabegaev, line 335: good catch!

src/porepy/numerics/ad/operators.py
```diff
@@ -64,70 +68,6 @@
...
-            cls.pow: "raising to the power of",
-            cls.rpow: "raising to the power of",
-            cls.evaluate: "evaluating",
-            cls.approximate: "approximating",
-            cls.void: "void",
-        }
-        return strings.get(value, "unknown")
-
-
 class Operator:
```
Yuriyzabegaev, line 72: Discussion (not critique): I am not fully convinced that having source and target for each operator as two distinct concepts is convenient. While it is technically correct (the best way of being correct), I have a feeling that most of the operators are self-mappings. These include:
- Scalar
- Variable
- MdVariable
- DenseArray (and time-dependent dense array)
- Function wrappers (not generally, but I can't come up with a single example where they are not used for point-wise functions)
- all arithmetic operations except matrix multiplication

while operators that have different source and target are:
- matrix multiplication
- sparse matrix
- projection

This potentially increases boilerplate, and I subjectively have a feeling that I need to write code that involves those from the first list quite more often. Would it be easier to make the listed operators explicitly self-mappings? They still can preserve getters for `source` and `target`.

The radical solution (that I would like to avoid) is to introduce a parent abstract class `SelfMappingOperator(Operator)`, and `NotSelfMappingOperator(Operator)` (for the lacking of a better name), that handle this logic in `__init__` and getters. I can see from here how ugly it looks.

The less radical alternative is to make `target=None`, and if None then `target = source`.  It can also lead to confusion. What do you think?

src/porepy/numerics/ad/operators.py
```diff
@@ -217,32 +148,21 @@ def __init__(
...
-    def domain_type(self) -> Literal["subdomains", "interfaces", "boundary grids"]:
-        """Type of domains where the operator is defined."""
-        return self._domain_type
+    def target(self) -> OperatorSpace:
+        """The algebraic target space of this operator."""
+        return self._target
 
     @property
-    def domains(self) -> GridLikeSequence:
+    def domains(self) -> list:
```
Yuriyzabegaev: Why change return type?

src/porepy/numerics/ad/operator_functions.py
```diff
@@ -99,11 +104,15 @@ def __call__(self, *args: pp.ad.Operator) -> pp.ad.Operator:
             "Operator functions must be called with at least 1 argument."
         )
 
+        # Space intentionally unspecified: the function application node only stores the
```
Yuriyzabegaev: I don't get this comment. Is it about name?

src/porepy/numerics/ad/operator_space.py
```diff
@@ -0,0 +1,233 @@
...
+                "Interface spaces must be defined on pp.MortarGrid objects."
+            )
+        if self.domain_type == DomainType.boundary_grids and not all(
+            isinstance(g, pp.BoundaryGrid) for g in self.grids
+        ):
+            raise ValueError(
+                "Boundary-grid spaces must be defined on pp.BoundaryGrid objects."
+            )
+
+    def __eq__(self, other: object) -> bool:
```
Yuriyzabegaev: It is a dataclass, why does it need `__eq__` and `__hash__` defined? Don't they come out of the box?

src/porepy/numerics/ad/operator_space.py
```diff
@@ -0,0 +1,233 @@
...
+        return (
+            self.domain_type == other.domain_type
+            and self.grids == other.grids
+            and self.dof_info == other.dof_info
+        )
+
+    def __hash__(self) -> int:
+        return hash((self.domain_type, self.grids, frozenset(self.dof_info.items())))
+
+    def num_dofs(self) -> int:
```
Yuriyzabegaev, line 120: There is definitely at least a single place in `equation_system.py` that does the same thing, probably in construct_equation_indexer. It would be so nice to replace it with this function call!

src/porepy/numerics/ad/operators.py
```diff
@@ -1421,51 +1506,34 @@ def id(self) -> int:
...
     def tags(self) -> dict[str, Any]:
         """A dictionary of tags associated with this variable."""
         return self._tags
 
+    @property
+    def domain(self) -> pp.Grid | pp.MortarGrid:
+        """Returns the subdomain or interface on which this variable is defined."""
+        # This method is added to the atomic variable for the purpose of filtering
+        # in the Indexer class.
+        return cast(pp.Grid | pp.MortarGrid, self.source.grids[0])
```
Yuriyzabegaev: It's probably `self.target.grids[0]` (I realize it is always a self-mapping).

Throughout the codebase, there are quite a few changes from `var.domain` to `var.domains[0]`. If you find the former is more expressive, maybe revert them back (I think an llm can do it pretty quickly). Or if you don't find it more expressive, we should reconsider the its usage in the indexers (because why should this stay only for the sake of indexers)?

I subjectively do find this more expressive (if you know your operator is an atomic variable, you can access its only domain).

src/porepy/numerics/ad/surrogate_operator.py
```diff
@@ -495,10 +512,15 @@ def __call__(self, domains: pp.GridLikeSequence) -> pp.ad.Operator:
...
         # This is for completeness reasons, when calling equations on empty list
         if len(domains) == 0:
-            return pp.wrap_as_dense_ad_array(np.zeros((0,)), name=self.name)
+            return pp.wrap_as_dense_ad_array(
+                np.zeros((0,)), name=self.name, grids=domains
+            )
         # On the boundary, this is a Time-Dependent dense array
         elif all(isinstance(grid, pp.BoundaryGrid) for grid in domains):
-            return pp.ad.TimeDependentDenseArray(self.name, domains)
+            # TODO: domains should be non-empty. Can we drop domain_type here?
```
Yuriyzabegaev: Good comment! Can't we always deduce domain type from domains?

src/porepy/numerics/ad/surrogate_operator.py
```diff
@@ -645,14 +668,14 @@ def num_dofs_on_grid(self, grid: pp.GridLike) -> int:
...
-            return self._dof_info.get("cells", 1) * grid.num_cells
+            return self._dof_info.get(GridEntity.cells, 1) * grid.num_cells
         elif isinstance(grid, pp.Grid):
             # NOTE cannot use default value of scalar, cell-wise, to not mess with
             # cases where the user defines only node- or face-wise dofs.
             return (
-                self._dof_info.get("cells", 0) * grid.num_cells
-                + self._dof_info.get("faces", 0) * grid.num_faces
-                + self._dof_info.get("nodes", 0) * grid.num_nodes
+                self._dof_info.get(GridEntity.cells, 0) * grid.num_cells
```
Yuriyzabegaev: can this routine become a public method of `Operator`? (counting operator's number of dofs)

src/porepy/numerics/vem/dual_elliptic.py
```diff
@@ -109,6 +112,16 @@ def ndof(self, sd: pp.Grid) -> int:
         """
         return sd.num_cells + sd.num_faces
 
+    def get_row_dof_info(
+        self, matrix_key: str = "", nd: int = 1
+    ) -> dict["GridEntity", int]:
+        return {}
```
Yuriyzabegaev: Could we add a single-line comment to explain why it's empty?

src/porepy/numerics/discretization.py
```diff
@@ -120,6 +123,50 @@ def assemble_matrix_rhs(
...
+
+        Raises:
+            ValueError: If the matrix_key is not recognized by this discretization.
+
+        Returns:
+            A mapping from :class:`~porepy.numerics.ad.GridEntity` to the number of
+            DOFs per grid entity that occupy the rows of the named matrix.
+
+        """
+        pass
```
Yuriyzabegaev: `pass` is not needed if a docstring is present (python counts the docstring as an expression). If this is intentional for the style sake, please ignore this comment!

Yuriyzabegaev (commented):
I have read through everything, great job!

I only looked at the tests less thoroughly. Is there a test that tries to combine (add / multiply / apply / etc) incompatible operators and catches the error early? (before the actual AD parsing). Sorry if I didn't see it.

See my comment about self-mapping operators, that's the only thing worth thinking about (and not sure that it's worth doing anything about).

Other than that, just minor comments, everything looks good!

src/porepy/models/geometry.py
keileg: It makes it possible to enforce consistency checks also on scalars. Whether it is worth the effort is case dependent, but I think it good to have the possibilty.

src/porepy/numerics/ad/equation_system.py
keileg: Well, it was you who found it, so I agree.

src/porepy/numerics/ad/grid_entity.py
keileg: Agreed, this is worth exploring. Will do.

src/porepy/numerics/ad/operator_functions.py
keileg: Will remove.

src/porepy/numerics/ad/operator_space.py
keileg: Will look into

src/porepy/numerics/ad/operator_space.py
keileg: Agree, I will explore this.

src/porepy/numerics/ad/operators.py
keileg: I have asked the same question, specifically to make target=None the default. Will explore.

src/porepy/numerics/ad/surrogate_operator.py
```diff
@@ -467,7 +484,7 @@ def __init__(
         self._name: str = name
         """See :meth:`name`."""
 
-        self._dof_info: dict[pp.ad.equation_system.GridEntity, int] = dof_info
+        self._dof_info: dict[GridEntity, int] = dof_info
```
keileg: Remove - this information should now be accessed from self.source

src/porepy/numerics/ad/surrogate_operator.py
keileg: I would think the whole num_dofs_on_grid to be a candidate. This is partly implemented, sort of, in the methods used for shape checks of Sparse and Dense Ad operators, so if we implement, we should unify with that usage as well.

src/porepy/models/metric.py
keileg: Should be explored.

src/porepy/numerics/ad/operator_space.py
keileg: Should be explored.

src/porepy/numerics/ad/grid_entity.py
keileg: I ended up implementing a dataclass GridEntities that maps from GridEntity to dof info. I'm not convinced about the name GridEntity, suggestions are most welcome.

For completeness, it was also tempting to get rid of the GridEntity enum, but that serves a different purpose than the dof info in GridEntities and needs to stay.

src/porepy/numerics/ad/operator_space.py
keileg: Deleted after making the dataclass frozen.

src/porepy/models/contact_mechanics.py
keileg: Fixed.

src/porepy/models/metric.py
keileg: Purged. Yahoo!

src/porepy/numerics/ad/operator_space.py
keileg: Improved documentation

src/porepy/numerics/ad/operators.py
keileg: Fixed

src/porepy/numerics/ad/operators.py
keileg: Agreed. Implemented.

src/porepy/numerics/ad/surrogate_operator.py
keileg: Yes, because we know domains is non-empty. Removed comment.

src/porepy/numerics/ad/operator_space.py
keileg: Implemented.

tests/numerics/ad/test_grid_operators.py
```diff
@@ -1,15 +1,21 @@
...
-    test_subdomain_projections: Operators for restriction and prolongation are checked
-        for both faces and cells;
-    test_mortar_projections_empty_list: Projections between empty lists of subdomains
-        and interfaces;
-    test_mortar_projections: Projections between mortar grids and subdomain grids;
-    test_boundary_grid_projection:  Tests are conducted on the boundary projection
-        operator and its inverse;
-    test_trace and test_divergence: Operators for discrete traces and divergences;
-    test_ad_discretization_class: test for AD discretizations.
+"""Tests for the AD representations of grid-related operators: projections between
```
keileg, line 1: I ended up refactoring this file as part of augmenting some of the tests contained. Summarized, the changes are:
1. Tests for subdomain, interface and boundary projections are grouped into separate classes.
2. Each of these sets of tests use parametrization and fixtures/common setups to a much larger extent.
3. The tests also cover the use of `OperatorSpaces`.
Only the last change is considers the core changes in this PR.

I have verified that the de facto test coverage did not decrease (in some places in was improved). In opinion, this file need not be a focal point for the review.

src/porepy/models/abstract_equations.py
```diff
@@ -310,11 +310,7 @@ def eliminate_locally(
         equ_name = f"elimination_of_{sec_var.name}_on_grids_{g_ids}"
         local_equ = sec_var - sec_expr(non_boundaries)
         local_equ.set_name(equ_name)
-        self.equation_system.set_equation(
```
keileg, line 313: @Yuriyzabegaev Update since laste time: `set_equations` no longer needs grids and dof-information, all is inferred from the operator.

src/porepy/models/constitutive_laws.py
```diff
@@ -370,12 +381,17 @@ def aperture(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
                 normal_jump = self.normal_component(
                     subdomains_of_dim
                 ) @ self.displacement_jump(subdomains_of_dim)
+                domain_and_range = OperatorSpace.from_domains(subdomains_of_dim)
```
keileg, line 384: Note: Dof-information for an operator defaults to a single cell unknown.

src/porepy/models/constitutive_laws.py
```diff
@@ -1334,7 +1378,9 @@ def diffusive_flux(
         # from half-faces to faces, using a mapping with signs, thereby taking the
         # difference between the two vector sources.
         vector_source_c_to_f = pp.ad.SparseArray(
-            hf_to_f @ d_vec @ cells_to_hf_3d @ cells_nd_to_3d
+            hf_to_f @ d_vec @ cells_to_hf_3d @ cells_nd_to_3d,
+            source=cell_vector_space,
```
keileg, line 1382: Sparse and Dense Ad arrays require source and target explicitly stated. Other operators are self-mappings by default.

src/porepy/models/constitutive_laws.py
```diff
@@ -1587,7 +1643,22 @@ def __transmissibility_matrix(
...
+            {pp.ad.GridEntity.cells: 9},
+            domain_type=pp.ad.DomainType.subdomains,
+        )
+        # The number of half-transmissibilities equals two times the number of internal
+        # faces plus the number of external faces. This cannot be expressed by the
+        # current OperatorSpaces, so we explicitly disable the compatibility checks
+        # here.
+        t_hf_inv = one / (
+            pp.ad.SparseArray(
+                d_n_by_dist, source=hf_source, target=pp.ad.OperatorSpace.waived()
```
keileg, line 1658: Note: This is one of few places where the waived space is used (the other cases are similarly related to the differentiable tpfa functions). A cost-benefit analysis made it very clear it is not worth the effort to extend the machinery.

src/porepy/numerics/ad/_operations.py
```diff
@@ -0,0 +1,259 @@
+"""The module contains the Operations enum, which is used to identify the operation
```
keileg, line 1: Moved the operations to a separate file out of `operators.py`. 

The functional change in this file is the method `infer_source_target` and its subfunctions. This dictates the rules for which combinations are permissible. I have reviewed this carefully in several rounds, but it is worth having a look.

src/porepy/numerics/ad/discretizations.py
```diff
@@ -43,7 +43,7 @@
 ]
 
 
-class Discretization(abc.ABC):
+class DiscretizationAd(abc.ABC):
```
keileg, line 46: Renamed to discretizationAd. We already have a Discretization class (the standard discretization).

src/porepy/numerics/ad/equation_system.py
```diff
@@ -360,22 +341,30 @@ def construct_equation_indexer(self) -> pp.ad.EquationSystemIndexer:
             offset = 0
 
             for domain in equation.domains:
-                dofs_info = self.equation_image_size_info[name]
+                dofs_info = dict(equation.target.dof_info)
```
keileg: `EquationSystem.eqaution_image_size_info` is gone.

src/porepy/numerics/ad/equation_system.py
```diff
@@ -982,18 +976,13 @@ def dofs_of(self, variables: VariableList) -> np.ndarray:
     def set_equation(
         self,
         equation: Operator,
-        grids: DomainList,
-        equations_per_grid_entity: dict[GridEntity, int],
+        equations_per_grid_entity: Optional[dict[GridEntity, int]] = None,
```
keileg: Input needed: Should we keep the option to pass equations_per_grid_entity for debugging purposes or let it go?

src/porepy/numerics/ad/grid_entity.py
```diff
@@ -0,0 +1,156 @@
...
+        faces: Degrees of freedom located at cell faces.
+        nodes: Degrees of freedom located at grid nodes (vertices).
+    """
+
+    cells = "cells"
+    faces = "faces"
+    nodes = "nodes"
+
+
+@dataclasses.dataclass(frozen=True, eq=False)
```
keileg: New dataclass (which behaves like a dictionary) to represent dof information.

src/porepy/numerics/ad/operator_space.py
keileg: Replaced.

tests/models/test_constitutive_laws.py
```diff
@@ -582,10 +582,9 @@ def matrix_permeability(self, subdomains: list[pp.Grid]) -> pp.ad.Operator:
         # Basis vector for the yy-component
         e_yy = self.e_i(subdomains, i=4, dim=tensor_dim)
 
-        return (
-            pp.wrap_as_dense_ad_array(all_vals, name="Constant_permeability_component")
-            + e_yy @ self.pressure(subdomains) ** 2
-        )
+        return pp.wrap_as_dense_ad_array(
+            all_vals, name="Constant_permeability_component", grids=subdomains
+        ) + e_yy @ self.pressure(subdomains) ** pp.ad.Scalar(2)
```
keileg, line 587: This illustrates how scalars now must be wrapped as pp.ad.Scalar (it used to be possible to circumvent this).

src/porepy/models/constitutive_laws.py
```diff
@@ -3107,22 +3225,33 @@ def fracture_stress(self, interfaces: list[pp.MortarGrid]) -> pp.ad.Operator:
...
+        # with contact_traction (also defined on fracture_subdomains) while the two
+        # still share that domain. Expand the cell-wise scalar characteristic traction
+        # to an nd-vector before scaling the local fracture traction.
+        scalar_to_nd = pp.ad.sum_projection_list(
+            self.basis(
+                fracture_subdomains,
+                dim=self.nd,
+                domain_type=pp.ad.DomainType.subdomains,
+            )
+        )
```
keileg, line 3239: Note the switched order here: The change is necessary to work for non-scalar characteristic quantities. I believe the old implementation could have failed in that case.

keileg:
@Yuriyzabegaev I have implemented several changes and improvements based on your suggestions. The PR should be ready for re-review now.

src/porepy/models/abstract_equations.py
Yuriyzabegaev: awesome!

src/porepy/numerics/ad/equation_system.py
Yuriyzabegaev: I vote for letting it go.

src/porepy/numerics/ad/grid_entity.py
Yuriyzabegaev: awesome!

src/porepy/numerics/ad/grid_entity.py
Yuriyzabegaev: I probably does not see the full picture, but why does it have to behave as a dict?

If there is a real reason somewhere in the code that it "must be a dict", then it's a serious pitfall, and what is a benefit of introducing this class?

I also see how it can be an LLM's overcomplication (considering how rigorously it's explained in the docstring how it behaves both as a dict and a struct), and if so, I suggest purging dict-like behavior.

src/porepy/numerics/ad/operator_space.py
Yuriyzabegaev: awesome!

src/porepy/numerics/ad/operator_space.py
Yuriyzabegaev: awesome!

src/porepy/numerics/ad/operators.py
Yuriyzabegaev: I see the change is made here. Awesome!

Yuriyzabegaev (approved):
I did a re-review, only reading the comments (and only looking at the code changes near the comments). Everything looks good from my perspective! Please only see my concern in the comment about the GridEntities class.

src/porepy/numerics/ad/grid_entity.py
keileg: I agree. Will do.

src/porepy/numerics/ad/surrogate_operator.py
keileg: Not really: this is an operator factory, not an operator, which has no source/target.

src/porepy/models/constitutive_laws.py
IvarStefansson: Would you say this is in principle doable in the future?

src/porepy/models/constitutive_laws.py
IvarStefansson: Is that a quirk of order of operation rules for ad?

src/porepy/numerics/ad/_operations.py
```diff
@@ -0,0 +1,259 @@
...
+            :class:`OperatorSpace` for the source and ``target`` is the inferred
+            :class:`OperatorSpace` for the target.
+
+        """
+        left_is_scalar = left.source.domain_type == DomainType.scalar
+        right_is_scalar = right.source.domain_type == DomainType.scalar
+
+        if self == Operations.matmul:
+            # left @ right: target(right) must equal source(left)
+            return self._process_matmul(left, right, right_is_scalar)
```
IvarStefansson: Is matmul with scalars allowed?

src/porepy/numerics/ad/_operations.py
```diff
@@ -0,0 +1,259 @@
...
+                        f" {left.target} vs {right.target}."
+                    )
+                    raise ValueError(s)
+                return (
+                    self._pick_source(left.source, right.source),
+                    self._pick_target(left.target, right.target),
+                )
+
+    def _process_matmul(
+        self, first, second, second_is_scalar: bool
```
IvarStefansson: Why use both left/right and first/second?

src/porepy/models/geometry.py
```diff
@@ -415,11 +418,27 @@ def wrap_grid_attribute(
...
+        array = pp.ad.DenseArray(vals, source=domain_and_range, target=domain_and_range)
         array.set_name(f"Array wrapping attribute {attr} on {len(grids)} grids.")
         return array
 
-    def basis(self, grids: Sequence[pp.GridLike], dim: int) -> list[pp.ad.Projection]:
+    def basis(
+        self,
+        grids: Sequence[pp.GridLike],
+        dim: int,
+        domain_type: Optional[pp.ad.DomainType] = None,
```
IvarStefansson, line 435: Why optional? Seems error prone to me.

src/porepy/models/geometry.py
```diff
@@ -485,6 +510,8 @@ def e_i(
             grids: List of grids on which the basis vector is defined.
             i: Index of the basis function. Note: Counts from 0.
             dim: Dimension of the functions.
+            domain_type: The type of domain (subdomains, interfaces, or boundary
+                grids) that *grids* represents. See :meth:`basis` for details.
```
IvarStefansson: I can't see more details in basis.

src/porepy/models/geometry.py
```diff
@@ -504,11 +531,34 @@ def e_i(
...
+            domain_type is None
+            and len(grids) > 0
+            and not (
+                all(isinstance(g, pp.Grid) for g in grids)
+                or all(isinstance(g, pp.MortarGrid) for g in grids)
+                or all(isinstance(g, pp.BoundaryGrid) for g in grids)
+            )
+        ):
+            # The grids mix subdomains, interfaces, and/or boundary grids, as is
+            # supported by this method. Mark the operator space as unclear.
```
IvarStefansson: Where do we need that?

src/porepy/models/geometry.py
```diff
@@ -803,7 +880,12 @@ def outwards_internal_boundary_normals(
         """
         if len(interfaces) == 0:
             # Special case if no interfaces.
-            return pp.ad.DenseArray(np.zeros(0))
+            space = pp.ad.OperatorSpace.from_domains(
+                interfaces,
+                {pp.ad.GridEntity.cells: self.nd},
```
IvarStefansson, line 869: Please explain why the space is nd even if we have no cells.

src/porepy/numerics/ad/_operations.py
```diff
@@ -0,0 +1,259 @@
...
+        return space.dof_info.is_unit_on_single_entity()
+
+    def _spaces_compatible(self, a: OperatorSpace, b: OperatorSpace) -> bool:
+        """Return True if a and b represent compatible operator spaces."""
+        if a == b:
+            # Spaces are equal. Fine.
+            return True
+        if len(a.grids) == 0 and len(b.grids) == 0:
+            # None of the spaces carry any grids, so they are considered compatible
+            # (even if they have different domain types).
```
IvarStefansson, line 220: Is that a good idea?

src/porepy/numerics/ad/_operations.py
```diff
@@ -0,0 +1,259 @@
...
+            # type, and one of them is a cellwise scalar, hence we can broadcast the
+            # result to the other space.
+            if self._can_broadcast(a) or self._can_broadcast(b):
+                return True
+        return False
+
+    def _pick_target(self, a: OperatorSpace, b: OperatorSpace) -> OperatorSpace:
+        """Pick the target space.
+
+        It assumed that the caller has already verified that the two spaces are
```
IvarStefansson:
```suggestion
        It is assumed that the caller has already verified that the two spaces are
```

src/porepy/numerics/ad/_operations.py
```diff
@@ -0,0 +1,259 @@
...
+        if self._can_broadcast(a) and not self._can_broadcast(b):
+            return b
+        if self._can_broadcast(b) and not self._can_broadcast(a):
+            return a
+        return a
+
+    def _pick_source(self, a: OperatorSpace, b: OperatorSpace) -> OperatorSpace:
+        """Pick the source space.
+
+        It assumed that the caller has already verified that the two spaces are
```
IvarStefansson:
```suggestion
        It is assumed that the caller has already verified that the two spaces are
```

src/porepy/numerics/ad/ad_utils.py
```diff
@@ -243,28 +254,25 @@ def uniquify_discretization_list(
...
             cls_obj_map[cls] = discr._discr
             cls_key_covered.append(key)
 
-            # Add new discretization with associated list of subdomains.
-            # Need a copy here to avoid assigning additional subdomains to this
-            # discretization (if not copy, this may happen if
-            # the key-discr combination is encountered a second time and the
-            # code enters the if part of this if-else).
-            grid_likes = discr.subdomains.copy() + discr.interfaces.copy()
+            # Add new discretization with associated list of grids. Conversoin to list
```
IvarStefansson:
```suggestion
            # Add new discretization with associated list of grids. Conversion to list
```

src/porepy/numerics/ad/grid_entity.py
```diff
@@ -0,0 +1,114 @@
...
+                    "GridEntity member (e.g. GridEntity.cells), not "
+                    f"{type(entity).__name__}."
+                )
+            kwargs[entity.value] = count
+        return cls(**kwargs)
+
+    @property
+    def present_entities(self) -> frozenset[GridEntity]:
+        """The grid entities that carry a nonzero number of DOFs."""
+        return frozenset(
```
IvarStefansson: Why frozenset?

src/porepy/numerics/ad/grid_operators.py
```diff
@@ -642,6 +699,33 @@ def secondary_to_mortar_avg(self) -> SparseArray:
...
+        """Construct the operator spaces for the interfaces and subdomains stored on
+        this instance.
+
+        Parameters:
+            sd_entity: The grid entity (cells or faces) the subdomain space is defined
+                on.
+
+        Returns:
+            A 2-tuple ``(intf_space, sd_space)`` with the operator space for the
+            interfaces, and for the subdomains.
```
IvarStefansson: Why this comma?

src/porepy/numerics/ad/operator_space.py
```diff
@@ -0,0 +1,231 @@
...
+    """
+
+    subdomains = "subdomains"
+    interfaces = "interfaces"
+    boundary_grids = "boundary_grids"
+    scalar = "scalar"
+    unclear = "unclear"
+    """Used for composits formed by operators with different domains."""
+    waived = "waived"
+    """Used to explicitly waive the domain/range check for an operator.  This should
```
IvarStefansson:
```suggestion
    """Used to explicitly waive the domain/range check for an operator. This should
```

src/porepy/numerics/ad/operators.py
```diff
@@ -217,32 +156,19 @@ def __init__(
...
+    def target(self) -> OperatorSpace:
+        """The algebraic target space of this operator."""
+        return self._target
 
     @property
-    def domains(self) -> GridLikeSequence:
+    def domains(self) -> tuple[GridLike, ...]:
         """List of domains where the operator is defined."""
-        return self._domains
+        return self._target.grids
```
IvarStefansson, line 171: Why is this target rather than source?

src/porepy/numerics/fv/fv_elliptic.py
```diff
@@ -64,6 +64,64 @@ def ndof(self, sd: pp.Grid) -> int:
         """
         return sd.num_cells
 
+    def get_row_dof_info(self, matrix_key: str = "", nd: int = 1) -> pp.ad.GridEntities:
+        """Return row DOF info for the named FVElliptic matrix.
```
IvarStefansson: I thought we reserved "DOF" for unknowns, not equation numbering.

tests/numerics/ad/test_grid_entity.py
```diff
@@ -0,0 +1,204 @@
...
+    def test_enum_has_three_members(self):
+        assert len(GridEntity) == 3
+
+    def test_construction_from_string(self):
+        """GridEntity('cells') == GridEntity.cells."""
+        assert GridEntity("cells") is GridEntity.cells
+        assert GridEntity("faces") is GridEntity.faces
+        assert GridEntity("nodes") is GridEntity.nodes
+
+    def test_accessible_via_pp_ad(self):
```
IvarStefansson: What's the value of this?

tests/numerics/ad/test_grid_entity.py
```diff
@@ -0,0 +1,204 @@
...
+        """GridEntity is accessible as pp.ad.GridEntity."""
+        assert pp.ad.GridEntity is GridEntity
+
+
+@pytest.fixture
+def simple_mdg(scope="module"):
```
IvarStefansson: Shouldn't this be
@pytest.fixture(scope="module")
def simple_mdg():
?

tests/numerics/ad/test_grid_entity.py
```diff
@@ -0,0 +1,204 @@
...
+    """Return a minimal MixedDimensionalGrid with two subdomains."""
+    mdg = pp.MixedDimensionalGrid()
+    g1 = pp.CartGrid([2, 2])
+    g2 = pp.CartGrid([3, 3])
+    mdg.add_subdomains([g1, g2])
+    return mdg, g1, g2
+
+
+class TestCreateVariables:
+    """Test that create_variables works with enum-keyed dof_info dicts and return
```
IvarStefansson:
```suggestion
    """Test that create_variables works with enum-keyed dof_info dicts and returns
```

tests/numerics/ad/test_grid_entity.py
```diff
@@ -0,0 +1,204 @@
...
+            ({}, GridEntities()),
+            ({GridEntity.cells: 2}, GridEntities(cells=2)),
+            (
+                {GridEntity.cells: 2, GridEntity.faces: 1},
+                GridEntities(cells=2, faces=1),
+            ),
+        ],
+        ids=["empty", "single", "double"],
+    )
+    def test_from_mapping_normalizes_dict(self, dof_info, expected):
```
IvarStefansson: What does "normalizes" mean in this context?

tests/numerics/ad/test_grid_operators.py
IvarStefansson, line 1: TLDR. This feels like out of a sensible PR scope TBH.

tests/numerics/ad/test_operator_space.py
```diff
@@ -0,0 +1,794 @@
...
+    ``DenseArray``/``SparseArray``, ``TimeDependentDenseArray``, ``MergedOperator``)
+    acquiring their ``source``/``target`` from their constructor arguments, including
+    the shape-consistency check between a grid-based space and the wrapped
+    array/matrix.
+  - Propagation through arithmetic and composition: ``source``/``target`` propagating
+    through operator arithmetic and composition (elementwise operations, ``@``,
+    ``sum_operator_list``), including the ``unclear``-domain sentinel and the errors
+    raised for incompatible spaces.
+
+Space checks for the grid operators themselves (``SubdomainProjections``,
```
IvarStefansson, line 19: Good comment!

tests/numerics/ad/test_operator_space.py
```diff
@@ -0,0 +1,794 @@
...
+
+
+class TestOperatorSpaceScalar:
+    def test_scalar_factory(self):
+        s = OperatorSpace.scalar()
+        assert s.domain_type == DomainType.scalar
+        assert s.grids == ()
+        assert s.dof_info == GridEntities()
+
+    def test_scalar_is_singleton_value(self):
```
IvarStefansson: Doesn't "singleton" imply identity?

tests/numerics/ad/test_operator_space.py
```diff
@@ -0,0 +1,794 @@
...
+        assert space.domain_type == DomainType.scalar
+
+    def test_subdomains(self, two_subdomains):
+        g1, g2 = two_subdomains
+        space = OperatorSpace.from_domains([g1, g2], {GridEntity.cells: 1})
+        assert space.domain_type == DomainType.subdomains
+        assert space.grids == (g1, g2)
+        assert space.dof_info == GridEntities(cells=1)
+
+    def test_interfaces(self, one_mortar):
```
IvarStefansson:
```suggestion
    def test_interface(self, one_mortar):
```

tests/numerics/ad/test_operator_space.py
```diff
@@ -0,0 +1,794 @@
...
+        operator is assumed to be a self-mapping)."""
+        space = OperatorSpace.from_domains([two_subdomains[0]], {GridEntity.cells: 1})
+        op_omitted = Operator(name="test", source=space)
+        assert op_omitted.target == space
+
+        op_explicit_none = Operator(name="test", source=space, target=None)
+        assert op_explicit_none.target == space
+
+    def test_set_source_target_in_init(self, two_subdomains):
+        space = OperatorSpace.from_domains([two_subdomains[0]], {GridEntity.cells: 1})
```
IvarStefansson: Wouldn't this test be stronger if using different target and source spaces?

tests/numerics/ad/test_operator_space.py
```diff
@@ -0,0 +1,794 @@
...
+
+class TestArraySpace:
+    """DenseArray and SparseArray both carry the explicit source/target OperatorSpace
+    passed to their constructor."""
+
+    @pytest.mark.parametrize(
+        "array_cls, make_data",
+        [
+            (DenseArray, _ones_for),
+            (SparseArray, lambda space: sps.eye(space.num_dofs(), format="csr")),
```
IvarStefansson: Isn't this eye_for?

tests/numerics/ad/test_operator_space.py
```diff
@@ -0,0 +1,794 @@
...
+            SparseArray(mat, source=self.cell_space, target=self.face_space)
+
+    @pytest.mark.parametrize(
+        "space",
+        [OperatorSpace.scalar(), OperatorSpace.unclear(), OperatorSpace.waived()],
+    )
+    def test_spaces_skip_check(self, space):
+        """Spaces that are not grid-based (scalar, unclear) skip the shape check."""
+        arr = np.ones(2)
+        DenseArray(arr, source=space, target=space)
```
IvarStefansson, line 495: So DenseArray can be scalar space? I thought that would be reserved for Scalar.

tests/numerics/ad/test_operators.py
```diff
@@ -29,124 +29,107 @@
...
+
+        # First check that the two copies have behaved as they should.
+        # The operators should be the same for all trees.
+        for item in ["operation", "source", "target"]:
+            assert getattr(self.c, item) == getattr(self.c_copy, item)
+            assert getattr(self.c, item) == getattr(self.c_deepcopy, item)
+
+        # The operator version of scalars and dense arrays calculates the hash based on
+        # the value of the underlying object, hence the comparison operator for
+        # pp.ad.Operator should evaluate for True for both the copy and the deepcopy.
```
IvarStefansson:
```suggestion
        # pp.ad.Operator should evaluate to True for both the copy and the deepcopy.
```

tests/numerics/ad/test_operators.py
```diff
@@ -29,124 +29,107 @@
...
+        # structures to be wrapped
+        self.a_val = 42
+        self.a = pp.ad.Scalar(self.a_val)
+        # Use an unclear operator space here since we do not care about the actual
+        # domain of the DenseArray (tests of this is carried out elsewhere).
+        space = pp.ad.OperatorSpace.unclear()
+        b_val = np.arange(3)
+        self.b = pp.ad.DenseArray(b_val, source=space, target=space)
+
+        # The combined operator, and two copies
```
IvarStefansson:
```suggestion
        # The combined operator and two copies.
```

tutorials/fluid_modeling.ipynb
IvarStefansson, line 1: Kill output diff?

tutorials/convergence_criteria.ipynb
IvarStefansson, line 1: Kill output diff?

IvarStefansson (commented):
Quite a few specific comments. Overall: Possibly unfair, but I'm not entirely sure what these almost 4k of net new code buy us. I note that substantive overhead is needed when defining operators etc. Let's discuss.

src/porepy/numerics/ad/operator_space.py
```diff
@@ -0,0 +1,231 @@
...
+            dof_info: Mapping from :class:`~porepy.numerics.ad.GridEntity` to the number
+                of DOFs per entity, or an existing
+                :class:`~porepy.numerics.ad.grid_entity.GridEntities`. If not given
+                (the default), one DOF per cell is assumed.
+            domain_type: If given, the returned space is forced to have this
+                domain type, and an empty ``domains`` sequence will *not* be interpreted
+                as a scalar space. Needed by grid operators whose domain type is known
+                from context even if they are constructed on an empty list of grids.
+                If not given (the default), the domain type is inferred from the grid
+                types found in ``domains``.
```
keileg: TODO: Raise if domain_type is not None and does not match the given domains (assuming domain list is non-empty.)

src/porepy/models/geometry.py
keileg: domain_type is only used if the grids list is empty, *and* the grid should not be assigned a scalar operator space.

TODO: Document this.

src/porepy/models/geometry.py
keileg: Removed misleading comment.

src/porepy/models/geometry.py
keileg: Not needed.

TODO: Remove this functionality. Check if traces of this flexibility are found in other parts of the code. Remove if possible, report back if not.

src/porepy/models/geometry.py
keileg: We can still do consistency checks on combined operators: If you combine this space with something that has an empty domain list but one cell dof ,the checks will raise. 

While this will add little to the existing code, which we by now trust have compatible operators, I would argue it is needed to safely manipulate operators (be it by dropping domains, terms etc.), as I believe we will do much more of during nonlinear solves.

src/porepy/numerics/ad/_operations.py
keileg: TODO: This is not allowed. Fix.

src/porepy/numerics/ad/_operations.py
keileg: TODO: Unify into left/right.

src/porepy/numerics/ad/_operations.py
keileg: You are right.

TODO: Fix, clean up any calling code giving issues here.

src/porepy/numerics/ad/grid_entity.py
keileg: TODO: Replace with tuple if functionally equivalent.

src/porepy/numerics/ad/operators.py
keileg: TODO: Switch?

tests/numerics/ad/test_grid_entity.py
keileg: TODO: Delete.

tests/numerics/ad/test_grid_operators.py
keileg: Yes, I had the same doubt and could (perhaps should) have done the refactoring in a separate PR (see my comment on line 1.)

tests/numerics/ad/test_operator_space.py
keileg: TODO: Implement.

tests/numerics/ad/test_operator_space.py
keileg: If the space is initialized with an empty domain list, a Scalar is assigned by default.

src/porepy/models/constitutive_laws.py
keileg: My feeling is, not unless we extend the OperatorSpace concept. With half-transmissibilities we get two items for each internal face, one for each external. Altertatively, we get one item per cell-face combination. Both can be implemented, but both will require processing grid topology is some sense.

src/porepy/models/constitutive_laws.py
keileg: Rather, when we tried to account for non-scalar quantities a year or two ago (I did that implementation), we did not fully succeed. AFAIK, this has never been tested, hence the QC on that scenario was not good enough?

src/porepy/numerics/fv/fv_elliptic.py
keileg: You are right.

TODO: Fix. If needed, follow up in similar methods in other discretization classes.

src/porepy/numerics/ad/operators.py
keileg: Terminology is mixed up here: The indexer uses 'domains' to identify the domains (as in subdomains/interfaces) of the range (as in the target space) of the operator. We will keep this target for now, or else several tests fail, but I will create an issue to clean up these concepts.

src/porepy/models/geometry.py
IvarStefansson: Feel free to resolve when TODO fixed. It seems to me that the current docstring deviates slightly from your comment.

IvarStefansson (approved):
Thanks. Feel free to merge when all comments are resolved.

PR #1756, opened 2026-08-27
Remove protocol from mro

IvarStefansson:
## Proposed changes

`PorePyModel` is a `typing.Protocol` under `TYPE_CHECKING` and an empty placeholder class
at runtime ([protocol.py:26-30](src/porepy/models/protocol.py#L26-L30)). 122 classes inherit it.

The placeholder is empty, so it can never shadow a method. But it is a real class, so C3 must
order it after every class that inherits it. This can mess up the MRO.

**Fix:** make the runtime placeholder contribute nothing to the MRO via PEP 560
`__mro_entries__` returning `()`. `class DarcysLaw(pp.PorePyModel)` then compiles to
`class DarcysLaw:`. mypy is untouched — it only reads the `else:` branch.

The PR also 
- Adds testing of the PorePyModel protocol
- Reclaims old usage of PorePyModel for existing mixins, removing explicit type declaration.

## Types of changes

- [ ] Other:

## Checklist

### DIFF
--- a/src/porepy/models/momentum_balance.py
+++ b/src/porepy/models/momentum_balance.py
@@ -197,7 +197,7 @@ class MomentumBalanceEquations(pp.BalanceEquation):
         )
 
 
-class AngularMomentumEquation:
+class AngularMomentumEquation(pp.PorePyModel):
     """Conservation equation for the angular momentum balance.
 
     Intended used for mechanics formulations that employ the two-point stress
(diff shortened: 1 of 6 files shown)


src/porepy/models/momentum_balance.py
```diff
@@ -923,7 +892,8 @@ def set_initial_values_primary_variables(self) -> None:
             - :meth:`ic_values_total_pressure`
 
         """
-        # Super call for compatibility with multi-physics.
+        # Super call for compatibility with multi-physics. The method is not part
```
Yuriyzabegaev: Should this method be a part of the protocol?

Yuriyzabegaev (approved):
This solution looks elegant. I hope it will make our lives easier.

src/porepy/models/momentum_balance.py
IvarStefansson: Yes, I guess that makes sense. Thanks!

keileg (approved):
I hope never to think of this again.

PR #1761, opened 2026-09-01
BUG: Use norm, not component sum, in damage history zero-skip test

IvarStefansson:
## Proposed changes

`damage_convolution_integral` drops a history term when the quantity returned alongside
the damage length falls below tolerance. Both length functions (isotropic and anisotropic) returned the *sum* of the
tangential components of the slip increment. In 3D that sum vanishes for an increment
such as `(a, -a)` while the increment itself does not, so a genuine contribution could be
discarded. The norm is returned instead, which vanishes exactly when the increment does.

Latent rather than active: dropping a term requires every cell's component sum to fall
below tolerance, so realistic 3D slip would not have triggered it. Fixed because the
guard is wrong, not because a failure was observed. 2D is unaffected, there being one
tangential component.

Adds a regression test, parametrised over the isotropic and anisotropic length
functions, which fails on the previous implementation and passes on this one.

## Types of changes

## Checklist

- [x] The documentation is up-to-date. — docstrings and the tolerance comment updated to match.
- [x] If new skipped tests have been introduced in this PR, `pytest` was run with the `--run-skipped` flag. — n/a, no skipped tests added.

### DIFF
--- a/src/porepy/models/fracture_damage.py
+++ b/src/porepy/models/fracture_damage.py
@@ -236,11 +236,11 @@ class FractureDamageEquations(pp.PorePyModel):
 
         Parameters:
             length_function: Function that takes (subdomains, time_step_index) and
-                returns (contribution, constant_part) tuple.
+                returns a (contribution, increment_norm) tuple.
             damage_coefficient_function: Function returning the damage coefficient
                 operator for the current time step.
             subdomains: List of fracture subdomains.
-            tolerance: Tolerance for checking if constant part is non-zero.
+            tolerance: Tolerance for checking if the increment norm is non-zero.
 
         Returns:
             Operator for the damage equation.
@@ -256,14 +256,14 @@ class FractureDamageEquations(pp.PorePyModel):
         for i in range(1, num_steps):
             # i = number of steps back in time.
             damage_coefficient_i = damage_coefficient.previous_timestep(i)
-            length_i, constant_part_i = length_function(subdomains, i)
-            # Provided the contribution is the product of the constant part and some
-            # variable, constant_part_i = 0 implies contribution_i = 0 regardless of the
-            # variable's value. The damage coefficient is also treated as constant (it
-            # is evaluated at the previous time step).
+            length_i, increment_norm_i = length_function(subdomains, i)
+            # A vanishing slip increment gives a vanishing contribution for both length
+            # functions, regardless of the current state they are evaluated against, so
+            # such terms can be dropped from the sum. Both factors are evaluated at a
+            # previous time step and are therefore constant.
             constant_value = cast(
                 np.ndarray,
-                (constant_part_i * damage_coefficient_i).value(self.equation_system),
+                (increment_norm_i * damage_coefficient_i).value(self.equation_system),
             )
             if np.any(np.abs(constant_value) > tolerance):  # tolerance for zero check
                 eq += length_i * damage_coefficient_i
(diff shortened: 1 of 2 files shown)


src/porepy/models/fracture_damage.py
```diff
@@ -501,10 +500,13 @@ def damage_length(
         f_abs = pp.ad.Function(pp.ad.abs, "abs_function")
         contribution = f_abs(max_1 - max_0)
         # If time_step_index > 0, we can safely disregard the contribution if the
-        # displacement increment is zero. Return increment for checking before adding
-        # the contribution.
+        # displacement increment is zero. Return the increment norm for checking before
+        # adding the contribution; the norm vanishes if and only if the increment does,
+        # whereas summing the tangential components would report an increment such as
```
keileg: Is this a comment on how the code changed rather than how it is?

src/porepy/models/fracture_damage.py
```diff
@@ -501,10 +500,13 @@ def damage_length(
         f_abs = pp.ad.Function(pp.ad.abs, "abs_function")
         contribution = f_abs(max_1 - max_0)
         # If time_step_index > 0, we can safely disregard the contribution if the
-        # displacement increment is zero. Return increment for checking before adding
-        # the contribution.
+        # displacement increment is zero. Return the increment norm for checking before
```
keileg: I would say this comment belongs elsewhere: This method should compute the length; how it is used is a modeling choice made in the caller (or somewhere else).

src/porepy/models/fracture_damage.py
```diff
@@ -407,25 +407,24 @@ def damage_length(
...
         u_t = nd_vec_to_tangential @ self.plastic_displacement_jump(subdomains)
         u_t_increment = u_t.previous_timestep(time_step_index) - u_t.previous_timestep(
             time_step_index + 1
         )
 
         f_norm = pp.ad.Function(partial(pp.ad.l2_norm, self.nd - 1), "norm_function")
 
-        contribution = f_norm(u_t_increment)
+        # The norm vanishes if and only if the increment does, which is what the caller
+        # needs in order to discard a history term. Summing the tangential components
```
keileg: Is this comment about what changed, not what the code is now?

src/porepy/models/fracture_damage.py
```diff
@@ -256,14 +256,14 @@ def damage_convolution_integral(
         for i in range(1, num_steps):
```
keileg, line 256: If this loop is costly in some sense, it should be possible to drop it if the variable length is approximately zero. Not sure if that is helpful.

tests/models/test_fracture_damage_unit.py
```diff
@@ -683,3 +683,30 @@ def test_anisotropic_3d_oblique_direction_dot_product(self):
             d * np.sqrt(2.0) * np.ones(nc),
             rtol=1e-10,
         )
+
+    @pytest.mark.parametrize(
+        "isotropic", [True, False], ids=["isotropic", "anisotropic"]
+    )
+    def test_3d_zero_check_uses_norm_not_component_sum(self, isotropic: bool):
```
keileg: Better: Only a zero jump makes the term disapear.  I would set up/focus with two jumps u_t = (0, 0) and (-a, a). The last comment ('A failure here...') is misleading - a failure indicates that something is wrong, but not necessarily reversion to zeros.

keileg (commented):
Minor comments, should not be much work.

keileg (approved):

src/porepy/models/fracture_damage.py
IvarStefansson: Sort of. But it's also the explanation of why the norm is needed. I lean towards keeping.

src/porepy/models/fracture_damage.py
IvarStefansson: Do you mean skip without the evaluation below? I guess a simple caching might do the work. If we want to do that, I think that's a separate PR, no?

src/porepy/models/fracture_damage.py
IvarStefansson: See other answer to same comment.

src/porepy/models/fracture_damage.py
keileg: I think I misread the code in the OP and that what I had in mind would not work. You may be right that caching may help, but my point was anyhow nother more than to say that this may be simplified in case the loop bothers you in some sense. So no need to act if you don't feel like it.

IvarStefansson: I'll close here and investigate separately. Might open an issue.

PR #1772, opened 2026-09-08
Accelerate face recognition in TetrahedralGrid

wmboon:
There was a bottleneck in recognizing unique faces of tetrahedral grids. It becomes apparent on grids with `h = 0.01`.

## Proposed changes

Instead of using `np.unique(..., axis=1)` on the sorted face-node array, this PR first encodes the array using the numpy multi-index functionality. This is only possible because we know it consists of triplets of integers, each within the known range `[0, num_nodes)`. After encoding, we apply `unique` on the 1D array (which is much more efficient), and then decode.

On my machine, with mesh size 0.01 on the unit cube, this change accelerated `pp.TetrahedralGrid.__init__()` from 15.3s to 3.5s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,12 @@ Format: each entry is one line, `PR: Summary of the change.`
 Keep summaries to 1-2 sentences and written for a PorePy user, not just the
 author.
 
+The lists are sorted on PR numbers.
+
 ## Unreleased
 
 ### Changes
+PR 1772: SPEED: Faster construction of grid topologies on 3d simplex grids.
 PR 1770: Simplify PR template by deferring to CONTRIBUTING.md for details on code style and conventions.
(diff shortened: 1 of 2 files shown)


keileg (approved):
Thanks. I can confirm a speedup of ~8, it seems, on the relevant code, and a speedup of `~35-40%` for `pp.TetraherdalGrid(points)`. Nice work!

One caveat: The encoding introduce a numbering system where the highest number scales as `num_nodes**3`. It turns out this will trigger an overflow for `int64` when the number of nodes is about `2.05e6`. So, I introduced a fallback to the old case for such grids. This is of course unfortunate, the bigger the grid, the more useful the fix, but it will have to do.

FYI, we have also introduced a changelog that we are trying to use. I took care of it for this PR. I imagine this can be of use for pyGeon as well.

PR #1776, opened 2026-09-15
SPEED: Optimized compute_geometry for tet grids

wmboon:
## Proposed changes

In our workflow (on the PyGeoN side), the only fine grids we consider in 3D are tetrahedral. For those grids, the `compute_geometry` function has unnecessary overhead because it is made for general polyhedral grids.

This PR introduces an optimized `_compute_geometry_3d_simplices` routine for tetrahedral grids. On my machine, it consistently reaches a speed-up factor 10 compared to `_compute_geometry_3d`. We fall back to the general implementation if the grid is not tetrahedral.


src/porepy/grids/grid.py
```diff
@@ -777,6 +787,55 @@ def bincount_nd(arr, weights):
...
+        self.face_areas = np.sqrt(dot(self.face_normals, self.face_normals))
+
+        # Finally, the volume of cell K can be computed by using the divergence theorem:
+        # 3|K| = int_K 3 dx = int_K div(x) dx = sum_F int_F n dot x ds
+        # where the final integral is over the faces F of K. Since the integrand is
+        # linear and each face is planar, we evaluate it exactly using the midpoint
+        # rule, i.e. in the face center.
+        self.cell_volumes = (
+            dot(self.face_centers, self.face_normals) @ self.cell_faces / 3
+        )
```
keileg, line 837:
```suggestion
        )
        if not np.all(self.cell_volumes > -1e-12): 
            raise ValueError("Some tetrahedra have negative volume") 
```
Mirror the sanity check in the polyhedral implementation. I have lost count of how many times that check has detected grids that are bad in way sense or another.

src/porepy/grids/grid.py
```diff
@@ -777,6 +787,55 @@ def bincount_nd(arr, weights):
         self.cell_centers = cell_centers
         self.cell_volumes = cell_volumes
 
+    def _compute_geometry_3d_simplices(self) -> None:
```
keileg, line 790: Having a test for this would be good. My suggestion would be to check for consistency with the general implementation. Something like


```python

pts = np.random.rand(3, 10)
g = pp.TetrahedralGrid(p)
g_copy = g.copy()
g.compute_geometry()

g_copy._compute_geometry_3d()
for attr in ['face_centers', 'face_areas' ...]:
  np.testing.assert_equal(getattr(g, attr), getattr(g_copy, attr)

```

src/porepy/grids/grid.py
```diff
@@ -777,6 +787,55 @@ def bincount_nd(arr, weights):
...
+        """Optimized geometry computation for tetrahedral grids. This function uses
+        several shortcuts that are only available for tet-grids.
+        """
+        # The face centers are the average of the 3 adjacent node coordinates.
```
keileg, line 794: Excellent documentation throughout. This is really appreciated.

keileg (commented):
Thanks! A few minor comments.

src/porepy/grids/grid.py
wmboon: Thanks!

src/porepy/grids/grid.py
wmboon: Makes sense, I added the sanity check.

src/porepy/grids/grid.py
wmboon: Test added, thanks for the suggestion

keileg (approved):
Thank you for the contribution, this will be helpful.

